PR #20, opened 2019-12-20
test: improve performance of createFullTransform util

marionebl:
* Introduce a virtual filesystem to `createFullTransform` util
* Mock `pkg.name` during transformation
* Avoid compilation of on-disk project as much as possible


src/__tests__/utils/transform.tsx
```diff
@@ -1,51 +1,120 @@
...
 import * as path from 'path';
+import * as resolveModule from 'resolve';
+import pkg from '../../../package.json';
+import { copy, mkdirp } from './memfs';
 
 const printer = ts.createPrinter();
 
+declare module 'resolve' {
+  export interface SyncOpts extends resolveModule.Opts {
+    /** how to read files synchronously (defaults to fs.readFileSync) */
```
madou: is this not covered in `@types/resolve`?

src/__tests__/utils/transform.tsx
```diff
@@ -1,51 +1,120 @@
...
-    });
-
-    const [rootFile] = files;
+    const fs = (Volume.fromJSON(
+      Object.entries(sources).reduce<Record<string, string>>(
+        (acc, [name, content]) => ({
+          ...acc,
+          [`/${name}.tsx`]: content,
+        }),
+        {}
```
madou: if we type this can we skip having to cast it to `unknown`?

src/__tests__/utils/transform.tsx
```diff
@@ -1,51 +1,120 @@
...
+    createMockModule(pkg.name, fs);
+
     const config: ts.CompilerOptions = {
       jsx: ts.JsxEmit.Preserve,
+      lib: ['/node_modules/typescript/lib/lib.esnext.full.d.ts'],
       module: ts.ModuleKind.ESNext,
+      moduleResolution: ts.ModuleResolutionKind.NodeJs,
       suppressImplicitAnyIndexErrors: true,
+      resolveJsonModule: true,
+      skipLibCheck: true,
```
madou: oh nice - would this have sped up the project creation?

src/__tests__/utils/transform.tsx
```diff
@@ -1,51 +1,120 @@
...
+
+    compilerHost.fileExists = file => fs.existsSync(file);
+
+    compilerHost.resolveModuleNames = (names, containingFile) => {
+      return names.map(name => {
+        return {
+          resolvedFileName: resolveModule.sync(name, {
+            basedir: path.dirname(containingFile),
+            extensions: ['.js', '.json', '.node', '.tsx', '.ts', '.d.ts'],
+            readFileSync: path => fs.readFileSync(path),
```
madou: so using the memfs we're like half using it? since some parts of ts still need the real fs? 🤔

src/__tests__/utils/transform.tsx
```diff
@@ -79,3 +148,22 @@ export const createTransform = (programTransformer: ProgramTransformer) => (
   const actual = ts.transform(sourceFile, [transformer]).transformed[0];
   return printer.printFile(actual).toString();
 };
+
+function createMockModule(name: string, fs: typeof Fs) {
+  mkdirp({ path: `/node_modules/${name}`, fs });
+
+  fs.writeFileSync(`/node_modules/${name}/index.js`, `module.exports = {};`);
+  fs.writeFileSync(
+    `/node_modules/${name}/index.d.ts`,
```
madou: is there any way around hardcoding this type def?

src/ts-transformer/css-prop/__tests__/index.test.tsx
```diff
@@ -159,7 +157,7 @@ describe('css prop transformer', () => {
       expect(actual).toInclude('<style>.test-class:hover{color:blue;}</style>');
     });
 
-    it.only('should transform object with object selector from variable', async () => {
+    it('should transform object with object selector from variable', async () => {
```
madou: hehe thanks, we should setup eslint

madou:
hey so i pulled down your branch to have a look - honestly everything looks great! if we can just rename some variables so its more obvious we aren't using `fs` would be fantastic.

other than that would be awesome to see how we can use the real source for the project instead of hardcoding a mock - but we can probably figure that out later than now.

if you want to update variables lmk - else i'm happy to merge this and make changes later :)

src/__tests__/utils/transform.tsx
```diff
@@ -1,51 +1,120 @@
...
+      lib: ['/node_modules/typescript/lib/lib.esnext.full.d.ts'],
       module: ts.ModuleKind.ESNext,
+      moduleResolution: ts.ModuleResolutionKind.NodeJs,
       suppressImplicitAnyIndexErrors: true,
+      resolveJsonModule: true,
+      skipLibCheck: true,
       target: ts.ScriptTarget.ESNext,
+      types: [],
       // Uncomment this if shit isn't working.
       // noEmitOnError: true,
```
madou: this can be true now all the time

src/__tests__/utils/transform.tsx
marionebl: Yes, I opened a PR to change it: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41175

src/__tests__/utils/transform.tsx
marionebl: Yes, about 10% for our use case according to my tests.

src/__tests__/utils/transform.tsx
marionebl: There are some differences between `typeof Fs` and `typeof Volume`, I'll have to fiddle a bit to make good typings for the memfs utilities so this can go away. I started work on https://github.com/marionebl/ts-transformer-testing-library to that end. I'd suggest we land this and then have a look on how to use the external lib with better typings and encapsulation.

src/__tests__/utils/transform.tsx
marionebl: :+1:

src/__tests__/utils/transform.tsx
marionebl: Sure, did this in my initial push towards an external helper lib: https://github.com/marionebl/ts-transformer-testing-library/blob/master/src/transform-file.ts#L82

If you think this important we can backport it.

marionebl:
> hey so i pulled down your branch to have a look - honestly everything looks great! if we can just rename some variables so its more obvious we aren't using fs would be fantastic.

Glad you like it

> other than that would be awesome to see how we can use the real source for the project instead of hardcoding a mock - but we can probably figure that out later than now.

There is tradeoff between faithfulness and performance here - maybe we can have specific (slow) tests that do that?

> if you want to update variables lmk - else i'm happy to merge this and make changes later :)

Will rename the variables.

src/__tests__/utils/transform.tsx
madou: awesome - for now move it to a resolve.d.ts under the `src/types` folder :)

src/__tests__/utils/transform.tsx
madou: oh awesome let's use that when you've fleshed it out more!

yeah sounds good

src/__tests__/utils/transform.tsx
madou: ah cool - nah not for now - let's just use your lib after it's ready :)

madou:
hey what's your twitter? let's stay connected on there :) would love to talk to you about how i can do things better with ts transformers etc too.

marionebl:
> hey what's your twitter? let's stay connected on there :) would love to talk to you about how i can do things better with ts transformers etc too.

Unfortunately I deleted my twitter handle 1-2 years ago - but you can find me e.g. on node-tooling.slack.com.

madou:
hmm actions aren't running for your branch 🤔

PR #21, opened 2019-12-20
Add CI scripts

marionebl:
Add basic Github actions


madou:
ah cool makes sense to use github actions :) 

if/when this gets put into the atlassian repo we'll have to rethink it - i believe the plan we're on atm doesn't support actions even with public repos

PR #25, opened 2019-12-23
test: use externalized helpers

marionebl:


src/ts-transformer/css-prop/__tests__/index.test.tsx
```diff
@@ -156,27 +159,25 @@ describe('css prop transformer', () => {
...
-            }}>
-            Hello, world!
-          </div>
-        `,
-        mixins: `
-          export const mixin = { color: 'blue' };
-        `,
-      });
+    it('should transform object with object selector from variable', () => {
+      const actual = transformer.addSource({
```
madou: when you add source does it mutate the local above (does every test now have a source?)

src/ts-transformer/css-prop/index.tsx
```diff
@@ -103,7 +103,7 @@ export default function cssPropTransformer(
       // We need to rebind after transforming. Why isn't this in the typed API?
       // @ts-ignore
       ts.bindSourceFile(completedNode, {});
-      console.log(completedNode);
+      // console.log(completedNode);
```
madou: nice catch we should delete it :D

src/ts-transformer/css-prop/__tests__/index.test.tsx
marionebl: Currently it does, I'll change the implementation to create new instances for each call to maintain proper encapsulation here.

madou:
hey @marionebl fyi some tests were breaking. the ones that were referencing the `react` module (just made a stub insted) - and then the mixin test (file name was wrong).

easy fix though :)

with react - do you suggest making a mock every time its needed? or should we/can we reference the `node_modules` one?

PR #26, opened 2020-01-01
fix: unbreak ts-loader compat by avoiding node mutations

### LINKED ISSUE #23: Fix transformer blowing up when transforming imported variables
See: https://github.com/madou/untitled-css-in-js-project/pull/22

It's blowing up when iterating through the JSX attributes (`className` doesn't have a parent = blows up).

Need to understand why - if it's a bug - and if we can work around it. It works fine in tests - blows up when running storybook/webpack/ts-loader though.

---

Found this https://github.com/longlho/ts-transform-css-modules - should be good reference.

Edit: Or not because it follows css files not typescript. Meh.

marionebl:
closes #23


madou:
cheers man you're a legend, will fix the tests after merging

Issue #50, opened 2020-01-10
Add ability to assert multiple properties at once in jest matcher


madou:
Right now we can use it like this:

```js
expect(getByText('hello world')).toHaveCompiledCss('font-size', '15px');
```

Would be great it we could..

```js
expect(getByText('hello world')).toHaveCompiledCss({
  fontSize: '15px',
  color: 'red',
});
```

madou:
Fixed in #54

PR #54, opened 2020-01-15
feat(jest-matcher): allow multiple properties assertion

vanya829:
Hi there.
[#50] I thought that this issue is interesting and simple enough to be a first-time issue.
Please have a look, if this PR is what you expected from the jest matcher.

### DIFF
--- a/src/class-names/__tests__/index.test.tsx
+++ b/src/class-names/__tests__/index.test.tsx
@@ -10,6 +10,6 @@ describe('class names component', () => {
       </ClassNames>
     );
 
-    expect(getByText('hello world')).toHaveCompiledCss({ 'font-size': '13px' });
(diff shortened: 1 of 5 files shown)


madou:
hi! i love it thanks for contributing! how are you going?

can we support both apis? so for example when we're interested in just one value i'll...

```
expect(xyz).toHaveCompiledCss('color', 'blue');
```

but when i'm interested in multiple and want to keep it to a single assertion...

```
expect(xyz).toHaveCompiledCss({
  fontSize: '12px',
  color: 'blue',
});
```

vanya829:
Thanks for such a fast reply. Update PR with the support of both apis.

madou:
awesome 🎉 

thanks for your first contribution!

PR #72, opened 2020-03-14
Adds small runtime to move compiled styles to the head

### LINKED ISSUE #61: Add styles to head when ran on the client
Smaller the better

Thinking something like replacing the `<style>` element we add for every component with an equivalent custom component.

```
const Style = (props) => {
  if (/* on the client */ && /* not yet in head */) {
    /* add styles to head */
    return null;
  }

  /* on the server */
  return <style data-compiled-id={props.id}>{props.children}</style>;
};
```

- should not add if its already there
- should only do the calculation once - if it re-renders do nothing
- should render style element on the server
- should be as small and fast as possible
- need to figure out what to do with the whole "no runtime" thing since this breaks it - but it's kind of needed
- should be FAST - probably needs to use the same browser APIs that styled/emotion use
- needs to be able to be removed via a transformer

---

Inspiration points:

- https://github.com/emotion-js/emotion/tree/master/packages/sheet
- https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/sheet/Rehydration.js#L55
- https://github.com/emotion-js/emotion/blob/master/packages/utils/src/index.js#L23
- https://github.com/emotion-js/emotion/blob/master/packages/cache/src/index.js#L71

madou:
Closes #61


package.json
```diff
@@ -5,10 +5,11 @@
   "license": "Apache-2.0",
   "scripts": {
     "clean": "workspaces-run -- rm -rf dist -- rm -f tsconfig.tsbuildinfo",
-    "start": "npx nodemon --exec \"start-storybook -p 6006 --ci\" --watch packages/ts-transform/ -e tsx",
+    "start": "yarn build && npx nodemon --exec \"start-storybook -p 6006 --ci\" --watch packages/ts-transform/ -e tsx",
+    "start:prod": "NODE_ENV=production yarn start",
     "start:inspect": "npx nodemon --exec \"node --inspect-brk node_modules/.bin/start-storybook -p 6006 --ci\" --watch packages/ts-transform/ -e tsx",
-    "test": "jest",
-    "test:cover": "jest --collectCoverage",
+    "test": "yarn build && jest --no-cache",
```
danieldelcore: > Note: the cache should only be disabled if you are experiencing caching related problems. On average, disabling the cache makes Jest at least two times slower.

danieldelcore: You alright with that or is it a left over from some testing?

madou: yeah - when playing with the transformer it gets cached too easily and ends up being... annoying. because then i have to clear it (when I remember that that's probably the issue hehe)

i figured the easy solution for now is to just not cache it.

madou:
thanks mate!

madou:
@obweger can i get admin back pls - and/or give dan write access to this repo

PR #133, opened 2020-04-16
chore: remove css hash prefix

danieldelcore:
Addresses: https://github.com/atlassian-labs/compiled-css-in-js/issues/122


packages/ts-transform/src/utils/hash.tsx
```diff
@@ -46,7 +46,7 @@ export function hash(str: string, seed: number = 0): string {
 }
 
 export const classNameHash = (css: string): string => {
-  return `${CLASS_NAME_PREFIX}-${hash(css)}`;
```
madou, line 49: hey we need this still because numbers aren't a valid starting character for a class

ideally we want to add this back and then remove the prefix from the style hash, let me find you the location to change

madou:
will reopen

PR #146, opened 2020-04-17
feat: adds csp nonce support

### LINKED ISSUE #69: Add csp nonce support
See: https://github.com/styled-components/styled-components/issues/887

madou:
Closes #69 

TODO:
- [x] Add option getter inside babel plugin

### DIFF
--- a/examples/class-names-dynamic-object.tsx
+++ b/examples/class-names-dynamic-object.tsx
@@ -2,7 +2,7 @@ import React, { useState } from 'react';
 import { ClassNames } from '@compiled/css-in-js';
 
 export default {
-  title: 'class names sdynamic object',
+  title: 'class names dynamic object',
 };
 
 export const objectLiteral = () => {
--- a/examples/class-names-dynamic-object.tsx
+++ b/examples/class-names-dynamic-object.tsx
@@ -2,7 +2,7 @@ import React, { useState } from 'react';
 import { ClassNames } from '@compiled/css-in-js';
 
 export default {
-  title: 'class names dynamic object',
+  title: 'class names sdynamic object',
 };
 
 export const objectLiteral = () => {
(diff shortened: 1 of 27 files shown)


examples/class-names-dynamic-object.tsx
```diff
@@ -2,7 +2,7 @@ import React, { useState } from 'react';
 import { ClassNames } from '@compiled/css-in-js';
 
 export default {
-  title: 'class names dynamic object',
+  title: 'class names sdynamic object',
```
danieldelcore, line 5: typo

madou: gottem

theKashey:
`nonce` is not supposed to be exposed at `window`. You should "hide" it, well, nobody explains how.
However `__webpack_nonce__` is not a global variable - it is a local variable, injected via _local module variable_.

I would propose use a module-in-the-middle to handle this stuff - https://github.com/thekashey/get-nonce
It makes debugging/mocking a bit easier as well.

theKashey:
In short - idea behind `nonce` - don't let some _non-trusted-stuff_ like ads or browser plugins inject code you do not expect. 
As a result exposing `nonce` as a global variable, or expecting that it could be read from a global variable, is ruing a whole idea.

madou:
hey thanks anton good point. i'll change the example to use something hardcoded for now.

when it comes to actual usage - it just sets a variable which webpack can then pick up and modify - that would work fine right? (so no changes to the implementation needed)

theKashey:
It's not about webpack. It is:
- about `csp` *headers* your server is sending (that is not webpack)
- or about `csp` *meta* you added into the `<head>`(that is not webpack)
- and about the `nonce` generation (that is not webpack)
- and the ability for different pieces to read and use `nonce` if it's required
- actually - `script nonce` and `style nonce` could be different, and the second it not that required.

So that's not so simple :(

madou:
thanks for the info!

for this library though i can't imagine needing to do anything more? other than ensuring there is good documentation around how to use nonce properly. and re-enforcing what you said in your last point

Issue #120, opened 2020-04-11
Jest matcher can't find styles when composing from multiple sources


madou:
It should look up both class names.

```
    const StyledDiv = styled.div`
      font-size: 12px;
    `;

    const { getByText } = render(<StyledDiv css={{ fontSize: 14 }}>Hello world</StyledDiv>);

    expect(getByText('Hello world')).toHaveCompiledCss('font-size', '14px');
```

```
pairing style element was not found
```

PR #147, opened 2020-04-17
refactor: jest matcher now supports multiple style tags

danieldelcore:
Closes: #120 

Styles coming from multiple sources are added to the head as separate style elements to the head.
This PR adds support for multiple style elements and class names. If we are seeking properties from two separate sources (css prop and styled api) the matcher is now able to pick it up. 

Bonus: we're now able to output all mounted css properties in the error message. 
  
image

### DIFF
--- a/packages/jest/src/__tests__/matchers.test.tsx
+++ b/packages/jest/src/__tests__/matchers.test.tsx
@@ -1,57 +1,88 @@
 import { render } from '@testing-library/react';
 import React from 'react';
-import '@compiled/css-in-js';
+import { styled } from '@compiled/css-in-js';
 
-it('should detect styles', () => {
-  const { getByText } = render(
-    <div
-      css={{
-        fontSize: '12px',
-      }}>
-      hello world
-    </div>
-  );
-
-  expect(getByText('hello world')).toHaveCompiledCss('font-size', '12px');
-});
-
-it('should detect missing styles', () => {
-  const { getByText } = render(<div css={{ fontSize: '12px' }}>hello world</div>);
-
-  expect(getByText('hello world')).not.toHaveCompiledCss('color', 'blue');
-});
-
-it('should detect multiple styles', () => {
-  const { getByText } = render(<div css={{ fontSize: '12px', color: 'blue' }}>hello world</div>);
-
-  expect(getByText('hello world')).toHaveCompiledCss({
-    fontSize: '12px',
-    color: 'blue',
+describe('toHaveCompliedCss', () => {
+  afterEach(() => {
+    document.getElementsByTagName('html')[0].innerHTML = '';
   });
-});
 
-it('should detect single missing styles', () => {
-  const { getByText } = render(<div css={{ fontSize: '12px', color: 'blue' }}>hello world</div>);
+  it('should detect styles', () => {
+    const { getByText } = render(
+      <div
+        css={{
+          fontSize: '12px',
+        }}>
+        hello world
+      </div>
+    );
 
-  expect(getByText('hello world')).not.toHaveCompiledCss({
-    zindex: '9999',
+    expect(getByText('hello world')).toHaveCompiledCss('font-size', '12px');
   });
-});
 
-it('should detect multiple missing styles', () => {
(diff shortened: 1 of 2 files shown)


packages/jest/src/__tests__/matchers.test.tsx
```diff
@@ -1,57 +1,88 @@
...
-
-  expect(getByText('hello world')).toHaveCompiledCss('font-size', '12px');
-});
+import { styled } from '@compiled/css-in-js';
 
-it('should detect missing styles', () => {
-  const { getByText } = render(<div css={{ fontSize: '12px' }}>hello world</div>);
+describe('toHaveCompliedCss', () => {
+  afterEach(() => {
+    document.getElementsByTagName('html')[0].innerHTML = '';
```
danieldelcore, line 7: Noticed that we need to clear the head after every test because styles are left behind after every test :(

madou: is that really a problem?

packages/jest/src/__tests__/matchers.test.tsx
```diff
@@ -1,57 +1,88 @@
...
+    expect(getByText('Hello world')).toHaveCompiledCss('font-size', '14px');
+  });
+
+  it('should find multiple styles composed from multiple sources', () => {
+    const StyledDiv = styled.div`
+      color: blue;
+      background-color: red;
+    `;
+
+    const { getByText } = render(<StyledDiv css={{ color: 'blue' }}>Hello world</StyledDiv>);
```
madou, line 81: can we change this to `color: red` so its different to the initial StyledDiv?

packages/jest/src/matchers.tsx
```diff
@@ -4,84 +4,74 @@ const kebabCase = (str: string) =>
...
-    // There wasn't a style element found within - let's check the head for it instead.
-    const styleElements = Array.from(document.head.querySelectorAll('style'));
-    for (const tag of styleElements) {
-      if (tag.innerHTML.includes(element.className)) {
-        styleElement = tag as HTMLStyleElement;
-        break;
-      }
-    }
-  }
+  const inlineStyleTag = element.parentElement && element.parentElement.querySelector('style');
```
madou, line 25: is this correct? there could be multiple style tags tied to a singe element (like the composed StyleDiv test above)

danieldelcore: Hey, im might be wrong here but i thought this selector was for style tags in the JSX pragma? That's why it attempts to get it from parent of the actual element?

packages/jest/src/__tests__/matchers.test.tsx
danieldelcore: yeah might make more sense to change the color on line 77. Good point!

packages/jest/src/__tests__/matchers.test.tsx
danieldelcore: maybe not, if we only attempt to get styles for the provided css classes 🤔

packages/jest/src/__tests__/matchers.test.tsx
danieldelcore: done 👍

packages/jest/src/matchers.tsx
madou: its for the SSR case/local dev

```
<StyledDiv css={{ color: 'blue' }}>
```

ends up making

```
<style>.a
<style>.b
<div class="a b">
```

does this work?

packages/jest/src/__tests__/matchers.test.tsx
danieldelcore: Updated code to narrow down to properties contained within the supplied classes

packages/jest/src/matchers.tsx
danieldelcore: Hey man, not sure if the matcher would work in server mode currently. Because currently we're looking at elements rather than html strings. Am i missing something?

madou: ah well, yeah, true

all good

danieldelcore: Probably wouldn't be too hard to get working in the future though. was just playing around with it then :)

PR #148, opened 2020-04-17
chore: remove matcher types from css in js

### LINKED ISSUE #135: Tests files are included in build bundles
**Describe the bug**
Tests files are included in build bundles

**To Reproduce**
`yarn build`

**Expected behavior**
__test__ should not be parsed at build time and directory should be excluded from final bundle 

**Screenshots**
Screen Shot 2020-04-16 at 1 59 54 pm

### LINKED ISSUE #139: Move jest matcher types out of css in js
```
declare global {
  namespace jest {
    interface Matchers<R> {
      toHaveCompiledCss(properties: { [key: string]: string }): R;
      toHaveCompiledCss(property: string, value: string): R;
    }
  }
}
```

Shouldn't be there... but make sure there is a way forward for consumers to get types working properly (documented!)

madou:
closes #139 
closes #135

### DIFF
--- a/packages/css-in-js/src/class-names/__tests__/index.test.tsx
+++ b/packages/css-in-js/src/class-names/__tests__/index.test.tsx
@@ -1,6 +1,7 @@
 import { render } from '@testing-library/react';
 import React from 'react';
 import { ClassNames } from '@compiled/css-in-js';
+import '@compiled/jest-css-in-js';
 
 describe('class names component', () => {
   it('should create css from object literal', () => {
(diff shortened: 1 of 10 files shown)


flexdinesh:
Do we need to add `import 'jest-extended';` in every test file? Can we add it somewhere in the entry?

madou:
> Do we need to add import 'jest-extended'; in every test file? Can we add it somewhere in the entry?

i tried to but i couldn't get it working. if you have any ideas lmk - can do a follow up! just wanted to be pragmatic for now.

PR #149, opened 2020-04-17
config: introduces eslint 

### LINKED ISSUE #38: Setup eslint
pls someone

danieldelcore:
Closes #38 

Introduces eslint:
- Fixes failing checks
- Adds lint check to test github workflow

### DIFF
--- a/examples/class-names-dynamic-object.tsx
+++ b/examples/class-names-dynamic-object.tsx
@@ -5,7 +5,7 @@ export default {
   title: 'class names dynamic object',
 };
 
-export const objectLiteral = () => {
+export const ObjectLiteral = () => {
   const [color, setColor] = useState('blue');
 
   return (
--- a/package.json
+++ b/package.json
@@ -12,6 +12,8 @@
     "test": "yarn build && jest --no-cache",
     "test:watch": "yarn build && jest --no-cache --watch",
     "test:cover": "yarn test --collectCoverage",
+    "lint": "eslint --config .eslintrc.js --ext tsx,ts ./packages/**/src ./examples",
+    "lint:fix": "yarn lint -- --fix",
     "build": "tsc --build packages",
     "build:inspect": "node --inspect-brk node_modules/typescript/lib/tsc.js --build packages",
     "build-storybook": "build-storybook",
@@ -35,7 +37,14 @@
     "@types/jest": "24.0.19",
     "@types/node": "^12.12.17",
     "@types/styled-components": "^5.0.1",
+    "@typescript-eslint/eslint-plugin": "^2.28.0",
+    "@typescript-eslint/parser": "^2.28.0",
     "babel-loader": "^8.0.6",
+    "eslint": "^6.8.0",
+    "eslint-config-prettier": "^6.10.1",
+    "eslint-plugin-prettier": "^3.1.3",
+    "eslint-plugin-react": "^7.19.0",
+    "eslint-plugin-react-hooks": "^3.0.0",
     "husky": "^3.1.0",
     "jest": "24.9.0",
     "jest-extended": "^0.11.2",
@@ -57,7 +66,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "pretty-quick --staged"
+      "pre-commit": "pretty-quick --staged && yarn lint:fix"
     }
   },
   "workspaces": [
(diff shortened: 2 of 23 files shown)


examples/class-names-dynamic-object.tsx
```diff
@@ -5,7 +5,7 @@ export default {
   title: 'class names dynamic object',
 };
 
-export const objectLiteral = () => {
+export const ObjectLiteral = () => {
```
danieldelcore, line 8: Not able to use hooks in 'non components' 
The rule checks to see if the component starts with an uppercase letter. It will throw otherwise

package.json
```diff
@@ -12,6 +12,8 @@
     "test": "yarn build && jest --no-cache",
     "test:watch": "yarn build && jest --no-cache --watch",
     "test:cover": "yarn test --collectCoverage",
+    "lint": "eslint --config .eslintrc.js --ext tsx,ts ./packages/**/src ./examples",
+    "lint:fix": "yarn lint -- --fix",
```
madou, line 16: can we add a pre-commit hook for this? + prettier

packages/ts-transform/src/utils/ast-node.tsx
```diff
@@ -1,5 +1,12 @@
 import * as ts from 'typescript';
 
+export const createNodeError = (message: string, node: ts.Node) => {
+  throw new Error(`@compiled/css-in-js => ${message}
+
+${node.getText()}
+`);
+};
+
```
danieldelcore, line 9: Moved this method to the top because of a rule that requires functions to be declared before use. Not sure if this is something you want to keep.

package.json
```diff
@@ -36,6 +38,11 @@
     "@types/node": "^12.12.17",
     "@types/styled-components": "^5.0.1",
     "babel-loader": "^8.0.6",
+    "eslint": "^6.8.0",
+    "eslint-config-prettier": "^6.10.1",
+    "eslint-plugin-prettier": "^3.1.3",
```
madou, line 43: do we need this?

package.json
```diff
@@ -74,5 +81,9 @@
         "react"
       ]
     }
-  ]
+  ],
+  "dependencies": {
+    "@typescript-eslint/eslint-plugin": "^2.28.0",
```
madou, line 86: dev dependency for consistency

packages/ts-transform/src/utils/css-builder.tsx
```diff
@@ -13,7 +13,7 @@ export const buildCss = (
   variableDeclarations: Declarations,
   context: ts.TransformationContext
 ): ToCssReturnType => {
-  let cssToPassThroughCompiler: string = '';
+  let cssToPassThroughCompiler = '';
```
danieldelcore, line 16: Removed annotation because it is 'trivially inferred'

packages/ts-transform/src/utils/hash.tsx
```diff
@@ -6,7 +6,7 @@ import { CLASS_NAME_PREFIX, CSS_VARIABLE_PREFIX } from '../constants';
  * @param str
  * @param seed
  */
-export function hash(str: string, seed: number = 0): string {
+export function hash(str: string, seed = 0): string {
```
danieldelcore, line 9: Same here

examples/class-names-dynamic-object.tsx
madou: hehe i was wondering!

package.json
danieldelcore: yeah probably, i'll have a look

package.json
danieldelcore: Up to you, we can remove it if you like

package.json
danieldelcore: yeah not sure how that would play with pretty-quick 🤔

package.json
danieldelcore: done :)

packages/ts-transform/src/utils/ast-node.tsx
madou: is ok

package.json
madou: im fine just using prettier defaults. i dont think we need eslint ones as well.

package.json
danieldelcore: nevermind, should be fine to call it after 👍 added

package.json
```diff
@@ -57,7 +66,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "pretty-quick --staged"
+      "pre-commit": "pretty-quick --staged && yarn lint:fix"
```
madou, line 69: run lint fix first just in case - then prettier will run over the fixed code

PR #150, opened 2020-04-17
Support calc property; FIX: Object property with suffix;

### LINKED ISSUE #129: Css prop suffix extraction not working as expected
```
<div css={{ marginLeft: `${heading.depth}rem` }}>
</div>
```

```
--var-5un9uz: 2;
margin-left: var(--var-5un9uz);
```

where'd `rem` go?

flexdinesh:
- Add support to use `calc` property in css-prop
- FIX: Usage of suffix in literals with object properties (`marginLeft: ${obj.prop}rem`)
- FIX: This change magically fixed a dormant bug where when more than one prop was used in BinaryExpression code paths, only the dynamic property was picked in the css

Closes #129

### DIFF
--- a/packages/ts-transform/src/css-prop/__tests__/index.test.tsx
+++ b/packages/ts-transform/src/css-prop/__tests__/index.test.tsx
@@ -361,7 +361,9 @@ describe('css prop transformer', () => {
         );
       `);
 
-      expect(actual).toInclude('.css-test{color:var(--var-test);}');
+      expect(actual).toInclude(
+        '.css-test{color:var(--var-test);text-transform:uppercase;font-weight:600;}'
+      );
       expect(actual).toInclude(
         '<span className="css-test" style={{ "--var-test": props.color }}>{props.children}</span>'
       );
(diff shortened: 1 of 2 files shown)


packages/ts-transform/src/css-prop/__tests__/index.test.tsx
```diff
@@ -361,7 +361,9 @@ describe('css prop transformer', () => {
         );
       `);
 
-      expect(actual).toInclude('.css-test{color:var(--var-test);}');
+      expect(actual).toInclude(
+        '.css-test{color:var(--var-test);text-transform:uppercase;font-weight:600;}'
```
madou, line 365: love this fix, silly i missed it back when 😅

packages/ts-transform/src/css-prop/__tests__/index.test.tsx
```diff
@@ -595,6 +597,52 @@ describe('css prop transformer', () => {
...
+        import '@compiled/css-in-js';
+        import React from 'react';
+
+        const fontSize = 20;
+
+        <div css={{ fontSize: \`calc(100% - \${fontSize}px)\` }}>hello world</div>
+      `);
+
+      expect(actual).toInclude('style={{ "--var-test": fontSize + "px" }}');
+      expect(actual).toInclude('.css-test{font-size:calc(100% - var(--var-test));}');
```
madou, line 627: this is wrong it needs to be

```
style={{ "--var-test": "calc100% - " + fontSize + "px" }}

.css-test{font-size:var(--var-test);}
```

packages/ts-transform/src/css-prop/__tests__/index.test.tsx
```diff
@@ -595,6 +597,52 @@ describe('css prop transformer', () => {
...
+        import React from 'react';
+
+        const heading = {
+          depth: 20
+        };
+
+        <div css={{ marginLeft: \`\${heading.depth}rem\` }}>hello world</div>
+      `);
+
+      expect(actual).toInclude('style={{ "--var-test": heading.depth + "rem" }}');
```
madou, line 612: good fix

packages/ts-transform/src/css-prop/__tests__/index.test.tsx
```diff
@@ -595,6 +597,52 @@ describe('css prop transformer', () => {
...
+
+        const heading = {
+          depth: 20
+        };
+
+        <div css={{ marginLeft: \`calc(100% - \${heading.depth}rem)\` }}>hello world</div>
+      `);
+
+      expect(actual).toInclude('style={{ "--var-test": heading.depth + "rem" }}');
+      expect(actual).toInclude('.css-test{margin-left:calc(100% - var(--var-test));}');
```
madou, line 643: this is wrong too

madou:
good job so far! but we're only half way there

getting close though!

packages/ts-transform/src/css-prop/__tests__/index.test.tsx
madou: im wrong!

packages/ts-transform/src/css-prop/__tests__/index.test.tsx
madou: no its me!

packages/ts-transform/src/utils/template-literal-to-css.tsx
```diff
@@ -18,9 +18,18 @@ const extractSuffix = (tail: string) => {
   if (tail[0] === '\n' || tail[0] === ';') {
     rest = tail;
   } else {
-    // Sometimes people forget to put a comma at the end.
-    // This handles that case.
-    let tailIndex = tail.indexOf(';') === -1 ? tail.indexOf('\n') : tail.indexOf(';');
+    // Sometimes people forget to put a semi-colon at the end.
+    let tailIndex;
+    if (tail.indexOf(')') !== -1) {
```
madou, line 23: can you leave a comment here explaining the use case its solving? (for calc, transform, etc)

flexdinesh: Added

madou:
love it

PR #153, opened 2020-04-18
feat: sourcemap support

### LINKED ISSUE #118: Investigate what we need to do for sourcemap support
See: https://github.com/emotion-js/emotion/issues/15 + https://github.com/emotion-js/emotion/blob/master/packages/babel-plugin-emotion/src/utils/source-maps.js

Example snapshots here: packages/babel-plugin-emotion/test/__snapshots__/source-map.test.js.snap

Essentially we want this to work (css declaration is tied to some JavaScript): 

image

madou:
to run locally

```
git clone git@github.com:atlassian-labs/compiled-css-in-js.git
cd compiled-css-in-js
git checkout source-maps
yarn
yarn start
```

then look at sourcemaps through storybook. you'll see generated inline sourcemaps in the head inside the inline `style` elements (in the story iframe).

---

closes #118 

so i mean i got some names showing up in dev tools. thats cool.

now when clicking on the link tho it still goes into inline. im not sure if type checking on/off is going to affect things. needs more testing.

also need to figure out if i need to link my css generated source maps with my javascript ones.. im missing something need to read up more about sourcemaps heh.

TODO:

- [x] figure out firefox (emotion doesn't work with it either, maybe it's a bug/missing feature in FF https://twitter.com/itsmadou/status/1251748381090705409?s=20)
- [x] figure out safari (edit: seems like it won't work, potentially: https://github.com/angular/angular-cli/issues/16256#issuecomment-566173626)
- [x] figure out how to get it working with hot module reloading (in chrome works first time. after updating and hot module reloads it doesn't flush the source map and goes to old source) refresh fixes it
- [x] put source maps behind dev flag
- [x] add option to turn sourcemaps off

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -1,5 +1,4 @@
 <meta
   http-equiv="Content-Security-Policy"
-  content=" object-src 'none'; style-src 'nonce-k0Mp1lEd'
-'unsafe-inline' 'unsafe-eval' 'strict-dynamic' https: http:; base-uri 'none';"
+  content=" object-src 'none'; style-src 'nonce-k0Mp1lEd' 'unsafe-eval' https: http:; base-uri 'none';"
 />
--- a/examples/class-names-dynamic-object.tsx
+++ b/examples/class-names-dynamic-object.tsx
@@ -2,7 +2,7 @@ import React, { useState } from 'react';
 import { ClassNames } from '@compiled/css-in-js';
 
 export default {
-  title: 'class names dynamics object',
+  title: 'class names dynamic object',
 };
 
 export const ObjectLiteral = () => {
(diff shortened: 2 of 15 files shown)


madou:
so this is getting close. like 

we have the file name correctly displayed in the style inspector:
image

and then the actual source map is there
image

but then how tf do we set the relationship? when i click "class-names-dynamic-object.tsx" in the style inspector it doesn't navigate me where i expect. do we need something to let a bundler pick it up? i don't know.

madou:
interestingly chrome works great

image

what's firefox missing?

madou:
chrome-smp

^ chrome working great. doesn't work on hot-module reload, but its good enough for now.

PR #156, opened 2020-04-19
feat: minimal css rendering on the server

### LINKED ISSUE #154: Render minimal CSS when server side rendering
Right now every component will render its own CSS. That's going to bloat. We should only render a style tag once - for every subsequent use it shouldn't render a style element.

link from mateusz: https://github.com/emotion-js/emotion/blob/10514a8635dcaa55b85c7bff90e2a9e14d1ba61f/packages/core/src/context.js
twitter thread: https://twitter.com/itsmadou/status/1251780172350545921?s=20

they essentially add a context provider/consumer to every emotion component which then notifies children if they have already rendered or not.

this is good because it remains "0 config".
this is not so good because every emotion component now has a context provider + consumer.
they also use a top level cache provider too get around sibling elements rendering extra markup. - but generally isn't that likely to happen if you're rendering an entire app.

is there an alternative path that:

- remains 0 config
- doesn't need every compiled component wrapped

something to think about.

### LINKED ISSUE #158: Clean up sheet
- Remove firefox hack
```
  // this weirdness brought to you by firefox
  for (let i = 0; i < document.styleSheets.length; i++) {
    if (document.styleSheets[i].ownerNode === styleElement) {
      return document.styleSheets[i] as CSSStyleSheet;
    }
  }
```
- Remove the try/catch
- Enforce speedy in production (see: https://emotion.sh/docs/ssr#puppeteer)
- Move this line into a server bundle if statement https://github.com/atlassian-labs/compiled-css-in-js/blob/10d15bd147c0057aa102ae22ad4acfcf81a804a6/packages/style/src/style.tsx#L33

madou:
Closes #154 Closes #158 

Uses context on the server and a singleton on the client.

TODO:

- [x] bring bundle size back to the allowed limit
- [x] See if we can get the bundle back down to `376 B` which was the previous value. Current size is `410 B`. Got it down to `323 B`!


nhardy:
Love what I'm seeing come out of this library so far.

Regarding the PR, as `inserted` is a singleton, I don't think this would work with any async rendering like `React.renderToNodeStream` since this cache would be shared between different requests.

`react-helmet` suffers a similar issue, which `react-helmet-async` solves by using a context Provider (which you've already discussed on Twitter).

madou:
hey Nathan, thanks! 

Yeah I've written some things down about it here https://github.com/atlassian-labs/compiled-css-in-js/issues/154

Yeah it looks like the singleton is a recipe for disaster on the server. Im not sure what the logistics are if we wrap everything with a context provider / consumer to keep things 0 config.

Need to think about it.

packages/style/src/provider/provider-browser.tsx
```diff
@@ -0,0 +1,37 @@
+import { ProviderComponent, UseCacheHook } from './types';
+
+if (typeof window === 'undefined') {
+  throw new Error(
+    `
+ ██████╗ ██████╗ ███╗   ███╗██████╗ ██╗██╗     ███████╗██████╗
```
theKashey, line 6: you definitely have to extract this as a constant.

madou: ehehee

yeah i'll make sure this won't affect bundle size outside of dev. will put a process.env infront of it.

packages/babel-plugin/src/__tests__/index.test.tsx
```diff
@@ -37,11 +37,11 @@ describe('babel plugin', () => {
 
     expect(output?.code).toMatchInlineSnapshot(`
       "import React from \\"react\\";
-      import { Style } from '@compiled/css-in-js';
+      import { CC, Style } from '@compiled/css-in-js';
```
madou, line 40: every component now wrapped with "CC" component - this will have context on server and noop on client (immediately return children)

PR #168, opened 2020-04-22
feat: add support to assert styles on states (hover, active) and media queries

### LINKED ISSUE #138: Improve jest matcher to assert on different states
- hover
- focus
- active

etc.

Right now it just does a string lookup - and if the thing you're looking for is there - anywhere! - it'll pass.

Ideally we update it to take on extra stuff. Including media queries!

```
expect(element).toHaveCompiledCss('font-size', '12px', { state: 'hover' });
```

cc @danieldelcore

ankeetmaini:
Fixes #138. Let me know if this approach looks okay.

### DIFF
--- a/packages/jest/src/index.tsx
+++ b/packages/jest/src/index.tsx
@@ -1,9 +1,15 @@
 declare global {
   namespace jest {
-    type MatchConf = Partial<Record<'state' | 'media', string>>;
     interface Matchers<R> {
-      toHaveCompiledCss(properties: { [key: string]: string }, matchFilter?: MatchConf): R;
-      toHaveCompiledCss(property: string, value: string, matchFilter?: MatchConf): R;
+      toHaveCompiledCss(
+        properties: { [key: string]: string },
+        matchFilter?: Partial<Record<'state' | 'media', string>>
+      ): R;
+      toHaveCompiledCss(
+        property: string,
+        value: string,
+        matchFilter?: Partial<Record<'state' | 'media', string>>
+      ): R;
     }
   }
 }
--- a/packages/jest/src/index.tsx
+++ b/packages/jest/src/index.tsx
@@ -1,15 +1,10 @@
+import { MatchFilter } from './types';
+
 declare global {
   namespace jest {
     interface Matchers<R> {
-      toHaveCompiledCss(
-        properties: { [key: string]: string },
-        matchFilter?: Partial<Record<'state' | 'media', string>>
-      ): R;
-      toHaveCompiledCss(
-        property: string,
-        value: string,
-        matchFilter?: Partial<Record<'state' | 'media', string>>
-      ): R;
+      toHaveCompiledCss(properties: { [key: string]: string }, matchFilter?: MatchFilter): R;
+      toHaveCompiledCss(property: string, value: string, matchFilter?: MatchFilter): R;
     }
   }
 }
--- a/packages/jest/src/index.tsx
+++ b/packages/jest/src/index.tsx
@@ -1,8 +1,9 @@
 declare global {
+  type MatchFilter = Partial<Record<'state' | 'media', string>>;
   namespace jest {
     interface Matchers<R> {
-      toHaveCompiledCss(properties: { [key: string]: string }): R;
-      toHaveCompiledCss(property: string, value: string): R;
+      toHaveCompiledCss(properties: { [key: string]: string }, matchFilter?: MatchFilter): R;
+      toHaveCompiledCss(property: string, value: string, matchFilter?: MatchFilter): R;
     }
   }
 }
--- a/packages/jest/src/index.tsx
+++ b/packages/jest/src/index.tsx
@@ -1,9 +1,9 @@
 declare global {
-  type MatchFilter = Partial<Record<'state' | 'media', string>>;
   namespace jest {
+    type MatchConf = Partial<Record<'state' | 'media', string>>;
     interface Matchers<R> {
-      toHaveCompiledCss(properties: { [key: string]: string }, matchFilter?: MatchFilter): R;
-      toHaveCompiledCss(property: string, value: string, matchFilter?: MatchFilter): R;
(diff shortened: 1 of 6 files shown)


madou:
hey mate thanks for throwing this PR up! I'll have a look

what's your thoughts on the API? I just threw something into the issue but I didn't really think about it.

packages/jest/src/matchers.tsx
```diff
@@ -15,15 +15,16 @@ const getMountedProperties = () =>
     )
     .join(' ');
 
-const containsClassNames = (classNames: string[], css: string) =>
-  classNames.reduce((accum, className) => (css.includes(`.${className}`) ? true : accum), false);
-
+// sorry but using ? was throwing TS off
```
madou, line 18: the ? after MatchFilter?

packages/jest/src/matchers.tsx
```diff
@@ -56,9 +57,14 @@ export function toHaveCompiledCss(
...
     }
 
-    if (containsClassNames(classNames, css)) {
-      foundStyles.push(...stylesToFind.filter(styleToFind => css.includes(styleToFind)));
-    }
+    classNames.forEach(c => {
+      let matcher = c;
+      if (state) matcher += `:${state}`;
+      if (media) matcher = `${media}{.*${matcher}`;
+      if (css.match(matcher)) {
```
madou, line 64: this means if there is no matcher defined it will just mark all styles it finds as assertable yeah? 

what's your thoughts on this? i can see two sides to it 

- it might give false positives so people think something is working when it isnt
- we treat matchers as narrowing and its ok that the default case asserts on every state

packages/jest/src/matchers.tsx
ankeetmaini: yes

packages/jest/src/matchers.tsx
ankeetmaini: Thanks for reviewing. I think a better approach would be to not do a text search and to convert this into an AST and then work with rules. This is done both in emotion-jest and styled-component jest matchers. 

Also, at least in `jest-styled-components` matchers are mandatory and it doesn't work without narrowing, which I like too! 

Let me know your thoughts and I can give it a spin.

madou: i like it - let's give it a spin

ankeetmaini:
@madou this is ready for review now!

packages/jest/src/index.tsx
```diff
@@ -1,8 +1,15 @@
...
-      toHaveCompiledCss(properties: { [key: string]: string }): R;
-      toHaveCompiledCss(property: string, value: string): R;
+      toHaveCompiledCss(
+        properties: { [key: string]: string },
+        matchFilter?: Partial<Record<'state' | 'media', string>>
+      ): R;
+      toHaveCompiledCss(
+        property: string,
+        value: string,
+        matchFilter?: Partial<Record<'state' | 'media', string>>
```
madou, line 11: can bring in `Pseudos` to make it a bit easier and then fallback to `string`?

```
import * as CSS from 'csstype';

type State = Pseudos;
```

should we rename `state` to something that encompasses other groups pseudo selectors as well? what's your thoughts

packages/jest/src/matchers.tsx
```diff
@@ -1,3 +1,10 @@
+import CSS from 'css';
+
+type MatchFilter = Partial<Record<'state' | 'media', string>>;
```
madou, line 3: this is duplicated in `src/index.tsx` maybe we can make a `types.tsx` file?

madou:
looks great man! really impressed. left a few comments 👍

packages/jest/src/matchers.tsx
```diff
@@ -15,15 +22,66 @@ const getMountedProperties = () =>
...
+          acc = [...acc, ...m.rules];
+        }
+        return acc;
+      }, []);
+    }
+    return ast.stylesheet?.rules.filter(r => (r.type = 'rule')); // omit media objects
+  };
+
+  const allRules = getAllRules();
+  const klass = state ? `.${className}:${state}` : `.${className}`;
```
madou, line 51: will we be able to assert or more complex selectors?

```
css={`
  > :first-child {
    color: red;
  }
`}

expect(element).toHaveCompiledCss('color', 'red', { state: '> :first-child' });
```

Does that work? If not should it?

packages/jest/src/index.tsx
ankeetmaini: Wow, state to 'selector' maybe? I'll try this and update

packages/jest/src/matchers.tsx
ankeetmaini: Wow, let me try this as well

packages/jest/src/index.tsx
ankeetmaini: @madou if we just use the value as Pseudos, TS will automatically suggest and add `:`, but then `> :first-child` this won't be a valid one.

Is there a way I can preserve the intellisense as well as adding string?

`Pseudos | string` is just `string`

madou: let me check it out locally and look

madou: found an issue https://github.com/microsoft/TypeScript/issues/29729

madou:
```
import { Pseudos } from 'csstype';

/**
 * {} is used to enabled the compiler to not reduce Target down to a string.
 * This keeps both Pseduos intellisense working as well as then allowing us to define anything.
 * See: https://github.com/microsoft/TypeScript/issues/29729#issuecomment-460346421
 */
type AnyTarget = string & { _?: never };
type Target = Pseudos | AnyTarget;

export interface MatchFilter {
  target?: Target;
  media?: string;
}
```

this fixes it

madou:
you've done a really great job!! ❤️

ankeetmaini:
Yay, thanks so much @madou :D

madou:
no worries! thank **you** for doing the work 🤗

PR #175, opened 2020-04-26
Replace stylis with postcss

### LINKED ISSUE #100: Move from stylis to a pre-processor with more appropriate compile time features
We don't need the library to small since it's done at comiled time.

## Stylus

Huge benefits, namely built in functions for free.

https://stylus-lang.com/docs/bifs.html

## PostCSS

Kind of a standard. Loads of plugins.

madou:
Replaces stylis with postcss.

It comes with a more powerful minifier which is great - however we need to do more.

TODO

- [x] Create style rules plugin to extract each css block as an element in an array
- [x] Actually wire everything up
- [x] Figure out how to get equivalent nesting from stylis (might have to roll our own plugin?)
- [x] Ensure correct implementation of orphaned leading pseudos

Closes #100


packages/ts-transform/src/utils/css-transform.tsx
```diff
@@ -0,0 +1,75 @@
...
+    .map(([creator]: any) => {
+      // replicate the `initializePlugin` behavior from https://github.com/cssnano/cssnano/blob/a566cc5/packages/cssnano/src/index.js#L8
+      return creator();
+    })
+    .filter((plugin: any) => {
+      return !asyncPluginsToExclude.includes(plugin.postcssPlugin);
+    });
+};
+
+const topLevelPseudos = plugin('top-level-pseudos', () => {
```
Andarist, line 21: I dont think this covers all cases yet:
- `.cls{span{:hover{}}}`
- `.cls{@media{:hover{}}}`
- `.cls{@supports{:hover{}}}`

madou: hey thanks for jumping in

does stylis (now your emotion plugin) transform these to

```
span.cls:hover {}
@media { .cls:hover { } }
@supports { .cls:hover { } }
```

i'll write some tests 😄

madou: gottem a539c05

Andarist: https://github.com/emotion-js/emotion/blob/5e39d81bf9177e68e78d23bd869c1592341d9768/packages/css/test/selectivity.test.js#L122

> span.cls:hover {} // <-- this one doesn't seem right to me tbh unless it was `&span` from the consumer

Should be:
```css
.cls span:hover {}
```

The "rule" how it works is actually pretty simple - orphaned leading pseudo always acts as `&:pseudo`. So the difference with and without this plugin is a single whitespace (or lack of it) between "parent" selector part (`&`) and the current selector containing pseudo as the first item.

madou: ah interesting - does this behaviour add it for any depth of nested orphaned leading pseudos?

Andarist: yes, the behavior is the same for all depths

madou: pushed up a fix. does it look good to you? 😄

ankeetmaini:
🎉

packages/ts-transform/src/utils/css-transform.tsx
Andarist: some of my test cases are kinda dumb/weird now 😅 i've added them because i was falling into some edge cases with my prior approach (before I've implemented `.parent` on stylis nodes which just made the final plugin **way** easier), wasn't feeling like removing them though

packages/ts-transform/src/utils/css-transform.tsx
```diff
@@ -0,0 +1,89 @@
...
+      if (rule.selector.includes(':')) {
+        const newSelector = rule.selector
+          .split(',')
+          .map(part => {
+            if (part.match(/^. /)) {
+              // If the selector has one characters with a space after it, e.g. "> :first-child" then return early.
+              return part;
+            }
+
+            if (part.includes(':')) {
```
Andarist, line 34: wouldnt it be easier and more performant to just check `part.charCodeAt(0) === 58` (58 is a char code for `:`)?

Andarist: you probably wouldnt need the previous if statements at all using this strategy

packages/ts-transform/src/utils/css-transform.tsx
```diff
@@ -0,0 +1,89 @@
...
+      return !asyncPluginsToExclude.includes(plugin.postcssPlugin);
+    });
+};
+
+const parentOrphenedPseudos = plugin('parent-orphened-pseudos', () => {
+  return root => {
+    root.walkRules(rule => {
+      if (rule.selector.includes(':')) {
+        const newSelector = rule.selector
+          .split(',')
```
Andarist, line 27: this will fail in edge cases like `[data-foo=","]`

packages/ts-transform/src/utils/css-transform.tsx
```diff
@@ -0,0 +1,89 @@
...
+          .map(part => {
+            if (part.match(/^. /)) {
+              // If the selector has one characters with a space after it, e.g. "> :first-child" then return early.
+              return part;
+            }
+
+            if (part.includes(':')) {
+              // If the selector starts with a colon or has a space and then a colon prepend an "&"!
+              return part.replace(/^:| :/g, '&:');
```
Andarist, line 36: do you know for what input you can receive a colon preceded with whitespace?

packages/ts-transform/src/utils/css-transform.tsx
madou: thanks for the catch

packages/ts-transform/src/utils/css-transform.tsx
madou: it was for one of the tests i saw - i'd have to remove this and see which test fails 😅

packages/ts-transform/src/utils/css-transform.tsx
madou: cheers :)

hows it go vs. checking `part[0] === ':'`?

packages/ts-transform/src/utils/css-transform.tsx
madou: wheres your plugin located? maybe i should just look at it 😺

Andarist: https://github.com/emotion-js/emotion/blob/5e39d81bf9177e68e78d23bd869c1592341d9768/packages/cache/src/stylis-plugins.js#L49

packages/ts-transform/src/utils/css-transform.tsx
Andarist: this is, of course, a hypothetical edge case - probably nobody would actually bump into this in their code 😅 but my general rule of thumb is to try to avoid this kind of things so I just don't have to deal with any potential issues. For dealing with contextualized strings things like `.split` or using regexps are always red flags for me to notice such possible edge cases

https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/1732454#1732454

packages/ts-transform/src/utils/css-transform.tsx
Andarist: could you check this out? would love to see when this happens, probably it's related to how postcss work and what output does it yield, but knowing the test case I could recheck if this is working as intended

packages/ts-transform/src/utils/css-transform.tsx
Andarist: it's equivalent - but `.charCodeAt` according to benchmarks is faster, why? no idea 😅maybe it's just because comparing numbers is more optimized than comparing strings? 🤷‍♂️

packages/ts-transform/src/utils/css-transform.tsx
madou: turns out i just had to parse away the unneeded whitespace and then its not a problem.

Andarist: Yes, rule elements have their computed selectors available in `.props` property. This holds already computed (expanded) selectors, so for this input:
```sass
a,b,c{d,e{color:hotpink;}}
```
it would have such props:
```js
['a d','b d','c d','a e','b e','c e']
```
if you want to get "selectors" only for the "current level" then you need to reparse `.value` which for the inner node here would hold `'d,e'` string.

madou: man that would be sweet - I don't think the current version of Post CSS does that....

PR #180, opened 2020-04-29
chore: configure global types for jest

### LINKED ISSUE #176: Figure out how to get jest matcher types working across the repo without needing to import them in every test
Both `jest-extended` and `@compiled/jest-css-in-js` are being imported everywhere. Annoying!

I tried creating a `types` folder and importing them that way but it didn't seem to work. Unsure what to do.

jdanil:
Using configuration suggested by `jest-extended`.
https://github.com/jest-community/jest-extended#typescript

Fixes #176.

### DIFF
--- a/packages/babel-plugin/src/__tests__/index.test.tsx
+++ b/packages/babel-plugin/src/__tests__/index.test.tsx
@@ -1,5 +1,4 @@
 import { transformSync, TransformOptions } from '@babel/core';
-import 'jest-extended';
 import compiledPlugin from '../index';
 
 const babelOpts: TransformOptions = {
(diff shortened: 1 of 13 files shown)


madou:
thanks mate will have a look 😄 

weird that the bundle size has increased. need to investigate.

jdanil:
I think it must be. I tried running it with just the `cross-env` change and bundlesize failed again with the same error.
I can't run bundlesize without the `cross-env` change so maybe I'll just try to back that part out (and just keep the jest changes) and see how it runs on CI.

madou:
i think we can exclude them from the bundle size. let me check your stuff out locally and test.

its weird its even brought in at all, though

jdanil:
Yeah it is. I couldn't find any mention of cross-env in the new bundles though so not really sure what was going on. Might look into it again if I get time to add to the testing docs.

madou:
@jdanil hey mate so your change works when running jest - but when using in the IDE we get this error

image

i think this is what happened last time i tried to fix this. any ideas?

PR #181, opened 2020-04-29
chore-fix: broken link in docs

gwyneplaine:
Link to consuming transformers section of ts-transform handbook was borked.


madou:
nice catch - thanks!

Issue #182, opened 2020-04-30
RFC: Strict mode


madou:
Relates to #18 #165.

- This mode would enforce that your CSS can't have any magic values.
- As well as restrict potential unsafe patterns that would blow out CSS bundle size when atomicifying.

- [ ] TODO fill in more info.

madou:
Closing this for now to clean up the issue view. This might come back with supplementary lint rules instead in the `@compiled/eslint-plugin` package.

Issue #18, opened 2019-12-19
RFC: Theming


madou:
Fresh eyes needed. We should also take inspiration from https://theme-ui.com/

We can even create a TS language service to tell us what themes are available with intellisense!

## Global theming

Define your tokens in your options:

```js
{
  base: {},
  default: {},
  [optionalotherthemes]: {} // every other theme needs to have the same keys from default
}
```

Then use them in your CSS sourced from default:

```js
css`
  color: primary; // or
  color: theme(primary);
`
```

Transforms to:

```css
color: #fff;
color: var(--var-123abc);
```

Hardcodes and use the variable so it works with/without a parent theme.

Prepare a theme provider for consumers (you should only ever create one of these):

```js
import { createThemeProvider } from '@compiled/css-in-js';

export const ThemeProvider = createThemeProvider();

// Optional "runtime" themes.
export const ThemeProvider = createThemeProvider({ ... });
```

Transforms to:

```js
import { _TP } from '@compiled/css-in-js';

const theme = { base: { ... }, default: { '--var-abc123': '#000', ... }, ... };
// optional runtime themes would be merged in and converted to css variables

export const ThemeProvider = (props) => (
  <_TP theme={props.theme}>
   {props.children(theme[props.mode])
  <_TP>
);
```

And then your consumers use like:

```js
import { ThemeProvider } from 'your-design-system';

const App = () => (
  <ThemeProvider theme="dark">
     {style => <div style={style}>...</div>
  </ThemeProvider>
);
```

## Component theming

Would be the same as above with one key difference - the theme isn't sourced from the config. It's sourced inline.

```js
import { createVariants } from '@compiled/css-in-js';

const useVariants = createVariants<'primary' | 'danger'>({
  default: { default: {}, [optionalotherthemes]: {} },
  [optionalotherthemes]: {},
});

<button style={useVariants('primary')} css={{ border: 'border', color: 'color' }}>
  blah
</button>
```

```js
import { useMode } from '@compiled/style;

// transformed to css variables
const variants = { default: {}, ... };

const useVariants = (variant: string) => {
  const mode = useMode();
  const defaultVariant = variants.default;

  return { 
    ...defaultVariant.default,
    ...defaultVariant[mode],
    ...variants[variant][mode],
  };
};
```

The type safety aspect is missing a little. Perhaps instead of using string literals `theme(primary)` and `variant(color)` we could import them?

```js
import { themeRefs, createVariants } from '@compiled/css-in-js'; 

[...]

PR #184, opened 2020-04-30
tokens 🤝 global theming

madou:
Investigating #18 #182

- introduces `createThemeProvider()` api
- introduces `useMode()` hook

## TODO

- [x] Enable nesting support in tokens json
- [ ] Fix theme inlining not working
- [ ] Fix setting of tokens being broken (see example storybook)
- [ ] Add limited support for runtime token overrides
- [ ] Rename from theme provider to token provider(?)

## Questions

- Does the tokens JSON look ok? The point is - define your `base` and then use those references in the modes.
- Do we think this is a problem if tokens are primarily built around `base` tokens and then themes - with `default` a mandatory theme?
- How should we handle runtime tokens? Tokens that aren't known until runtime, use case could be overriding tokens, for example Confluence changing `primary` to another color.
I was imagining this could be something like:

```
// Don't have to pass all properties in
<ThemeProvider overrides={{ default: { primary: 'red' } }}>
```

And if you want to add an entire new theme:

```
// Have to pass all properties in
<ThemeProvider overrides={{ batman: { primary: 'red' } }}>
```

- How could we enforce "must use base tokens" even during runtime? Currently they get compiled away. Do we add it back?

## Usage

1. Create a tokens json

```json
{
  "base": {
    "r300": "#FF5630",
    "b400": "#0052CC"
  },
  "default": {
    "primary": "b400"
  },
  "dark": {
    "primary": "r300"
  }
}
```

- define your base tokens
- use these tokens in `default` and others
- mode definitions can only tokens from `base` - anything else results in an error
- other modes mandate they have the same shape as `default`
- nested tokens also possible - see next:

<details><summary>Nested tokens</summary>
We can support nested tokens we just need to do some extra work because the transformed appearance needs to be flat.

```js
{
  "base": {
    "r300": "#FF5630",
    "b400": "#0052CC"
  },
  "default": {
    "borderRadius": "3px",
    "colors": {
      "primary": "b400"
     }
  },
  "dark": {
    // Don't need to define everything here purely an additive thing.
    "colors": {
      "primary": "r300"
     }
  }
}
```

Transforms top level tokens that would be CSS variables to:

```js
const tokens = {
  default: { '--borderRadiusHash': '3px', '--colorsPrimaryHash': '#0052CC' },
  dark: {  '--colorsPrimaryHash': '#FF5630' },
};
```

And then transform the token access to:

```js
const theme = {
  borderRadius: 'var(--borderRadiusHash)',
  colors: { primary: 'var(--colorsPrimaryHash)' }
};

theme.borderRadius;
theme.colors.primary;
````
</details>

2. Turn on tokens

```json
"options": {
  "tokens": "./tokens.json"
}
```

3. Create and use your provider

```jsx
const { ThemeProvider, theme } = createThemeProvider();

export ThemeProvider;
export theme;
```

- `ThemeProvider` ends up being a component with react context
- `theme` ends up being an object to access tokens.

```jsx
import { ThemeProvider } from './theme';

const App = ({ children }) => (
  <ThemeProvider theme="default">
    {style => <div style={style}>{children}</div>
  </ThemeProvider>
);
```

4. Use tokens in a component

```jsx
import { styled } from '@compiled/css-in-js';
import { theme } from './theme';

export const StyledDiv = styled.div`
  color: ${theme.primary};
`;
```

Ends up being inlined and looking like this CSS, uses default css variable so it works without the theme provider as well.

```css
color: var(--cc-1tivpv1,#0052CC);
```

If we want to provide basic support for IE11 we can also do this (but let's not):

```css
color: #0052CC;
color: var(--cc-1tivpv1,#0052CC);
```


packages/ts-transform/src/__tests__/tokens.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+              b400: '#0052CC',
+            },
+            default: {
+              primary: 'b400',
+            },
+          },
+        }
+      )
+    ).toThrowError(
+      new Error(`You've defined hard-coded colors which is not allowed in strict mode.
```
gwyneplaine, line 113: this is sick 🎉

madou:
Here's what I'm thinking about for `css` prop variants https://codesandbox.io/s/compiled-variants-typing-cfts2?file=/src/App.tsx

i'll come back with something for styled api as well

madou:
@gwyneplaine yo chime in sometime about the theming api :) 

i think global api is probably a little easier so i might remove all the variant stuff for now - but would be keen for a look over that too

i quite like the css prop api because it can be typed + transformed easily and its pretty easy to use api - everything else is 🤷

examples/theming.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export default {
+  title: 'theming',
+};
+
+const ThemeProvider = createThemeProvider();
+
+const Thing = styled.div({
+  fontSize: '20px',
+  color: 'theme(primary)',
```
madou, line 12: @gwyneplaine one thing im on the fence about is the global theming api isn't typed right now.

any thoughts how we could make it? since the tokens are sourced from a json file we don't have... an easy way to do it i think..

packages/ts-transform/src/__tests__/tokens.test.tsx
madou: the whole concept of strict has been removed for now

PR #178, opened 2020-04-28
feat: inline const variables in CSS

ankeetmaini:
Fix #167

PR #185, opened 2020-04-30
fix: destructured variable used inside string template

### LINKED ISSUE #186: Throws when destructured variable is used in CSS template string
Added a failing test case in #185 

Will submit a PR.

ankeetmaini:
`collectDeclarations` returns undefined for the const variable which throws in the generation of CSS.

Add a failing test. Blocks #178 and fixes #186


madou:
gj :)

Issue #103, opened 2020-04-04
Change Style import to import directly from @compiled/style


madou:
It means for people who can't rely on treeshaking they still get the smallest bundle size. We can also potentially be in trouble for name clashes, so throw an underscore in there for good measure.

Before

```
import { Style } from '@compiled/css-in-js';
```

After

```
import _Styl from '@compiled/style';
```

flexdinesh:
@madou this would mean 

```
import React from 'react';
import { Style } from '@compiled/css-in-js';
export const ColoredText = React.forwardRef(
  ({ as: C = 'span', ...props }, ref) => (
    <>
      <Style hash="css-zd46j1">{['.css-zd46j1{color:#ff5630;}']}</Style>
      <C
        {...props}
        ref={ref}
        className={
          'css-zd46j1' + (props.className ? ' ' + props.className : '')
        }
      />
    </>
  )
);
```

will now become 

```
import React from 'react';
import _Styl from '@compiled/style';
export const ColoredText = React.forwardRef(
  ({ as: C = 'span', ...props }, ref) => (
    <>
      <_Styl hash="css-zd46j1">{['.css-zd46j1{color:#ff5630;}']}</_Styl>
      <C
        {...props}
        ref={ref}
        className={
          'css-zd46j1' + (props.className ? ' ' + props.className : '')
        }
      />
    </>
  )
);
```

Instead of opting for underscrore, what do you think about making it as a plugin option, so consumers can choose a different name if and when there's a name collision?

madou:
yeah exactly dinesh. i think we could introduce a option but it should just work out of the box IMO with something very unlikely to colide.

gwyneplaine:
@madou  keen to pick this up to familiarize myself with the repo.  I'm assuming this applies to the CC export as well? 

*Before*
```type=js
  import React from \\"react\\";
  import { CC, Style } from '@compiled/css-in-js';
  const StyledDiv = /*#__PURE__*/React.forwardRef(({
      as: C = \\"div\\",
      ...props
    }, ref) => <CC><Style hash=\\"1x3e11p\\">{[\\".cc-1x3e11p{font-size:12px}\\"]}</Style><C {...props} ref={ref} className={\\"cc-1x3e11p\\" + (props.className ? \\" \\" + props.className : \\"\\")} /></CC>);
```

*After*
```type=js
  import React from \\"react\\";
  import _Style, { CC } from '@compiled/style';
[...]

PR #190, opened 2020-05-03
replace css in js imports with style imports

gwyneplaine:
PR for #103 
Remaining issues with this PR (failing tests): 
- *css-in-js_1 is undefined*: Failing tests in css-in-js package, running that test-suite gives us a few errors about `css-in-js_1` being undefined. Looks like the common-js bundle transpiles the exports from @compiled/style down to `style_1`, I've made changes in ts-transform constants to account for this, but for reasons I do not know the css-in-js test suite still fails (do i need to run a fresh build?)

- *dangling semi-colons for removed nodes* , in this implementation we're inserting a sibling node via `visit-source-file-ensure-style-import`. on subsequent invocations of this fn, we remove the `compiled-css-in-js` import if there are no more namedImports. Oddly removals after an insertion of a sibling node via `[newNode, node]` returns a dangling semi-colon while doing so without insertions does not. 

@madou  if you have insight into any of the above, let me know. 

**Aside** We should probably separate `visit-source-file-ensure-style-import` into two separate operations now: 
* The first checks if an import from `@compiled/style` exists in the sourceFile, if it doesn't it inserts `import { CC, CS } from @compiled/style` if it does, then it does nothing. 
* The second is basically the existing logic, filtering out namedImports marked for removal from "@compiled/css-in-js", until no namedImports exist (at which point the node is removed)

^ would have done this in this PR, but a) wanted an MVP of this and b) to  check that there wasn't anything blocking us from doing this.

### DIFF
--- a/packages/ts-transform/src/utils/visit-source-file-ensure-style-import.tsx
+++ b/packages/ts-transform/src/utils/visit-source-file-ensure-style-import.tsx
@@ -40,7 +40,6 @@ export const visitSourceFileEnsureStyleImport = (
 
       // if it exists already, then just remove the removedImports from the importstatement.
 
-      const defaultImport = node.importClause && node.importClause.name;
       let namedImports: ts.ImportSpecifier[] = [];
 
       // remove namedImports that have been flagged for removal.
@@ -63,7 +62,7 @@ export const visitSourceFileEnsureStyleImport = (
         node,
         /* decorators */ undefined,
         /* modifiers */ undefined,
-        ts.createImportClause(defaultImport, ts.createNamedImports(namedImports)),
+        ts.createImportClause(undefined, ts.createNamedImports(namedImports)),
         node.moduleSpecifier
       );
       // if we hit an import we should check if any sibling nodes are
--- a/packages/ts-transform/src/utils/visit-source-file-ensure-style-import.tsx
+++ b/packages/ts-transform/src/utils/visit-source-file-ensure-style-import.tsx
@@ -34,11 +34,6 @@ export const visitSourceFileEnsureStyleImport = (
       node.moduleSpecifier.text === COMPILED_PKG
     ) {
       log.log('ensuring style export is defined');
-      // if a style import doesn't exist, add it as a sibling to this node
-      // idealy this should occur in a separate transform, possibly as an independent transform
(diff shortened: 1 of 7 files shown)


madou:
i'll look at this over the weekend :)

madou:
ok so checking things out

1. we can return `undefined` to remove a node
2. `packages/css-in-js/src/__tests__/browser.test.tsx` is passing `packages/css-in-js/src/__tests__/ssr.test.tsx` is failing - digging into it its because we are importing `import { styled, CC } from '@compiled/css-in-js';` at the top and then the import is being removed.

replacing

```diff
-import { styled, CC } from '@compiled/css-in-js';
+import { styled } from '@compiled/css-in-js';
+import { CC } from '@compiled/style';
```

makes it work.

perhaps lets remove all the style exports inside `css-in-js`? then we don't have to worry about it anymore.

gwyneplaine:
@madou  tests fixed, I'm gonna leave the removal of CC CS and CT from css-in-js for separate PR

madou:
look at this guy, getting shit done 😎 

will look tomorrow

packages/ts-transform/src/utils/visit-source-file-ensure-style-import.tsx
```diff
@@ -3,61 +3,92 @@ import * as log from './log';
...
       node.moduleSpecifier.text === COMPILED_PKG
     ) {
       log.log('ensuring style export is defined');
+      // if a style import doesn't exist, add it as a sibling to this node
+      // idealy this should occur in a separate transform, possibly as an independent transform
+      // before any of the three processes start.
+
+      // if it exists already, then just remove the removedImports from the importstatement.
 
       const defaultImport = node.importClause && node.importClause.name;
```
madou, line 43: nit: I think we can get rid of this now there isn't a default import on css-in-js

madou:
took me a little bit but i see what you're doing - looks good! we definitely need to clean up the code a bit soon but since it's tested it should be okay.

madou:
@gwyneplaine let's finish this!

gwyneplaine:
@madou soz for delay, fixed up the remaining nit, will have a good look at that theming PR tomorrow 👍

PR #210, opened 2020-05-30
chore(css-transform): replace postcss selector parsing with official library

### LINKED ISSUE #200: Replace post css selector parsing with the official library
Right now we're parsing it with a `string.split` - we should move over to the official parsing library instead: https://github.com/postcss/postcss-selector-parser

Code to change: https://github.com/atlassian-labs/compiled-css-in-js/blob/master/packages/ts-transform/src/utils/css-transform.tsx#L30

Now what we have works - so there's no real pressing need to change it. But this is a good first issue I think 😄.

---

That said looking at some of the issues on there are interesting (like: https://github.com/postcss/postcss-selector-parser/issues/204) - but we probably have the same issue ATM tbh.

pgmanutd:
This fixes #200

### DIFF
--- a/packages/ts-transform/src/utils/css-transform.tsx
+++ b/packages/ts-transform/src/utils/css-transform.tsx
@@ -3,7 +3,7 @@ import autoprefixer from 'autoprefixer';
 import cssnano from 'cssnano-preset-default';
 import nested from 'postcss-nested';
 import whitespace from 'postcss-normalize-whitespace';
-import parser from 'postcss-selector-parser';
+import selectorParser from 'postcss-selector-parser';
 
 const minify = () => {
   const preset = cssnano();
@@ -20,42 +20,42 @@ const minify = () => {
     });
 };
 
-const removeSpacesBeforeSelector = (selector: parser.Node) => (selector.spaces.before = '');
-const doesCombinatorTypePrecedesSelector = (selector: parser.Node) => {
+const isPreviousSelectorCombinatorType = (selector: selectorParser.Node) => {
   const previousSelector = selector.prev();
 
   return previousSelector && previousSelector.type === 'combinator';
 };
-const prependNestingTypeToSelector = (selector: parser.Node) => {
+const prependNestingTypeToSelector = (selector: selectorParser.Node) => {
   const { parent } = selector;
 
   if (parent) {
-    const nesting = parser.nesting({});
+    const nesting = selectorParser.nesting();
 
     parent.insertBefore(selector, nesting);
   }
 };
+const stringifySelectorParserRoot = (parserRoot: selectorParser.Root) =>
+  parserRoot.reduce<string[]>((memo, selector) => [...memo, String(selector)], []).join(',\n');
 const parentOrphenedPseudos = plugin('parent-orphened-pseudos', () => {
   return (root) => {
     root.walkRules((rule) => {
-      const parserRoot = parser((selectors) => {
-        selectors.walk((selector) => {
-          removeSpacesBeforeSelector(selector);
+      const { selector: ruleSelector } = rule;
 
-          if (selector.type === 'pseudo') {
-            if (doesCombinatorTypePrecedesSelector(selector)) {
-              return;
-            }
+      if (!ruleSelector.includes(':')) {
+        return;
+      }
 
-            prependNestingTypeToSelector(selector);
+      const selectorParserRoot = selectorParser((selectors) => {
+        selectors.walkPseudos((selector) => {
+          if (isPreviousSelectorCombinatorType(selector)) {
+            return;
           }
-        });
-      }).astSync(rule.selector);
 
-      // TODO: Remove any typecasting once https://github.com/postcss/postcss-selector-parser/pull/224 gets merged
-      rule.selector = (parserRoot as any)
-        .reduce((memo: parser.Node[], selector: parser.Node) => [...memo, String(selector)], [])
-        .join(',\n');
+          prependNestingTypeToSelector(selector);
+        });
+      }).astSync(ruleSelector, { lossless: false });
+
+      rule.selector = stringifySelectorParserRoot(selectorParserRoot);
     });
   };
 });
--- a/packages/ts-transform/src/utils/css-transform.tsx
+++ b/packages/ts-transform/src/utils/css-transform.tsx
@@ -3,6 +3,7 @@ import autoprefixer from 'autoprefixer';
 import cssnano from 'cssnano-preset-default';
 import nested from 'postcss-nested';
 import whitespace from 'postcss-normalize-whitespace';
+import parser from 'postcss-selector-parser';
 
 const minify = () => {
   const preset = cssnano();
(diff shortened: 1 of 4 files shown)


packages/ts-transform/src/utils/css-transform.tsx
```diff
@@ -19,33 +20,42 @@ const minify = () => {
     });
 };
 
-const COMMA_CHAR_CODE = 58;
+const removeSpacesBeforeSelector = (selector: parser.Node) => (selector.spaces.before = '');
+const doesCombinatorTypePrecedesSelector = (selector: parser.Node) => {
```
madou, line 24:
```suggestion
const isPreviousSelectorCombinatorType = (selector: parser.Node) => {
```

packages/ts-transform/src/utils/css-transform.tsx
```diff
@@ -19,33 +20,42 @@ const minify = () => {
...
+  const previousSelector = selector.prev();
 
+  return previousSelector && previousSelector.type === 'combinator';
+};
+const prependNestingTypeToSelector = (selector: parser.Node) => {
+  const { parent } = selector;
+
+  if (parent) {
+    const nesting = parser.nesting({});
```
madou, line 33:
```suggestion
    const nesting = parser.nesting();
```

I don't think we need the empty object according to api docs https://github.com/postcss/postcss-selector-parser/blob/master/API.md#parsernestingprops

packages/ts-transform/src/utils/css-transform.tsx
```diff
@@ -19,33 +20,42 @@ const minify = () => {
...
+
+    parent.insertBefore(selector, nesting);
+  }
+};
 const parentOrphenedPseudos = plugin('parent-orphened-pseudos', () => {
   return (root) => {
     root.walkRules((rule) => {
-      if (rule.selector.includes(':')) {
```
madou, line 27: should we keep something like this so we don't parse every selector unnecessarily?

packages/ts-transform/src/utils/css-transform.tsx
```diff
@@ -19,33 +20,42 @@ const minify = () => {
...
-          .replace(/\s+/g, ' ')
-          .split(', ')
-          .map((part) => {
-            if (part.match(/^. /)) {
-              // If the selector has one characters with a space after it, e.g. "> :first-child" then return early.
-              return part;
-            }
+      const parserRoot = parser((selectors) => {
+        selectors.walk((selector) => {
+          removeSpacesBeforeSelector(selector);
```
madou, line 43: nit: if we don't need this to make our code work we can probably just remove this? i had it so it made the silly string splitting easier 😄

packages/ts-transform/src/utils/css-transform.tsx
```diff
@@ -19,33 +20,42 @@ const minify = () => {
...
-          .join(',\n');
+            prependNestingTypeToSelector(selector);
+          }
+        });
+      }).astSync(rule.selector);
 
-        rule.selector = newSelector;
-      }
+      // TODO: Remove any typecasting once https://github.com/postcss/postcss-selector-parser/pull/224 gets merged
+      rule.selector = (parserRoot as any)
```
madou, line 56: can we type cast this to something better than `any`? `string[]` or something?

packages/ts-transform/src/utils/css-transform.tsx
```diff
@@ -3,6 +3,7 @@ import autoprefixer from 'autoprefixer';
 import cssnano from 'cssnano-preset-default';
 import nested from 'postcss-nested';
 import whitespace from 'postcss-normalize-whitespace';
+import parser from 'postcss-selector-parser';
```
madou, line 6: nit: can we rename this to `selectorParser`?

```suggestion
import selectorParser from 'postcss-selector-parser';
```

madou:
This is great - good job!

packages/ts-transform/src/utils/css-transform.tsx
pgmanutd: I know, but typings were throwing error. Typings seems to be messed up. Options were marked required 😞

madou: oh 🤔 

can we copy their latest types from that PR and just make our own typedef?

pgmanutd: sure

packages/ts-transform/src/utils/css-transform.tsx
pgmanutd: `if (selector.type === 'pseudo')` I added this pseudo check. It's same right.

packages/ts-transform/src/utils/css-transform.tsx
pgmanutd: hmm..If I remove this, tests got failed because of leading spaces with following error:

Screen Shot 2020-05-31 at 4 46 22 PM

madou: thats ok just update snapshots

packages/ts-transform/src/utils/css-transform.tsx
madou: not really because we've already parsed the selectors by this point 🤔

pgmanutd: hmm..so I was thinking of avoiding string checks..I think you mean that we should only pass those selectors to `postcss-selector-parser` which includes `:`, but in this case I think it will be difficult to assign value back to `rule.selector` because `postcss-selector-parser` will only parse those selectors which includes `:`. Have to check.

Actually, I referred this example: https://github.com/postcss/postcss-selector-parser/blob/master/API.md#synchronous-error-example

pgmanutd: I think I can assign back to `rule.selector` 👍

pgmanutd: Sorry my bad 😞

packages/ts-transform/src/utils/css-transform.tsx
pgmanutd: @madou This looks bad if don't remove spaces before pseudo selectors:

Screen Shot 2020-05-31 at 9 30 49 PM

packages/ts-transform/src/utils/css-transform.tsx
madou: all good!

packages/ts-transform/src/utils/css-transform.tsx
madou: ! no it isn't 🤔 

let's leave it as is for now 😄 

fyi emotions implementation of this is here https://github.com/emotion-js/emotion/blob/5e39d81bf9177e68e78d23bd869c1592341d9768/packages/cache/src/stylis-plugins.js#L49 if we need some inspiration

pgmanutd: There is a `lossless` settings under [ProcessorOptions](https://github.com/postcss/postcss-selector-parser/blob/master/API.md#processoroptions). If we set it to false, things works as expected. No need to manually remove the spaces 🤘 

hmm..Emotion one is bit different it seems..anyways `lossless` solves our problem 😃

packages/ts-transform/src/utils/css-transform.tsx
```diff
@@ -19,33 +20,42 @@ const minify = () => {
...
-        rule.selector = newSelector;
+      const { selector: ruleSelector } = rule;
+
+      if (!ruleSelector.includes(':')) {
+        return;
       }
+
+      const selectorParserRoot = selectorParser((selectors) => {
+        selectors.walkPseudos((selector) => {
+          if (isPreviousSelectorCombinatorType(selector)) {
```
madou, line 50: hmm did we not need to check if the selector is a pseudo anymore? 🤔

pgmanutd: `selectors .walkPseudos`?

madou: OH

cool!!!

PR #217, opened 2020-06-11
Babel next plumbing

madou:
Ok confirmed we're going with a Babel plugin. We need to get this PR into a shape we're happy to merge and then we'll start working on features in styled components API.


packages/babel-plugin/src/next.tsx
```diff
@@ -0,0 +1,60 @@
...
+        visitStyledPath(path, state);
+      },
+      CallExpression(path, state) {
+        if (!state.compiledImportFound) {
+          return;
+        }
+
+        visitStyledPath(path, state);
+      },
+      JSXOpeningElement(path, state) {
```
pgmanutd, line 51: @madou: Does spread operator works here 🤔? Something like this:

```
import '@compiled/css-in-js';

const props = {
  css: 'font-size:12px;'
}

const MyDiv = () => {
  return <div {...props}>hello</div>
};
```

pgmanutd: Didn't find `JSXSpreadAttribute` 🤔.

madou: hi - yeah currently that isn't supported. i don't see that as a very common use case. do you think it's useful to add it in?

my thoughts: do the bare minimum and then add more if it has wide spread usage.

needing a static `css` prop directly on the jsx element seems like the right path for me.

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -0,0 +1,164 @@
...
+    name: string;
+    expression: t.Expression;
+  }[];
+}
+
+const getInterpolation = <TNode extends {}>(expression: TNode | undefined, state: State) => {
+  if (t.isIdentifier(expression) && state.declarations) {
+    const declaration = state.declarations[expression.name];
+    if (t.isVariableDeclaration(declaration) && declaration.kind === 'const') {
+      // We only want to pick out variable declarations that are constants
```
lol-russo, line 25: might be an obvious answer - but why?

packages/babel-plugin/src/next.tsx
pgmanutd: Nah!! not that much important for `css` prop tbh. But yups for other props we should definitely consider it.

> my thoughts: do the bare minimum and then add more if it has wide spread usage.

^^ I am sold 😛

packages/babel-plugin/src/next.tsx
```diff
@@ -0,0 +1,60 @@
...
+  return {
+    inherits: require('babel-plugin-syntax-jsx'),
+    visitor: {
+      ImportDeclaration(path, state) {
+        if (path.node.source.value === '@compiled/css-in-js') {
+          state.compiledImportFound = true;
+          path.node.specifiers = path.node.specifiers
+            .filter(
+              (specifier) =>
+                specifier.local.name !== 'styled' && specifier.local.name !== 'ClassNames'
```
lol-russo, line 20: Okay so no matter what import you use (i.e. what flavour of CSS-in-JS you like) we are going to filter out that import and replace it with `CC` and `CS`

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -0,0 +1,164 @@
+import * as t from '@babel/types';
+import kebabCase from '@compiled/ts-transform-css-in-js/dist/utils/kebab-case';
+import { addUnitIfNeeded } from '@compiled/ts-transform-css-in-js/dist/utils/css-property';
```
lol-russo, line 3: All these utils in the ast and css builders coming from the TS Transform package feels like an interesting package boundary to me. I guess it's an implementation detail that doesn't impact users at all so it doesn't matter too much, it's only a dev dependency so it doesn't really matter if installing the babel plugin also installs the TS transformer.

madou:
Remember when I mentioned extracting internals to another package? This is
what I was talking about 😁

packages/babel-plugin/src/utils/css-builders.tsx
madou: lets add some more context around why we are doing this.

packages/babel-plugin/src/next.tsx
```diff
@@ -0,0 +1,60 @@
...
+      VariableDeclaration(path, state) {
+        if (!state.declarations) {
+          state.declarations = {};
+        }
+
+        if (!t.isIdentifier(path.node.declarations[0].id)) {
+          return;
+        }
+
+        const declarationName = path.node.declarations[0].id.name;
```
madou, line 34: add comment explaining why we are doing this and potential pitfalls of the current implementation (name clashes)

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -0,0 +1,164 @@
...
+  let css = '';
+
+  node.properties.forEach((prop) => {
+    if (t.isObjectProperty(prop)) {
+      // Don't use prop.value directly as it extracts constants from identifiers if needed.
+      const propValue = getInterpolation(prop.value, state);
+      const key = prop.key.name || prop.key.value;
+      let value = '';
+
+      if (t.isStringLiteral(propValue)) {
```
madou, line 59: lets add comments with example code for each block to help visualize whats happening

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+const normalizeContentValue = (value: string) => {
+  if (value.charAt(0) !== '"' && value.charAt(0) !== "'") {
+    return `"${value}"`;
+  }
+
+  return value;
+};
+
+const extractObjectExpression = (node: t.ObjectExpression, state: State): CSSOutput => {
```
madou, line 48: add jsdoc comments to each method explaining what its responsibility is for

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -0,0 +1,164 @@
...
+  let variables: CSSOutput['variables'] = [];
+  let css = '';
+
+  node.properties.forEach((prop) => {
+    if (t.isObjectProperty(prop)) {
+      // Don't use prop.value directly as it extracts constants from identifiers if needed.
```
madou, line 54: make this comment better

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -0,0 +1,164 @@
...
+      } else if (t.isObjectExpression(propValue)) {
+        const result = extractObjectExpression(propValue, state);
+        css += `${key} { ${result.css} }`;
+        variables = variables.concat(result.variables);
+        return;
+      } else if (t.isTemplateLiteral(propValue)) {
+        // eslint-disable-next-line @typescript-eslint/no-use-before-define
+        const result = extractTemplateLiteral(propValue, state);
+        value = result.css;
+        variables = variables.concat(result.variables);
```
madou, line 72: lets add a comment saying why we arent treating this like a mixin

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -0,0 +1,164 @@
...
+      } else if (t.isExpression(propValue)) {
+        const variableName = `--var-${hash(generate(propValue).code)}`;
+        variables.push({ name: variableName, expression: propValue });
+        value = `var(${variableName})`;
+      } else {
+        throw new Error(`Not supported.`);
```
madou, line 78: change this to use a stack frame error (maybe in next PR)

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -0,0 +1,164 @@
...
+        t.isObjectExpression(declaration.declarations[0].init)
+      ) {
+        // The declaration is in this module - let's use it!
+        const declarationValue = declaration.declarations[0].init;
+        const result = extractObjectExpression(declarationValue, state);
+        css += result.css;
+        variables = variables.concat(result.variables);
+        return;
+      } else {
+        throw new Error(`Declaration for ${prop.argument.name} could not be found.`);
```
madou, line 95: see if we can throw stack frame error (in next PR)

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  if (t.isTemplateLiteral(node)) {
+    return extractTemplateLiteral(node, state);
+  }
+
+  if (t.isObjectExpression(node)) {
+    return extractObjectExpression(node, state);
+  }
+
+  throw new Error('Unsupported node.');
```
madou, line 163: throw stack frame error (in another PR)

packages/babel-plugin/src/styled/__tests__/index.test.tsx
```diff
@@ -0,0 +1,889 @@
...
+
+      const ListItem = styled.div\`
+        font-size: \${props => props.color}px;
+      \`;
+    `);
+
+    expect(actual).toInclude('"--var-test-propscolor": (props.color || "") + "px"');
+  });
+
+  xit('should add react default import if missing', () => {
```
madou, line 101: delete me

PR #230, opened 2020-08-03
feat: add codemod for styled-components and emotion

pgmanutd:
This closes #179 

😓 uff!! Didn't get much time since past couple of weeks. Anyways here you go 😅. 
Just wait for another PR. Added `cli` in that for running codemods or doing other stuff. Though user can run codemods directly using `jscodeshift` as mentioned in `README` of this PR. So basically this PR is independent of that PR.

PR #231, opened 2020-08-03
Add cli package with codemods preset

pgmanutd:
This PR is basically an extension of PR #230. Added `cli` with `codemod` preset. We can add as many as presets as we want. It becomes easy for users now to use this `cli` to run codemods or doing other stuff.

Use this commit hash (05540874d0ab4e46e8241d33fdd71b461c4bb26e) to review things related to `cli`.


madou:
Let's come back to this when its in the sprint 👍

PR #232, opened 2020-08-12
Remove typescript transformer

madou:
- Moves common utils to `@compiled/utils`
- Moves css funcs to `@compiled/css`
- Creates stub function and tests for class names api
- Deletes `@compiled/ts-transformer-css-in-js`


packages/babel-plugin/package.json
```diff
@@ -19,17 +19,15 @@
     "src"
   ],
   "dependencies": {
-    "@babel/core": "^7.9.0",
+    "@babel/core": "^7.10.2",
```
madou, line 22: decided to make sure they were all aligned onto the same babel version

packages/babel-plugin/src/class-names/index.tsx
```diff
@@ -0,0 +1,5 @@
+import * as t from '@babel/types';
+import { NodePath } from '@babel/core';
+import { State } from '../types';
+
+export const visitClassNamesPath = (_: NodePath<t.JSXOpeningElement>, __: State) => {};
```
madou, line 5: stub for when we work on class names pi

packages/babel-plugin/src/css-prop/__tests__/index.test.tsx
```diff
@@ -1,14 +1,16 @@
 import { transformSync } from '@babel/core';
-import babelNext from '../../next';
+import babelPlugin from '../../index';
 
-jest.mock('@compiled/ts-transform-css-in-js/dist/utils/hash');
+jest.mock('@compiled/utils', () => {
+  return { ...jest.requireActual('@compiled/utils'), hash: () => 'hash-test' };
```
madou, line 5: because its a whole package now and not a single file we need to mock it like this. i couldnt get the `__mocks__` folder working - would take suggestions if you have any.

packages/babel-plugin/src/index.tsx
```diff
@@ -1,54 +1,60 @@
-import * as ts from 'typescript';
-import { parse as babelParse, transformSync, ParserOptions } from '@babel/core';
-import transformer from '@compiled/ts-transform-css-in-js';
-import { TransformerOptions } from '@compiled/ts-transform-css-in-js/dist/types';
+import * as t from '@babel/types';
```
madou, line 1: this hasnt changed just replaced `index` with `next`

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -1,10 +1,10 @@
 import template from '@babel/template';
 import * as t from '@babel/types';
-import { hash } from '@compiled/ts-transform-css-in-js/dist/utils/hash';
-import { transformCss } from '@compiled/ts-transform-css-in-js/dist/utils/css-transform';
-import { unique } from '@compiled/ts-transform-css-in-js/dist/utils/array';
 import traverse, { Scope, NodePath } from '@babel/traverse';
+import { hash, unique } from '@compiled/utils';
```
madou, line 4: wooo packages!

packages/css/src/array.tsx
```diff
@@ -11,3 +11,7 @@ export const unique = <TArrItem extends {}>(
     return acc;
   }, [] as TArrItem[]);
 };
+
+export const flatten = <TArr extends {}>(...assignments: TArr[][]): TArr[] => {
```
madou, line 15: moved this func into here from `functional-programming` file

packages/css/src/hash.tsx
```diff
@@ -44,11 +41,3 @@ export function hash(str: string, seed = 0): string {
 
   return (h >>> 0).toString(36);
 }
-
-export const classNameHash = (css: string): string => {
```
madou, line 48: removed because we dont use it anymore

packages/css/src/kebab-case.tsx
```diff
@@ -1,6 +1,6 @@
 const KEBAB_REGEX = /[A-Z\u00C0-\u00D6\u00D8-\u00DE]/g;
 
-export default function kebabCase(str: string) {
+export function kebabCase(str: string) {
```
madou, line 3: named > default

packages/css/src/source-maps.tsx
```diff
@@ -15,21 +9,21 @@ const getFileName = (sourceFile: ts.SourceFile): string => {
...
  */
-export function getSourceMap(
+export function buildSourceMap(
   offset: {
     line: number;
     character: number;
   },
-  sourceFile: ts.SourceFile,
-  context: ts.TransformationContext
+  code: string,
```
madou, line 17: this file not currently used but i make it not need TS anymore for when we implement this again.

packages/css/src/string-interpolations.tsx
```diff
@@ -86,5 +86,3 @@ export const cssBeforeInterpolation = (css: string): BeforeInterpolation => {
     variablePrefix,
   };
 };
-
-export const inline = (str?: string) => (str ? str : '');
```
madou, line 90: unused

packages/tsconfig.json
```diff
@@ -2,9 +2,10 @@
   "files": [],
   "references": [
     { "path": "babel-plugin/tsconfig.json" },
+    { "path": "css/tsconfig.json" },
```
madou, line 5: needed to add them in

packages/babel-plugin/src/css-prop/__tests__/index.test.tsx
pgmanutd: Sometimes I feel manual mocks makes me jump between files and folders, which I really don't like. Mocking where your tests are gives me clear picture where things are coming from. I like being explicit than implicitly mocking things using manual mocks. It's like having a pure function with all variables defined inside it (or coming from arguments). I can be wrong here 🤔

madou: agree! for me it was a trade-off for writing the same mock every time vs. placing it in __mocks__ folder 😄

pgmanutd: Yups right 😉

PR #234, opened 2020-08-13
Jest storybook babel

madou:


.storybook/preview-head.html
```diff
@@ -1,4 +1,4 @@
-<meta
+<!-- <meta
```
madou, line 1: turning this off until the feature is added back in

PR #235, opened 2020-08-13
chore: adds jsdoc to babel plugin

madou:


packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -4,33 +4,70 @@ import traverse, { Scope, NodePath } from '@babel/traverse';
...
 const buildCssVariablesProp = (
   variables: CSSOutput['variables'],
   transform = (expression: t.Expression) => expression
 ): (t.ObjectProperty | t.SpreadElement)[] => {
-  return unique(variables, (item) => item.name).map((variable) => {
-    return t.objectProperty(t.stringLiteral(variable.name), transform(variable.expression));
+  return unique(
+    // Make sure all defined CSS variables are unique
+    variables,
+    // We consider their uniqueness based on their name
```
lol-russo, line 52: Is this enough to consider them unique? It's valid css to have two variables with the same name in the same scope, the cascade just takes over and the last one defined is used. If we use a `find` for example, this could return something other than what the user intended

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -13,11 +13,24 @@ export interface CSSOutput {
   }[];
 }
 
+/**
+ * Will look in an expression and return the actual value.
+ * If the expression is an identifier node (a variable) and a constant,
+ * it will return the variable reference.
+ *
+ * E.g: If there was a identifier called `color` that is set somewhere as `const color = 'blue'`,
+ * passing the `color` identifier to this function would return `'blue'`.
```
lol-russo, line 22: 👏 this answers my question from the first time I reviewed this! much more approachable :)

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -32,6 +45,11 @@ const getInterpolation = <TNode extends {}>(expression: TNode | undefined, state
   return expression;
 };
 
+/**
+ * Will normalize the value of a `content` CSS property to ensure it has quotations around it.
+ * This is done to replicate both how Styled Components behaves,
+ * while not breaking how Emotion handles it.
```
lol-russo, line 51: 👏 always worth explicitly calling out when we are handling lib differences

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -107,38 +141,58 @@ const extractObjectExpression = (node: t.ObjectExpression, state: State): CSSOut
...
 
+/**
+ * Extracts CSS data from a template literal node.
+ *
+ * @param node Node we're interested in extracting CSS from.
+ * @param state Babel state - should house options and meta data used during the transformation.
+ */
 const extractTemplateLiteral = (node: t.TemplateLiteral, state: State): CSSOutput => {
   let variables: CSSOutput['variables'] = [];
+  // quasis are the string pieces of the template literal - the parts around the interpolations.
```
lol-russo, line 152: ❤️

packages/css/src/css-transform.tsx
```diff
@@ -22,20 +26,30 @@ const minify = () => {
...
+    .join(',\n');
+};
+
+/**
+ * Parent orphened pseudos PostCSS plugin.
+ * This plugin will move child nested orphened pseudos to the parent declaration.
+ *
+ * E.g: `.class { &:hover {} }` will become `.class:hover {}`
+ */
 const parentOrphenedPseudos = plugin('parent-orphened-pseudos', () => {
```
lol-russo, line 53: nit: spelling error should be "orphaned"

packages/babel-plugin/src/utils/ast-builders.tsx
madou: in this case its fine because it will always be the closest name.

```
import '@compiled/core';

let color = 'black';

export default function Component () {
  let color = 'red';
  return <div css={{ color }} />
}
```

99% case last reference wins - which is OK i think?

PR #236, opened 2020-08-13
feat: add React imports if missing in the module

lol-russo:
- should add react *  namespaced import if missing
- should add react default import if it only has named imports
- should do nothing if react default import or namespaced import is already defined

### DIFF
--- a/packages/babel-plugin/src/index.tsx
+++ b/packages/babel-plugin/src/index.tsx
@@ -1,4 +1,5 @@
 import * as t from '@babel/types';
+import template from '@babel/template';
 import { declare } from '@babel/helper-plugin-utils';
 import { importSpecifier } from './utils/ast-builders';
 import { visitCssPropPath } from './css-prop';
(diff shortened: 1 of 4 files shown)


packages/babel-plugin/src/styled/__tests__/index.test.tsx
```diff
@@ -100,27 +100,25 @@ describe('styled component transformer', () => {
...
+  it('should add react default import if missing', () => {
     const actual = transform(`
       import { styled } from '@compiled/core';
       const ListItem = styled.div\`
         font-size: 20px;
       \`;
     `);
-
-    expect(actual).toInclude('import React from "react";');
+    expect(actual).toInclude(`${transform("import * as React from 'react'")}`);
```
lol-russo, line 110: I found it a bit weird to add the `transform` here as well but otherwise the strings didn't match 🤔

packages/babel-plugin/src/index.tsx
```diff
@@ -11,16 +12,58 @@ export default declare<State>((api) => {
   return {
     inherits: require('babel-plugin-syntax-jsx'),
     visitor: {
+      Program(path, state) {
+        const importDeclarations = path.node.body.filter((node) => t.isImportDeclaration(node));
+        if (
+          importDeclarations.some(
+            (declaration: any) => declaration.source.value === '@compiled/core'
```
lol-russo, line 19: these `any`s are not ideal, but if I use what I believe is the correct type `t.ImportDeclaration` I get a TS error that I'm finding it a little hard to unpack 😓 

```
Argument of type '(declaration: t.ImportDeclaration) => boolean' is not assignable to parameter of type '(value: Statement, index: number, array: Statement[]) => unknown'.
  Types of parameters 'declaration' and 'value' are incompatible.
    Type 'Statement' is not assignable to type 'ImportDeclaration'.
      Type 'BlockStatement' is missing the following properties from type 'ImportDeclaration': specifiers, source, importKind
```

packages/babel-plugin/src/index.tsx
```diff
@@ -11,16 +12,58 @@ export default declare<State>((api) => {
   return {
     inherits: require('babel-plugin-syntax-jsx'),
     visitor: {
+      Program(path, state) {
+        const importDeclarations = path.node.body.filter((node) => t.isImportDeclaration(node));
```
madou, line 16:
```
         const importDeclarations = path.node.body.filter((node): node is t.ImportDeclaration => t.isImportDeclaration(node));
```

packages/babel-plugin/src/styled/__tests__/index.test.tsx
```diff
@@ -147,7 +145,7 @@ describe('styled component transformer', () => {
     expect(actual).toInclude('ListItem.displayName = "ListItem";');
   });
 
-  xit('should do nothing if react default import is already defined', () => {
+  it('should do nothing if react default import is already defined', () => {
     const actual = transform(`
       import React from 'react';
```
madou, line 150: good to have another test for `import * as React from 'react';`

madou:
changes coming in means this will need re-work - going to close sorry :)

PR #255, opened 2020-08-25
feat: adds support for inlining constant property access

### LINKED ISSUE #247: Inline constant object values into css
```
const obj = { color: 'red' };

styled.div`
  color: ${obj.color};
`;

.cc {
  color: red; // no dynamic value
}
```

madou:
Closes #247

### DIFF
--- a/packages/babel-plugin/src/css-prop/__tests__/index.test.tsx
+++ b/packages/babel-plugin/src/css-prop/__tests__/index.test.tsx
@@ -454,9 +454,14 @@ describe('css prop', () => {
           colors: {
             light: {
               primary: '#fff',
-            }
+            },
+            dark: {
+              primary: 'black',
+            },
           }
         };
+
+        theme.colors.light.primary;
 
         <div
           css={\`
--- a/packages/babel-plugin/src/css-prop/__tests__/index.test.tsx
+++ b/packages/babel-plugin/src/css-prop/__tests__/index.test.tsx
@@ -427,6 +427,48 @@ describe('css prop', () => {
       expect(actual).toInclude('.cc-hash-test{font-size:20px}');
     });
 
+    it('should inline constant object property value', () => {
+      const actual = transform(`
+        import '@compiled/core';
+        import React from 'react';
+
+        const colors = { error: 'red' };
+
+        <div
+          css={\`
+          color: \${colors.error};
+        \`}>
+          hello world
+        </div>
+      `);
+
+      expect(actual).toInclude('.cc-hash-test{color:red}');
+    });
+
+    it('should inline nested constant object property value', () => {
+      const actual = transform(`
+        import '@compiled/core';
+        import React from 'react';
+
+        const theme = {
+          colors: {
+            light: {
+              primary: '#fff',
+            }
+          }
+        };
+
+        <div
+          css={\`
+          color: \${theme.colors.light.primary};
+        \`}>
+          hello world
+        </div>
+      `);
+
+      expect(actual).toInclude('.cc-hash-test{color:#fff}');
+    });
+
     it('should transform binary expression', () => {
       const actual = transform(`
         import '@compiled/core';
@@ -639,12 +681,44 @@ describe('css prop', () => {
       expect(actual).toInclude('.cc-hash-test{font-size:20px}');
     });
 
+    it('should inline constant object property value', () => {
+      const actual = transform(`
+        import '@compiled/core';
+        import React from 'react';
+
+        const colors = { error: 'red' };
+
+        <div css={{ color: colors.error }}>hello world</div>
+      `);
+
+      expect(actual).toInclude('.cc-hash-test{color:red}');
+    });
+
+    it('should inline nested constant object property value', () => {
+      const actual = transform(`
+        import '@compiled/core';
+        import React from 'react';
+
+        const theme = {
+          colors: {
+            light: {
+              primary: '#fff',
+            }
+          }
+        };
+
+        <div css={{ color: theme.colors.light.primary }}>hello world</div>
+      `);
+
+      expect(actual).toInclude('.cc-hash-test{color:#fff}');
+    });
+
     it('should persist suffix of dynamic property value from objects into inline styles', () => {
       const actual = transform(`
         import '@compiled/core';
         import React from 'react';
 
-        const heading = {
+        let heading = {
           depth: 20
         };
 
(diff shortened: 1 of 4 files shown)


packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -0,0 +1,106 @@
...
+ * ['colors', 'primary']
+ * ```
+ *
+ * Would result in returning the `red` string literal node.
+ * If the value is not found `undefined` will be returned.
+ *
+ * @param expression - Member expression node.
+ * @param path - Path string array.
+ */
+export const getValueFromObjectExpression = (
```
madou, line 72: i did a quick spike previously to see if `path.evaluate()` would simplify things when accessing an object expression - turns out it doesn't return anything (just bails out) - so no worries.

packages/babel-plugin/src/css-prop/__tests__/index.test.tsx
```diff
@@ -639,12 +681,44 @@ describe('css prop', () => {
...
+      expect(actual).toInclude('.cc-hash-test{color:#fff}');
+    });
+
     it('should persist suffix of dynamic property value from objects into inline styles', () => {
       const actual = transform(`
         import '@compiled/core';
         import React from 'react';
 
-        const heading = {
+        let heading = {
```
pgmanutd, line 721: Are we checking something with `let` instead of `const`? 🤔

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -28,20 +33,38 @@ export interface CSSOutput {
...
+  if (t.isIdentifier(expression)) {
+    const binding = state.declarations[expression.name];
+
+    if (t.isVariableDeclaration(binding) && binding.kind === 'const') {
+      value = binding.declarations[0].init;
+    }
+  }
+
+  if (t.isMemberExpression(expression)) {
+    const bindingName = getMemberExpressionIdentifier(expression);
```
pgmanutd, line 51: Instead of `getMemberExpressionIdentifier `, `getMemberExpressionPath `, `getValueFromObjectExpression` can't we do something like this:

```typescript
let bindingName;
let bindingKey;

traverse(expression, {
  MemberExpression: {
    enter(path) {
      if (t.isIdentifier(path.node.object)) {
       // Will be theme
        bindingName = path.node.object.name;
      }
    },
    exit(path) {
      if (t.isIdentifier(path.node.property)) {
       // Will be colors, light, primary. But as we are mutating, we will take last value, i.e. primary
        bindingKey = path.node.property.name;
      }
    }
  }
});

// and then probably

const binding = state.declarations[bindingName];
let value;

traverse(binding.declarations[0].init, {
  ObjectProperty(path) {
      if (t.isIdentifier(path.node.key, { name: bindingKey })) {
         value = path.node.value;
      }
    }
});
```

Not sure, I can be wrong here 😞 (I am new to babel plugin world). Just thought of leveraging the power of babel traverse instead of our own tree traversal. 🤔 This might save some code and will be faster than `getValueFromObjectExpression` where we have loop inside another loop. 🤔

packages/babel-plugin/src/css-prop/__tests__/index.test.tsx
madou: if its const we will inline - if its let (or var) we won't

so i changed this so the test was still the same :P

packages/babel-plugin/src/utils/css-builders.tsx
madou: ohh not a bad idea - let me play with it. the only downside i see atm is:

1. it won't traverse the top level node (any ideas how we can without wrapping it in something?)
2. we need access to the parent path - so need to update a lot of wrapping code (which we will definitely do in the later PR for accessing the binding instead of the declaration stuff)

pgmanutd: oh right, `traverse` needs a parent node.

🤔 I think this should work

```typescript
traverse(binding.declarations[0].init, {
  ObjectProperty(path) {
      if (t.isIdentifier(path.node.key, { name: bindingKey })) {
         value = path.node.value;
      }
    }
});
```

because `binding.declarations[0].init` will be `ObjectExpression` and we are looking for child node i.e. `ObjectProperty`.

For other one, we can skip exit path and pick `primary` color from `MemberExpression` `property` key and `theme` will come when we traverse children or may be checking `object` key when we have case like `theme.primary`. Something like this:
```typescript
const bindingKey = t.isMemberExpression(expression) && t.isIdentifier(expression.property) && expression.property.name;
let bindingName = t.isMemberExpression(expression) && t.isIdentifier(expression.object) && expression.object.name;

traverse(expression, {
 //  Will not run in case of `theme.primary`. But will run when we have `theme.colors.light` and so on
  MemberExpression(path) {
    if (t.isIdentifier(path.node.object)) {
      // Will be theme.
      bindingName = path.node.object.name;
    }
  }
});
```

pgmanutd: And regarding parent path, I found this:

```typescript
traverse(expression, {
  noScope: true, // <====== this is the magic key, it says there is no scope, so no parent path is required
  MemberExpression(path) {
    if (t.isIdentifier(path.node.object)) {
      bindingName = path.node.object.name;
    }
  }
});
```

madou: ohh!! cool will have a play!! good job

madou: sounds like we will only need parent path (and scope) for when we want to run evaluate.

pgmanutd: Seems like. Because it might need to find the binding (if it's a variable) in that scope. But rest of the traversal can be done without scope and parent path right.

madou: okay so i've made some changes - one current problem is we need more information that just a single identifier to get the property value - because you can have the same identifier name in different parts of the object expression. i'll have a dig and see if theres anything in the scope/path we can use.

madou:
@pgmanutd i've made some changes but for the most part the underlying architecture is the same - at the point just kind of fiddling with the bits

if you have any suggestions let me know else let's move forward for now and we can come back if this proves to be a problem - would love to get stuck into the binding refactor and that might open up some avenues for improvement here

pgmanutd (approved):
Awesome man

PR #258, opened 2020-08-27
feat: [#257] add logic for picking arrow functions body

### LINKED ISSUE #257: Styled component interpolation function behaviour
See: [https://github.com/compiled/compiled-css-in-js/blob/0954b67b19530685d4505a0570e2404eeb78193c/packages/babel-plugin/src/styled/__tests__/index.test.tsx#L111](https://github.com/compiled/compiled-css-in-js/blob/0954b67b19530685d4505a0570e2404eeb78193c/packages/babel-plugin/src/styled/__tests__/index.test.tsx#L111)

Short circuiting is already implemented - what we need to do is update the code so we pick out the arrow functions body. Right now the output is:

`"--var-hash-test\":((props=>props.color)||\"\")`

It should be:

`"--var-hash-test\":(props.color||\"\")`

Think about other tests to ensure this implementation works.

This implementation is also partially implemented here: [https://github.com/compiled/compiled-css-in-js/blob/023d2f5afefa659a7ab0787af1ea0ab648f3a89f/packages/babel-plugin/src/utils/ast-builders.tsx#L136](https://github.com/compiled/compiled-css-in-js/blob/023d2f5afefa659a7ab0787af1ea0ab648f3a89f/packages/babel-plugin/src/utils/ast-builders.tsx#L136)

We want to update the code so it matches more use cases/scenarios.

We should also handle the case when the arrow function has a body with brackets, see test here: [https://github.com/compiled/compiled-css-in-js/blob/0954b67b19530685d4505a0570e2404eeb78193c/packages/babel-plugin/src/styled/__tests__/index.test.tsx#L333](https://github.com/compiled/compiled-css-in-js/blob/0954b67b19530685d4505a0570e2404eeb78193c/packages/babel-plugin/src/styled/__tests__/index.test.tsx#L333)

```typescript
styled.div`
  color: ${props => props.color};
  color: ${props => { return props.color }};
`;
```

We aren’t worrying about function declarations, e.g. `function () {}`

Update styled ast function to handle the above scenarios: [https://github.com/compiled/compiled-css-in-js/blob/023d2f5afefa659a7ab0787af1ea0ab648f3a89f/packages/babel-plugin/src/utils/ast-builders.tsx#L136](https://github.com/compiled/compiled-css-in-js/blob/023d2f5afefa659a7ab0787af1ea0ab648f3a89f/packages/babel-plugin/src/utils/ast-builders.tsx#L136)

**Testing:**

A few of these tests will be able to just enable and they will pass after fixing.

For the most part the wiring up of this is done - it’s a matter of uncommenting tests and seeing what needs to be changed.

pgmanutd:
closes #257

### DIFF
--- a/packages/babel-plugin/src/styled/__tests__/index.test.tsx
+++ b/packages/babel-plugin/src/styled/__tests__/index.test.tsx
@@ -108,7 +108,7 @@ describe('styled component transformer', () => {
     expect(actual).toInclude('<CS nonce={__webpack_nonce__}hash="hash-test">');
   });
 
-  xit('should shortcircuit props with suffix to a empty string to avoid undefined in css', () => {
+  it('should shortcircuit props with suffix to a empty string to avoid undefined in css', () => {
     const actual = transform(`
       import { styled } from '@compiled/core';
 
@@ -117,7 +117,7 @@ describe('styled component transformer', () => {
       \`;
     `);
 
-    expect(actual).toInclude('"--var-test-propscolor": (props.color || "") + "px"');
+    expect(actual).toInclude('"--var-hash-test":(props.color||"")+"px"');
   });
 
   xit('should add react default import if missing', () => {
@@ -234,7 +234,7 @@ describe('styled component transformer', () => {
       expect(actual).toInclude('.cc-hash-test{font-size:12}');
     });
 
-    xit('should not pass down invalid html attributes to the node when property has a suffix', () => {
+    it('should not pass down invalid html attributes to the node when property has a suffix', () => {
       const actual = transform(`
         import { styled } from '@compiled/core';
         const ListItem = styled.div\`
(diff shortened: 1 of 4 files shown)


packages/babel-plugin/src/styled/__tests__/index.test.tsx
```diff
@@ -339,8 +339,51 @@ describe('styled component transformer', () => {
         \`;
       `);
 
-      expect(actual).toInclude('.css-test{color:var(--var-test-propscolor)}');
-      expect(actual).toInclude('"--var-test-propscolor": (() => { return props.color; })() }}');
+      expect(actual).toInclude('.cc-hash-test{color:var(--var-hash-test)}');
+      expect(actual).toInclude('"--var-hash-test":(()=>{return props.color;})()}}');
+    });
+
+    it('should transform an arrow function with a body into an IIFE by preventing passing down invalid html attributes to the node', () => {
```
madou, line 346: great addition

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -76,6 +76,49 @@ const styledStyleProp = (
...
+ *
+ * @param node Node of type either BlockStatement or Expression
+ */
+export const wrapNodeInIIFE = (node: t.BlockStatement | t.Expression) =>
+  t.callExpression(t.arrowFunctionExpression([], node), []);
+
+const tryWrappingBlockStatementInIIFE = (node: t.BlockStatement | t.Expression) =>
+  t.isBlockStatement(node) ? wrapNodeInIIFE(node) : node;
+
+const pickArrowFunctionExpressionBody = (node: t.ArrowFunctionExpression) => {
```
madou, line 91: nit: we can move this into `ast.tsx` when my PR lands

madou (approved):
would be great for us to add some storybook examples to make sure theory and practice are aligning. if you have 5 minutes today let's add some before merging?

pgmanutd:
Yups sure. Thanks. Will just add few examples 😄

madou (approved):

PR #260, opened 2020-08-28
chore: removes use for declarations map and uses get binding on the parent path scope instead

madou:


packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -113,29 +113,28 @@ export const getKey = (node: t.Expression) => {
...
   let value: t.Node | undefined | null = undefined;
 
   if (t.isIdentifier(expression)) {
-    const binding = state.declarations[expression.name];
-
-    if (t.isVariableDeclaration(binding) && binding.kind === 'const') {
-      value = binding.declarations[0].init;
+    const binding = meta.parentPath.scope.getBinding(expression.name);
+    if (binding && binding.kind === 'const' && t.isVariableDeclarator(binding.path.node)) {
+      value = binding.path.node.init;
```
pgmanutd, line 125: ah!! finally no more index access `binding.declarations[0]`. Binding gives us `VariableDeclarator` directly 🎉

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -113,29 +113,28 @@ export const getKey = (node: t.Expression) => {
...
     const { accessPath, bindingIdentifier } = getMemberExpressionMeta(expression);
-    const binding = state.declarations[bindingIdentifier.name];
+    const binding = meta.parentPath.scope.getBinding(bindingIdentifier.name);
 
     if (
-      t.isVariableDeclaration(binding) &&
+      binding &&
       binding.kind === 'const' &&
-      t.isObjectExpression(binding.declarations[0].init)
+      t.isVariableDeclarator(binding.path.node) &&
```
pgmanutd, line 134: nit: should we extract this into a function? 🤔 We have used it twice atleast

```typescript
const isBindedVariableDeclarator = (binding, kind = 'const') => binding && binding.kind === kind && t.isVariableDeclarator(binding.path.node);
```

and probably for this too `binding.path.node.init`

pgmanutd:
Damn!! loved these changes ❤️. Finally we are getting rid of `state.declarations`. Less states means less bugs 😛

pgmanutd:
Looks good. Go ahead with merging @madou. Just check one review comment I left. If you think it's not important atm, no worries then.

pgmanutd (approved):

packages/babel-plugin/src/utils/ast.tsx
madou: good idea

madou:
thanks! will merge - just rebased from master

good job with your examples!!

PR #261, opened 2020-08-28
fix: fixes hard coded usage of styled import specifier

### LINKED ISSUE #256: Remove hardcoded 'styled' check for 'local.name'
Replace `local.name` with `imported.name` [here](https://github.com/atlassian-labs/compiled/blob/023d2f5afefa659a7ab0787af1ea0ab648f3a89f/packages/babel-plugin/src/index.tsx#L20). 

Why? because when codemoded, original code

`import styledFunction from '@emotion/styled';` 

will be converted to

`import { styled as styledFunction } from '@compiled/css-in-js';`

Please think of other scenarios we can have here.

madou:
Closes #256

### DIFF
--- a/packages/babel-plugin/src/index.tsx
+++ b/packages/babel-plugin/src/index.tsx
@@ -1,5 +1,6 @@
 import { declare } from '@babel/helper-plugin-utils';
 import template from '@babel/template';
+import * as t from '@babel/types';
 import { importSpecifier } from './utils/ast-builders';
 import { visitCssPropPath } from './css-prop';
 import { visitStyledPath } from './styled';
@@ -13,7 +14,7 @@ export default declare<State>((api) => {
     visitor: {
       Program: {
         exit(path, state) {
-          if (state.compiledImportFound && !path.scope.getBinding('React')) {
+          if (state.compiledImports && !path.scope.getBinding('React')) {
             // React is missing - add it in at the last moment!
             path.unshiftContainer('body', template.ast(`import * as React from 'react'`));
           }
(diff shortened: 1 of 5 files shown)


madou:
@pgmanutd please review :)

pgmanutd:
> @pgmanutd please review :)

packages/babel-plugin/src/index.tsx
```diff
@@ -13,39 +14,52 @@ export default declare<State>((api) => {
...
-            )
+            .filter((specifier) => {
+              if (!state.compiledImports || !t.isImportSpecifier(specifier)) {
+                return false;
+              }
+
+              if (specifier.imported.name === 'styled') {
+                state.compiledImports.styled = specifier.local.name;
+                // Remove the import
+                return false;
```
pgmanutd, line 37: nit: Are we concerned about loop performance? Coz we can do `.find` and `.filter` separately for readability may be. Otherwise this looks cool 🎉

packages/babel-plugin/src/index.tsx
```diff
@@ -13,39 +14,52 @@ export default declare<State>((api) => {
...
+              // Keep the import
+              return true;
+            })
+            // Add on the util imports we're going to use later in the transform.
             .concat([importSpecifier('CC'), importSpecifier('CS')]);
         }
       },
       TaggedTemplateExpression(path, state) {
-        if (!state.compiledImportFound) {
+        if (!state.compiledImports?.styled) {
```
pgmanutd, line 48: Optional chaining is dope. I remember from coffescript days how we used to do this. It was awesome. This is the one I am talking about http://coffeescript.org/#existential-operator

pgmanutd (dismissed):

packages/babel-plugin/src/index.tsx
madou: thanks!

honestly we might be concerned - we'll find out later when we roll out to Jira 😓

packages/babel-plugin/src/index.tsx
```diff
@@ -13,39 +14,52 @@ export default declare<State>((api) => {
...
+          state.compiledImports = {};
+
           path.node.specifiers = path.node.specifiers
-            .filter(
-              (specifier) =>
-                specifier.local.name !== 'styled' && specifier.local.name !== 'ClassNames'
-            )
+            .filter((specifier) => {
+              if (!state.compiledImports || !t.isImportSpecifier(specifier)) {
+                return false;
```
madou, line 31: @pgmanutd we might want to return `true` just in case - i'll update the code

madou: done - check the latest commit

packages/babel-plugin/src/index.tsx
```diff
@@ -13,39 +14,55 @@ export default declare<State>((api) => {
...
           }
         },
       },
       ImportDeclaration(path, state) {
-        if (path.node.source.value === '@compiled/core') {
-          state.compiledImportFound = true;
-          path.node.specifiers = path.node.specifiers
-            .filter(
-              (specifier) =>
-                specifier.local.name !== 'styled' && specifier.local.name !== 'ClassNames'
```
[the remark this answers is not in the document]
madou: i figured since its not implemented we just remove it for now

pgmanutd (approved):

PR #262, opened 2020-08-28
fix: imports react at the end of the transformation if its missing

### LINKED ISSUE #241: Import react module when not defined
If `React` is found within the module - don't do anything. If it isn't found - add import:

```
```

Add ability to turn it off (either by config - or by finding if the JSX babel transform to automatically add the import is turned on).

madou:
Closes #241 

Built on what you did @lol-russo but simplified it a lot. The basic premise is "does `React` exist somewhere in the scope - yes? Do nothing. No? Add it.

### DIFF
--- a/packages/babel-plugin/src/index.tsx
+++ b/packages/babel-plugin/src/index.tsx
@@ -1,4 +1,5 @@
 import { declare } from '@babel/helper-plugin-utils';
+import template from '@babel/template';
 import { importSpecifier } from './utils/ast-builders';
 import { visitCssPropPath } from './css-prop';
 import { visitStyledPath } from './styled';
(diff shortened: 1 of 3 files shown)


madou:
@pgmanutd please review :)

packages/babel-plugin/src/styled/__tests__/index.test.tsx
```diff
@@ -176,7 +176,7 @@ describe('styled component transformer', () => {
       \`;
     `);
 
-    expect(actual).toInclude("import React from 'react';");
+    expect(actual).toInclude("import React from'react';");
```
marcins, line 179: Is the lack of space here after `from` correct? (is that actually valid JavaScript?). Seems to be some of that above as well on line 143.

packages/babel-plugin/src/index.tsx
```diff
@@ -10,6 +11,14 @@ export default declare<State>((api) => {
   return {
     inherits: require('babel-plugin-syntax-jsx'),
     visitor: {
+      Program: {
+        exit(path, state) {
```
lol-russo, line 15: So much cleaner! Nice 👍

packages/babel-plugin/src/styled/__tests__/index.test.tsx
madou: It's valid. It's just the test code compressing everything so we have consistent white space

lol-russo (approved):

PR #263, opened 2020-08-28
feat: adds support for composing styles to styled api

### LINKED ISSUE #246: Array and identifier css for styled api/css prop
```
styled.div([])
styled.div(identifier)
css={[]}
css={identifier}
```

madou:
Closes #246 

So this was mostly already implemented in CSS prop - I did a bit of refactoring to move the logic to `css-builder`.

### DIFF
--- a/packages/babel-plugin/package.json
+++ b/packages/babel-plugin/package.json
@@ -19,6 +19,7 @@
     "src"
   ],
   "dependencies": {
+    "@compiled/css": "0.4.7",
     "@babel/core": "^7.10.2",
     "@babel/generator": "^7.10.2",
     "@babel/helper-plugin-utils": "^7.10.2",
--- a/packages/babel-plugin/src/styled/__tests__/index.test.tsx
+++ b/packages/babel-plugin/src/styled/__tests__/index.test.tsx
@@ -49,6 +49,22 @@ describe('styled component transformer', () => {
     `);
   });
 
+  it('should compose CSS from multiple sources', () => {
+    const actual = transform(`
+      import { styled } from '@compiled/core';
+
+      const styles = { fontSize: 12 };
+
+      const ListItem = styled.div([
+        styles,
+        \`color: blue;\`,
+        { fontWeight: 500 }
+      ]);
+    `);
+
+    expect(actual).toInclude('.cc-hash-test{font-size:12px;color:blue;font-weight:500}');
+  });
+
   it('should not pass down invalid html attributes to the node', () => {
     const actual = transform(`
       import { styled } from '@compiled/core';
--- a/packages/babel-plugin/src/styled/index.tsx
+++ b/packages/babel-plugin/src/styled/index.tsx
@@ -32,7 +32,7 @@ const extractStyledDataFromNode = (
     t.isMemberExpression(node.callee) &&
     t.isIdentifier(node.callee.object) &&
     node.callee.object.name === 'styled' &&
-    t.isObjectExpression(node.arguments[0]) &&
+    t.isExpression(node.arguments[0]) &&
     t.isIdentifier(node.callee.property)
   ) {
     const tagName = node.callee.property.name;
--- a/packages/babel-plugin/src/utils/css-builders.tsx
+++ b/packages/babel-plugin/src/utils/css-builders.tsx
@@ -184,5 +184,37 @@ export const buildCss = (node: t.Expression, meta: Metadata): CSSOutput => {
     return extractObjectExpression(node, meta);
   }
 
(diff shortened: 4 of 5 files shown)


madou:
@pgmanutd please review :)

packages/babel-plugin/package.json
```diff
@@ -19,6 +19,7 @@
     "src"
   ],
   "dependencies": {
+    "@compiled/css": "0.4.7",
```
madou, line 22: this was missing from dependencies - we should add it before release.. :)

packages/babel-plugin/src/styled/index.tsx
```diff
@@ -32,7 +32,7 @@ const extractStyledDataFromNode = (
     t.isMemberExpression(node.callee) &&
     t.isIdentifier(node.callee.object) &&
     node.callee.object.name === 'styled' &&
-    t.isObjectExpression(node.arguments[0]) &&
+    t.isExpression(node.arguments[0]) &&
```
madou, line 35: we widen the typecheck so any expression can be passed through (such as arrays, identifiers, and so on)

pgmanutd:
> @pgmanutd please review :)

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -184,5 +184,37 @@ export const buildCss = (node: t.Expression, meta: Metadata): CSSOutput => {
...
   }
 
+  if (t.isIdentifier(node)) {
+    const binding = meta.parentPath.scope.getBinding(node.name);
+    if (
+      !binding ||
+      !t.isVariableDeclarator(binding.path.node) ||
+      !t.isExpression(binding.path.node.init)
+    ) {
+      throw new Error();
```
pgmanutd, line 194: Should we throw some meaning full error here with some description? 🤔

packages/babel-plugin/src/styled/index.tsx
pgmanutd: What all we have to support here? `isExpression` will also return true for expression of any type including `ArrowFunctionExpression`. Should we limit this 🤔 ? Do we have to handle nested arrays? Didn't dig deep but something like this:

```typescript
    const ListItem5 = styled.div([
        styles,
        `color: blue;`,
        { fontWeight: 500 },
       [`background-color: red`]
      ]);
```

pgmanutd: I think we did handle. There is a recursion in `buildCss` 😅

pgmanutd: We can add tests for nested arrays thought. They will get flattened with recursion.

packages/babel-plugin/src/styled/__tests__/index.test.tsx
```diff
@@ -49,6 +49,22 @@ describe('styled component transformer', () => {
     `);
   });
 
+  it('should compose CSS from multiple sources', () => {
+    const actual = transform(`
+      import { styled } from '@compiled/core';
+
+      const styles = { fontSize: 12 };
+
+      const ListItem = styled.div([
```
pgmanutd, line 58: nice ❤️

packages/babel-plugin/src/styled/index.tsx
pgmanutd: > `isExpression` will also return true for expression of any type including `ArrowFunctionExpression`

ahh!! we are throwing error for unsupported node in `buildCss` 🎉

packages/babel-plugin/src/utils/css-builders.tsx
madou: we probably should - if it's ok with you i'll make another issue to come back and improve all these errors in one pass?

packages/babel-plugin/src/styled/index.tsx
madou: yeah!

madou: > Do we have to handle nested arrays?

i think it could work but its not officially supported atm - at least from the perspective of the style typing. maybe for now we leave it?

pgmanutd: Yeah no worries man. I was just curious to know. Loved the work ❤️

packages/babel-plugin/src/utils/css-builders.tsx
pgmanutd: yeah no worries man. we can do it later 😄

pgmanutd (dismissed):

pgmanutd (approved):

PR #264, opened 2020-08-29
feat: evaluates constant expressions

### LINKED ISSUE #248: Evaluate simple expressions that resolve to a constant value
```
css={{ fontSize: 2 + 2 }}

.cc {
  font-size: 4; // reduced to 4
}
```

handle:
- multiplication
- addition
- subtraction
- division

on:

- strings
- numbers
- identifiers that resolve to a constant string or number

Use babel path method `path.evaluate`

madou:
Closes #248

### DIFF
--- a/packages/babel-plugin/src/css-prop/__tests__/index.test.tsx
+++ b/packages/babel-plugin/src/css-prop/__tests__/index.test.tsx
@@ -146,6 +146,34 @@ describe('css prop', () => {
     expect(actual).toInclude('.cc-hash-test{font-size:var(--var-hash-test)}');
   });
 
+  it('should bail out evaluating expression referencing a mutable identifier', () => {
+    const actual = transform(`
+      import '@compiled/core';
+      import React from 'react';
+
+      let mutable = 2;
+      const dontchange = mutable;
+
+      <div css={{ fontSize: dontchange }}>hello world</div>
+    `);
+
+    expect(actual).toInclude('.cc-hash-test{font-size:var(--var-hash-test)}');
+  });
+
+  it('should bail out evaluating expression that references multiple mutable expressions', () => {
+    const actual = transform(`
+      import '@compiled/core';
+      import React from 'react';
+
+      let mutable = false;
+      const dontchange = mutable ? 1 : 2;
+
+      <div css={{ fontSize: dontchange }}>hello world</div>
+    `);
+
+    expect(actual).toInclude('.cc-hash-test{font-size:var(--var-hash-test)}');
+  });
+
   it('should bail out evaluating binary expression referencing a mutable identifier', () => {
     const actual = transform(`
       import '@compiled/core';
--- a/packages/babel-plugin/src/css-prop/__tests__/index.test.tsx
+++ b/packages/babel-plugin/src/css-prop/__tests__/index.test.tsx
@@ -139,6 +139,7 @@ describe('css prop', () => {
       import React from 'react';
 
       let mutable = 2;
+      mutable = 1;
 
       <div css={{ fontSize: mutable }}>hello world</div>
     `);
@@ -146,13 +147,14 @@ describe('css prop', () => {
     expect(actual).toInclude('.cc-hash-test{font-size:var(--var-hash-test)}');
   });
 
-  it('should bail out evaluating identifier expression referencing a mutable identifier', () => {
+  it('should bail out evaluating identifier expression referencing a mutated identifier', () => {
     const actual = transform(`
       import '@compiled/core';
       import React from 'react';
 
       let mutable = 2;
       const dontchange = mutable;
+      mutable = 3;
 
       <div css={{ fontSize: dontchange }}>hello world</div>
     `);
@@ -160,13 +162,14 @@ describe('css prop', () => {
     expect(actual).toInclude('.cc-hash-test{font-size:var(--var-hash-test)}');
   });
 
-  it('should bail out evaluating expression that references a constant expression referencing a mutable expression', () => {
+  it('should bail out evaluating expression that references a constant expression referencing a mutated expression', () => {
     const actual = transform(`
       import '@compiled/core';
       import React from 'react';
 
       let mutable = false;
       const dontchange = mutable ? 1 : 2;
+      mutable = true;
 
       <div css={{ fontSize: dontchange }}>hello world</div>
     `);
@@ -174,12 +177,13 @@ describe('css prop', () => {
     expect(actual).toInclude('.cc-hash-test{font-size:var(--var-hash-test)}');
   });
 
-  it('should bail out evaluating a binary expression referencing a mutable identifier', () => {
+  it('should bail out evaluating a binary expression referencing a mutated identifier', () => {
     const actual = transform(`
       import '@compiled/core';
       import React from 'react';
 
       let mutable = 2;
+      mutable = 3;
 
       <div css={{ fontSize: mutable * 2 }}>hello world</div>
     `);
@@ -303,6 +307,7 @@ describe('css prop', () => {
       import '@compiled/core';
 
       let hello = true;
+      hello = false;
 
       <div css={{ color: hello ? 'red' : 'blue', fontSize: 10 }}>hello world</div>
     `);
@@ -338,6 +343,7 @@ describe('css prop', () => {
       import {useState} from 'react';
 
       let N30 = 'gray';
+      N30 = 'blue';
 
       <div css={{
         backgroundImage: \`linear-gradient(45deg, \${N30} 25%, transparent 25%),
@@ -358,7 +364,9 @@ describe('css prop', () => {
       import '@compiled/core';
 
       let hello = true;
+      hello = false;
       let color = hello ? 'red' : 'blue' ;
+
       <div css={{ color }}>hello world</div>
     `);
 
@@ -371,7 +379,9 @@ describe('css prop', () => {
       import '@compiled/core';
 
       let hello = true;
+      hello = false;
       let colorsz = hello ? 'red' : 'blue' ;
+
       <div css={{ color: colorsz }}>hello world</div>
     `);
 
@@ -471,6 +481,7 @@ describe('css prop', () => {
         import React from 'react';
 
         let fontSize = 20;
+        fontSize = 19;
 
         <div css={\`font-size: \${fontSize}px;color:red;\`}>hello world</div>
       `);
@@ -626,6 +637,8 @@ describe('css prop', () => {
         import React from 'react';
 
         let sidenav = true;
+        sidenav = false;
+
         <div
           css={\`
             display: grid;
@@ -738,7 +751,9 @@ describe('css prop', () => {
         import '@compiled/core';
 
         let x = 1;
+        x = 1;
         let y = '2px';
+        y = '2px';
 
         <div
           css={\`
@@ -766,6 +781,7 @@ describe('css prop', () => {
 
         const bg = 'blue';
         let cl = 'red';
+        cl = 'red';
 
         <div css={{ background: bg, color: cl, textDecoration: 'none', }}>hello world</div>
       `);
@@ -821,6 +837,19 @@ describe('css prop', () => {
       expect(actual).toInclude('.cc-hash-test{color:#fff}');
     });
 
+    it('should not exhaust the stack when an identifier references itself', () => {
+      expect(() => {
+        transform(`
+        import '@compiled/core';
+        import React from 'react';
+
+        let heading = heading || 20;
+
+        <div css={{ marginLeft: \`\${heading.depth}rem\`, color: 'red' }}>hello world</div>
+      `);
+      }).not.toThrow();
+    });
+
     it('should persist suffix of dynamic property value from objects into inline styles', () => {
       const actual = transform(`
         import '@compiled/core';
@@ -829,6 +858,7 @@ describe('css prop', () => {
         let heading = {
           depth: 20
         };
+        heading = {};
 
         <div css={{ marginLeft: \`\${heading.depth}rem\`, color: 'red' }}>hello world</div>
       `);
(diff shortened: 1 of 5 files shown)


madou:
@pgmanutd please review :)

packages/babel-plugin/src/css-prop/__tests__/index.test.tsx
```diff
@@ -226,7 +291,7 @@ describe('css prop', () => {
     const actual = transform(`
       import '@compiled/core';
 
-      const hello = true;
+      let hello = true;
```
madou, line 294: these all changed to mutable else they are evaluated to a static representation :)

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -1,7 +1,38 @@
...
+ * @param node
+ * @param parentPath
+ */
+export const getPathOfNode = <TNode extends {}>(
+  node: TNode,
+  parentPath: NodePath
+): NodePath<TNode> => {
+  let foundPath: NodePath | null = null;
+
+  traverse(
```
madou, line 16: can we think of any better way to do this?

pgmanutd:
> @pgmanutd please review :)

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -99,6 +130,77 @@ export const getKey = (node: t.Expression) => {
...
+ * Will recursively traverse a path and its identifiers to find all bindings.
+ * If any of those bindings are mutable `true` will be returned.
+ * Else `false`.
+ *
+ * @param path
+ */
+export const doesPathReferenceAnyMutableIdentifiers = (path: NodePath<any>): boolean => {
+  if (path.isIdentifier()) {
+    const binding = path.scope.getBinding(path.node.name);
+    if (!binding || !t.isVariableDeclarator(binding.path.node) || binding.kind !== 'const') {
```
pgmanutd, line 143: nit: `binding.kind !== 'const'` can be `!binding.constant`

packages/babel-plugin/src/utils/ast.tsx
pgmanutd: Its difficult I think 😞. I mean we are playing with nodes at each level and not having reference to its path info. We are only keeping reference of top level path in `meta.parentPath`. The way you did @madou is not that bad. We have access to `parentPath`. We can traverse through it and find the node. Do we need to wrap node in `t.expression?` Can't we traverse through `parentPath.node` and check the reference of **node** in `enter` like `path.node === node` and return that **path** 🤔. I think reference will be same 🤔.

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -99,6 +130,77 @@ export const getKey = (node: t.Expression) => {
...
+      return true;
+    }
+
+    return doesPathReferenceAnyMutableIdentifiers(
+      getPathOfNode(binding.path.node.init as t.Expression, binding.path)
+    );
+  }
+
+  let mutable = false;
+  path.traverse({
```
pgmanutd, line 153: Damn these traversal are dope. ❤️

packages/babel-plugin/src/utils/ast.tsx
madou: yeah that's the worst part i find of this TBH

`traverse` will only traverse children - not the immediate passed in node - so i used this as a trick to get around it.

we could do `path.parentPath` instead maybe? 🤔

pgmanutd (commented):
Looks good 🎉 . 

nit: Just verify once those traversal logics (`getPathOfNode` and `doesPathReferenceAnyMutableIdentifiers`). May be with more tests (use cases) or storybook examples if possible.

madou:
i'll add some storybooks and fix the nits :)

packages/babel-plugin/src/utils/ast.tsx
madou: interestingly tests start failing when making this change - maybe they're for two different things?

packages/babel-plugin/src/utils/ast.tsx
pgmanutd: 😞 we will get best out of babel as we go on. No worries. We just need to make sure to have as many as tests possible right now.

packages/babel-plugin/src/utils/ast.tsx
pgmanutd: 😱 It was given here so I thought https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md#bindings

pgmanutd: ohh got it. It refers to variable which is not assigned again. Even if it is `let`

pgmanutd: Can we make good use of 

```typescript
 constant: false,
constantViolations: [path]
```

somehow?

madou:
added some examples - let me know if you can think of any missing tests as well and i can add them

packages/babel-plugin/src/utils/ast.tsx
madou: hmm let me look at what `constantViolations` pumps out - wonder what it's for?

pgmanutd: I think this one:

```typescript
function scopeOne() {
  var ref1 = "This is a constant binding";

  becauseNothingEverChangesTheValueOf(ref1);

  function scopeTwo() {
    var ref2 = "This is *not* a constant binding";
    ref2 = "Because this changes the value";
  }
}
```

madou: OH! so we can safely use `binding.constant` - it basically returns `true` if it isnt mutated - and false otherwise, maybe?

sounds good! we should use it! let me test

pgmanutd: yups. I think its for that purpose only.

madou: yeah thats exactly what happens. awesome

madou:
yeah good catch on the constant - im going to make the changes and use it.

madou:
alright so i've made some changes - i have one outstanding test to fix which is `should not exhaust the stack when an identifier references itself`

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -113,15 +236,12 @@ export const getKey = (node: t.Expression) => {
  * @param expression Expression we want to interrogate.
  * @param state Babel state - should house options and meta data used during the transformation.
  */
-export const getInterpolation = <TNode extends {}>(
-  expression: TNode | undefined,
-  meta: Metadata
-) => {
+export const getInterpolation = (expression: t.Expression, meta: Metadata): t.Expression => {
```
madou, line 239: i removed the generic because it was causing more type pain than it was worth

packages/babel-plugin/src/css-prop/__tests__/index.test.tsx
```diff
@@ -122,6 +122,75 @@ describe('css prop', () => {
...
+      import '@compiled/core';
+      import React from 'react';
+
+      <div css={{ fontSize: 8 * 2 }}>hello world</div>
+    `);
+
+    expect(actual).toInclude('.cc-hash-test{font-size:16px}');
+  });
+
+  it('should bail out evaluating expression referencing a mutable identifier', () => {
```
pgmanutd, line 136: Do we have test like this somewhere where there is no constant violation 🤔?

```typescript
it('test statement', () => {
    const actual = transform(`
      import '@compiled/core';
      import React from 'react';

      let notMutatedAgain = 20;

      <div css={{ fontSize: notMutatedAgain }}>hello world</div>
    `);

    expect(actual).toInclude('.cc-hash-test{font-size:20px}');
  });
```

madou: not anymore - i'll add one

madou: done

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -99,6 +130,98 @@ export const getKey = (node: t.Expression) => {
...
+    const refPath = binding.referencePaths[i];
+    const innerBinding = refPath.scope.getOwnBinding(path.node.name);
+    if (!innerBinding) {
+      continue;
+    }
+
+    if (!t.isVariableDeclarator(innerBinding.path.node) || !innerBinding.constant) {
+      return true;
+    }
+  }
```
pgmanutd, line 157: nit: Can be replaced with a function to remove duplicate functionality if possible. Not sure. I prefer duplicate code over wrong abstraction any day. See if its really duplicate, otherwise lets skip abstraction part.

If we want, may be something like this

```typescript
function someFunction(path, getBindingFn) {
    const binding = path.scope[getBindingFn(path.node.name)];

    if (!binding) {
      return false;
    }

    if (!t.isVariableDeclarator(binding.path.node) || !binding.constant) {
      return true;
    }

    return false;
}
```

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -99,6 +130,98 @@ export const getKey = (node: t.Expression) => {
...
+ * Returns `true` if an identifier or any paths that reference the identifier are mutated.
+ * @param path
+ */
+const isIdentifierReferencesMutated = (path: NodePath<t.Identifier>): boolean => {
+  const binding = path.scope.getBinding(path.node.name);
+  if (!binding) {
+    return false;
+  }
+
+  if (!t.isVariableDeclarator(binding.path.node) || !binding.constant) {
```
pgmanutd, line 143: Which case is this one `!t.isVariableDeclarator(binding.path.node)`?

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -99,6 +130,98 @@ export const getKey = (node: t.Expression) => {
...
+  }
+
+  let mutated = false;
+  path.traverse({
+    Identifier(innerPath) {
+      const result = isIdentifierReferencesMutated(path);
+      if (result) {
+        mutated = true;
+        // No need to keep traversing - let's stop!
+        innerPath.stop();
```
pgmanutd, line 180: ahh!! nice ❤️

packages/babel-plugin/src/utils/ast.tsx
madou: this forces all bindings to be

`blah = 'ok';`

variable declarator^

functions declarations could be other things we aren't handling atm. will tackle them when we get to mixins.

packages/babel-plugin/src/utils/ast.tsx
madou: i dont think its worth it for now to be honest, the two are slightly different

would recommend just leaving it for now

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -99,6 +130,98 @@ export const getKey = (node: t.Expression) => {
...
+    return false;
+  }
+
+  if (!t.isVariableDeclarator(binding.path.node) || !binding.constant) {
+    return true;
+  }
+
+  for (let i = 0; i < binding.referencePaths.length; i++) {
+    const refPath = binding.referencePaths[i];
+    const innerBinding = refPath.scope.getOwnBinding(path.node.name);
```
madou, line 149: i have to wonder if this should be getBinding..

madou: yeah still passes, pushing change

packages/babel-plugin/src/utils/ast.tsx
pgmanutd: yeeee duplicate code over wrong abstraction wins 😛 hurrayy 🎉

packages/babel-plugin/src/utils/ast.tsx
pgmanutd: yeah..thats what I thought of..that we are not handling those cases as of now..!! Just wanted to confirm 😄

packages/babel-plugin/src/utils/ast.tsx
pgmanutd: yeah..`getBinding` is better. It will traverse more right. We can have test cases for this if we want.

madou: any link for the differences?

pgmanutd: https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md#scope

See this

madou: dont really see anything going in depth around the get binding functions though 🤔

pgmanutd: tests tests tests..they might help..!! 😛

pgmanutd: May be something like this

```typescript
function cube(n) {
  function square(n) {
    return n * n;
  }

  return n * square(n);
}
```

madou: if this ends up being called as `cube(2)` it would just bail out to evaluate it as a dynamic variable :)

madou:
@pgmanutd what do we think is missing :)

pgmanutd:
> @pgmanutd what do we think is missing :)

Nah nothing. I thought you were investigating `getOwnBinding`. Looks good anyways 🎉

pgmanutd (approved):

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -1,7 +1,38 @@
 import * as t from '@babel/types';
```
theKashey, line 1: Oh how I wish this file to be a separate package 🙏

madou: if we made it another package in compiled would that work for you? want to put a PR up?

theKashey: 🤔 `@compiled` or a `secret-org-name`?

PR #266, opened 2020-09-03
First pass at module traversal

madou:
Spiking out module traversal - the things we want answers to

## Answers

- [x] fs read and babel parse found modules
- [x] return imported nodes and be able to inline them
- [x] find relative modules `./colors`
- [x] find node modules `@atlaskit/theme`
- [x] create tests we can use when actually implementing this
- [x] simple css objects
- [x] css objects referencing local identifiers
- [x] css objects referencing imported identifiers
- [x] working storybook examples
- [x] simple css strings
- [x] css strings referencing local identifiers
- [x] css strings referencing imported identifiers
- [x] throw when can't statically evaluate over module boundaries (stop gap measure until we implement it properly)

## Todo follow up
- [ ] implement basic caching strategy

## Dropped

- [ ] find absolute modules `jira/my-app`
- [ ] find custom absolute modules `~src/button`

## Notes

- Investigate potential solutions/workarounds for mixins found in other modules where they reference identifiers that can't be statically evaluated - currently it will generate invalid code. We can error - or we can handle it (by importing those modules into the owning module - but that's a decent amount of effort)
- Do we need to handle dynamic imports in some fashion?
- How do we want to handle caching?


packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -128,6 +130,29 @@ export const getInterpolation = <TNode extends {}>(
     const { accessPath, bindingIdentifier } = getMemberExpressionMeta(expression);
     const binding = meta.parentPath.scope.getBinding(bindingIdentifier.name);
 
+    if (
+      binding &&
+      binding.path.isImportDefaultSpecifier() &&
+      binding.path.parentPath.isImportDeclaration()
+    ) {
+      const moduleImportName = binding.path.parentPath.node.source.value;
+      const modulePath = require.resolve(meta.state.cwd + '/' + moduleImportName);
```
pgmanutd, line 139: Where we are setting `meta.state.cwd` 🤔?

madou: babel sets it - we override it in the test https://github.com/atlassian-labs/compiled/pull/266/files#diff-7162f4ecb36d5e9e278ec6b3e335af9eR13

still need to figure out what the best path is for all of this, will flesh it out more when i start doing fake node modules

pgmanutd: Are we covering imports with aliases or paths like we have in webpack or tsconfig? 🤔

for eg:
`import fn from '@root/utils/fn';`

or

`import fn from 'src/utils/fn';`

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -128,6 +130,29 @@ export const getInterpolation = <TNode extends {}>(
...
+      const moduleCode = fs.readFileSync(modulePath, 'utf-8');
+      const ast = parse(moduleCode, { sourceType: 'module' });
```
pgmanutd, line 141: Should we cache this `ast` here because we might end up reading same `modulePath` many times for eg.

**module C** is required by both **module A & B**. In this case we might end up parsing **module C** twice. 🤔

packages/babel-plugin/src/utils/ast.tsx
madou: probably. any suggestions? it's really only a problem in jira - in AF its just relative + node modules

packages/babel-plugin/src/utils/ast.tsx
madou: oh definitely - just not bothered in the spike

madou: let me write up a description of what we want to solve in the spike

packages/babel-plugin/src/utils/ast.tsx
madou: the pragmatic approach can also be "solve this later" if we need

pgmanutd: hmm..may be we can pass something as babel option. Like we have this in `eslint-plugin-import`
https://github.com/benmosher/eslint-plugin-import#resolvers

packages/babel-plugin/src/__tests__/module-traversal.test.tsx
```diff
@@ -0,0 +1,199 @@
...
+jest.mock('@compiled/utils', () => {
+  return { ...jest.requireActual('@compiled/utils'), hash: () => 'hash-test' };
+});
+
+const transform = (code: string) => {
+  return transformSync(code, {
+    configFile: false,
+    babelrc: false,
+    compact: true,
+    cwd: process.cwd(),
```
madou, line 13: we can delete this (babel defaults to `process.cwd()`

packages/babel-plugin/src/__tests__/stubs/reexport.js
```diff
@@ -0,0 +1,7 @@
+import { colors } from 'module-a';
+
+export const reexport = colors.primary;
+
+export const objectReexport = {
+  foo: colors.primary,
```
madou, line 6: nit: change this to something other than primary

madou:
add some storybook examples actually showing this working

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -241,20 +343,24 @@ export const getInterpolation = (expression: t.Expression, meta: Metadata): t.Ex
...
+
+    if (resolvedBinding && resolvedBinding.constant && t.isObjectExpression(resolvedBinding.node)) {
+      const objectValue = getValueFromObjectExpression(
+        resolvedBinding.node,
+        accessPath
+      ) as t.Expression;
+      // We recursively call get interpolation until it not longer returns an identifier or member expression
+      value = getInterpolation(objectValue, resolvedBinding.meta);
     }
   }
```
madou, line 366: add a comment on line 371? talking about potential perf pitfalls of the current recursive solution and to watch out for later - we may need to optimize this code path later.

madou:
update from jira: they don't use any "custom" absolute paths - they just rely on being a monorepo. cool - that makes things easy for us.

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -222,6 +226,107 @@ export const tryEvaluateExpression = (
...
+      constant: binding.constant,
+      meta,
+    };
+  }
+
+  if (binding && binding.path.parentPath.isImportDeclaration()) {
+    const moduleImportName = binding.path.parentPath.node.source.value;
+    const isRelative = moduleImportName.startsWith('.');
+
+    if (!meta.state.filename) {
```
pgmanutd, line 262: nit: Should we move this above a bit before `const moduleImportName = binding.path.parentPath.node.source.value;`?

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -222,6 +226,107 @@ export const tryEvaluateExpression = (
...
+      node: binding.path.node.init as t.Node,
+      path: binding.path,
+      constant: binding.constant,
+      meta,
+    };
+  }
+
+  if (binding && binding.path.parentPath.isImportDeclaration()) {
+    const moduleImportName = binding.path.parentPath.node.source.value;
+    const isRelative = moduleImportName.startsWith('.');
```
pgmanutd, line 260: What happens if file starts with `.` 🤔? Can we use something like this https://nodejs.org/api/path.html#path_path_isabsolute_path

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -222,6 +226,107 @@ export const tryEvaluateExpression = (
...
+        ...meta,
+        parentPath: foundParentPath,
+        state: {
+          ...meta.state,
+          file: ast,
+          filename,
+        },
+      },
+    };
+  }
```
pgmanutd, line 325: ❤️

madou:
good comments! fixed

madou:
@pgmanutd can you think of any other scenarios we need to handle? (other than caching)

madou:
i've added in error handling if the found CSS can't be statically evaluated as a stop gap atm - we should properly implement it (so it pulls imports into the owning module) later on - you need to kind of do that in your mixin story so that's a start :)

pgmanutd:
> i've added in error handling if the found CSS can't be statically evaluated as a stop gap atm - we should properly implement it (so it pulls imports into the owning module) later on - you need to kind of do that in your mixin story so that's a start :)
> 
> i think for now we don't add any caching - let's do it in a follow up PR so we have more focus. if you can please review @pgmanutd!

hehe no worries. Looks good atm.

pgmanutd (approved):

PR #273, opened 2020-09-08
Refactor css file locations

madou:
Moving some files around.


packages/css/src/plugins/minify.tsx
```diff
@@ -0,0 +1,20 @@
+import cssnano from 'cssnano-preset-default';
+
+/**
+ * PostCSS `cssnano` plugin to minify CSS.
+ * It deliberately excludes some plugins so it is forced to be synchronous.
+ */
+export const minify = () => {
```
pgmanutd, line 7: nit: Shouldn't this be a default export? 🤔

pgmanutd (approved):
Nice man. Loved the changes ❤️

packages/css/src/plugins/minify.tsx
madou: ah personal preference - if we use named exports/imports they're always named as the original author expected :)

Issue #272, opened 2020-09-08
Investigate baked in styles to start as atomic declarations


madou:
## Summary

1. Refactor baked in styles to start in atomic form
1. Create a tiny runtime to ensure the last atomic definition of a particular _group_ wins
1. Class names will be hashed and made as short as possible

**These changes would**

1. Enable overrides to work without needing to battle specificity wars
1. Get us a step closer to extracted atomic CSS
1. Make writing the CSS extraction plugin easier because we don't need to re-write class names again
1. Ensure overrides work across different major versions of the library

**Caveats of this change are**

1. Bundle size will go up slightly (which is offset when we end up extracting to CSS)
1. Source maps wouldn't work well anymore

## Background

For background, currently baked in styles looks like this:

```
const MyComponent = styled.div`
  font-size: 12px;
  color: blue;
`;
```

Which would generate this:

```
.cc-abc111 {
  font-size: 12px;
  color: blue;
}

<div class="cc-abc111" />
```

By itself - this works fantastically. But what happens if you want to override something? Well...

```
const WrappedMyComponent = styled(MyComponent)`
  font-size: 15px;
`;
```

Which would then generate this:

```
.cc-abc222 {
  font-size: 15px;
}

<div class="cc-abc111 cc-abc222" />
```

And then we are at the bane of specificity wars to see where the override may, or may not, be applied. This is a problem - and it's a problem for any API not just the styled API.

And what's worse is we can't fundamentally fix it with the current CSS output - we can work around it but there will be small edge cases - such as what happens when you render both the original, and the overridden component - which styles win? It will depend who renders first.

### Styles start in atomic form

Instead of creating single declarations we start with atomic declarations. Looking back at the example above, let's rewrite it with this in mind.

```
const MyComponent = styled.div`
  font-size: 12px;
  color: blue;
`;
```

Which would generate this:

```
.cc-font-size-12 {
  font-size: 12px;
}

.cc-color-blue {
  color: blue;
}

<div class="cc-font-size-12 cc-color-blue" />
```

And what does it look like if we override?

```
const WrappedMyComponent = styled(MyComponent)`
  font-size: 15px;
`;
```

Which would then generate this:

```
.cc-font-size-15 {
  font-size: 15px;
}

<div class="cc-font-size-12 cc-color-blue cc-font-size-15" />
```

Ok so then we have three class names now assigned to this component. But hopefully you see something we can do with this information...

### Runtime to handle atomic declarations

[...]

PR #274, opened 2020-09-09
[SPIKE] Atomic baked in styles

madou:
This is a spike which refactors the internals to start as atomic declarations. It's rough - and we'll probably throw it away and re-write it afterwards. For now we're just using it to gather data. #272 

It does work though!

Before (overrides cancelled):

image

After (overrides persisted):

image

## Observations

- Bundle size gets bigger
- Gzip keeps bundle size in a reasonable state
- Atomic declarations naturally solve the overriding problem

## Todo

- [x] Create a complex example
- [x] Get bundle size (mingzip) before
- [x] Get bundle size (mingzip) after
- [x] Investigate alternate hashing solutions
- [x] Investigate ensuring atomic styles within a single module
- [ ] Ordering of the runtime stylesheet to ensure pseduo selector order


pgmanutd:
So what I was saying, we can get numbers around how many unique atomic classes we have and how much bundle size they are increasing.

**For eg.** 
Suppose we have 50 unique atomic classes and 50 components are using min. 5 (and max 10) of them at the same time. This might add additional bundle size of suppose `~2KB`. And in future we might have 200-300 unique classes and 1000 of components. So it might give us rough estimate of what we can expect in future right. I know this sounds weird, but hehe 🤷‍♂️ 😛

And what if we can combine similar classes and create one for them? I know this deviates us from being atomified, but we can get smart and reduce bundlesize if we see similar classes are being used many times (above a minimum threshold) by components. And we can also let other classes being atomified if they are not used many times.

**For eg.**
```typescript
expect(actual.sheets.join('\n')).toMatchInlineSnapshot(`
      ".cc-v5rl1c-h2mmj6{position:relative}
      .cc-cy1bws-1dk0nxm{text-transform:capitalize}
      .cc-1aqc269-16l8ovn:after{content:\\"›\\"}
      .cc-p8hyfu-stnw88:after{position:absolute}
      .cc-1982bvg-lgv52v:after{right:-2rem}"
    `);
```   

will become 

```typescript
expect(actual.sheets.join('\n')).toMatchInlineSnapshot(`
      ".cc-abcde-fghijk{position:relative;text-transform:capitalize}
      .cc-abcde-fghijk:after{content:\\"›\\";position:absolute;right:-2rem}
    `);
```

madou:
Running with these settings `{ nonce: '"k0Mp1lEd"', minify: true }`:

```
npx babel examples/complex-example.tsx --out-file .tmp/compiled.js
npx babel .tmp/compiled.js --presets minify --out-file .tmp/compiled.min.js
npx gzip-size .tmp/compiled.min.js
```

## Atomic baked

```
➜  compiled git:(master) ✗ npx babel examples/complex-example.tsx --out-file .tmp/compiled.js
➜  compiled git:(master) ✗ npx babel .tmp/compiled.js --presets minify --out-file .tmp/compiled.min.js
➜  compiled git:(atomic-baked-in-styles) ✗ npx gzip-size .tmp/compiled.min.js
3.55 kB
```

## Previous baked

```
➜  compiled git:(master) ✗ npx babel examples/complex-example.tsx --out-file .tmp/compiled.js
➜  compiled git:(master) ✗ npx babel .tmp/compiled.js --presets minify --out-file .tmp/compiled.min.js
➜  compiled git:(master) ✗ npx gzip-size .tmp/compiled.min.js
2.9 kB
```

## Difference

```
0.65kB
```

madou:
Current extract spike:

```
➜  compiled git:(atomic-baked-in-styles) ✗ npx babel examples/complex-example.tsx --out-file .tmp/compiled.js
➜  compiled git:(atomic-baked-in-styles) ✗ npx babel .tmp/compiled.js --presets minify --out-file .tmp/compiled.min.js
➜  compiled git:(atomic-baked-in-styles) ✗ npx gzip-size .tmp/compiled.min.js
3.86 kB
```

It's actually bigger. Let me keep testing.

madou:
Update, after fixing some things and turning mangling on I get it a smaller bundle size (but still not as small as before):

```
➜  compiled git:(atomic-baked-in-styles) ✗ npx babel examples/complex-example.tsx --out-file .tmp/compiled.js
➜  compiled git:(atomic-baked-in-styles) ✗ npx minify .tmp/compiled.js --out-file .tmp/compiled.min.js --mangle.topLevel
➜  compiled git:(atomic-baked-in-styles) ✗ npx gzip-size .tmp/compiled.min.js
3.37 kB
```

Difference: `0.47kB`

madou:
Reducing the hash down we get this:

```
➜  compiled git:(atomic-baked-in-styles) ✗ npx babel examples/complex-example.tsx --out-file .tmp/compiled.js
➜  compiled git:(atomic-baked-in-styles) ✗ npx minify .tmp/compiled.js --out-file .tmp/compiled.min.js --mangle.topLevel
➜  compiled git:(atomic-baked-in-styles) ✗ npx gzip-size .tmp/compiled.min.js
3.33 kB
```

Difference: `0.43kB`

madou:
Reducing hashes even further we get:

```
➜  compiled git:(atomic-baked-in-styles) ✗ npx babel examples/complex-example.tsx --out-file .tmp/compiled.js
➜  compiled git:(atomic-baked-in-styles) ✗ npx minify .tmp/compiled.js --out-file .tmp/compiled.min.js --mangle.topLevel
➜  compiled git:(atomic-baked-in-styles) ✗ npx gzip-size .tmp/compiled.min.js
3.15 kB
```

Difference: `0.25kB`

I think this is in the realm of "okay". I'll push my changes up and tomorrow morning I'll write up a Confluence page with next steps.

madou:
TLDR the generated file looks like:

```
const a = ".1xum1fqe{padding:8px 12px}";
const b = ".1ylxu67f{color:#fff}";
const c = ".1k611j28{background-color:transparent}";
const d = ".pruiglyw{border:none}";
const e = ".14w9dlk8{font-size:14px}";
const f = ".1y9u1l7b{border-radius:3px}";
const g = ".ysyxz9vh:hover{background:rgba(116,118,121,0.6)}";

const Button = ({
  children
}) => /*#__PURE__*/React.createElement(_core.CC, null, /*#__PURE__*/React.createElement(_core.CS, {
  nonce: "k0Mp1lEd"
}, [a, b, c, d, e, f, g]), /*#__PURE__*/React.createElement("button", {
  className: (0, _core.ax)("1xum1fqe 1ylxu67f 1k611j28 pruiglyw 14w9dlk8 1y9u1l7b ysyxz9vh 1k9tz9vh")
}, children));
```

Multiple declarations get reused in the same file.

packages/css/src/plugins/atomicify.tsx
```diff
@@ -28,8 +28,8 @@ const normalizeSelector = (str?: string) => {
 
 const atomicifyDecl = (node: Declaration, opts: Opts = {}) => {
   const normalizedSelector = normalizeSelector(opts.parentSelector);
-  const group = hash(`${join(opts.parentAtRule, opts.parentSelector)}-${node.prop}`);
-  const className = `cc-${group}-${hash(node.value)}`;
+  const group = hash(`${join(opts.parentAtRule, opts.parentSelector)}${node.prop}`).slice(0, 4);
+  const className = `${group}${hash(node.value).slice(0, 4)}`;
```
CraigCav, line 32: It looks like this change would  break the ax runtime - which splits the class name on the `-` in order to identify the group (for dedupe/handling overridden styles).

packages/babel-plugin/src/css-prop/__tests__/object-literal.test.tsx
```diff
@@ -32,17 +32,23 @@ describe('css prop object literal', () => {
     expect(actual).toInclude('style={{"--var-hash-test":cl}}');
   });
 
-  it('should inline constant variable', () => {
+  it.only('should inline constant variable', () => {
```
pgmanutd, line 35: oops 😛

packages/babel-plugin/src/css-prop/__tests__/string-literal.test.tsx
```diff
@@ -30,7 +30,7 @@ describe('css prop string literal', () => {
     expect(actual).toInclude('style={{"--var-hash-test":(fontSize||"")+"px"}}');
   });
 
-  it('should persist suffix of constant value', () => {
+  it.only('should persist suffix of constant value', () => {
```
pgmanutd, line 33: hehe oops 😛

pgmanutd (commented):
Will review in morning. Very tired 😞. Already looked once, will review in detail tmr morning.

madou:
no rush @pgmanutd

packages/css/src/plugins/atomicify.tsx
madou: good catch! yeah you're right - we'd be changing `ax` to get the first 4 characters instead

packages/babel-plugin/src/css-prop/__tests__/string-literal.test.tsx
madou: its a spike! DW 😂

alexreardon:
I am not completely sold on this yet (i haven't thought it through)

That said - i think compiled gives us the flexibility to change our mind on this approach as we go

madou:
let's catch up today and talk through it alex

madou:
@alexreardon so I've been doing some testing this morning - gzip does quite well at reducing size when multiple duplicate atomic declarations are inside a single bundle.

**Original**

This is the current ADS example.

```bash
➜  compiled git:(atomic-baked-in-styles) npx gzip-size .tmp/compiled.min.js
3.16 kB
```

**Duplicated declarations**

Takes the ADS example and duplicates the declarations.

```bash
➜  compiled git:(atomic-baked-in-styles) npx gzip-size .tmp/compiled.min.js
3.13 kB
```

_It's smaller_. This makes sense because of how gzip can optimize duplicated lengths of strings in a single file. I have to wonder if we took this to an entire app what would this look like.

---

When we talk around the alternate strategy of compiling overrides, so for example:

```
const StyledDiv = styled.div`
  color: red;
  font-size: 12px;
`;

const WrappedStyled = styled(StyledDiv)`
  color: blue;
`;
```

Which would then produce these styles for wrapped style:

```
.hash { color: blue; font-size: 12px }
```

It does paint a good story for using the less boilerplate strings compared to the atomic declarations approach. But - where can this break expectations, or make things harder for us?

1. Overriding styles on wrapping components
1. Overriding already baked components 
1. Conditionally applying styles

**Overriding styles on wrapping components**

This is more insidious when we consider this would naturally happen when using CSS prop, but it can happen for all APIs. 

```jsx
const Component = ({ className }) => <div css={{ fontSize: 12, className }} />
export const WrappedComponent = (props) => <Component {...props} />;
```
```jsx
import WrappedComponent from './component';

styled(WrappedComponent)`
  font-size: 20px;
`;
```

We now need to keep traversing until we _might_ find a component. It's a lot more traversal.

**Overriding already baked components** 

When considering consuming packages from NPM - they're already baked. We will need to traverse already transpiled code and try to analyze the defined styles. This will be more code path to handle these scenarios.

**Conditionally applying styles**

Take this code.

```jsx
const StyledDiv = styled.div`
  color: red;
  font-size: 12px;
`;

const WrappedStyled = () => {
  const [isBig] = useState(false);
  return <StyledDiv css={[isBig && { fontSize: 20 }]} />
};
```

Fundamentally this code can't work. We wouldn't be able to support this API.

---

Atomic declarations solve this problem quite naturally - with the biggest bonus being all the above cases now work as expected. But - as we've seen - at the cost of bundle size increasing. I think we need more data to understand the true ramifications of this - however:

1. gzipping the example with duplicated declarations has shown to be quite effective
1. the bundle size increase will be made irrelevant when we extract to CSS

Finally I don't think we should underestimate the benefits of _behaviour_ being the same when baked and extracted. We will inevitably find bugs or fundamental blockers in our implementation if they diverge (like this spike has been aiming to solve 😄).

We have flexibility and potential to optimize the further we go on - if we go down this path we can continually evaluate and adjust as needed.

packages/css/src/plugins/atomicify.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { hash } from '@compiled/utils';
+
+interface Opts {
+  parentSelector?: string;
+  parentAtRule?: string;
+  callback?: (className: string) => void;
+}
+
+const join = (...str: (string | undefined)[]) => {
+  if (!str) {
```
pgmanutd, line 11: nit: Will this be undefined at any point of time since its a rest parameter.

packages/runtime/src/__tests__/ax.test.tsx
```diff
@@ -0,0 +1,15 @@
+import ax from '../ax';
+
+describe('ax', () => {
+  it('should remove duplicates', () => {
```
pgmanutd, line 4: ahh! I was looking for this duplicate one. Nice ❤️

packages/runtime/src/ax.tsx
```diff
@@ -0,0 +1,11 @@
+export default function ax(className: string) {
+  const found: any = {};
+
+  className.split(' ').forEach((cls) => {
+    const index = cls.charAt(0) === '_' ? 5 : 4;
```
pgmanutd, line 5: nit: 😛 5 and 4 looks like magic numbers.

pgmanutd (commented):
@madou: you can ignore nits if we are not merging this into master. Anyways this is just a spike right. nice work ❤️.

madou:
yeah exactly we will close this and break it down into smaller discreet chunks

- PostCSS atomic plugin
- atomic ax runtime + update ast builder to use ax runtime
- update style component to use not add based on stylesheets instead of hash
- update ast builder to use ax runtime
- hoist style declarations to the top of the module

madou:
Spike completed. Follow up tasks to follow.

PR #282, opened 2020-09-14
fix: fixes evaluation blowing up when referencing local func args

### LINKED ISSUE #281: Babel compilation error when css prop values refer to function arguments
**Describe the bug**
The babel plugin fails to compile when a css prop contains a reference to a function argument, either destructured or primitive arguments. 

**To Reproduce**

```
import React from 'react';

import '@compiled/core';

export const DestructuredComp = ({ foo }) => {
  return (
    <>
      <span css={{ fontSize: foo }} />
    </>
  );
};
```

**Expected behaviour**
No errors

**Actual behaviour**

```
RangeError: /.../packages/design-system/lozenge/src/Lozenge/Test.tsx: Maximum call stack size exceeded
    at Object.exports.getInterpolation (/.../node_modules/@compiled/babel-plugin/dist/utils/ast.js:318:11)
    at Object.exports.getInterpolation (/.../node_modules/@compiled/babel-plugin/dist/utils/ast.js:323:29)
    at Object.exports.getInterpolation (/.../node_modules/@compiled/babel-plugin/dist/utils/ast.js:323:29)
    at Object.exports.getInterpolation (/.../node_modules/@compiled/babel-plugin/dist/utils/ast.js:323:29)
    at Object.exports.getInterpolation (/.../node_modules/@compiled/babel-plugin/dist/utils/ast.js:323:29)
    at Object.exports.getInterpolation (/.../node_modules/@compiled/babel-plugin/dist/utils/ast.js:323:29)
    at Object.exports.getInterpolation (/.../node_modules/@compiled/babel-plugin/dist/utils/ast.js:323:29)
    at Object.exports.getInterpolation (/.../node_modules/@compiled/babel-plugin/dist/utils/ast.js:323:29)
    at Object.exports.getInterpolation (/.../node_modules/@compiled/babel-plugin/dist/utils/ast.js:323:29)
    at Object.exports.getInterpolation (/.../node_modules/@compiled/babel-plugin/dist/utils/ast.js:323:29) {
  code: 'BABEL_TRANSFORM_ERROR'
}
```

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

madou:
Closes #281

### DIFF
--- a/packages/babel-plugin/src/utils/ast.tsx
+++ b/packages/babel-plugin/src/utils/ast.tsx
@@ -260,7 +260,9 @@ export const resolveBindingNode = (
   binding: Binding | undefined,
   meta: Metadata
 ): PartialBindingWithMeta | undefined => {
-  if (!binding) {
+  if (!binding || binding.path.isObjectPattern()) {
+    // Bail early if there is no binding or its a node that we don't want to resolve
+    // such as an destructured args from a function.
     return undefined;
   }
 
--- a/packages/babel-plugin/src/utils/ast.tsx
+++ b/packages/babel-plugin/src/utils/ast.tsx
@@ -268,11 +268,11 @@ export const resolveBindingNode = (
 
   if (t.isVariableDeclarator(binding.path.node)) {
     return {
+      meta,
       node: binding.path.node.init as t.Node,
       path: binding.path,
       constant: binding.constant,
       source: 'module',
-      meta,
     };
   }
 
@@ -374,6 +374,10 @@ export const getInterpolation = (expression: t.Expression, meta: Metadata): t.Ex
   if (t.isIdentifier(expression)) {
     const binding = meta.parentPath.scope.getBinding(expression.name);
     const resolvedBinding = resolveBindingNode(binding, meta);
+    if (binding?.path.node === expression) {
+      // We resolved to the same node - bail out!
+      return expression;
+    }
 
     if (resolvedBinding && resolvedBinding.constant) {
       // We recursively call get interpolation until it not longer returns an identifier or member expression
--- a/packages/babel-plugin/src/__tests__/expression-evaluation.test.tsx
+++ b/packages/babel-plugin/src/__tests__/expression-evaluation.test.tsx
@@ -109,4 +109,76 @@ describe('import specifiers', () => {
 
     expect(actual).toInclude('.cc-hash-test{font-size:var(--var-hash-test)}');
   });
+
+  it('should not blow up when referencing local destructured args in arrow func', () => {
+    expect(() => {
+      transform(
+        `
+        import '@compiled/core';
+
+        export const DestructuredComp = ({ foo, color }) => {
+          return (
+            <>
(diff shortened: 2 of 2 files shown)


packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -260,7 +260,9 @@ export const resolveBindingNode = (
   binding: Binding | undefined,
   meta: Metadata
 ): PartialBindingWithMeta | undefined => {
-  if (!binding) {
+  if (!binding || binding.path.isObjectPattern()) {
```
pgmanutd, line 263: 🤔 Can there be more use cases like this? Just thinking 🤔 If yes, then this **or** condition will go up 😞

madou: most likely tbh - but im sure its a finite list

pgmanutd: > most likely tbh - but im sure its a finite list
> 
> can we think of any other scenarios that we've missed (when it comes to bindings referencing something in scope)? i can't think of any.

I just pulled the branch. There are 2 things:

1. It's still throwing "Maximum call stack size exceeded" error for this use case 😞
```typescript
import '@compiled/core';

export const DestructuredComp = (props) => {
  const { foo, color } = props;

  return (
    <>
      <span css={{ fontSize: foo, color, backgroundColor: 'blue' }} />
    </>
  );
};
```
2. We are skipping the resolving part right, but what if its coming from a variable outside the function 🤔? May be from a variable declarator or import. Should we also skip in that case 🤔? Something like this:

```typescript
import '@compiled/core';

const { foo } = { foo: 10 };

export const DestructuredComp = ({ color }) => {
  return (
    <>
      <span css={{ fontSize: foo, color, backgroundColor: 'blue' }} />
    </>
  );
};
```

madou: awesome insight man, i'll do some testing this morning and fix anything that breaks. we should still be supporting proper de-structuring - perhaps we don't right now?

good catch for the maximum call stack exceed - i'll write a text

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -372,6 +374,10 @@ export const getInterpolation = (expression: t.Expression, meta: Metadata): t.Ex
   if (t.isIdentifier(expression)) {
     const binding = meta.parentPath.scope.getBinding(expression.name);
     const resolvedBinding = resolveBindingNode(binding, meta);
+    if (binding?.path.node === expression) {
```
pgmanutd, line 377: hehe 😛. Breaking the circular cycle with reference check. Nice.

pgmanutd (approved):

packages/babel-plugin/src/utils/ast.tsx
madou: yeah, elegant!

Issue #286, opened 2020-09-17
PostCSS atomic plugin


madou:
This plugin should take a CSS declaration and outut it as atomic declarations.

**Input**

```css
display: block;
color: blue;
```

**Output**

```css
.{group}{value} {
  display: block;
}
.{group}{value} {
  color: blue;
}
```

Look at the spike implementation here: https://github.com/atlassian-labs/compiled/pull/274/files#diff-3d0e3948c146d9c40c8c173d78fda7e0

Each declaration should be made up of two hashes, a group, and a value, each 4 characters long.

```
.{grouphash}{valuehash}
```

We’re going to use the group as a unique identifier to make overrides work correctly.

A group is defined as:

```
{atrules}{selectors}{propertyname}
```

A value is defined as:

```
{value}
```

**At rules**

Ensure they're combined.

```css
display: block;

@media (min-width: 500px) {
  display: flex;
  color: red;
}
```

```css
.{group}{value} { display: block }
@media (min-width: 500px) {
  .{group}{value} { display: flex }
  .{group}{value} { color: red }
}
```

**Autoprefixing**

Properties that need autoprefixing shouldn't create duplicate atomic entries.

```css
display: grid;
```

```css
.{group}{value} {
  display:-ms-grid;
  display:grid
}
```

**Duplicate declarations**

We should remove any fallback values and lean on autoprefixer from above to do the work if its needed.

```css
display: block;
display: flex;
```

```css
.{group}{value} {
  display:flex
}
```

**Testing notes**

We need to make sure all kinds of combinations of selectors work.
 
- [x] top level property
- [x] properties in nested selectors
- [x] top level property inside an at rule
- [x] properties in nested selectors in an at rule
- [x] should callback with all created class names to be used later on
- [x] ensure if any property that needs prefixing has it in the atomic declaration

## Follow up tasks

We need to keep this ticket in a reasonable size.

**Shorthands**

Shorthand properties need to be blown up. For a list see: https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties. This should be its own PostCSS plugin.

```diff
-background: black;
+background-color: black;
```

Q: This means that when values are omitted should we generate their initial values as per spec? Probably. But we could also not.

**Nested at rules**

We also need to support nested at rules, see: https://css-tricks.com/can-you-nest-media-and-support-queries/. We'll implement it later.

TxHawks:
[...]

PR #291, opened 2020-09-18
Add ax runtime

madou:
Adds the ax runtime. This still works without the atomic CSS implementation - examples still work as expected.

### DIFF
--- a/packages/babel-plugin/src/utils/ast-builders.tsx
+++ b/packages/babel-plugin/src/utils/ast-builders.tsx
@@ -326,11 +326,11 @@ export const buildCompiledComponent = (opts: CompiledOpts) => {
       ? classNameProp.value.expression
       : classNameProp.value;
 
-    const newClassNameValue = t.isStringLiteral(classNameExpression)
-      ? joinExpressions(t.stringLiteral(className), classNameExpression)
-      : conditionallyJoinExpressions(t.stringLiteral(className), classNameExpression);
-
-    classNameProp.value = t.jsxExpressionContainer(newClassNameValue);
+    classNameProp.value = t.jsxExpressionContainer(
+      t.callExpression(t.identifier('ax'), [
+        t.arrayExpression([t.stringLiteral(className), classNameExpression as t.Expression]),
+      ])
+    );
   } else {
     // No class name - just push our own one.
     opts.node.openingElement.attributes.push(
--- a/packages/babel-plugin/src/utils/ast-builders.tsx
+++ b/packages/babel-plugin/src/utils/ast-builders.tsx
@@ -197,9 +197,12 @@ const styledTemplate = (opts: {
   }, ref) => (
     <CC>
       <CS ${nonceAttribute} hash="${opts.hash}">{%%cssNode%%}</CS>
-      <C {...props} style={%%styleProp%%} ref={ref} className={"${
-        opts.className
-      }" + (props.className ? " " + props.className : "")} />
+      <C
+        {...props}
+        style={%%styleProp%%}
+        ref={ref}
+        className={ax(["${opts.className}", props.className])}
+      />
     </CC>
   ));
 `,
--- a/packages/runtime/src/ax.tsx
+++ b/packages/runtime/src/ax.tsx
@@ -1,9 +1,9 @@
 /**
  * Joins classes together and ensures atomic declarations of a single group exist.
- * This function is made to be fast and small - so it takes some liberties with its API.
- *
- * Atomic declarations take the form of `_{group}{value}`,
+ * Atomic declarations take the form of `_{group}{value}` (always prefixed with an underscore),
  * where both `group` and `value` are hashes **four characters long**.
+ * Class names can be of any length,
+ * this function can take both atomic declarations and class names.
  *
  * Input:
  *
--- a/packages/runtime/src/ax.tsx
+++ b/packages/runtime/src/ax.tsx
@@ -1,0 +1,36 @@
+/**
+ * Joins classes together and ensures atomic declarations of a single group exist.
+ * This function is made to be fast and small - so it takes some liberties with its API.
+ *
+ * Atomic declarations take the form of `_{group}{value}`,
+ * where both `group` and `value` are hashes **four characters long**.
+ *
+ * Input:
+ *
+ * ```
+ * ax(['_aaaabbbb', '_aaaacccc'])
+ * ```
+ *
+ * Output:
+ *
+ * ```
+ * '_aaaacccc'
+ * ```
+ *
+ * @param classes
+ */
+export const ax = (classes: (string | undefined | false)[]): string => {
+  const found: Record<string, string> = {};
+
+  for (let i = 0; i < classes.length; i++) {
+    const cls = classes[i];
+    if (!cls) {
+      continue;
+    }
+
+    const group = cls.slice(0, cls.charCodeAt(0) === 95 ? 5 : undefined);
+    found[group] = cls;
+  }
+
+  return Object.values(found).join(' ');
+};
--- a/packages/babel-plugin/src/__tests__/index.test.tsx
+++ b/packages/babel-plugin/src/__tests__/index.test.tsx
@@ -33,14 +33,14 @@ describe('babel plugin', () => {
 
     expect(output?.code).toMatchInlineSnapshot(`
       "import * as React from 'react';
-      import { CC, CS } from '@compiled/core';
+      import { ax, CC, CS } from '@compiled/core';
       const MyDiv = React.forwardRef(({
         as: C = \\"div\\",
         style,
         ...props
       }, ref) => <CC>
             <CS hash=\\"1x3e11p\\">{[\\".cc-1x3e11p{font-size:12px}\\"]}</CS>
-            <C {...props} style={style} ref={ref} className={\\"cc-1x3e11p\\" + (props.className ? \\" \\" + props.className : \\"\\")} />
+            <C {...props} style={style} ref={ref} className={ax([\\"cc-1x3e11p\\", props.className])} />
           </CC>);"
     `);
   });
@@ -59,7 +59,7 @@ describe('babel plugin', () => {
 
     expect(output?.code).toMatchInlineSnapshot(`
       "import * as React from 'react';
-      import { CC, CS } from '@compiled/core';
+      import { ax, CC, CS } from '@compiled/core';
 
       const MyDiv = () => {
         return <CC>
--- a/packages/babel-plugin/src/__tests__/module-imports.test.tsx
+++ b/packages/babel-plugin/src/__tests__/module-imports.test.tsx
@@ -71,9 +71,9 @@ describe('import specifiers', () => {
     `);
 
     expect(actual).toMatchInlineSnapshot(`
-      "import*as React from'react';import{ThemeProvider,CC,CS}from'@compiled/core';const ListItem=React.forwardRef(({as:C=\\"div\\",style,...props},ref)=><CC>
+      "import*as React from'react';import{ThemeProvider,ax,CC,CS}from'@compiled/core';const ListItem=React.forwardRef(({as:C=\\"div\\",style,...props},ref)=><CC>
             <CS hash=\\"hash-test\\">{[\\".cc-hash-test{font-size:20px}\\"]}</CS>
-            <C{...props}style={style}ref={ref}className={\\"cc-hash-test\\"+(props.className?\\" \\"+props.className:\\"\\")}/>
+            <C{...props}style={style}ref={ref}className={ax([\\"cc-hash-test\\",props.className])}/>
           </CC>);"
     `);
   });
--- a/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
+++ b/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
@@ -119,7 +119,7 @@ describe('css prop behaviour', () => {
       <div className="foobar" css={{}}>hello world</div>
     `);
 
-    expect(actual).toInclude('className={"cc-hash-test"+(" "+"foobar")}');
+    expect(actual).toInclude('className={ax(["cc-hash-test","foobar"])}');
   });
 
   it('should pass through spread props', () => {
@@ -165,7 +165,7 @@ describe('css prop behaviour', () => {
       <div className={className} css={{}}>hello world</div>
     `);
 
-    expect(actual).toInclude('className={"cc-hash-test"+(className?" "+className:"")}');
+    expect(actual).toInclude('className={ax(["cc-hash-test",className])}');
   });
 
   it('should pick up array composition', () => {
@@ -219,7 +219,7 @@ describe('css prop behaviour', () => {
       <div css={{}} className={getFoo()}>hello world</div>
     `);
 
-    expect(actual).toInclude('className={"cc-hash-test"+(getFoo()?" "+getFoo():"")}');
+    expect(actual).toInclude('className={ax(["cc-hash-test",getFoo()])}');
   });
 
   it('should allow inlined expressions as property values', () => {
--- a/packages/babel-plugin/src/index.tsx
+++ b/packages/babel-plugin/src/index.tsx
@@ -45,7 +45,7 @@ export default declare<State>((api) => {
             return true;
           })
           // Add on the util imports we're going to use later in the transform.
-          .concat([importSpecifier('CC'), importSpecifier('CS')]);
+          .concat([importSpecifier('ax'), importSpecifier('CC'), importSpecifier('CS')]);
       },
       TaggedTemplateExpression(path, state) {
         if (!state.compiledImports?.styled) {
--- a/packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
+++ b/packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
@@ -25,9 +25,9 @@ describe('styled component behaviour', () => {
     `);
 
     expect(actual).toMatchInlineSnapshot(`
-      "import*as React from'react';import{ThemeProvider,CC,CS}from'@compiled/core';const ListItem=React.forwardRef(({as:C=\\"div\\",style,...props},ref)=><CC>
+      "import*as React from'react';import{ThemeProvider,ax,CC,CS}from'@compiled/core';const ListItem=React.forwardRef(({as:C=\\"div\\",style,...props},ref)=><CC>
             <CS hash=\\"hash-test\\">{[\\".cc-hash-test{font-size:20px}\\"]}</CS>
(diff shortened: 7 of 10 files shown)


packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -326,11 +329,11 @@ export const buildCompiledComponent = (opts: CompiledOpts) => {
       ? classNameProp.value.expression
       : classNameProp.value;
 
-    const newClassNameValue = t.isStringLiteral(classNameExpression)
-      ? joinExpressions(t.stringLiteral(className), classNameExpression)
-      : conditionallyJoinExpressions(t.stringLiteral(className), classNameExpression);
-
-    classNameProp.value = t.jsxExpressionContainer(newClassNameValue);
+    classNameProp.value = t.jsxExpressionContainer(
```
madou, line 332: this means that we conditionally apply `ax` only when we **know** it can accept outside class names.

packages/runtime/src/ax.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * Output:
+ *
+ * ```
+ * '_aaaacccc'
+ * ```
+ *
+ * @param classes
+ */
+export const ax = (classes: (string | undefined | false)[]): string => {
+  const found: Record<string, string> = {};
```
pgmanutd, line 23: nit: Should we use `Map` instead of object? 🤔

packages/runtime/src/ax.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  for (let i = 0; i < classes.length; i++) {
+    const cls = classes[i];
+    if (!cls) {
+      continue;
+    }
+
+    const group = cls.slice(0, cls.charCodeAt(0) === 95 ? 5 : undefined);
```
pgmanutd, line 31: `95` is underscore right? Got confused at first 😅

pgmanutd: Are we assuming `cls ` should always have length of `8`?

packages/runtime/src/ax.tsx
pgmanutd: `Object.values(found).join(' ')` will become `found.values().join(' ')`

packages/runtime/src/ax.tsx
madou: no it can be more - we just make assumptions that we'll take a slice of it if it starts with underscore

pgmanutd: >  * Atomic declarations take the form of `_{group}{value}`, * where both `group` and `value` are hashes **four characters long**.

This gave me that impression. No worries 😄

madou: ah maybe we can rephrase

packages/runtime/src/ax.tsx
madou: let me run a quick test

packages/babel-plugin/src/utils/ast-builders.tsx
madou: @pgmanutd i have to wonder if we should keep it though

if we don't what happens if someone defines a declaration like:

```
display: block;
display: flex;
```

We'd have to remove the block declaration - either at runtime or build. Builds possible though.

packages/runtime/src/ax.tsx
madou: map values an iterable not an array

madou: map is much slower 

https://perf.link/#eyJpZCI6ImRka2pxM3B2Y2ZrIiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImNvbnN0IGF4MSA9IChjbGFzc2VzKSA9PiB7XG4gIGNvbnN0IGZvdW5kID0ge307XG5cbiAgZm9yIChsZXQgaSA9IDA7IGkgPCBjbGFzc2VzLmxlbmd0aDsgaSsrKSB7XG4gICAgY29uc3QgY2xzID0gY2xhc3Nlc1tpXTtcbiAgICBpZiAoIWNscykge1xuICAgICAgY29udGludWU7XG4gICAgfVxuXG4gICAgY29uc3QgZ3JvdXAgPSBjbHMuc2xpY2UoMCwgY2xzLmNoYXJDb2RlQXQoMCkgPT09IDk1ID8gNSA6IHVuZGVmaW5lZCk7XG4gICAgZm91bmRbZ3JvdXBdID0gY2xzO1xuICB9XG5cbiAgcmV0dXJuIE9iamVjdC52YWx1ZXMoZm91bmQpLmpvaW4oJyAnKTtcbn07XG5cbmNvbnN0IGF4MiA9IChjbGFzc2VzKSA9PiB7XG4gIGNvbnN0IGZvdW5kID0gbmV3IE1hcCgpO1xuXG4gIGZvciAobGV0IGkgPSAwOyBpIDwgY2xhc3Nlcy5sZW5ndGg7IGkrKykge1xuICAgIGNvbnN0IGNscyA9IGNsYXNzZXNbaV07XG4gICAgaWYgKCFjbHMpIHtcbiAgICAgIGNvbnRpbnVlO1xuICAgIH1cblxuICAgIGNvbnN0IGdyb3VwID0gY2xzLnNsaWNlKDAsIGNscy5jaGFyQ29kZUF0KDApID09PSA5NSA%2FIDUgOiB1bmRlZmluZWQpO1xuICAgIGZvdW5kLnNldChncm91cCwgY2xzKTtcbiAgfVxuXG4gIHJldHVybiBBcnJheS5mcm9tKGZvdW5kLnZhbHVlcygpKS5qb2luKCcgJyk7XG59OyIsInRlc3RzIjpbeyJuYW1lIjoiVGVzdCBDYXNlIiwiY29kZSI6ImF4MShbJ2ZvbycsICdfMTIzMTIzJywgJ18xMjMxMjMnLCAnXzIzMjMyMzIzJ10pOyIsInJ1bnMiOlsxMDAwLDEwMDAsMTcwMDAsMjAwMCwxMDAwLDEzNDAwMCwxODAwMCwyMDAwLDMwMDAsMTAwMCw1OTAwMCwyMDAwLDIwMDAsMTU4MDAwLDMwODAwMCwyMTAwMCwxMDAwMCwyMDAwLDcwMDAsOTgwMDAsMTAwMCwwLDIwMDAsMTkyMDAwLDEwNzAwMCwyMDAwLDEwMDAsMjAwMCwyMDAwLDc4MDAwLDIwMDAsMjg2MDAwLDEwMDAwLDkwMDAsMTA5MDAwLDIyNDAwMCwxMDAwLDY2MDAwLDEwMDAwLDMwMDAsMzAwMCwzMDAwLDIwMDAsMzAwMCwxNjgwMDAsMzAwMCwzMDAwLDIwMDAsMzAwMCwzMTAwMCwzMDAwLDQwMDAsMTcwMDAsMzAwMCwzMDAwLDYwMDAsMjAwMCwyMDAwLDE2ODAwMCwxMDAwMCwyMDAwLDMwMDAsMTAwMDAsMjAwMCwwLDcxMDAwLDQwMDAsMzAwMCwzNDAwMCwyMDAwLDEwMDAwLDY3MDAwLDE2MDAwLDMwMDAsMjAwMCwyMDAwLDIwMDAsMjAwMCwzMDAwLDEwMDAsMjAwMCwzMDAwLDIwMDAsMjAwMCwyMDAwLDMwMDAsMjAwMCwyNTAwMCwyMDAwLDMwMDAsMzAwMCwzMDAwLDMwMDAsMjAwMCwyMDAwLDIwMDAsMzAwMCwxNTUwMDAsMTY4MDAwLDIwMDBdLCJvcHMiOjMwMjEwfSx7Im5hbWUiOiJUZXN0IENhc2UiLCJjb2RlIjoiYXgyKFsnZm9vJywgJ18xMjMxMjMnLCAnXzEyMzEyMycsICdfMjMyMzIzMjMnXSk7IiwicnVucyI6WzEwMDAsMTAwMCwzMDAwLDIwMDAsMTAwMCwxMDAwMCwzMDAwLDIwMDAsMzAwMCwxMDAwLDQwMDAsMjAwMCwyMDAwLDEwMDAwLDQyMDAwLDQwMDAsMzAwMCwyMDAwLDMwMDAsMTEwMDAsMTAwMCwwLDEwMDAsMjAwMDAsMTAwMDAsMjAwMCwxMDAwLDIwMDAsMjAwMCw1MDAwLDEwMDAsNTcwMDAsMzAwMCwzMDAwLDkwMDAsMjUwMDAsMTAwMCw0MDAwLDMwMDAsMjAwMCwzMDAwLDIwMDAsMjAwMCwyMDAwLDE2MDAwLDIwMDAsMzAwMCwyMDAwLDMwMDAsMzAwMCwzMDAwLDMwMDAsMzAwMCwyMDAwLDIwMDAsMzAwMCwyMDAwLDIwMDAsMTcwMDAsMzAwMCwyMDAwLDIwMDAsMzAwMCwyMDAwLDAsMzAwMCwzMDAwLDIwMDAsMzAwMCwxMDAwLDMwMDAsNTAwMCwzMDAwLDIwMDAsMjAwMCwyMDAwLDIwMDAsMjAwMCwyMDAwLDEwMDAsMTAwMCwyMDAwLDEwMDAsMjAwMCwyMDAwLDMwMDAsMTAwMCwzMDAwLDEwMDAsMzAwMCwyMDAwLDMwMDAsMzAwMCwxMDAwLDIwMDAsMTAwMCwzMDAwLDEyMDAwLDE3MDAwLDIwMDBdLCJvcHMiOjQ1MDB9XSwidXBkYXRlZCI6IjIwMjAtMDktMThUMDc6NDI6NDQuOTc5WiJ9

pgmanutd: Yeah 😅 hehe.`values()` will be an iterable. Forgot. hehe. But I just checked. If we remove `join('  ')` then `Map` is faster in most of the cases.

https://perf.link/#eyJpZCI6ImRka2pxM3B2Y2ZrIiwidGl0bGUiOiJGaW5kaW5nIG51bWJlcnMgaW4gYW4gYXJyYXkgb2YgMTAwMCIsImJlZm9yZSI6ImNvbnN0IGF4MSA9IChjbGFzc2VzKSA9PiB7XG4gIGNvbnN0IGZvdW5kID0ge307XG5cbiAgZm9yIChsZXQgaSA9IDA7IGkgPCBjbGFzc2VzLmxlbmd0aDsgaSsrKSB7XG4gICAgY29uc3QgY2xzID0gY2xhc3Nlc1tpXTtcbiAgICBpZiAoIWNscykge1xuICAgICAgY29udGludWU7XG4gICAgfVxuXG4gICAgY29uc3QgZ3JvdXAgPSBjbHMuc2xpY2UoMCwgY2xzLmNoYXJDb2RlQXQoMCkgPT09IDk1ID8gNSA6IHVuZGVmaW5lZCk7XG4gICAgZm91bmRbZ3JvdXBdID0gY2xzO1xuICB9XG5cbiAgcmV0dXJuIE9iamVjdC52YWx1ZXMoZm91bmQpO1xufTtcblxuY29uc3QgYXgyID0gKGNsYXNzZXMpID0%2BIHtcbiAgY29uc3QgZm91bmQgPSBuZXcgTWFwKCk7XG5cbiAgZm9yIChsZXQgaSA9IDA7IGkgPCBjbGFzc2VzLmxlbmd0aDsgaSsrKSB7XG4gICAgY29uc3QgY2xzID0gY2xhc3Nlc1tpXTtcbiAgICBpZiAoIWNscykge1xuICAgICAgY29udGludWU7XG4gICAgfVxuXG4gICAgY29uc3QgZ3JvdXAgPSBjbHMuc2xpY2UoMCwgY2xzLmNoYXJDb2RlQXQoMCkgPT09IDk1ID8gNSA6IHVuZGVmaW5lZCk7XG4gICAgZm91bmQuc2V0KGdyb3VwLCBjbHMpO1xuICB9XG5cbiAgcmV0dXJuIGZvdW5kLnZhbHVlcygpO1xufTsiLCJ0ZXN0cyI6W3sibmFtZSI6IlRlc3QgQ2FzZSIsImNvZGUiOiJheDEoWydmb28nLCAnXzEyMzEyMycsICdfMTIzMTIzJywgJ18yMzIzMjMyMyddKTsiLCJydW5zIjpbMjk3MDAwLDM2MjAwMCw0NDMwMDAsMzQ2MDAwLDIxNjAwMCwxNzQwMDAsNDA0MDAwLDIwNTAwMCw0ODkwMDAsMjc3MDAwLDIwMTAwMCwxOTQwMDAsNTk4MDAwLDg0MDAwLDQ5OTAwMCw5MDAwLDYyNjAwMCwxNjUwMDAsMTEwMDAsMTAwMCwzNjkwMDAsMTgwMDAwLDEwMDAsMTcxMDAwLDYwMzAwMCwzNDgwMDAsMzY5MDAwLDI0OTAwMCw3MDAwLDY1MDAwMCwyNTEwMDAsNjQzMDAwLDU3NzAwMCw0NzkwMDAsMzI4MDAwLDY4MDAwMCwyODYwMDAsNjAwMDAsMTM1MDAwLDQ0NzAwMCwxMzkwMDAsMjcwMDAsNDQwMDAwLDE5NzAwMCwxMDMwMDAsNDE2MDAwLDkzMDAwLDIzMDAwMCw1MjYwMDAsMzYzMDAwLDE2NzAwMCwyNTkwMDAsNDAxMDAwLDYxNDAwMCwyNjQwMDAsMzc1MDAwLDE0MjAwMCwxMzkwMDAsMTcwMDAsODgwMDAsNDY3MDAwLDQwNzAwMCwxNzgwMDAsMzQyMDAwLDM0MDAwLDQ4MzAwMCwyODUwMDAsMjE0MDAwLDExNDAwMCwyMDAwLDg0MDAwLDE5MTAwMCw0MTIwMDAsMjA1MDAwLDkwMDAwLDM1NjAwMCw1NDQwMDAsNzAzMDAwLDQ1ODAwMCwzODYwMDAsMTAwMCw0MDIwMDAsMzAwMCwzNzAwMCw2MDAwMCwxMzkwMDAsNDQ5MDAwLDI0OTAwMCw0MjIwMDAsNjk4MDAwLDQ0NzAwMCwyNzQwMDAsOTkwMDAsNTk4MDAwLDEwMDAsMTc3MDAwLDQ4ODAwMCw1NTUwMDAsNTQyMDAwLDI2NzAwMF0sIm9wcyI6MjkyOTcwfSx7Im5hbWUiOiJUZXN0IENhc2UiLCJjb2RlIjoiYXgyKFsnZm9vJywgJ18xMjMxMjMnLCAnXzEyMzEyMycsICdfMjMyMzIzMjMnXSk7IiwicnVucyI6WzE1MTAwMCwyNzAwMCwyNDgwMDAsMzU5MDAwLDI1OTAwMCw2NDAwMDAsMTAwMCw0MTAwMDAsNDkxMDAwLDUzODAwMCwzMTgwMDAsNDE4MDAwLDM3MDAwLDg0MDAwLDc0MDAwLDMwODAwMCw3MTIwMDAsMzUxMDAwLDc4MzAwMCw1MzgwMDAsNTk4MDAwLDE5ODAwMCwyMzQwMDAsODAzMDAwLDQ0NzAwMCwyOTUwMDAsMTkwMDAwLDU4MTAwMCw3MzMwMDAsMjA2MDAwLDY1NzAwMCw4MDAwLDUzNzAwMCw2NTAwMCwxNzgwMDAsMjA0MDAwLDQ1NDAwMCwzMDUwMDAsMjU2MDAwLDU0MTAwMCwyMTAwMCwzMDAwMCw0OTgwMDAsNjQxMDAwLDE4OTAwMCw1OTIwMDAsMzc4MDAwLDM3MTAwMCw0MjQwMDAsMTQ3MDAwLDM1OTAwMCw2NDgwMDAsMzk2MDAwLDY1NjAwMCw2NzcwMDAsMjk4MDAwLDE0NjAwMCw3MjQwMDAsMTAwMCwyODAwMCwxNjcwMDAsMTA4MDAwLDY1MTAwMCwzMjAwMCw1MzcwMDAsNTYxMDAwLDgwMDAsNTg2MDAwLDQ1NjAwMCw0ODgwMDAsNTA0MDAwLDc5MDAwLDU3MDAwLDE2MDAwMCw0OTAwMCwxNzgwMDAsMjU1MDAwLDU3MDAwMCwxMDcwMDAsNDQ0MDAwLDE1NDAwMCw0MDAwLDYzMjAwMCwzNDAwMCwzNDQwMDAsNDMwMDAsMzUwMDAwLDE2NzAwMCwyOTcwMDAsNzM2MDAwLDE4NDAwMCwxMjQwMDAsNTAwMCw2OTMwMDAsNjc3MDAwLDIyNjAwMCw2MTEwMDAsMjQ5MDAwLDkzMDAwLDY5OTAwMF0sIm9wcyI6MzM3ODAwfV0sInVwZGF0ZWQiOiIyMDIwLTA5LTE4VDA4OjA0OjEwLjQ1NVoifQ%3D%3D

pgmanutd: I think `Map` is only preferred when keys order matters. Coz it maintains the order of keys which is costly.

packages/babel-plugin/src/utils/ast-builders.tsx
pgmanutd: 🤔

packages/runtime/src/ax.tsx
pgmanutd: If performance matters, probably object is preferred. `Map` will be slower because of maintaining insertion order (+ increase in space complexity) and `.get()` increases a function call.

madou: perf will matter here because its going to be a hot path

order only matters on input - output it can be in any order 😄

madou: will take any suggestions to improve perf!!

packages/babel-plugin/src/utils/ast-builders.tsx
madou: i updaated #286 with this info (see duplicate declarations)

packages/runtime/src/ax.tsx
pgmanutd: hehe it's completely fine. If perf matters lets go with object. Coz insertion doesn't matter in this case. 🎉 😄

packages/babel-plugin/src/utils/ast-builders.tsx
pgmanutd: cool. got it. 

> let the last declaration win and then let auto prefixer take over

^^ works mate

packages/runtime/src/ax.tsx
pgmanutd: it was just a **nit** in the first place though hehe 😛

pgmanutd (approved):

packages/runtime/src/ax.tsx
madou: cool no wakkas

PR #293, opened 2020-09-18
Replace component hashing with sheet hashing

### LINKED ISSUE #288: Update stylesheet runtime to cache based on stylesheets instead of hash
Currently we pass a `hash` to every CS component to use for caching at runtime.

Now with moving to atomic declarations this doesn’t make sense because each declaration may or may not have already been added to the head (or rendered on the server).

- Remove `hash` from ast
- Remove `hash` from style component

Internally in the style component iterate through each sheet (children) - if it hasn’t been added - render it/move to head - else don’t.

madou:
Just smashed this out before going out haha, let me know what you think.

Closes #288

### DIFF
--- a/packages/runtime/src/sheet.tsx
+++ b/packages/runtime/src/sheet.tsx
@@ -22,7 +22,7 @@ function createStyleElement(opts: StyleSheetOpts): HTMLStyleElement {
  *
  * @param opts StyleSheetOpts
  */
-export const createStyleSheet = (opts: StyleSheetOpts) => {
+export default function createStyleSheet(opts: StyleSheetOpts) {
   const speedy = process.env.NODE_ENV === 'production',
     styleElements: HTMLStyleElement[] = [];
   let tagCount = 0;
@@ -51,4 +51,4 @@ export const createStyleSheet = (opts: StyleSheetOpts) => {
 
     tagCount++;
   };
-};
+}
--- a/package.json
+++ b/package.json
@@ -76,7 +76,15 @@
   "size-limit": [
     {
       "path": "./packages/runtime/dist/browser/index.js",
-      "limit": "299B",
+      "limit": "80B",
+      "import": "{ ax }",
+      "ignore": [
+        "react"
+      ]
+    },
+    {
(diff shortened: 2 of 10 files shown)


pgmanutd:
> Just smashed this out before going out haha, let me know what you think.
> 
> Closes #288

🙄 🙄 oooo..you are so fast. Slow down a bit hehehe 😛

madou:
image

Bundle size went up i'll see if i can reduce it next week

pgmanutd:
Looking into it Michael 👀. Give me sometime 😄

packages/runtime/src/sheet.tsx
```diff
@@ -22,7 +22,7 @@ function createStyleElement(opts: StyleSheetOpts): HTMLStyleElement {
  *
  * @param opts StyleSheetOpts
  */
-export const createStyleSheet = (opts: StyleSheetOpts) => {
+export default function createStyleSheet(opts: StyleSheetOpts) {
```
pgmanutd, line 25: nit: hehe. I thought you like named export 😛. Probably if we are using default export, it should match filename right 🤔?

pgmanutd (approved):
lgtm

packages/runtime/src/sheet.tsx
madou: only did this to reduce file size down haha - it reduces it a few bytes

pgmanutd: 😱 wooo, never guessed that. Damn. Nice.

pgmanutd: And btw how it reduced the size 🤔 thinking

madou: when minified default imports reduce down to `a` - something small

named imports cant be reduced it seems (you probably can but requires more complex minification)

pgmanutd: ohh right. hehe. I totally forgot minification here 😛 ahh another drawback of using named export then hehehehe 😛

PR #294, opened 2020-09-18
feat: add basic in memory caching for module boundary traversal

### LINKED ISSUE #292: Implement in memory cache for module boundary traversal
This ticket will build on what we’ve done previously (#266). Module traversal has been implemented and is working - we need to cache it now.

For now we probably want to go down an inmemory cache - but implement it in such a way that we can write to disk if needed later on.

pgmanutd:
This closes #292 

> @madou: I need early feedback on this one. Creating a draft PR for now. It's still WIP. I might add few more things.

Right now we are saving cache in memory. I left some **TODO's** where we can add file related caching. What I am thinking 🤔, probably we can create a mapping json file with content hash keys and module number as values. These modules will have their own json files with ast info in it.

For eg.

`.cache/compiled/module-mapping.json` will have following content:

```json
{
  "8haj1j": 1,
  "34fgj4": 2
}
```

Each hash value will have separate json file with ast info in it

`.cache/compiled/1.json` will have following content

```json
{
  "type": "File",
  "start": 0,
  "end": 2831,
  "loc": {
    "start": {
      "line": 1,
      "column": 0
    },
    "end": {
      "line": 141,
      "column": 0
    }
  }
...
```
and so on with other modules.

Each key is hash of **module path** and **module code**. Right now hash of **module path** and **module code** will be new when **module code** changes and new entry will be created in `module-mapping.json` file. We can optimize it further with separate hash for both **module path** and **module code** and update its content only when **module code** changes instead of creating new entry in `module-mapping.json`. Didn't give much thought on this one, but this should work.

I did some benchmarking, and here is the result when my laptop was in idle state:

```bash
❯ yarn perf
yarn run v1.22.5
$ ts-node --files ./src/__perf__/module-traversal-caching
#initial-run x 173 ops/sec ±5.23% (72 runs sampled)
#with-cache x 231 ops/sec ±2.28% (76 runs sampled)
#no-cache x 208 ops/sec ±2.41% (77 runs sampled)
Fastest is #with-cache
✨  Done in 24.18s.
```

You can see `#with-cache` is faster than `#no-cache` when we are caching only 2 modules with limited code. Probably it can be much faster than that in real world scenario. I also creating one test suite for `#initial-run` for consistent subsequent results. Even if you swap `#with-cache` position with `#no-cache`, result is same:

```bash
❯ yarn perf
yarn run v1.22.5
$ ts-node --files ./src/__perf__/module-traversal-caching
#initial-run x 173 ops/sec ±4.93% (70 runs sampled)
#no-cache x 202 ops/sec ±1.95% (74 runs sampled)
#with-cache x 234 ops/sec ±2.10% (76 runs sampled)
Fastest is #with-cache
✨  Done in 24.19s.
```


packages/babel-plugin/src/__perf__/module-traversal-caching.tsx
```diff
@@ -0,0 +1,66 @@
...
+};
+
+suite
+  .add('#initial-run', () => {
+    transform(
+      `
+      import '@compiled/core';
+      import React from 'react';
+
+      import { colorMixin } from '../__tests__/stubs/objects';
```
pgmanutd, line 26: Can we move these stubs outside of `__tests__` folder? Probably in `__fixtures__` folder under `src` to make them more reusable.

madou: Sounds good

packages/babel-plugin/src/__tests__/module-traversal.test.tsx
```diff
@@ -5,6 +5,15 @@ jest.mock('@compiled/utils', () => {
   return { ...jest.requireActual('@compiled/utils'), hash: () => 'hash-test' };
 });
 
+jest.mock('../utils/Cache', () => {
+  const { Cache } = jest.requireActual('../utils/Cache');
+
+  Cache.getUniqueKey = (key: string, namespace?: string) =>
```
madou, line 11: should we add this to `__mocks__` folder?

packages/babel-plugin/src/index.tsx
```diff
@@ -11,6 +12,9 @@ export default declare<State>((api) => {
 
   return {
     inherits: require('babel-plugin-syntax-jsx'),
+    pre() {
+      this.cache = new Cache(this.opts);
```
madou, line 16: can this be placed in the module level? rationale is: this means we will only cache for the current file we're traversing. if we import the same module in another file - we will pay the cost again

packages/babel-plugin/src/utils/Cache.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+export class Cache {
+  _options: CacheOptions & typeof defaultOptions;
+  _cache: InstanceType<typeof Map>;
+
+  constructor(options?: CacheOptions) {
+    // TODO: Replace `this._instanceKey` with `this.#instanceKey` once we upgrade to typescript >= 3.8
+    this._options = { ...defaultOptions, ...options };
+
+    // TODO: add logic for reading and writing saved cached file to this._cache here
```
madou, line 23: nit: instead of us leaving TODOs everywhere could we place it in a github issue instead? just because things will probably change between now and then and it will end up being more noise in the short term.. esp. if theres some bugs to fix here in the short term

packages/babel-plugin/src/utils/Cache.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  _tryDeletingLRUCachedValue() {
+    if (this._cache.size >= this._options.maxSize) {
+      const key = this._cache.keys().next().value;
+
+      this._cache.delete(key);
+    }
+  }
+
+  _moveLastInQueue<T>(uniqueKey: string, cacheValue: T) {
```
madou, line 52: why do we delete then set it again?

packages/babel-plugin/src/utils/__tests__/Cache.test.tsx
```diff
@@ -0,0 +1,118 @@
...
+describe('#Cache', () => {
+  beforeAll(() => {
+    Cache.getUniqueKey = (key: string, namespace?: string) =>
+      namespace ? `${namespace}----${key}` : key;
+  });
+
+  const setup = (options: CacheOptions = {}) => {
+    const cache = new Cache(options);
+
+    return { cache };
```
madou, line 16: nit: are you expecting to add more to this return object? might as well just return it immediately :P

packages/babel-plugin/src/utils/__tests__/Cache.test.tsx
```diff
@@ -0,0 +1,118 @@
...
+    expect(cache.getSize()).toBe(0);
+    expect(value).toBe(10);
+  });
+
+  it('should return original cached value when accessed with original key', () => {
+    const { cache } = setup();
+
+    const data = { namespace: 'namespace', cacheKey: 'cacheKey', value: () => 10 };
+
+    cache.load(data);
```
madou, line 54: why do we have three calls here? can we add one test that asserts that the lazy `value` only gets called once? :)

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -290,7 +290,12 @@ export const resolveBindingNode = (
       extensions: ['.js', '.jsx', '.ts', '.tsx'],
     });
     const moduleCode = fs.readFileSync(modulePath, 'utf-8');
```
madou, line 292: can we move this in the cache too please

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -290,7 +290,12 @@ export const resolveBindingNode = (
...
-    const ast = parse(moduleCode, { sourceType: 'module', sourceFilename: modulePath });
+
+    const ast = meta.state.cache.load({
+      namespace: 'module-traversal',
+      cacheKey: `${moduleCode}//# modulePath=${modulePath}`,
```
madou, line 296: we can cache baseed on module path instead of using the code i think?

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -290,7 +290,12 @@ export const resolveBindingNode = (
...
+      value: () => parse(moduleCode, { sourceType: 'module', sourceFilename: modulePath }),
+    });
 
     let foundNode: t.Node | undefined = undefined;
     let foundParentPath: NodePath | undefined = undefined;
```
madou, line 301: we still pay for the traversal cost below - is there any chance we can also cache the result of the traversal? (keyed on export identifier + file path)

packages/babel-plugin/src/index.tsx
pgmanutd: `pre()` should before babel start transforming. Check [this](https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md#-pre-and-post-in-plugins). And if we import same module with same absolute path, it will bring the cached value. Moreover, I also wanted to set `this.opts` initially. I can pass `opts` on every function call but thought this looks better. What you say 🤔?

packages/babel-plugin/src/utils/Cache.tsx
pgmanutd: Oh! Right. jsdoc is better approach. Actually we are moving things last in queue which are accessed frequently.

packages/babel-plugin/src/utils/__tests__/Cache.test.tsx
pgmanutd: Hehe. just convention I was following, if we have something more to return in future. Cool. Lets make the change 😄

packages/babel-plugin/src/utils/__tests__/Cache.test.tsx
pgmanutd: I mean, I was testing the behaviour instead of implementation detail. So if something is cached, it will be returned from cache. Probably I can pass different value with same key and check if value is same as initial? 🤔

packages/babel-plugin/src/utils/ast.tsx
pgmanutd: 🤔 What if module code changes between the builds? When we go for file caching?

packages/babel-plugin/src/utils/ast.tsx
pgmanutd: 🤔 What if module code changes between the builds? When we go for file caching?

packages/babel-plugin/src/utils/ast.tsx
pgmanutd: Yeah, we can 😄. Will look into this if we can also cache traversal thing too.

packages/babel-plugin/src/index.tsx
madou: > `pre()` should before babel start transforming right

per file unfortunately - not before the entire run

packages/babel-plugin/src/utils/__tests__/Cache.test.tsx
madou: fair but we haven't tested the behaviour that caching is actually working right? we could set value to a jest mock and assert how many  times its called?

packages/babel-plugin/src/utils/ast.tsx
madou: we could use last modified date maybe? 🤔 

only a problem for file caching at least

pgmanutd: 🤔 hmm. Let me try something.

packages/babel-plugin/src/utils/ast.tsx
pgmanutd: 🤔 hmm. Let me try something.

packages/babel-plugin/src/utils/Cache.tsx
pgmanutd: https://github.com/atlassian-labs/compiled/issues/297

pgmanutd:
@madou: Damn!! new perf numbers looks amazing
- when I moved cached instance out of `pre()`
- when I started caching module code
- when I started caching module ast traversal

> We will go for file modification time instead of file code. We will append modification time along with module path as `cacheKey`. What you say? 😄

**Previous numbers were:**
```bash
❯ yarn perf
yarn run v1.22.5
$ ts-node --files ./src/__perf__/module-traversal-caching
#initial-run x 173 ops/sec ±5.23% (72 runs sampled)
#with-cache x 231 ops/sec ±2.28% (76 runs sampled)
#no-cache x 208 ops/sec ±2.41% (77 runs sampled)
Fastest is #with-cache
✨  Done in 24.18s.
```

**New numbers are:**
```bash
❯ yarn perf
yarn run v1.22.5
$ ts-node --files ./src/__perf__/module-traversal-caching
#initial-run x 173 ops/sec ±5.19% (70 runs sampled)
#with-cache x 292 ops/sec ±1.85% (79 runs sampled)
#no-cache x 210 ops/sec ±0.93% (78 runs sampled)
Fastest is #with-cache
✨  Done in 24.53s.
```

madou (dismissed):
Looks great!

> We will go for file modification time instead of file code. We will append modification time along with module path as cacheKey. What you say?

Sure 😄

madou (approved):

PR #295, opened 2020-09-20
Hoist style rules to top of the module

### LINKED ISSUE #289: Hoist style rules to the top of the module
Move each stylesheet declaration to the top of the module.

If a stylesheet has already been found - do nothing - but reference the identifier in the ast.

See: https://github.com/atlassian-labs/compiled/pull/274/files?file-filters[]=.json&file-filters[]=.lock&file-filters[]=.tsx&hide-deleted-files=true#diff-f497fa51195bedd5e35d547e1401d219R229

Each declaration should be given a unique, short name. Start with a and work your way up. You should be able to use the inbuilt utils in Babel to get a unique name.

madou:
Closes #289

### DIFF
--- a/packages/babel-plugin/src/styled/index.tsx
+++ b/packages/babel-plugin/src/styled/index.tsx
@@ -124,12 +124,12 @@ export const visitStyledPath = (
   const cssOutput = buildCss(styledData.cssNode, meta);
 
   path.replaceWith(
-    buildStyledComponent({
-      ...meta.state.opts,
-      cssOutput,
-      tag: styledData.tag,
-      parentPath: path as NodePath,
-      scope: path.scope,
-    })
+    buildStyledComponent(
+      {
+        cssOutput,
+        tag: styledData.tag,
+      },
+      meta
+    )
   );
 };
--- a/packages/babel-plugin/src/utils/ast-builders.tsx
+++ b/packages/babel-plugin/src/utils/ast-builders.tsx
@@ -8,6 +8,31 @@ import { Tag } from '../types';
 import { CSSOutput } from './css-builders';
 import { pickFunctionBody } from './ast';
 import { CompiledOpts, CompiledTemplateOpts, StyledOpts, StyledTemplateOpts } from './types';
+import { Metadata } from '../types';
+
+/**
+ * Hoists a sheet to the top of the module if its not already there.
+ * Returns the referencing identifier.
+ *
+ * @param sheet
+ */
+const hoistSheet = (sheet: string, meta: Metadata): t.Identifier => {
+  if (meta.state.sheets[sheet]) {
+    return meta.state.sheets[sheet];
+  }
+
+  const sheetIdentifier = meta.parentPath.scope.generateUidIdentifier('a');
+  const parent = meta.parentPath.findParent((path) => path.isProgram()).get('body') as NodePath[];
+  const path = parent.filter((path) => !path.isImportDeclaration())[0];
+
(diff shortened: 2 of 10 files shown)


pgmanutd:
Looking into it Michael 👀. Give me sometime 😄

madou:
@pgmanutd after we merge my two open PRs I think theres an opportunity to improve the function APIs for ast-builder - they're a little.. wild atm.

Thinking to type them something like:

```js
function compiledComponent(node: t.Node, css: CSSOutput, meta: Metadata) {}
function styledComponent(tag: Tag, css: CSSOutput, meta: Metadata) {}
```

packages/babel-plugin/src/styled/index.tsx
```diff
@@ -124,12 +124,12 @@ export const visitStyledPath = (
   const cssOutput = buildCss(styledData.cssNode, meta);
 
   path.replaceWith(
-    buildStyledComponent({
-      ...meta.state.opts,
-      cssOutput,
-      tag: styledData.tag,
-      parentPath: path as NodePath,
-      scope: path.scope,
```
pgmanutd, line 132: Are we not passing updated `parentPath` and `scope` now 🤔?

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -1,43 +1,38 @@
...
 import { transformCss } from '@compiled/css';
 import isPropValid from '@emotion/is-prop-valid';
-
-import { PluginOptions, Tag } from '../types';
-
+import { Tag } from '../types';
 import { CSSOutput } from './css-builders';
 import { pickFunctionBody } from './ast';
+import { CompiledOpts, CompiledTemplateOpts, StyledOpts, StyledTemplateOpts } from './types';
+import { Metadata } from '../types';
```
pgmanutd, line 11: nit: Two `types` files confused me at first 😅? Can we do something about these 🤔?

packages/babel-plugin/src/styled/index.tsx
madou: we are just through metadata instead

packages/babel-plugin/src/utils/ast-builders.tsx
madou: oops

any suggestions? main thing i wanted to improve was the size of types is getting a bit crazy atm

packages/babel-plugin/src/styled/index.tsx
pgmanutd: hmmm I just thought we are passing updated `parentPath` and `scope` separately. No worries, if we are now passing these through meta. This new api looks more clean 🎉

packages/babel-plugin/src/utils/ast-builders.tsx
pgmanutd: Yeah they do. `{ CompiledOpts, CompiledTemplateOpts, StyledOpts, StyledTemplateOpts }` seems more like `ASTUtilsTypes` something. Maybe we can have something better naming than this.

pgmanutd: Are we making this change @madou? Otherwise this looks good.

madou: let's make the change in a follow up PR 👍 (will modify the APIs of above funcs too)

pgmanutd (approved):
lgtm

PR #298, opened 2020-09-21
fix: adds missing behaviour to ax func

madou:
It was missing a piece of behaviour.

Adds more code - any thoughts on how to shrink it down?

### DIFF
--- a/packages/runtime/src/ax.tsx
+++ b/packages/runtime/src/ax.tsx
@@ -28,8 +28,13 @@ export const ax = (classes: (string | undefined | false)[]): string => {
       continue;
     }
 
-    const group = cls.slice(0, cls.charCodeAt(0) === 95 ? 5 : undefined);
-    found[group] = cls;
+    const groups = cls.split(' ');
+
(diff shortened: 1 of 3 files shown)


pgmanutd:
> It was missing a piece of behaviour.
> 
> Adds more code - any thoughts on how to shrink it down?

I tried few things; extracting out code, using reduce but nothing worked 😞. Size remained at `105B`

packages/runtime/src/ax.tsx
```diff
@@ -28,8 +28,13 @@ export const ax = (classes: (string | undefined | false)[]): string => {
       continue;
     }
 
-    const group = cls.slice(0, cls.charCodeAt(0) === 95 ? 5 : undefined);
-    found[group] = cls;
+    const groups = cls.split(' ');
+
+    for (let x = 0; x < groups.length; x++) {
```
pgmanutd, line 33: hehe `x` 😛

pgmanutd (approved):

PR #302, opened 2020-09-24
Expand shorthand plugin

### LINKED ISSUE #296: Expand shorthand rules to long form rules
This should be a new PostCSS plugin that gets ran before the atomic plugin. This shouldn't expand every  shorthand - only ones that would make sense to override. Be pragmatic.

**In**

```css
margin: 0 auto;
```

**Out**

```css
margin-top: 0;
margin-bottom: 0;
margin-left: auto;
margin-right: auto;
```

**Links/prior art**

- https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties
- https://github.com/kripod/css-shorthand-expanders
- https://github.com/kapetan/css-shorthand-expand
- https://github.com/modulz/stitches/blob/canary/packages/core/src/shorthand-parser/index.ts
- https://github.com/ChromeDevTools/devtools-frontend/blob/e11be3c4ad0438fda453a7179a601b9f74ffada0/front_end/sdk/CSSMetadata.js

madou:
## Overview

Closes #296 

There are two reasons we want to expand shorthands:

1. Reduce the potential for total amount of defined CSS
1. Enable overrides to work in a consistent matter

This plugin should keep both in mind, if a property will end up creating more defined CSS **and** is also not a common property to override, we need to figure out what the best path forward is. Quick options are:

- Don't expand the property
- Only expand the minimum amount of properties
- Later - test in Jira - see what the size difference is expanded and not

## Todo

- [x] Size util
- [x] Color util
- [x] Correct tests describes
- [x] Go through each expander are x-check them against the spec

## Todo in a follow up PR

- [ ] Wiring up plugin

## Expansions

**Fully expanded**

- [x] flex-flow
- [x] margin
- [x] outline
- [x] overflow
- [x] padding
- [x] place-content
- [x] place-items
- [x] place-self
- [x] text-decoration
- [x] flex

**Partially expanded**

- [x] background

**Ordered shorthand**

These shorthands aren't common to override - however we will order them alphabetically to ensure minimal duplicates.
We will see if we need to expand these at a later date (comparing before/after size in Jira).

- [x] column-rule
- [x] list-style
- [x] columns
- [x] background

**Needs further thought**

These shorthands have the possibility to blow out to a large amount of properties. For now we are leaving them.

- border
- border-bottom
- border-color
- border-left
- border-radius
- border-right
- border-style
- border-top
- border-width

**Not initially**

These require either complete or partial ordering and/or aren't common to override. For now we are leaving them.

- animation
- transition
- font
- grid-column
- grid-row
- grid-template
- grid
- grid-area
- offset


packages/css/src/plugins/property-expander.tsx
```diff
@@ -0,0 +1,315 @@
+import { plugin, Declaration } from 'postcss';
```
pgmanutd, line 1: @madou: This file might get too big 😞 What should we do? Split in different files depending on property type? What you think 🤔? Tests are already touching 600 lines 😨

madou: Yeah we should

madou:
@pgmanutd let's talk through this and see if we need to change the scope

pgmanutd:
> @pgmanutd let's talk through this and see if we need to change the scope

Sure. No worries 😃 Happy to

packages/css/src/plugins/expand-shorthands/utils.tsx
```diff
@@ -0,0 +1,53 @@
...
+ */
+export const globalValues = ['inherit', 'initial', 'unset'];
+
+/**
+ * Returns `true` if the node is a color,
+ * else `false`.
+ *
+ * @param node
+ */
+export const isColor = (node: Node) => {
```
madou, line 14: flesh this out to cover all cases of color

packages/css/src/plugins/expand-shorthands/utils.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Returns `true` if the node is a width,
+ * else `false`.
+ *
+ * @param node
+ */
+export const isWidth = (node: Node) => {
+  if (node.type === 'numeric') {
+    if (['px', 'rem', 'em', '%'].includes(node.unit)) {
```
madou, line 30: flesh this out to cover all cases of width

packages/css/src/plugins/expand-shorthands/flex-flow.tsx
```diff
@@ -0,0 +1,54 @@
...
+  let directionValue = '';
+  let wrapValue = '';
+
+  /**
+   * Extracts values from a node and mutates variables in scope.
+   * If it returns `true` we should bail out and return no nodes.
+   *
+   * @param node
+   */
+  const extactValues = (node: Node): boolean => {
```
pgmanutd, line 21: hehe. Got confused with so many if's 😛

packages/css/src/plugins/expand-shorthands/padding.tsx
```diff
@@ -0,0 +1,15 @@
+import { ConversionFunction } from './types';
+
+/**
+ * https://developer.mozilla.org/en-US/docs/Web/CSS/padding
+ */
+export const padding: ConversionFunction = (value) => {
+  const [top, right = top, bottom = top, left = right] = value.nodes;
```
pgmanutd, line 7: Nice ❤️

packages/css/src/plugins/expand-shorthands/text-decoration.tsx
```diff
@@ -0,0 +1,55 @@
...
+  let colorValue = '';
+  let styleValue = '';
+
+  /**
+   * Extracts values from a node and mutates variables in scope.
+   * If it returns `true` we should bail out and return no nodes.
+   *
+   * @param node
+   */
+  const extractValues = (node: Node): boolean => {
```
pgmanutd, line 22: nit: Ifs and switches are getting quite complex in `flex-flow.tsx`, `flex.tsx`, `outline.tsx`, `text-decoration.tsx` etc files. I also got confused. Can we simplified it a bit, if possible?

packages/css/src/plugins/expand-shorthands/text-decoration.tsx
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    return false;
+  };
+
+  if (extractValues(left) || extractValues(middle) || extractValues(right)) {
+    return [];
+  }
+
+  lineValue.sort(); // Ensure the sorting is always in the same order.
```
pgmanutd, line 47: Q: Should we sort if `lineValue.length > 1`?

packages/css/src/plugins/expand-shorthands/text-decoration.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const resolvedLineValue = lineValue.length ? lineValue.join(' ') : 'initial';
```
pgmanutd, line 48: Q: Should we read `initial` from `globalValues`?

packages/css/src/plugins/expand-shorthands/utils.tsx
```diff
@@ -0,0 +1,78 @@
...
+  if (node.type === 'word' && node.isColor) {
+    return true;
+  }
+
+  if (node.type === 'func' && node.isColor) {
+    return true;
+  }
+
+  return false;
+};
```
pgmanutd, line 24: nit: Can be simplified to 
```suggestion
const { isColor, type } = node;

return isColor && ['word', 'func'].includes(type);
```

packages/css/src/plugins/expand-shorthands/utils.tsx
```diff
@@ -0,0 +1,78 @@
...
+export const getWidth = (node: Node) => {
+  if (node.type === 'numeric') {
+    return `${node.value}${node.unit}`;
+  }
+
+  if (node.type === 'word') {
+    return node.value;
+  }
+
+  return undefined;
```
pgmanutd, line 77: Q: This should be `0` or `auto` or `undefined`? 🤔

packages/css/src/plugins/expand-shorthands/text-decoration.tsx
madou: i'll see what i can do

packages/css/src/plugins/expand-shorthands/text-decoration.tsx
madou: IMO its fine to leave it as it - less code = better 😄

packages/css/src/plugins/expand-shorthands/text-decoration.tsx
madou: i think it's fine as is TBH

packages/css/src/plugins/expand-shorthands/utils.tsx
madou: cool!

packages/css/src/plugins/expand-shorthands/utils.tsx
madou: undefined or throw an error TBH 😄 for now ive just gone with undefined

what do you think?

packages/css/src/plugins/expand-shorthands/utils.tsx
madou: update: i can't change it to this because of type restrictions (is color is only on two types of nodes)

i slightly made the code better though

packages/css/src/plugins/expand-shorthands/utils.tsx
madou: update: actually undefined is useful because we end up returning invalid (ie empty rules) if it returns undefined.

packages/css/src/plugins/expand-shorthands/text-decoration.tsx
madou: this is kind of hard to be honest - mainly because of:

1. the order of values can be anything
2. needing to handle invalid usage (doubling up values, missing values, invalid values..)

any suggestions for improvement? would love to make the code easier to reason about

packages/css/src/plugins/expand-shorthands/text-decoration.tsx
```diff
@@ -0,0 +1,55 @@
...
+   * @param node
+   */
+  const extractValues = (node: Node): boolean => {
+    if (node && node.type === 'word') {
+      if (lineValues.includes(node.value)) {
+        if (lineValue.length === 0) {
+          lineValue.push(node.value);
+        } else if (!lineValue.includes(node.value)) {
+          lineValue.push(node.value);
+        } else {
```
pgmanutd, line 29: hmmm..looks hard though..hehehe..Following looks repeated `lineValue.push(node.value);`. But its ok to skip this one.

```suggestion
        if (lineValue.length === 0 || !lineValue.includes(node.value)) {
          lineValue.push(node.value);
        } else {
```

And its ok if refactoring is not possible. Lets merge this one right now. We can anyhow comeback to this piece of code later. Because code evolves as time goes one. And "I don't like premature abstraction" because "code duplication is far cheaper than the wrong abstraction" 😌

packages/css/src/plugins/expand-shorthands/text-decoration.tsx
pgmanutd: Check below [comment](https://github.com/atlassian-labs/compiled/pull/302/files#r506591806) @madou

pgmanutd:
@madou: Is this PR completed? I can see one pending task ==> "Go through each expander are x-check them against the spec". Otherwise it looks good. Just check once if all the scenarios are covered in the tests.

packages/css/src/plugins/expand-shorthands/text-decoration.tsx
madou: done

madou:
done the nits, going through the spec now

madou:
done have a look @pgmanutd

packages/css/src/plugins/expand-shorthands/outline.tsx
```diff
@@ -75,8 +75,8 @@ export const outline: ConversionFunction = (value) => {
   }
 
   return [
-    { prop: 'outline-color', value: colorValue || 'initial' },
-    { prop: 'outline-style', value: styleValue || 'initial' },
-    { prop: 'outline-width', value: widthValue || 'initial' },
+    { prop: 'outline-color', value: colorValue || 'currentColor' },
```
pgmanutd, line 78: Did we intentionally made `colorCode` camel case to fit in [this issue](https://github.com/atlassian-labs/compiled/issues/318)? Because in `text-decoration.tsx` we are setting it as lower case => `{ prop: 'text-decoration-color', value: colorValue || 'currentcolor' }`

madou: i noticed it because the spec mentions both `currentColor` and `currentcolor`... and ideally they should be just be one value 🤔 

i'll update the code to be all lower case for now

madou: done

pgmanutd (approved):
lgtm

PR #304, opened 2020-09-28
feat: handle function calls outside css apis

### LINKED ISSUE #285: Handle function calls when not inlined in object or template literal (both for css prop and styled api)
We have to handle cases like below which can be inlined when statically evaluated.

1. In this test case, `colorMixin` `CallExpression` should be evaluated & inlined instead of `colors.color` reference. Output should be whatever is the value of `colors.color`.
```typescript
 it('should reference property access expression from another module', () => {
    const result = transform(
      `
      import '@compiled/core';
      import { colorMixin } from './stubs/objects';

      const colors = colorMixin();

      <div css={\`:hover { color: \${colors.color}; }\`} />
    `
    );

    expect(result).toInclude('.cc-hash-test:hover{color:var(--var-hash-test)}');
    expect(result).toInclude('style={{"--var-hash-test":colors.color}}');
  });
```

2. Handle function calls like below. This should evaluate `fontSize` and inline it.
```typescript
const fontSize = (() => 12)();

const objectStyles = {
  fontSize,
  color: colors.danger,
  backgroundColor: () => colors.primary,
};

<div css={{ ... objectStyles }} />
```

Cases like below are already handled for both css prop and styled api when function calls are inlined.

```typescript
const fontSize = (() => 12)();

const mixin = () => 12;

<div css={{ fontSize: mixin() }} />
<div css={{ fontSize: mixin }} />
```

**Tech details**
Check `if` `else` case of `evaluate-expression.tsx` file. We probably have to handle `CallExpression` there.

pgmanutd:
This closes #285


packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -334,7 +347,7 @@ export const resolveBindingNode = (
 
   if (binding.path.parentPath.isImportDeclaration()) {
     if (!meta.state.filename) {
-      throw new Error('Filename is needed for module traversal.');
+      return;
```
pgmanutd, line 350: Removed this error because it was getting thrown when styles are coming from `React.useState`. I thought its better to return early instead of throwing an error. This is such use case:

```typescript
import React from 'react';
import '@compiled/core';

const [fontSize] = React.useState('10px');

const Component = () => <div css={{ fontSize }}>hello world</div>;
```

madou: nice!

packages/babel-plugin/src/__tests__/module-traversal.test.tsx
```diff
@@ -283,19 +283,44 @@ describe('module traversal', () => {
...
-  it('should reference property access expression from another module', () => {
+  it('should inline css from a directly called & assigned function mixin referencing an identifier from another module', () => {
     const result = transform(
       `
       import '@compiled/core';
       import { colorMixin } from '../__fixtures__/mixins/objects';
 
       const colors = colorMixin();
 
-      <div css={\`:hover { color: \${colors.color}; }\`} />
```
madou, line 294: i don't see a test remaining for this usage - can we add it back pls?

packages/babel-plugin/src/css-prop/__tests__/string-literal.test.tsx
```diff
@@ -371,4 +371,18 @@ describe('css prop string literal', () => {
...
+        import '@compiled/core';
+        import React from 'react';
+
+        const color = 'red';
+        const mixin = () => ({ color });
+
+        <div css={\`color: \${mixin().color};\`}>hello world</div>
+      `);
+
+    expect(actual).toInclude('.cc-hash-test{color:red}');
```
madou, line 386: this is cool

packages/babel-plugin/src/utils/evaluate-expression.tsx
```diff
@@ -141,6 +239,11 @@ export const evaluateExpression = (
     ({ value, meta: updatedMeta } = traverseMemberExpression(expression, updatedMeta));
   } else if (t.isFunction(expression)) {
     ({ value, meta: updatedMeta } = traverseFunction(expression, updatedMeta));
+  } else if (t.isCallExpression(expression)) {
+    ({ value, meta: updatedMeta } = evaluateExpression(
+      expression.callee as t.Expression,
```
madou, line 244: do we need to cast this? could we check if `callee` is an expression?

```
if (t.isCallExpression(expression) && t.isExpression(expression.callee) {}
```

packages/babel-plugin/src/__tests__/module-traversal.test.tsx
pgmanutd: 🤔 You mean object and string literal tests along with this module traversal test?

packages/babel-plugin/src/utils/ast.tsx
pgmanutd: Yups sure. Will add the comment.

Hardcoded 🤔 hmm..just thinking what should we harcode 🤔. Should we hardcode only most used packages like React?

packages/babel-plugin/src/utils/evaluate-expression.tsx
pgmanutd: Yeah right 😅

madou:
will finish review tmr

packages/babel-plugin/src/__tests__/module-traversal.test.tsx
pgmanutd: ohh got it. hehe. you mean this string literal test. 😛

packages/babel-plugin/src/utils/ast.tsx
pgmanutd: I mean what all we will hardcode? May be we can have whitelisting of packages we allow eg. `@atlaskit/*`, `@material-ui/*`, `antd*`, `@adobe/react-spectrum` etc. Few popular design systems or libraries from where theming/coloring are coming from?

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -69,6 +75,12 @@ export const getMemberExpressionMeta = (expression: t.MemberExpression) => {
     MemberExpression(path) {
       if (t.isIdentifier(path.node.object)) {
         bindingIdentifier = path.node.object;
+        originalBindingType = bindingIdentifier.type;
+      } else if (t.isCallExpression(path.node.object)) {
+        if (t.isIdentifier(path.node.object.callee)) {
+          bindingIdentifier = path.node.object.callee;
+          originalBindingType = path.node.object.type;
```
madou, line 82: when we want to handle args later - what will we do? any major changes needed?

packages/babel-plugin/src/utils/ast.tsx
madou: for now let's not worry about it - but keep it in the back of our mind

madou (approved):
lgtm

packages/babel-plugin/src/utils/ast.tsx
pgmanutd: 🤔 In such case we just have to capture arguments which are being passed when we are checking for `t.isCallExpression` and if those arguments are used in return statement we will evaluate them. I think we can put them somewhere in the state, probably in metadata and use them later when evaluating return statement? Its just matter of saving arguments somewhere 🤔

PR #309, opened 2020-10-06
feat: add object destructuring support

### LINKED ISSUE #283: Destructuring generates invalid CSS
Currently we don't support destructuring "officially" so it's resulting in broken CSS. We need to:

1. handle one or more levels of destructuring
1. handle the destructuring coming from an identifier or an object

Test cases:

```jsx
import '@compiled/core';

const { foo, color } = { foo: 14, color: 'blue' };

function Component() {
  return (
    <span css={{ fontSize: foo, color: color, backgroundColor: 'blue' }} />
  );
};
```

```jsx
import '@compiled/core';

const obj = { foo: 14, color: 'blue' };
const { foo, color } = obj;

function Component() {
  return (
    <span css={{ fontSize: foo, color: color, backgroundColor: 'blue' }} />
  );
};
```

pgmanutd:
This closes #283 

- add variable declaration object destructuring support
- will work for referenced identifiers too
- will work across different modules
- fallback to variable if it can't be resolved
- refactor #resolveBindingNode to accept reference name instead of binding.
It will get binding by reference name now and resolves its node

### DIFF
--- a/packages/babel-plugin/src/__tests__/expression-evaluation.test.tsx
+++ b/packages/babel-plugin/src/__tests__/expression-evaluation.test.tsx
@@ -187,4 +187,77 @@ describe('import specifiers', () => {
       );
     }).not.toThrow();
   });
+
+  it('handles object destructuring', () => {
+    const actual = transform(`
+      import '@compiled/core';
+      import React from 'react';
+
+      const { foo, color } = { foo: 14, color: 'blue' };
+
+      function Component() {
+        return (
+          <span css={{ fontSize: foo, color: color, backgroundColor: 'blue' }} />
+        );
+      };
+    `);
+
+    expect(actual).toInclude('.cc-hash-test{font-size:14px;color:blue;background-color:blue}');
+  });
+
+  it('handles the destructuring coming from an identifier', () => {
+    const actual = transform(`
+      import '@compiled/core';
+      import React from 'react';
+
+      const obj = { foo: 14, color: 'blue' };
+      const { foo, color } = obj;
+
+      function Component() {
+        return (
+          <span css={{ fontSize: foo, color: color, backgroundColor: 'blue' }} />
+        );
+      };
+    `);
+
+    expect(actual).toInclude('.cc-hash-test{font-size:14px;color:blue;background-color:blue}');
+  });
+
+  it('handles the destructuring coming from a referenced identifier', () => {
+    const actual = transform(`
+      import '@compiled/core';
+      import React from 'react';
+
+      const obj = { foo: 14, color: 'blue' };
+      const bar = obj;
+      const { foo, color } = bar;
+
+      function Component() {
+        return (
+          <span css={{ fontSize: foo, color: color, backgroundColor: 'blue' }} />
+        );
+      };
+    `);
+
+    expect(actual).toInclude('.cc-hash-test{font-size:14px;color:blue;background-color:blue}');
+  });
+
+  it('handles the function call destructuring coming from a referenced identifier', () => {
+    const actual = transform(`
+      import '@compiled/core';
+      import React from 'react';
+
+      const obj = { foo: () => ({ bar: 14 }), color: 'blue' };
+      const bar = obj;
+      const { foo, color } = bar;
+
+      function Component() {
+        return (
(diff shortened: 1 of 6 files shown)


pgmanutd:
@madou: This handles object destructuring coming from variable (or a referenced variable) only. I think we can skip array destructuring as of now.

packages/babel-plugin/src/__tests__/module-traversal.test.tsx
```diff
@@ -338,4 +338,19 @@ describe('module traversal', () => {
...
     expect(result).toInclude('.cc-hash-test:hover{padding-top:10px}');
   });
+
+  it('should inline css when destructuring an identifier from another module', () => {
+    const result = transform(
+      `
+      import '@compiled/core';
+      import { spacingMixin } from '../__fixtures__/mixins/objects';
+
+      const { padding: { top } } = spacingMixin;
```
madou, line 348: good test! was wondering about this scenario

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -316,35 +316,94 @@ const findNamedExportModuleNode = (
...
       meta,
-      node: binding.path.node.init as t.Node,
+      node,
       path: binding.path,
       constant: binding.constant,
       source: 'module',
     };
   }
 
+  if (binding.path.isObjectPattern()) {
```
madou, line 402: could we add this to L381?

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -316,35 +316,94 @@ const findNamedExportModuleNode = (
...
  * When wanting to do futher traversal on the resulting `node` make sure to use the output `meta` as well.
  * The `meta` will be for the resulting file it was found in.
  *
- * @param path
+ * @param referenceName Reference name for which `binding` to be resolved
+ * @param meta Plugin metadata
  */
 export const resolveBindingNode = (
-  binding: Binding | undefined,
+  referenceName: string,
```
madou, line 376: good change tbh!

packages/babel-plugin/src/utils/evaluate-expression.tsx
```diff
@@ -31,13 +31,7 @@ const traverseIdentifier = (expression: t.Identifier, meta: Metadata) => {
...
   let updatedMeta: Metadata = meta;
 
-  const binding = updatedMeta.parentPath.scope.getBinding(expression.name);
-  const resolvedBinding = resolveBindingNode(binding, updatedMeta);
-
-  if (binding?.path.node === expression) {
-    // We resolved to the same node - bail out!
-    return { value: expression, meta: updatedMeta };
-  }
+  const resolvedBinding = resolveBindingNode(expression.name, updatedMeta);
```
madou, line 34: :chefkiss:

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -316,35 +316,94 @@ const findNamedExportModuleNode = (
   };
 };
 
+const resolveObjectPatternValueNode = (
```
madou, line 319: can we add a jsdoc please and some example input/output

packages/babel-plugin/src/__tests__/expression-evaluation.test.tsx
```diff
@@ -191,4 +191,77 @@ describe('import specifiers', () => {
...
+      import '@compiled/core';
+      import React from 'react';
+
+      const obj = { foo: () => ({ bar: 14 }), color: 'blue' };
+      const bar = obj;
+      const { foo, color } = bar;
+
+      function Component() {
+        return (
+          <span css={{ fontSize: foo().bar, color: color, backgroundColor: 'blue' }} />
```
madou, line 260: this is great

madou:
looks great @pgmanutd - just a few nits

packages/babel-plugin/src/utils/ast.tsx
pgmanutd: Oh yeah..hehe right

pgmanutd:
@madou: Done with the changes. 😃

madou (approved):
lgtm

PR #314, opened 2020-10-12
fix: refactors browser check to be a node check to enable ssr to work in jsdom environments

### LINKED ISSUE #250: Style component doesn't work well in a browser like node environment
Currently we check for the existence of `window` to see if we're in a browser environment or not (see: https://github.com/atlassian-labs/compiled/blob/master/packages/runtime/src/provider/index.tsx#L4).

Unfortunately this isn't resilient enough for how Jira and AFP run their SSR environments (which is a "browser like" environment, in node).

So - we need to flip this and check for the existence of `process` or something of the like. Instead of "are we in brower" - it should be "are we in node".

madou:
Closes #250

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     },
     {
       "path": "./packages/runtime/dist/browser/index.js",
-      "limit": "200B",
+      "limit": "313B",
       "import": "{ CC, CS }",
       "ignore": [
         "react"
--- a/package.json
+++ b/package.json
@@ -15,9 +15,9 @@
     "lint": "eslint --config .eslintrc.js --ext tsx,ts ./packages/**/src ./examples",
     "lint:fix": "yarn lint -- --fix",
     "build": "yarn build:main && yarn build:browser",
-    "build:main": "TYPEOF_WINDOW='typeof window' ttsc --build packages/tsconfig.json",
+    "build:main": "ttsc --build packages/tsconfig.json",
     "build:dead-code-elimination": "cd test/dead-code-elimination && ttsc",
-    "build:browser": "TYPEOF_WINDOW='\"object\"' ttsc --build packages/tsconfig.browser.json",
+    "build:browser": "IS_NODE_EXPRESSION='false' ttsc --build packages/tsconfig.browser.json",
     "build:inspect": "node --inspect-brk node_modules/typescript/lib/tsc.js --build packages",
     "build-storybook": "build-storybook",
     "preversion": "git-branch-is master && yarn clean && yarn build && yarn test",
@@ -85,7 +85,7 @@
     },
     {
       "path": "./packages/runtime/dist/browser/index.js",
-      "limit": "312B",
+      "limit": "200B",
       "import": "{ CC, CS }",
       "ignore": [
         "react"
@@ -93,7 +93,7 @@
     },
     {
       "path": "./test/dead-code-elimination/dist/index.js",
-      "limit": "680 B",
+      "limit": "600B",
       "ignore": [
         "react"
       ]
--- a/packages/core/src/__tests__/browser.test.tsx
+++ b/packages/core/src/__tests__/browser.test.tsx
@@ -2,16 +2,20 @@ import React from 'react';
 import { render } from '@testing-library/react';
 import { styled } from '@compiled/core';
 
+jest.mock('../../../runtime/dist/is-node', () => ({
+  isNodeEnvironment: () => false,
+}));
+
 describe('browser', () => {
   it('should not render styles inline', () => {
     const StyledDiv = styled.div`
       font-size: 12px;
     `;
 
-    const { getByText } = render(<StyledDiv>hello world</StyledDiv>);
+    const { baseElement } = render(<StyledDiv>hello world</StyledDiv>);
 
-    expect(getByText('hello world').outerHTML).toMatchInlineSnapshot(
-      `"<div class=\\"_36l61fwx\\">hello world</div>"`
+    expect(baseElement.innerHTML).toMatchInlineSnapshot(
+      `"<div><div class=\\"_36l61fwx\\">hello world</div></div>"`
     );
   });
 
--- a/packages/jest/src/matchers.tsx
+++ b/packages/jest/src/matchers.tsx
@@ -16,14 +16,6 @@ const removeSpaces = (str?: string) => str && str.replace(/\s/g, '');
 const mapProperties = (properties: Record<string, any>) =>
   Object.keys(properties).map((property) => `${kebabCase(property)}:${properties[property]}`);
 
-const getMountedProperties = () =>
-  Array.from(document.styleSheets)
-    .map((sheet) =>
-      // @ts-ignore
-      sheet.cssRules.map((rule: CSSRule) => rule.style.cssText)
-    )
-    .join(' ');
-
 const onlyRules = (rules?: StyleRules['rules']) => rules?.filter((r) => r.type === 'rule');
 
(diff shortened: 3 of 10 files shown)


package.json
```diff
@@ -85,15 +85,15 @@
     },
     {
       "path": "./packages/runtime/dist/browser/index.js",
-      "limit": "312B",
+      "limit": "200B",
```
madou, line 88: im confused here - we actually remove more code now, currently in master this is:

```
➜  compiled git:(master) yarn bundlesize
yarn run v1.22.4
$ yarn build && yarn build:dead-code-elimination && size-limit
$ yarn build:main && yarn build:browser
$ TYPEOF_WINDOW='typeof window' ttsc --build packages/tsconfig.json
$ TYPEOF_WINDOW='"object"' ttsc --build packages/tsconfig.browser.json
$ cd test/dead-code-elimination && ttsc
✔ Adding to empty webpack project

  ./packages/runtime/dist/browser/index.js
  Size limit: 105 B
  Size:       105 B with all dependencies, minified and gzipped

  ./packages/runtime/dist/browser/index.js
  Size limit: 312 B
  Size:       312 B with all dependencies, minified and gzipped

  ./test/dead-code-elimination/dist/index.js
  Size limit: 680 B
  Size:       550 B with all dependencies, minified and gzipped

✨  Done in 35.52s.
```

uhhh?

packages/core/src/__tests__/browser.test.tsx
```diff
@@ -2,16 +2,20 @@ import React from 'react';
 import { render } from '@testing-library/react';
 import { styled } from '@compiled/core';
 
+jest.mock('../../../runtime/dist/is-node', () => ({
+  isNodeEnvironment: () => false,
```
madou, line 6: not my proudest code. because this isn't official api in the runtime we need to target dist directly.

packages/core/src/__tests__/browser.test.tsx
```diff
@@ -2,16 +2,20 @@ import React from 'react';
...
     const StyledDiv = styled.div`
       font-size: 12px;
     `;
 
-    const { getByText } = render(<StyledDiv>hello world</StyledDiv>);
+    const { baseElement } = render(<StyledDiv>hello world</StyledDiv>);
 
-    expect(getByText('hello world').outerHTML).toMatchInlineSnapshot(
-      `"<div class=\\"_36l61fwx\\">hello world</div>"`
+    expect(baseElement.innerHTML).toMatchInlineSnapshot(
```
madou, line 17: fixed this so it doesnt return false positives

package.json
```diff
@@ -15,9 +15,9 @@
     "lint": "eslint --config .eslintrc.js --ext tsx,ts ./packages/**/src ./examples",
     "lint:fix": "yarn lint -- --fix",
     "build": "yarn build:main && yarn build:browser",
-    "build:main": "TYPEOF_WINDOW='typeof window' ttsc --build packages/tsconfig.json",
```
madou, line 18: i updated the lib so we dont have to add in the default expression

packages/jest/src/matchers.tsx
```diff
@@ -89,9 +81,10 @@ export function toHaveCompiledCss(
 ): jest.CustomMatcherResult {
   const [property, value, matchFilter = DEFAULT_MATCH_FILTER] = args;
   const properties = typeof property === 'string' ? { [property]: value } : property;
-  const inlineStyleTag = element.parentElement && element.parentElement.querySelector('style');
-  const styleElements: HTMLStyleElement[] =
-    inlineStyleTag != null ? [inlineStyleTag] : Array.from(document.head.querySelectorAll('style'));
+  const styleElements: HTMLStyleElement[] = [
+    ...Array.from(document.body.querySelectorAll('style')),
```
madou, line 85: updated this because it wasnt picking up all style elements

packages/runtime/src/__tests__/style-ssr.test.tsx
```diff
@@ -1,23 +1,16 @@
-/**
- * @jest-environment node
- */
 import React from 'react';
 import { renderToStaticMarkup } from 'react-dom/server';
 import Style from '../style';
 
 describe('<Style />', () => {
   it('should render style as children on the server', () => {
-    const result = renderToStaticMarkup(<Style hash="a">{[`.a { display: block; }`]}</Style>);
```
madou, line 10: not sure how this remained haha

packages/runtime/src/__tests__/style-ssr.test.tsx
```diff
@@ -1,23 +1,16 @@
...
+    const result = renderToStaticMarkup(<Style>{[`.a { display: block; }`]}</Style>);
 
     expect(result).toInclude('<style>.a { display: block; }</style>');
   });
 
   it('should render style as children on the server with nonce', () => {
-    const result = renderToStaticMarkup(
-      <Style hash="a" nonce="1234">
```
madou, line 17: get out hash!

madou:
looks like the browser might be rendering styles inline - will look at a fix tmr.

albertogasparin (approved):
LGTM 👍

pgmanutd:
> looks like the browser might be rendering styles inline - will look at a fix tmr.

Yeah..Chromatic catched it 😛

package.json
madou: @pgmanutd ok its getting `true` instead of `false` - ill investigate

madou: fixed in ts-transform-define

pgmanutd: ahhhh `Boolean('false')` and `Boolean('true')` both will be true always.

package.json
```diff
@@ -85,15 +85,15 @@
     },
     {
       "path": "./packages/runtime/dist/browser/index.js",
-      "limit": "312B",
+      "limit": "313B",
```
pgmanutd, line 88: hehe `1B` is acceptable 😛

packages/runtime/src/is-node.tsx
```diff
@@ -0,0 +1,18 @@
...
+ *   // This code will be removed from the browser bundle
+ * }
+ * ```
+ */
+export const isNodeEnvironment = () => {
+  return (
+    Object.prototype.toString.call(typeof process !== undefined ? process : null) ===
+    '[object process]'
+  );
+};
```
pgmanutd, line 18: ❤️

pgmanutd:
lgtm. Should we test this with JIRA source code once 🤔? Just to make sure things are working as expected.

madou:
@albertogasparin any suggestions to how we can test this? we can public a test package component would just need to test with with SSR in Jira - possible locally?

albertogasparin:
I think testing it locally should not be a big deal, assuming it works fine with emotion/sc being there too it's just a matter of creating a dummy div and add the plugin. I suggest we merge this one and if there is more tweaking to do just sort it out later.
cc @marcins

pgmanutd (approved):

packages/runtime/src/is-node.tsx
alexreardon: This might be a little more resilient: 

```ts
const isNodeEnvironment = (): boolean => {
 typeof global !== 'undefined' && globalThis === global;
}
```

alexreardon: Looks like `globalThis` is well supported: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis (note minimum node version of 12.x)

I worry that the current approach might break in a future node release

PR #315, opened 2020-10-12
using globalThis to detect if running in node

alexreardon:
From here: https://github.com/atlassian-labs/compiled/pull/314/files#r503583214

This will require consumers be running on node `> 12.x`

### DIFF
--- a/packages/runtime/src/is-node.tsx
+++ b/packages/runtime/src/is-node.tsx
@@ -11,18 +11,5 @@
  * ```
  */
 export const isNodeEnvironment = (): boolean => {
-  if (typeof global === 'undefined') {
-    return false;
-  }
-
-  if ((globalThis as any) !== global) {
-    return false;
-  }
-
-  // what if a user did this: `global = globalThis`
-  if (typeof window !== 'undefined' && (globalThis as any) === window) {
-    return false;
-  }
-
-  return true;
+  return process?.release?.name === 'node';
 };
--- a/packages/runtime/src/is-node.tsx
+++ b/packages/runtime/src/is-node.tsx
@@ -20,7 +20,7 @@ export const isNodeEnvironment = (): boolean => {
   }
 
   // what if a user did this: `global = globalThis`
-  if (typeof window !== 'undefined' && globalThis === window) {
+  if (typeof window !== 'undefined' && (globalThis as any) === window) {
     return false;
   }
 
--- a/packages/runtime/src/is-node.tsx
+++ b/packages/runtime/src/is-node.tsx
@@ -10,9 +10,6 @@
  * }
  * ```
  */
-export const isNodeEnvironment = () => {
-  return (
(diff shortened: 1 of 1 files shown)


madou:
can we do anything to ensure consumers are on node 12? throw during run?

alexreardon:
It looks like your minimum node version is not defined for the project
You will want to do something like this:

```json
{ 
  "engines" : { 
    "node" : ">=12.0.0" 
   },
  "engineStrict" : true,
}
```
https://stackoverflow.com/questions/29349684/how-can-i-specify-the-required-node-js-version-in-package-json

You are using `nvm` but that only supports local running of the repo

alexreardon:
I can add a PR for that too if you want 😊

madou:
yes please

what happens if someone in the browser wants to be clever and sets:

```js
global = globalThis
```

🤔

alexreardon:
I've added another check. 

But i've found that `globalThis` is writable. So you could do this:

```
globalThis = whatever
```

So there is some level of risk in using globalThis.

That said, I would think that sort of behaviour is unlikely

alexreardon:
I think this way of detecting node is a bit more future proof (i'm not sure if the current toString approach is public API). 

That said, it would be fine to keep the current approach

madou:
hey looks like a test is failing

alexreardon:
I can dive in and fix if you are happy to proceed with this approach to `node` sniffing

alexreardon:
I ran into a bit of a wall here because `global` can be created on the client and `globalThis` can be re-assigned

madou (approved):
lgtm

@pgmanutd how about you? this relates to this https://nodejs.org/api/process.html#process_process_release

pgmanutd:
> lgtm
> 
> @pgmanutd how about you? this relates to this https://nodejs.org/api/process.html#process_process_release

@madou: What about 🤔
```ts
window.process = {release: {name: 'node'}};
// or
global.process = {release: {name: 'not node'}};
```

madou:
@pgmanutd interestingly this code should never be executed in the browser anyway because it of the `browser` entrypoint 😄 

so I think it's reasonable safe TBH

pgmanutd:
> @pgmanutd interestingly this code should never be executed in the browser anyway because it of the `browser` entrypoint 😄
> 
> so I think it's reasonable safe TBH

🤔 but `global.process = {release: {name: 'not node'}};` can be executed in node environment 🤔 
Anyways this works. Was just worried about this case ^^

alexreardon:
I'm not too worried about people overriding `process`. I'm more worried about that `toString()` behaviour changing

madou (approved):
lgtm

pgmanutd:
> I'm not too worried about people overriding `process`. I'm more worried about that `toString()` behaviour changing

❤️ nice. Works for me. 😉

pgmanutd (approved):

PR #320, opened 2020-10-19
feat: wire up expand plugin

madou:


packages/jest/src/__tests__/matchers.test.tsx
```diff
@@ -158,7 +164,7 @@ describe('toHaveCompliedCss', () => {
       </div>
     );
     const el = getByText('hello world');
-    expect(el).toHaveCompiledCss('background', 'red', { media: 'screen', target: ':hover' });
+    expect(el).toHaveCompiledCss('background-color', 'red', { media: 'screen', target: ':hover' });
```
madou, line 167: now the matcher will fail if you assert a shorthand - potentially. what should we do about this? (in another PR of course)

two thoughts are:

1. nothing - leave it as is and document the gotcha
2. convert the matcher to use post css and run it through the expand plugin 

🤔

pgmanutd: How about disabling it by default and let the user configure it in babel-plugin i.e. make it configurable 🤔? 
```typescript
expandShorthands: true/false
```

Probably we can also make other postcss plugins configurable (which are configurable)

madou: we need to be careful about this kind of configuration because we want to ensure a consistent experience when it comes to interop between usages (e.g. atlassian frontend vs. jira)

for now let's not?

pgmanutd: Yeah just a suggestion 😃 I mean in case people really don't care about shorthands..if we are using compiled outside of Atlassian.
Like below where order doesn't matter. Both will work. But this plugin will reorder it.
```typescript
expect(actual).toInclude(`{border:1px solid black}`);  => expect(actual).toInclude(`{border:1px black solid}`);
```

pgmanutd: I am in favour of "convert the matcher to use post css and run it through the expand plugin". Because we are giving our own matcher `toHaveCompiledCss`. What you say 🤔?

madou: i'll make a issue :)

madou: raised https://github.com/atlassian-labs/compiled/issues/321

packages/babel-plugin/src/css-prop/__tests__/object-literal.test.tsx
```diff
@@ -439,7 +442,7 @@ describe('css prop object literal', () => {
       `);
 
     expect(actual).toInclude(`{color:blue}`);
-    expect(actual).toInclude(`{border:1px solid blue}`);
+    expect(actual).toInclude(`{border:1px blue solid}`);
```
pgmanutd, line 445: hehe..ordered 😛 What I was saying. Both works.

madou: you're right 😄 it orders it so even if you define it both ways - it ends up being a single declaration instead of multiple

pgmanutd (approved):
lgtm

PR #324, opened 2020-10-20
Update stylesheet runtime to order psuedo state based on lvfha

### LINKED ISSUE #290: Update stylesheet runtime to order psuedo state atomic declarations based on lvfha
Atomic declarations means user land ordering of pseudo states no longer is possible. We are going to bake in the order based on LVFHA:

Top to bottom, bottom takes precedence:

- `:link`
- `:visited`
- `:focus`
- `:hover`
- `:active`

This ticket will modify the `stylesheet` runtime (used only when styles are baked in - once extracted this runtime is removed) to ensure the ordering.

One idea  I have for this is to build the class name with knowledge of the pseudo state so we don't have to do much at runtime.

Guiding principles:

1. less is more
1. code short as possible

pgmanutd:
This closes #290


package.json
```diff
@@ -85,7 +85,7 @@
     },
     {
       "path": "./packages/runtime/dist/browser/index.js",
-      "limit": "313B",
+      "limit": "697B",
```
pgmanutd, line 88: Help me here @madou. I can decrease it a bit by inlining function body where it is being used.

packages/runtime/src/cache.tsx
```diff
@@ -0,0 +1,74 @@
...
+  styleElement: HTMLStyleElement,
+  inserted: Record<string, true>
+) {
+  let styleNodeChild = styleNode.firstChild;
+
+  while (styleNodeChild) {
+    const css = styleNodeChild.textContent?.trim();
+
+    if (css) {
+      inserted[css] = true;
```
pgmanutd, line 42: We can also mark `inserted` `true` by checking if sheet is in `document.querySelector(style[data-compiled-${bucket}]` when its being rendered (on client) but that will become a performance bottleneck. This will run only once. Not sure how much we can rely on text nodes here.

packages/runtime/src/style.tsx
```diff
@@ -23,24 +24,38 @@ export default function Style(props: StyleProps) {
...
-    if (isNodeEnvironment()) {
-      return <style nonce={props.nonce}>{rules}</style>;
-    } else {
-      // Keep re-assigning over ternary because it's smaller
-      stylesheet = stylesheet || createStyleSheet(props);
-      rules.forEach(stylesheet);
+      return buckets
+        .filter((bucket) => !!rulesGroupedByBucket[bucket])
+        .map((bucket) => (
+          <style key={bucket} {...{ [getCompiledAttr(bucket)]: '' }} nonce={props.nonce}>
```
madou, line 50: using spread here ends up adding boilerplate code to support it - ideally let's come up with an alternate solution

madou (commented):
hey mate went over it, few more thoughts around the SSR code path right now that we can touch on this arvo in our catch up

madou:
can we also get a production build of storybook running locally to test that `insertRule` flow is working?

packages/runtime/src/cache.tsx
```diff
@@ -0,0 +1,74 @@
...
+export function addBucketToHead(bucket: Bucket) {
+  if (!bucketsAddedToHead[bucket]) {
+    const bucketIndex = buckets.indexOf(bucket);
+    let nextBucketFromCache = null;
+
+    // Find next bucket before which we will add our current bucket element
+    for (let i = bucketIndex + 1; i < buckets.length; i++) {
+      nextBucketFromCache = bucketsCache[buckets[i]] as HTMLStyleElement;
+
+      if (document.head.contains(nextBucketFromCache)) {
```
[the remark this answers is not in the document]
madou: we could replace this with:

```
bucketsAddedToHead[nextBucketFromCache]
```

pgmanutd:
@madou: I just need to add JSdocs. Right now looking possibility to reduce bundle size. 😄

packages/tsconfig.options.json
```diff
@@ -13,7 +13,7 @@
     "module": "es6",
     "moduleResolution": "node",
     "composite": true,
-    "target": "es5",
+    "target": "es6",
```
pgmanutd, line 16: @madou: How about this for excluding helpers like `__assign` to reduce bundlesize?

madou: how much does it reduce by

packages/runtime/src/buckets-utils.tsx
```diff
@@ -0,0 +1,44 @@
...
+  visited: 'v',
+  'focus-within': 'fw',
+  focus: 'f',
+  'focus-visible': 'fv',
+  hover: 'h',
+  active: 'a',
+};
+
+export const getBucket = (sheet: string): Bucket => {
+  if (sheet.charCodeAt(0) === 64) {
```
madou, line 19: can we add some comments here saying what we're doing

packages/runtime/src/buckets-utils.tsx
```diff
@@ -0,0 +1,44 @@
...
+    return 'm';
+  }
+
+  if (sheet.charCodeAt(10) === 58) {
```
madou, line 23: can we add some comments here saying what we're doing

packages/runtime/src/buckets-utils.tsx
```diff
@@ -0,0 +1,44 @@
...
+};
+
+export const getCompiledAttr = (bucket: Bucket) => `data-c${bucket}`;
+
+export const groupByBucket = <T extends string, U extends T[]>(sheets: U) => {
+  return sheets.reduce((accum, sheet) => {
+    const bucket = getBucket(sheet);
+    const bucketValue = accum[bucket];
+
+    accum[bucket] = bucketValue ? [...bucketValue, sheet] : [sheet];
```
madou, line 40: nit: potentially using concat here would be smaller

packages/runtime/src/buckets-utils.tsx
pgmanutd: Yeah JSDocs and comments are still pending 😅

packages/runtime/src/style.tsx
```diff
@@ -24,38 +24,35 @@ export default function Style(props: StyleProps) {
...
+    if (isNodeEnvironment()) {
+      const rulesGroupedByBucket = groupByBucket(sheets);
 
       return buckets
         .filter((bucket) => !!rulesGroupedByBucket[bucket])
         .map((bucket) => (
-          <style key={bucket} {...{ [getCompiledAttr(bucket)]: '' }} nonce={props.nonce}>
+          <style
+            key={bucket}
+            {...(bucket && { [getCompiledAttr(bucket)]: '' })}
```
madou, line 46: nit: we could also change this so instead of an attr we just use the bucket name?

```
<style data-c={bucket} /> 
```

packages/tsconfig.options.json
madou: unsure tbh because it will leave around const let and stuff which are potential de-op paths

what if we just not use spread/rest for now?

packages/runtime/src/style.tsx
pgmanutd: ah cool. This is what I was about to say when you asked **for the point of using es6 vs. es5 - we probably want to stay targeting es5 for now but leaving esm.. what do you think** 😅
That we might need es6 for object spread and can we do something like this. This is not a nit 😅. Its a change that should be in place hehe

packages/tsconfig.options.json
pgmanutd: Yeah yeah. I am removing `es6` and targeting `es5` only and bringing this change
```ts
<style data-c={bucket} /> 
```

package.json
```diff
@@ -77,15 +77,15 @@
   "size-limit": [
     {
       "path": "./packages/runtime/dist/browser/index.js",
-      "limit": "105B",
+      "limit": "114B",
```
madou, line 80: hmm why did this go up?

pgmanutd: `es6` hehe. I am moving it back to `es5` And removing all spread stuff. We will target `es5` only.

package.json
```diff
@@ -85,7 +85,7 @@
     },
     {
       "path": "./packages/runtime/dist/browser/index.js",
-      "limit": "313B",
+      "limit": "507B",
```
pgmanutd, line 88: @madou: Gone up by `194B` 😞. See if you can find anything 🤔.

pgmanutd:
@madou: One edge case is still left.

Styles like this

```ts
const StyledLink = styled.a`
  :hover,
  :focus {
    color: yellow;
  }
`;
```

ends up with this under `Style` component
```ts
{
  nonce: 'k0Mp1lEd',
  children: [
    '._30l31gy6:hover, ._f8pj1gy6:focus{color:yellow}'
  ]
}
```

If you remember we discussed that  classes like this `'._30l31gy6:hover, ._f8pj1gy6:focus{color:yellow}',` will end up under style bucket (There is a way. I can check for comma and space and get the first pseudo and put it under that bucket. Looks bit hacky). Can't we break 
```ts
 ._30l31gy6:hover, ._f8pj1gy6:focus{color:yellow}
```

into 

```ts
._30l31gy6:hover{color:yellow}
._f8pj1gy6:focus{color:yellow}
```

It will be easier for runtime to group it under different buckets if we break it using our atomicify plugin. Though bundlesize will go up a bit.

package.json
madou: i'll get it down to 450B 💪

madou:
@pgmanutd I actually don't think we need to solve the double selector!

Take this code:

```
const StyledLink = styled.a`
  :hover,
  :focus {
    color: yellow;
  }
`;
```

Let's talk through what happens.

1. Styles are generated and moved to the catch all bucket

```
<style data-c="s">._30l31gy6:hover, ._f8pj1gy6:focus{color:yellow}</style>
```

2. Happy days - it works as expected. It doesn't actually matter that they "aren't in order" because they can both be applied at the same time.

3. Another component is generated, except this time just using the single selector:

```
const AnotherLink = styled.a`
  :focus {
    color: yellow;
  }
`;
```

4. Styles are generated and moved to the focus bucket

```
<style data-c="f">._f8pj1gy6:focus{color:yellow}</style>
```

5. What happens? Well... exactly what we expect! Focusing on both StyledLink and AnotherLink turns the text yellow. Hovering over StyledLink turns its text yellow.

Can you think of any other edge cases? It looks like we can remove the hacky code!

pgmanutd:
> @pgmanutd I actually don't think we need to solve the double selector!
> 
> Take this code:
> 
> ```js-jsx
> const StyledLink = styled.a`
>   :hover,
>   :focus {
>     color: yellow;
>   }
> `;
> ```
> 
> Let's talk through what happens.
> 
> 1. Styles are generated and moved to the catch all bucket
> 
> ```js-jsx
> <style data-c="s">._30l31gy6:hover, ._f8pj1gy6:focus{color:yellow}</style>
> ```
> 
> 1. Happy days - it works as expected. It doesn't actually matter that they "aren't in order" because they can both be applied at the same time.
> 2. Another component is generated, except this time just using the single selector:
> 
> ```js-jsx
> const AnotherLink = styled.a`
>   :focus {
>     color: yellow;
>   }
> `;
> ```
> 
> 1. Styles are generated and moved to the focus bucket
> 
> ```js-jsx
> <style data-c="f">._f8pj1gy6:focus{color:yellow}</style>
> ```
> 
> 1. What happens? Well... exactly what we expect! Focusing on both StyledLink and AnotherLink turns the text yellow. Hovering over StyledLink turns its text yellow.
> 
> Can you think of any other edge cases? It looks like we can remove the hacky code!

What about this case:

```ts
<a
  href="https://atlassian.design"
  css={{
    display: 'flex',
    fontSize: '50px',
    ':link': {
      color: 'grey',
    },
    ':focus': {
      color: 'red',
    },
    ':active, :hover': {
      color: 'blue',
    },
  }}>
  Atlassian Design System
</a>
```

`':active, :hover':` will end up in style bucket and breaks lvfha ordering. Link should be blue when active but it will remain red (focus will take precedence). I know current solution for double selectors is not perfect but if its possible we can split them into 2 sheets using atomicify plugin. Downside is bundlesize will go up. Or may be create a card and we can investigate later? What you say?

madou:
hmm 🤔 good point. i think let's make a ticket and come back to it - i'll raise one

pgmanutd:
> hmm 🤔 good point (however would they end up in the style bucket? or would they end up in the active bucket?). i think let's make a ticket and come back to it - i'll raise one
> 
> here: #329

Hehe. Ticket is better for now. Anyways I am out today and we have public holiday on Monday. So will resume on Tuesday. I can see that you have merged few files and dropped `data-c` attribute. Nice Nice. Do let me know what is left. I can do that on Tuesday. Happy holidays btw 😃

madou:
no worries!! enjoy yourself - im just going through seeing what i can do to make the bundle smaller

madou:
@pgmanutd i got the bundle size down to `449B` - you're welcome 😉

pgmanutd:
> @pgmanutd i got the bundle size down to `449B` - you're welcome 😉

😱 😱 Damn!! Awesome man. Loved it. 2 clever techniques. One combining both `buckestCache` and `bucketsAddedToHead` and other making mapping keys small. Awesome. So `450B` was just a random number 😅?

pgmanutd:
> @pgmanutd i got the bundle size down to `449B` - you're welcome 😉

I am giving you `16B` more 😅. Check [this commit](https://github.com/atlassian-labs/compiled/pull/324/commits/e3337fe6b6706c92a61b50cef9fcfcade9d7ce70) and see if it makes sense. 😄

madou:
haha nice! looks great

the `return (x = true)` is wild but makes perfect sense!

madou (approved):

PR #325, opened 2020-10-21
fix: fixes doubling up nesting selector

madou:
Previously  when re-parenting dangling pseudos who already had a nested selector - it would incorrectly double up the selectors.

```
&& > * {
  margin-bottom: 1rem;

  &:last-child {
    margin-bottom: 0;
  }
}
```

Would end up transforming to:

```diff
&& > * {
  margin-bottom: 1rem;

+  &&:last-child
-  &:last-child {
    margin-bottom: 0;
  }
}
```

The code to generate a double selector was also not working:

```
.hash& > * { // <-- second & not replaced
  margin-bottom: 1rem;

  &:last-child {
    margin-bottom: 0;
  }
}
```

### DIFF
--- a/packages/css/src/plugins/__tests__/atomicify-rules.test.tsx
+++ b/packages/css/src/plugins/__tests__/atomicify-rules.test.tsx
@@ -42,6 +42,16 @@ describe('atomicify rules', () => {
     expect(result).toMatchInlineSnapshot(`"._q4hxglyw{-ms-user-select:none;user-select:none}"`);
   });
 
+  it('should double up class selector when two nesting selectors are found', () => {
+    const result = transform`
+      && {
+        display: block;
+      }
+    `;
+
+    expect(result).toMatchInlineSnapshot(`"._1e0c1ule._1e0c1ule{display:block}"`);
+  });
+
   it('should autoprefix atomic rules with multiple selectors', () => {
     process.env.BROWSERSLIST = 'Edge 16';
 
--- a/packages/css/src/plugins/__tests__/atomicify-rules.test.tsx
+++ b/packages/css/src/plugins/__tests__/atomicify-rules.test.tsx
@@ -200,6 +200,24 @@ describe('atomicify rules', () => {
     expect(secondActual).toEqual(expected);
   });
 
+  it('should double up selectors when using parent selector', () => {
+    const actual = transform`
+      && > * {
+        margin-bottom: 1rem;
+      }
+
+      && > *:last-child {
+        margin-bottom: 0;
+      }
+    `;
+
+    expect(actual.split('}').join('}\n')).toMatchInlineSnapshot(`
+      "._14rh1j6v._14rh1j6v > *{margin-bottom:1rem}
+      ._it8pidpf._it8pidpf > *:last-child{margin-bottom:0}
+      "
+    `);
+  });
+
   it('should atomicify a rule when its selector has a nesting at the end', () => {
     const actual = transform`
       :first-child & {
--- a/packages/css/src/plugins/atomicify-rules.tsx
+++ b/packages/css/src/plugins/atomicify-rules.tsx
@@ -69,7 +69,7 @@ const normalizeSelector = (selector: string | undefined) => {
  * @param parentClassName
  */
 const replaceNestingSelector = (selector: string, parentClassName: string) => {
-  return selector.replace(/ &/g, ` .${parentClassName}`);
+  return selector.replace(/&/g, `.${parentClassName}`);
 };
 
 /**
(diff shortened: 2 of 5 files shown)


packages/css/src/utils/__tests__/css-transform.test.tsx
```diff
@@ -132,7 +151,7 @@ describe('leading pseduos in css', () => {
     `
     );
 
-    expect(actual.join('\n')).toMatchInlineSnapshot(`"._12t41q9v >:first-child{color:hotpink}"`);
+    expect(actual.join('\n')).toMatchInlineSnapshot(`"._gkte1q9v > :first-child{color:hotpink}"`);
```
madou, line 154: this changed because the selector changed from

```
 >:first-child
```

to

```
 > :first-child
```

madou: this is another vector we need to make sure we normalize - currently the two could be different in userland but semantically the same - but be considered to be different thus generating a different group hash...

madou: made bug here https://github.com/atlassian-labs/compiled/issues/326

packages/css/src/plugins/atomicify-rules.tsx
```diff
@@ -69,7 +69,7 @@ const normalizeSelector = (selector: string | undefined) => {
  * @param parentClassName
  */
 const replaceNestingSelector = (selector: string, parentClassName: string) => {
-  return selector.replace(/ &/g, ` .${parentClassName}`);
```
pgmanutd, line 72: I had suspicion about this ` .${parentClassName}` spacing earlier. But I thought we had enough test cases to cover this. 😞   No worries. Lets add as many as test cases we can think of. What you say? 😃

madou: yeah definitely! i will make another test file

madou:
one outstanding fix im doing push soon

packages/css/src/plugins/__tests__/atomicify-rules.test.tsx
```diff
@@ -42,6 +42,16 @@ describe('atomicify rules', () => {
...
   });
 
+  it('should double up class selector when two nesting selectors are found', () => {
+    const result = transform`
+      && {
+        display: block;
+      }
+    `;
+
+    expect(result).toMatchInlineSnapshot(`"._1e0c1ule._1e0c1ule{display:block}"`);
```
pgmanutd, line 52: How this will end up in buckets 🤔 in case we have something like `._1e0c1ule:hover._1e0c1ule:hover`? It might break below logic 😞:

```ts
if (sheet.charCodeAt(10) === 58) {
  const openBracketIndex = sheet.indexOf('{');
  const name = sheet.slice(11, openBracketIndex);

  return pseudosMap[name] || '';
}
```

madou: ahh! yeah thats a problem isn't it. as well as selectors that have multiples

```
._1e0c1ule:focus, ._1e0c1ule:active {
  color: red;
}
```

HMM

pgmanutd: It will end up in style bucket 😞

pgmanutd: Can't we split this into two and always make sure things end up like this => `._1e0c1ule:focus{`

```ts
._1e0c1ule:focus {
  color: red;
}

._1e0c1ule:active {
  color: red;
}
```

madou: _we could_ but i'd prefer not creating more baked in styles if we can avoid it, we're already going pretty nuts on it

for this perhaps just place it in the "lowest bucket" and it should be OK? so for this instance it would go in the focus bucket.

pgmanutd: `._1e0c1ule:focus,._1e0c1ule:active {` will end up in style bucket. Let me change the logic and see if I can fix this one up.

packages/css/src/plugins/parent-orphaned-pseudos.tsx
```diff
@@ -32,7 +32,7 @@ export const parentOrphanedPseudos = plugin('parent-orphened-pseudos', () => {
     root.walkRules((rule) => {
       const { selector: ruleSelector } = rule;
 
-      if (!ruleSelector.includes(':')) {
+      if (!ruleSelector.startsWith(':')) {
```
madou, line 35: this is technically a regression but because we have the nesting logic bled into the atomicify plugin it doesnt break

my follow up PR fixes this

pgmanutd (approved):
lgtm

PR #331, opened 2020-10-24
chore: reduce ax bundlesize

madou:
Initially I had an implementation that did `join(' ').split(' ')` which allowed me to get the size down to `50B` but it was 30% slower than ax in master.

This PR has similar perf to ax in master.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
   "size-limit": [
     {
       "path": "./packages/runtime/dist/browser/index.js",
-      "limit": "75B",
+      "limit": "99B",
       "import": "{ ax }",
       "ignore": [
         "react"
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
   "size-limit": [
     {
       "path": "./packages/runtime/dist/browser/index.js",
-      "limit": "105B",
+      "limit": "75B",
       "import": "{ ax }",
       "ignore": [
         "react"
--- a/packages/runtime/src/ax.tsx
+++ b/packages/runtime/src/ax.tsx
@@ -19,23 +19,22 @@
  *
  * @param classes
  */
-export const ax = (classes: (string | undefined | false)[]): string => {
-  const found: Record<string, string> = {};
+export default function ax(classNames: (string | undefined | false)[]): string {
+  const atomicGroups: Record<string, string> = {};
+  let i = -1;
 
-  for (let i = 0; i < classes.length; i++) {
-    const cls = classes[i];
-    if (!cls) {
+  while (++i < classNames.length) {
+    if (!classNames[i]) {
       continue;
     }
 
-    const groups = cls.split(' ');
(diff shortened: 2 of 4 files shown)


packages/runtime/src/ax.tsx
```diff
@@ -19,23 +19,22 @@
...
-      const group = className.slice(0, className.charCodeAt(0) === 95 ? 5 : undefined);
-      found[group] = className;
+    while (++x < groups.length) {
+      atomicGroups[groups[x].slice(0, 5)] = groups[x];
     }
   }
 
-  return Object.values(found).join(' ');
-};
+  return Object.values(atomicGroups).join(' ');
```
pgmanutd, line 39: Q:  We are targeting `es5` right 🤔? This is for `Object.values`.

package.json
```diff
@@ -77,7 +77,7 @@
   "size-limit": [
     {
       "path": "./packages/runtime/dist/browser/index.js",
-      "limit": "105B",
+      "limit": "75B",
```
pgmanutd, line 80: Q: Ir went down because of `for => while` + removal of ternary thing?

packages/runtime/src/ax.tsx
```diff
@@ -19,23 +19,22 @@
...
       continue;
     }
 
-    const groups = cls.split(' ');
+    const groups = (classNames[i] as string).split(' ');
+    let x = -1;
 
-    for (let x = 0; x < groups.length; x++) {
-      const className = groups[x];
-      const group = className.slice(0, className.charCodeAt(0) === 95 ? 5 : undefined);
```
pgmanutd, line 35: We have removed `_` check. This might fail in following scenario:

```ts
// Will return 'foobar2 bar` but should be 'foobar1 foobar2 bar` 
ax(['foobar1 foobar2', 'bar']);
```

packages/runtime/src/ax.tsx
madou: its supported in everything except for IE11 which is okay

packages/runtime/src/ax.tsx
madou: hmm yeah true i'll add it back

madou: added back in - size back up to 99B but at least its under 100!

packages/runtime/src/ax.tsx
```diff
@@ -19,23 +19,22 @@
...
-      found[group] = className;
+    while (++x < groups.length) {
+      atomicGroups[groups[x].slice(0, groups[x].charCodeAt(0) === 95 ? 5 : undefined)] = groups[x];
```
pgmanutd, line 35: nit: hehe, Someone raised one issue around magic number 95. We can create a local variable and use it if possible.

madou: done

pgmanutd (approved):
lgtm

PR #336, opened 2020-10-31
chore: simplify css var name

madou:
This ends up producing smaller bundlers because `var` isn't needed anymore.

**Before**

```css
--var-1As3op: blue;
```

**After**

Prefixed with an underscore because `--1` isn't a valid CSS variable name.

```css
--_1As3op: blue;
```


pgmanutd (approved):
lgtm

PR #337, opened 2020-11-02
feat: add codemod for sc and emotion

pgmanutd:
This closes #239 
This closes #179

PR #338, opened 2020-11-02
feat: cli package with codemods preset

pgmanutd:
This PR is basically an extension of PR #337. Added `cli` with `codemod` preset. We can add as many as presets as we want. Its now easy for users now to run codemods using this `cli`.

Use this commit hash (dad0c1321bf58c83c85a24bd8c9cd5342f994166) to review things related to `cli`.


packages/cli/README.md
```diff
@@ -0,0 +1,15 @@
...
+npx @compiled/cli
+
+# or
+
+npx @compiled/cli --preset codemods
+```
+
+## Available presets
+
+1. [codemods](./src/presets/codemods)
```
madou, line 15: this is cool!

packages/cli/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "directory": "packages/cli"
+  },
+  "main": "./dist/main.js",
+  "module": "./dist/main.js",
+  "types": "./dist/main.d.ts",
+  "scripts": {
+    "start": "ts-node --files src/index.tsx"
+  },
+  "bin": {
+    "compiled-css-in-js": "./dist/index.js"
```
madou, line 21: `compiled-css-in-js` ?

packages/cli/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "@compiled/core": "0.4.7",
+    "chalk": "^4.1.0",
+    "enquirer": "^2.3.6",
+    "exec-sh": "^0.3.4",
+    "glob": "^7.1.6",
+    "jscodeshift": "^0.11.0",
+    "meow": "^8.0.0"
+  },
+  "devDependencies": {
+    "ts-node": "^9.0.0"
```
madou, line 37: should this be a dependency?

packages/cli/src/__tests__/main.test.tsx
```diff
@@ -0,0 +1,111 @@
+jest.mock('enquirer', () => ({
```
madou, line 1: nit: defining these here it will still get hoisted below imports wont it? might as well define below them?

packages/cli/src/__tests__/main.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+import main from '../main';
+
+describe('main', () => {
+  beforeEach(() => {
+    jest.spyOn(global.console, 'log').mockImplementation();
+    jest.spyOn(global.console, 'warn').mockImplementation();
+  });
+
+  afterEach(() => {
+    (global.console.log as jest.Mock).mockReset();
```
madou, line 22: nit: would be good to define a test util

```
function asJestMock<T>(obj: T): jest.Mock<T> {
  return obj as jest.Mock<T>;
}

asJestMock(global.console.warn).mockReset();
```

packages/cli/src/__tests__/main.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    return {
+      cli,
+    };
+  };
+
+  it('should warn when invalid preset is provided as cli input', async () => {
+    const { cli } = await setup('invalid-using-cli');
+
+    expect(AutoComplete).not.toHaveBeenCalledTimes(1);
```
madou, line 51: do we know of a better way to test this outside of using mocks for global + auto complete? seems a bit wild 🤔

packages/cli/src/cli.tsx
```diff
@@ -0,0 +1,33 @@
+import meow from 'meow';
+import chalk from 'chalk';
+
+import presets from './presets';
+
+const cli = meow(
+  `
+  Usage
+    ${chalk.cyan('$ npx @compiled/cli [options]')}
```
madou, line 9: nit: is including `$` a bit too cute?

packages/cli/src/isDev.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+declare global {
+  namespace NodeJS {
+    interface Process {
+      [REGISTER_INSTANCE]?: object;
+    }
+  }
+}
+
+const isDev = typeof process[REGISTER_INSTANCE] === 'object';
```
madou, line 14: should we rename to `isRunningInTsNode` ?

packages/cli/src/presets/codemods/__tests__/codemods.test.tsx
```diff
@@ -0,0 +1,188 @@
...
+            message: '--extensions',
+            hint: `default: ${chalk.cyan('js')}`,
+          },
+          {
+            name: 'ignorePattern',
+            message: '--ignore-pattern',
+          },
+          {
+            name: 'others',
+            message: 'other cli options',
```
madou, line 106: should these just be their entries? are there other options outside of these?

packages/cli/src/presets/codemods/__tests__/codemods.test.tsx
```diff
@@ -0,0 +1,188 @@
...
```
madou, line 177: same nit here for other options

packages/cli/src/presets/codemods/utils/transforms.tsx
```diff
@@ -0,0 +1,18 @@
+import path, { ParsedPath } from 'path';
+import glob from 'glob';
+
+import isDev from '../../../isDev';
+
+const basePath = isDev
+  ? path.join(process.cwd(), '..', 'core', 'src', 'codemods')
+  : path.join(process.cwd(), 'node_modules', '@compiled', 'core', 'dist', 'codemods');
```
madou, line 8: what would happen if we're running in a monorepo (with and without workspaces?)

packages/cli/src/presets/index.tsx
```diff
@@ -0,0 +1,12 @@
+import { Result } from '../cli';
+
+import codemods from './codemods';
+
+type Presets = {
```
madou, line 5: use Interface for object declarations please 😄

packages/cli/types/enquirer/index.d.ts
```diff
@@ -0,0 +1,15 @@
+import { Prompt } from 'enquirer';
+
+declare module 'enquirer' {
```
madou, line 3: oh this https://www.npmjs.com/package/@types/inquirer doesn't cover it? 😢

packages/core/src/codemods/emotion-to-compiled/README.md
```diff
@@ -2,11 +2,28 @@
...
+```bash
+npx @compiled/cli --preset codemods
+# and follow the instructions
+```
+
+How instructions looks like (when run without `--preset codemods`. Please run it with `--preset codemods` for fast access):
+
+![emotion-to-compiled cli](./assets/emotion-to-compiled-cli.gif)
+
+_NOTE: Use `npx @compiled/cli` instead of `npx compiled-css-in-js` (which is shown in image)_
```
madou, line 25: 😆 just make a new gif!

packages/core/src/codemods/styled-components-to-compiled/README.md
```diff
@@ -2,11 +2,28 @@
...
+### Using `@compiled/cli`
+
+```bash
+npx @compiled/cli --preset codemods
+# and follow the instructions
+```
+
+How instructions looks like (when run without `--preset codemods`. Please run it with `--preset codemods` for fast access):
+
+![styled-components-to-compiled cli](./assets/styled-components-to-compiled-cli.gif)
```
madou, line 23: same comment here re: gif

packages/cli/package.json
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "bin": {
+    "compiled-css-in-js": "./dist/index.js"
+  },
+  "files": [
+    "dist",
+    "src"
+  ],
+  "dependencies": {
+    "@compiled/core": "0.4.7",
```
madou, line 28: make sure this has this dependency declared in the tsconfig project references

packages/cli/package.json
pgmanutd: OOPS!!

packages/cli/package.json
pgmanutd: oh!! I am using `isDev` in `transform`. Right. Will move 😅

packages/cli/src/presets/codemods/__tests__/codemods.test.tsx
pgmanutd: Yups options like `--verbose=0 --version --help --silent` which are not mandatory and not super important

madou: do we think its worthwhile including them here? could we just omit?

if they're worthwhile - i'd say just list them - else omit.

packages/cli/package.json
pgmanutd: Its already there 😅

madou:
need to fix up merge conflicts in this 😄 

probably easier to just cherry pick the last commit onto a fresh version of master

packages/cli/package.json
madou: did we need to update this?

packages/react/src/codemods/emotion-to-compiled/README.md
```diff
@@ -2,13 +2,30 @@
...
+### Using `@compiled/cli`
+
+```bash
+npx @compiled/cli --preset codemods
+# and follow the instructions
+```
+
+How instructions looks like (when run without `--preset codemods`. Please run it with `--preset codemods` for fast access):
+
+![emotion-to-compiled cli](./assets/emotion-to-compiled-cli.gif)
```
madou, line 25: can we update this please 

- new gif (or just remove for now)
- do we need to mention updating instructions? when i run `npx @compiled/cli` and then we release a new version, what happens?
- remove `caveats` section for now

packages/react/src/codemods/emotion-to-compiled/README.md
```diff
@@ -2,13 +2,30 @@
 
 > Codemod for easy migration from emotion.
 
+## Usage
+
+### Using `jscodeshift`
```
madou, line 7: if we have cli perhaps we just remove this as an option? any reason to keep it?

madou:
is this ready for review?

pgmanutd:
> is this ready for review?
 
Wasn't well this Friday and over the weekend. Give me a day @madou

madou:
ok 😄

packages/cli/src/presets/codemods/utils/transforms.tsx
pgmanutd: **With**: It will pick `node_modules/@compiled/core/dist/codemods` from  root folder. If anyone is running inside any package, it will still work because of symlink to `@compiled/react` in root folder.

**Without**: In this case, packages will not be hoisted to root `node_modules` and this cli will install `@compiled/react` to the folder from from where it is running.

Q: What will happen if we use `npx @compiled/cli`? `process.cwd()` will be temp directory where `@compiled/cli` will be installed or the path where we are executing `npx @compiled/cli`? 🤔

packages/cli/src/__tests__/main.test.tsx
pgmanutd: hmm. 🤔 testing the behavior and not mocking things. I love this. But still testing cli is bit wild right now. Mocking is very safe in this case, but it will be wonderful if we can test the exact behaviour. But sometimes cli's introduce so many side effects and that might messed up our unit testing.
Can we skip for now? 😛 I ll keep this suggestion in mind. hehe. May be one open source library in future 😉

madou: np

packages/cli/src/presets/codemods/utils/transforms.tsx
madou: ideally it should work with `npx` because not many people would want to install the cli directly - it would be like use once and throw away

will npx work atm?

madou:
lgtm let's just double check npx behaviour

what are we expecting ppl to use? install and use or use npx?

pgmanutd:
> lgtm let's just double check npx behaviour
> 
> what are we expecting ppl to use? install and use or use npx?

Second will work. I prefer first one. Using `npx`. How can we test `npx` locally? 🤔. Can we merge this, release a nightly and check? Any idea? Not sure if answer [here](https://stackoverflow.com/questions/54261978/how-can-i-have-my-npx-script-run-commands-within-the-same-package-repo) really works.

madou (approved):
lgtm lets test

can you run npx on a local folder? `npx ./blah`

pgmanutd:
> lgtm lets test
> 
> can you run npx on a local folder? `npx ./blah`

Tried with `npx ../compiled/packages/cli/dist/index.js` and its working. Both `compiled` and `atlassian-frontend` are in the same folder 😃. Its taking `codemods` from `atlassian-frontend/node_modules/@compiled/react/dist/codemods`. Rest we can test after nightly release. What you say?

PR #339, opened 2020-11-03
test: add ssr hydration tests

pgmanutd:
This test ensures that ssr hydration works correctly with compiled.

### DIFF
--- a/packages/core/src/__tests__/server-side-hydrate.test.tsx
+++ b/packages/core/src/__tests__/server-side-hydrate.test.tsx
@@ -4,7 +4,7 @@ import ReactDOM from 'react-dom';
 import { styled } from '@compiled/core';
 
 describe('server side hydrate', () => {
-  it('should ssr then hydrate element correctly', () => {
+  it('it should not log unexpected warnings when hydrating SSRing', () => {
     jest.spyOn(global.console, 'error').mockImplementation(() => {});
 
     const StyledDiv = styled.div`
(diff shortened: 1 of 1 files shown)


packages/core/src/__tests__/server-side-hydrate.test.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { renderToString } from 'react-dom/server';
+import ReactDOM from 'react-dom';
+import { styled } from '@compiled/core';
+
+describe('server side hydrate', () => {
+  it('should ssr then hydrate element correctly', () => {
```
madou, line 7: could we rename this to `it should not log unexpected warnings when hydrating SSRing`

packages/core/src/__tests__/server-side-hydrate.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+    delete process.release.name;
+
+    ReactDOM.hydrate(element, elem);
+
+    // Restore the release name after hydration is complete
+    process.release.name = releaseName;
+
+    const mockCalls = (console.error as jest.Mock).mock.calls.filter(
+      ([f, s]) =>
+        !(f === 'Warning: Did not expect server HTML to contain a <%s> in <%s>.%s' && s === 'style')
```
madou, line 32: can we add a comment here explaining that we will get mismatches for the style rendering but this is expected

packages/core/src/__tests__/server-side-hydrate.test.tsx
```diff
@@ -4,7 +4,7 @@ import ReactDOM from 'react-dom';
 import { styled } from '@compiled/core';
 
 describe('server side hydrate', () => {
-  it('should ssr then hydrate element correctly', () => {
+  it('it should not log unexpected warnings when hydrating SSRing', () => {
```
madou, line 7: remove the it :P

pgmanutd: fuck

madou (approved):

PR #343, opened 2020-11-03
feat: expose babel plugin through core to simplify consumption experience

madou:
I wanted to do this via the `exports` field in package.json but it wouldn't work for me, so I did it the old school way. This also adds back `README.md` to packages in npm.

**Before**

```
npm i @compiled/core @compiled/babel-plugin
```

```js
module.exports = {
  plugins: [['@compiled/babel-plugin'],
};
```

**After**

```
npm i @compiled/core
```

```js
module.exports = {
  plugins: [['@compiled/core/babel-plugin']],
};
```

The CLA bot seems to be having problems..

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,5 +1,5 @@
 module.exports = {
-  plugins: [['@compiled/babel-plugin', { nonce: '"k0Mp1lEd"' }]],
+  plugins: [['@compiled/core/babel-plugin', { nonce: '"k0Mp1lEd"' }]],
   presets: [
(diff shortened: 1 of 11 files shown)


pgmanutd:
Nice. Do you still want to keep `exports` in `package.json`? May be for future reference or should we create an issue for this? I think `Node 13+` supports `exports` without experimental flag.

madou:
I don't think there is any harm adding it, I'll do it.

pgmanutd (approved):
lgtm

PR #354, opened 2020-11-09
browser/server providers now use isNodeEnvironment

### LINKED ISSUE #353: SSR checks in runtime are using old window check

Currently in the Server Provider https://github.com/atlassian-labs/compiled/blob/master/packages/runtime/src/provider/provider-server.tsx#L4 the check to see if window is "undefined" can fail if consumers use more exotic forms of Server Side Rendering that emulate a browser environment.

I propose exporting a function in which consumers can pass in a predicate which returns true when being Server Side Rendered. Giving consumers more flexibility with the technology they use to Server Side Render their apps.
Something like:
```
// index.js
import { isSSR } from @compiled/core;
import App from './App';
isSSR(() => !(window && window.navigator));
export default () => <App />;
```
Compiled would then use the returned value of this function as the source of truth for all SSR related code branches.

MatthewCharlton:
Updated the `provider-browser` and `provider-server` files to use the `isNodeEnvironment` check as per the discussion here - https://github.com/atlassian-labs/compiled/issues/353

Closes #353

### DIFF
--- a/packages/runtime/src/__tests__/provider.test.tsx
+++ b/packages/runtime/src/__tests__/provider.test.tsx
@@ -1,7 +1,7 @@
-const mockFunction = jest.fn();
+const mockNodeEnvFunction = jest.fn();
 
 jest.mock('../is-node', () => ({
-  isNodeEnvironment: mockFunction,
+  isNodeEnvironment: mockNodeEnvFunction,
 }));
 
 const OLD_ENV = process.env;
@@ -12,33 +12,26 @@ describe('Provider index.tsx', () => {
     process.env = { ...OLD_ENV }; // make a copy
     process.env.NODE_ENV = 'development';
   });
+
   afterAll(() => {
     process.env = OLD_ENV; // restore old env
   });
+
   it('returns the browser provider when isNodeEnvironment returns false', () => {
-    mockFunction.mockImplementation(() => false);
-    const loadComp = (path, name = 'default') => {
-      let comp = undefined;
-      try {
-        comp = require(path)[name];
-      } catch (_) {}
-      return comp;
-    };
-    const BrowserProvider = loadComp('../provider/provider-browser');
-    const Provider = loadComp('../provider');
+    mockNodeEnvFunction.mockImplementation(() => false);
+
+    const BrowserProvider = require('../provider/provider-browser').default;
+    const Provider = require('../provider').default;
+
     expect(Provider).toEqual(BrowserProvider);
   });
+
   it('returns the server provider when isNodeEnvironment returns true', () => {
-    mockFunction.mockImplementation(() => true);
-    const loadComp = (path, name = 'default') => {
-      let comp = undefined;
-      try {
-        comp = require(path)[name];
-      } catch (_) {}
-      return comp;
-    };
-    const ServerProvider = loadComp('../provider/provider-server');
-    const Provider = loadComp('../provider');
+    mockNodeEnvFunction.mockImplementation(() => true);
+
+    const ServerProvider = require('../provider/provider-server').default;
+    const Provider = require('../provider').default;
+
     expect(Provider).toEqual(ServerProvider);
   });
 });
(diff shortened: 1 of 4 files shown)


packages/runtime/src/provider/provider-server.tsx
```diff
@@ -1,7 +1,8 @@
 import React, { useRef, useContext, createContext } from 'react';
+import { isNodeEnvironment } from '../is-node';
 import { ProviderComponent, UseCacheHook } from './types';
 
-if (process.env.NODE_ENV === 'development' && typeof window !== 'undefined') {
+if (process.env.NODE_ENV === 'development' && isNodeEnvironment()) {
```
madou, line 5: you have these back to front

if this code runs on the browser - it should throw because it should only run on the server

vice versa for the other.

can you add some tests so we don't accidentally regress this in the future please? 🙏

packages/runtime/src/__tests__/provider.test.tsx
```diff
@@ -0,0 +1,126 @@
+const mockFunction = jest.fn();
```
madou, line 1: ah, right. bit of a wild thing to do for this, hehe.

packages/runtime/src/__tests__/provider.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+  });
+});
+
+describe('Browser Provider', () => {
+  beforeEach(() => {
+    jest.resetModules(); // most important - it clears the cache
+    process.env = { ...OLD_ENV }; // make a copy
+    process.env.NODE_ENV = 'development';
+  });
+  afterAll(() => {
```
madou, line 93: nit: can we have spaces between all function blocks

```
describe(() => {
  beforeEach();

  it();

  it();
});
```

packages/runtime/src/__tests__/provider.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+    process.env = OLD_ENV; // restore old env
+  });
+  it('should throw when "isNodeEnvironment" returns true', () => {
+    mockFunction.mockImplementation(() => true);
+
+    const loadComp = (path, name = 'default') => {
```
madou, line 99: do we need this function or we could just call require as is?

```
expect(() => require('../provider/provider-browser')).toThrowErrorMatchingSnapshot();
```

would simplify a lot of the test code

packages/runtime/src/__tests__/provider.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+  it('returns the browser provider when isNodeEnvironment returns false', () => {
+    mockFunction.mockImplementation(() => false);
+    const loadComp = (path, name = 'default') => {
+      let comp = undefined;
+      try {
+        comp = require(path)[name];
+      } catch (_) {}
+      return comp;
+    };
+    const BrowserProvider = loadComp('../provider/provider-browser');
```
madou, line 27: can we use require here instead of the load comp wrapper?

madou (commented):
great work! left some comments 😄

packages/runtime/src/__tests__/provider.test.tsx
MatthewCharlton: Didn't know you could pass in a function to expect. I thought it needed a result to test. Nice to know.
I've cleaned it up as per your comment.

packages/runtime/src/__tests__/provider.test.tsx
MatthewCharlton: Done

madou (approved):
sooo great!! good job matt!

packages/runtime/src/__tests__/provider.test.tsx
DarkPurple141: surely a lint rule we can add for this!

madou: want to add a pr pls

PR #355, opened 2020-11-10
fix: add null check when unhandled

pgmanutd:
We are not supporting complex `MemberExpression` object as of now. Only `Identifier` and `CallExpression` are supported as explained below:

 1. Member expression `foo.bar.baz` will return the `foo` identifier along with `originalBindingType` as `Identifier`.
 2. Member expression with function call `foo().bar.baz` will return the `foo` identifier along with `originalBindingType` as `CallExpression`.

### DIFF
--- a/packages/babel-plugin/src/__tests__/expression-evaluation.test.tsx
+++ b/packages/babel-plugin/src/__tests__/expression-evaluation.test.tsx
@@ -276,4 +276,19 @@ describe('import specifiers', () => {
       '{background-color:blue}',
     ]);
   });
+
+  it('should not blow up when member expression object is other than "Identifier" or "Call Expression"', () => {
+    expect(() => {
+      transform(`
+      import '@compiled/core';
+      import React from 'react';
+
+      function Component() {
+        return (
+          <span css={{ width: [{ bar: 10 }][0].bar }} />
+        );
+      };
+    `);
+    }).not.toThrow();
+  });
 });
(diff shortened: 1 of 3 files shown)


packages/babel-plugin/src/__tests__/expression-evaluation.test.tsx
```diff
@@ -276,4 +276,19 @@ describe('import specifiers', () => {
...
+      import '@compiled/core';
+      import React from 'react';
+
+      function Component() {
+        return (
+          <span css={{ width: [{ bar: 10 }][0].bar }} />
+        );
+      };
+    `);
+    }).not.toThrow();
```
madou, line 292: what comes out instead? a CSS variable/dynamic value?

pgmanutd: Let me check

pgmanutd: It should be a variable.

pgmanutd:
```ts
import{ax,CC,CS}from'@compiled/core';import React from'react';const _="._1bsb11sr{width:var(--_1mrga3f)}";function Component(){return<CC>
  <CS>{[_]}</CS>
  {<span className={ax(["_1bsb11sr"])}style={{"--_1mrga3f":[{bar:10}][0].bar}}/>}
</CC>;};
```

pgmanutd: I mean same variable as input. No conversion.

madou (approved):
lgtm

PR #361, opened 2020-11-11
Class names api

### LINKED ISSUE #342: ClassNames component
We need to re-implement the `ClassNames` component.

**1. Static**

```jsx
<ClassNames>
  {({ css }) => children({ className: css({ color: 'blue' }) })}
</ClassNames>
```

Becomes

```jsx
<CC>
  <CS>{['._123afs21 { color: blue }']}</CS>
  {children({ className: '_123afs21' })}
</CC>
```

**2. Dynamic**

```jsx
<ClassNames>
  {({ css, style }) => children({ className: css({ color: props.blue }), style })}
</ClassNames>
```

Becomes

```jsx
<CC>
  <CS>{['._123afs21 { color: var(--colorblue) }']}</CS>
  {children({ className: '_123afs21', style: { '--colorblue': props.blue } })}
</CC>
```

**3. Ax helper**

`ax` is already always imported so this should be easy to implement.

```jsx
<ClassNames>
  {({ css, ax, style }) => children({ className: ax([css({ color: props.blue })]), style })}
</ClassNames>
```

```jsx
import { ax } from '@compiled/core';

<CC>
  <CS>{['._123afs21 { color: var(--colorblue) }']}</CS>
  {children({ className: ax(['_123afs21']), style: { '--colorblue': props.blue } })}
</CC>
```

**Edge cases/testing notes**

- Static usage
- Dynamic usage
- TypeScript types for entire API (component, props)
- Destructured props - same name `({ css, style })`
- Destructured props - changed name `({ css: cs, style: stl })`
- Object access props `props.css, props.style`
- Array css prop `css([``, {}, myFunc(), mixin])`
- Template literal css prop `css```
- Object css prop `css({})`
- Identifier css prop `css(mixin)`
- Function css prop `css(myFunc())`

**Ignore**

- Reassigned variables `const css = props.css; css({})`

madou:
Closes #342 

**TODO**

- [x] Fix tests
- [x] Double check TS API
- [x] Destructured props - same name `({ css, style })`
- [x] Object access props `props.css`

**Later**

- [ ] Ax api in class names props https://github.com/atlassian-labs/compiled/issues/373
- [ ] Object access props `props.style` https://github.com/atlassian-labs/compiled/issues/374
- [ ] Object access props invalid (`dontexist.css`, `dontexist.style`) https://github.com/atlassian-labs/compiled/issues/374
- [ ] Destructured props - changed name `({ css: cs, style: stl })` https://github.com/atlassian-labs/compiled/issues/374


pgmanutd:
oooo love it...looking forward to it ❤️ Let me know if you want me to jump onto something 😛

packages/babel-plugin/src/class-names/__tests__/behaviour.test.tsx
```diff
@@ -22,86 +22,226 @@ describe('class names behaviour', () => {
...
+      const Component = ({
+        children,
+        color
+      }) => <CC>
+          <CS>{[_]}</CS>
+          {<div style={{
+          \\"--_1ylxx6h\\": color
+        }} className={\\"_syaz1aj3\\"} />}
+        </CC>;"
+    `);
```
pgmanutd, line 245: Uncommon case. See if we want to handle it. I checked. Its not working;

```ts
<ClassNames>
{(arg) => {
  const { css } = arg;

  return (
    <div
      className={css`
        font-size: ${props.size}px;
        color: red;
      `}>
      hello world
    </div>
  );
}}
</ClassNames>
```

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -44,7 +44,7 @@ export const getPathOfNode = <TNode extends {}>(
  * @param node
  * @param parentPath
  */
-export const buildCodeFrameError = (error: string, node: t.Node, parentPath: NodePath) => {
+export const buildCodeFrameError = (error: string, node: t.Node, parentPath: NodePath<any>) => {
```
pgmanutd, line 47: Was it giving error somewhere?

packages/babel-plugin/src/index.tsx
```diff
@@ -67,13 +68,19 @@ export default declare<State>((api) => {
...
-            state.compiledImports.styled = specifier.node.local.name;
-
-            // Remove specifier
-            specifier.remove();
-          }
+          (['styled', 'ClassNames'] as const).forEach((apiName) => {
+            if (
+              state.compiledImports &&
+              specifier.node &&
+              specifier.node.imported.name === apiName
```
pgmanutd, line 75: nit:
```suggestion
              state.compiledImports &&
              specifier.node?.imported.name === apiName
```

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -211,7 +211,7 @@ const styledTemplate = (opts: StyledTemplateOpts, meta: Metadata): t.Node => {
     }
   )({
     styleProp,
-    cssNode: t.arrayExpression(opts.sheets.map((sheet) => hoistSheet(sheet, meta))),
+    cssNode: t.arrayExpression(unique(opts.sheets).map((sheet) => hoistSheet(sheet, meta))),
```
pgmanutd, line 214: Q: How we figured out that we have to used `unique`? Any tests was failing? 🤔

packages/babel-plugin/src/class-names/index.tsx
```diff
@@ -1,6 +1,56 @@
 import * as t from '@babel/types';
 import { NodePath } from '@babel/core';
-import { State } from '../types';
+import { transformCss } from '@compiled/css';
+import { pickFunctionBody, buildCodeFrameError } from '../utils/ast';
+import { compiledTemplate, buildCssVariablesProp } from '../utils/ast-builders';
+import { buildCss, CSSOutput } from '../utils/css-builders';
+import { Metadata } from '../types';
+
+const extractStyles = (path: NodePath<t.Expression>) => {
```
pgmanutd, line 9: nit: JSDocs 😉

packages/babel-plugin/src/class-names/index.tsx
```diff
@@ -13,4 +63,55 @@ import { State } from '../types';
...
+  // First pass to replace all usages of `css({})`
+  path.traverse({
+    Expression(path) {
+      const styles = extractStyles(path);
+      if (!styles) {
+        // Nothing to do - skip.
+        return;
+      }
+
+      const builtCss = buildCss(styles, meta);
```
pgmanutd, line 88: Nice. API's are getting simpler now. Reusing existing utilities 😄

packages/babel-plugin/src/class-names/index.tsx
```diff
@@ -13,4 +63,55 @@ import { State } from '../types';
...
+export const visitClassNamesPath = (path: NodePath<t.JSXElement>, meta: Metadata) => {
+  if (
+    t.isJSXIdentifier(path.node.openingElement.name) &&
+    path.node.openingElement.name.name !== meta.state.compiledImports?.ClassNames
+  ) {
+    // We aren't interested in this element. Bail out!
+    return;
+  }
+
+  const children = getJsxChildrenAsFunction(path);
```
pgmanutd, line 75: Why we extracted children so early in the flow? 🤔 Can't we move it closer to its usage before line 155?

packages/babel-plugin/src/class-names/index.tsx
```diff
@@ -13,4 +63,55 @@ import { State } from '../types';
...
+      if (!styles) {
+        // Nothing to do - skip.
+        return;
+      }
+
+      const builtCss = buildCss(styles, meta);
+      const transformed = transformCss(builtCss.css);
+
+      variables = variables.concat(builtCss.variables);
+      sheets = sheets.concat(transformed.sheets);
```
pgmanutd, line 92: Can't we `push` here instead of `concat`? It will be faster.

pgmanutd:
```ts
const a = [1, 2, 3, 4, 5, 6];

a.push(...[1, 2, 3, 4, 5, 6, 7, 8, 9])
```

is faster than 

```ts
let a = [1, 2, 3, 4, 5, 6];

a.concat([1, 2, 3, 4, 5, 6, 7, 8, 9])
```

packages/babel-plugin/src/class-names/__tests__/behaviour.test.tsx
madou: nice, yeah i'll add this to the follow up ticket!

packages/babel-plugin/src/utils/ast.tsx
madou: it was yeah, complaining that a specific path wasnt a subclass of `t.Node` (when it was a `t.Expression`)

packages/babel-plugin/src/utils/ast-builders.tsx
madou: there was a test failing in class names behaviour https://github.com/atlassian-labs/compiled/blob/14ff96ac20d05d06fb9efea073a4896e358546d2/packages%2Fbabel-plugin%2Fsrc%2Fclass-names%2F__tests__%2Fbehaviour.test.tsx#L89

packages/babel-plugin/src/class-names/index.tsx
madou: nice! will do now

packages/babel-plugin/src/class-names/index.tsx
madou: yeah, amusingly not much code needed

packages/babel-plugin/src/class-names/index.tsx
madou: nice will do, always keep forgetting you can push multiple!

packages/babel-plugin/src/class-names/__tests__/behaviour.test.tsx
madou: found a edge case that isnt statically extracted

```js
      import { ClassNames } from '@compiled/react';

      const ListItem = () => (
        <ClassNames>
          {({ css }) => {
            const color = 'blue';
            return <div className={css({ fontSize: '20px', color })}>hello, world!</div>;
            // color ends up being a css variable
          }}
        </ClassNames>
      );
```

madou: made issue

madou:
pushed up @pgmanutd lmk how you go

pgmanutd (approved):
lgtm. Just test this with website once. See if everything works.

PR #364, opened 2020-11-13
chore: adds runtime benchmarking

madou:
Interestingly single string performs almost twice as fast. I'll do some tests to see what performs the best.

```
ax(...args: (string | undefined)[]);
ax(arr: (string | undefined)[]);
ax(first: string, second: string | undefined);
```


pgmanutd:
Are we memoizing the inputs. I did some benchmarking of `ax` and found out if we memoize based on `classnames.string()`, it definitely runs faster.

pgmanutd:
BTW how we are getting such performance boost as in image below 🤔? I built `lozenge`, though it didn't include prod build of React, compiled and emotion 😛 

image

madou:
> BTW how we are getting such performance boost as in image below 🤔? I built `lozenge`, though it didn't include prod build of React, compiled and emotion 😛
> 
> image

i believe compiled runs much better in dev - in prod its a different story which we've both seen

> Are we memoizing the inputs? I did some benchmarking of `ax` and found out if we memoize based on `classnames.string()`, it definitely runs faster.

not yet. only testing with calling the current ax differently 😄

madou:
added three different implementations, the static arg one blew everything else out of the water by... a hilarious amount

```
➜  compiled git:(runtime-benchmarks) yarn benchmark:ax
yarn run v1.22.4
$ cd benchmarks && yarn ax
$ ts-node  --files ./ax-benchmark
Start ax benchmarking

==>  axArr array x 204,776 ops/sec ±0.43% (89 runs sampled)
==>  axArr single string x 400,833 ops/sec ±1.24% (91 runs sampled)
==>  axArgs array x 206,797 ops/sec ±0.36% (90 runs sampled)
==>  axArgs single string x 394,438 ops/sec ±1.17% (91 runs sampled)
==>  axArgs static x 1,236,954 ops/sec ±0.91% (90 runs sampled)

Fastest is axArgs static
✨  Done in 32.82s.
➜  compiled git:(runtime-benchmarks) ✗
```

madou:
i made one slight mistake with the static func, updated here and it's not _that_ much better

```
➜  compiled git:(runtime-benchmarks) ✗ yarn benchmark:ax
yarn run v1.22.4
$ cd benchmarks && yarn ax
$ ts-node  --files ./ax-benchmark
Start ax benchmarking

==>  axArr array x 203,226 ops/sec ±0.40% (90 runs sampled)
==>  axArr single string x 398,883 ops/sec ±1.18% (91 runs sampled)
==>  axArgs array x 199,046 ops/sec ±0.30% (90 runs sampled)
==>  axArgs single string x 392,187 ops/sec ±0.29% (90 runs sampled)
==>  axArgs static x 405,991 ops/sec ±1.27% (92 runs sampled)

Fastest is axArgs static
✨  Done in 32.63s.
```

the main thing will be ensuring the majority of class names are in a single string instead of individual elements in the  array, however

madou:
current opportunities

## ax()

- [ ] memoize call (key based on `arr.join('')`)
- [ ] change transform to group as much as possible (from `['a', 'b', 'c']` to `['abc']`
- [ ] use for loops
- [ ] use for in instead of object values

## style component (ssr)

- [ ] order pseudos at build time instead of runtime
- [ ] for loops instead of filter
- [ ] memo re-renders (should never re-render)
- [ ] memoize initial render based on a hash (will need to hash sheets at build time and add as static prop to use, like previously)

--- 

will update more as i progress.

pgmanutd:
This is WIP PR right?

benchmarks/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@compiled/benchmarks",
+  "version": "1.0.0",
```
pgmanutd, line 3: `1.0.0` how 🤔?

benchmarks/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "private": true,
+  "scripts": {
+    "all": "yarn ax && yarn ssr && yarn stylesheet",
+    "ax": "NODE_ENV=production && ts-node ./src/ax/benchmark",
+    "ssr": "NODE_ENV=production && ts-node ./src/ssr/benchmark",
+    "stylesheet": "NODE_ENV=production && ts-node ./src/stylesheet/benchmark"
+  },
+  "devDependencies": {
```
pgmanutd, line 11: `dev` or `dep` 🤔 ?

benchmarks/package.json
madou: version doesnt matter tbh

benchmarks/package.json
madou: doesnt rly matter

pgmanutd (commented):
Can we add a READEME file for all the benchmarks with some numbers?

pgmanutd (approved):
lgtm

pgmanutd (approved):

PR #365, opened 2020-11-16
Runtime optimizations

madou:
This PR optimizes some of the HOT paths for the runtime. @pgmanutd can you see any more opportunities?

## Changes

**ax func**

- [x] change transform to group classNames as much as possible (from `['a', 'b', 'c']` to `['abc']`
- [x] use for loops
- [x] use for in instead of `Object.values`
- [x] short circuit if array length is 1

**stylesheet func**

- [x] use for loops

**style component**

- [x] for loops instead of filter
- [x] memo re-renders (should never re-render)

**css pkg**

- [x] remove duplicates in post css plugin (needed because of `ax` short circuit)

### DIFF
--- a/examples/compiled.tsx
+++ b/examples/compiled.tsx
@@ -1,41 +1,46 @@
 import * as React from 'react';
 import '@compiled/react';
+import { InteractionTaskArgs, PublicInteractionTask } from 'storybook-addon-performance';
+import { fireEvent, findAllByText } from '@testing-library/dom';
 
 export default {
   title: 'benchmarks | compiled',
 };
 
-export const Static = () => (
-  <span
-    css={{
-      backgroundColor: 'rgb(66, 82, 110)',
-      borderRadius: 3,
-      boxSizing: 'border-box',
-      color: 'rgb(255, 255, 255)',
-      display: 'inline-block',
-      fontSize: '11px',
-      fontWeight: 700,
-      lineHeight: 1,
-      maxWidth: '100%',
-      padding: '2px 0 3px 0',
-      textTransform: 'uppercase',
-      verticalAlign: 'baseline',
-    }}>
-    <span
-      css={{
-        display: 'inline-block',
-        verticalAlign: 'top',
-        overflow: 'hidden',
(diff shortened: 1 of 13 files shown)


madou:
```
  ./packages/runtime/dist/browser/index.js
  Package size limit has exceeded by 35 B
  Size limit: 99 B
  Size:       134 B with all dependencies, minified and gzipped
  
  ./packages/runtime/dist/browser/index.js
  Package size limit has exceeded by 18 B
  Size limit: 433 B
  Size:       451 B with all dependencies, minified and gzipped
```

pgmanutd:
- [x] Tried reverse while loop in `ax.tsx` and `style.tsx`. Reverse loop generates less instructions. But I think modern browsers are smart enough. (**Result**: No perf gain)
- [x] Group-by in babel-plugin (**Result**: Same perf coz we have to set `inserted[sheet]` for every atomic sheet)
- [ ] Thinking of reducing loop for atomic classes (big perf win). May be group by in babel-plugin and insert style rules (which are grouped by babel-plugin) directly instead of loop 🤔. Not sure if this is possible. Drawback is duplicate atomic classes in style tag.

madou:
> Thinking of reducing loop for atomic classes (big perf win)

insert style rule needs CSS rules, so we can't give it a list of rules as a single entry, e.g:

```
".a {} .b {} .c {}"
```

it would blow up

we also reuse atomic rules in scope to not create more than we need - i'd say let's not try to optimize this path for now.

pgmanutd:
> > Thinking of reducing loop for atomic classes (big perf win)
> 
> insert style rule needs CSS rules, so we can't give it a list of rules as a single entry, e.g:
> 
> ```
> ".a {} .b {} .c {}"
> ```
> 
> it would blow up
> 
> we also reuse atomic rules in scope to not create more than we need - i'd say let's not try to optimize this path for now.

Yups right. I was also skeptical if this is possible or not. I was just testing if there is way to reduce that loop count. But unfortunately no. Did testing last night but to no avail.

pgmanutd:
Cleaning and reopening another PR

PR #372, opened 2020-11-18
feat: adds comment to transformed files

madou:
Outstanding question: How can we mock `package.json` globally so we don't have to keep updating tests when new versions are released? Tried using `moduleNameMapper` to no avail.


pgmanutd:
> Outstanding question: How can we mock `package.json` globally so we don't have to keep updating tests when new versions are released? Tried using `moduleNameMapper` to no avail.

Let me see if anything we can do something about this. This will break tests everytime we release new version.

pgmanutd:
Only way (kind of hacky) I figured out is doing something like this:

```ts
const compiledBabelVersion = process.env.COMPILED_BABEL_VERSION || pkgJson.version;
```

and setting `process.env.COMPILED_BABEL_VERSION` in jest setup file.

madou:
> Only way (kind of hacky) I figured out is doing something like this:
> 
> ```ts
> const compiledBabelVersion = process.env.COMPILED_BABEL_VERSION || pkgJson.version;
> ```
> 
> and setting `process.env.COMPILED_BABEL_VERSION` in jest setup file.

it's shit but i'll do it, haha!

pgmanutd (approved):
lgtm

PR #376, opened 2020-11-19
fix: adds new line after comment

madou:


packages/babel-plugin/src/__tests__/index.test.tsx
```diff
@@ -8,6 +8,19 @@ const babelOpts: TransformOptions = {
 };
 
 describe('babel plugin', () => {
+  it('should not comment file if not transformation occured', () => {
```
pgmanutd, line 11: nit:
```suggestion
  it('should not comment file if no transformation occurred', () => {
```

packages/babel-plugin/src/index.tsx
```diff
@@ -50,6 +50,9 @@ export default declare<State>((api) => {
             'leading',
             ` ${parseFilename(state.filename)} generated by ${pkgJson.name} v${version} `
           );
+
+          // Add a line break the comment
+          path.unshiftContainer('body', t.noop());
```
pgmanutd, line 55: oooo `babel` utilities `t.noop`. Nice.

pgmanutd (approved):
lgtm

packages/babel-plugin/src/index.tsx
madou: yeah you find wild things going through the api.. not documented anywhere of course

PR #379, opened 2020-11-22
feat: conditional css rules for css prop

### LINKED ISSUE #278: Conditionally applied style rules
Because of our early pivot to atomic CSS at "baked in" time - we're empowered to support this kind of behaviour - which will work very consistently!

**Style declarations**

You can conditionally apply mixins like so:

```jsx
<div 
  css={{
    color: 'blue',
    ...props.isPrimary && { fontSize: 15, fontWeight: bold },
  }}
/>

<div 
  css={`
    color: blue;
    ${props.isPrimary && { fontSize: 15, fontWeight: bold }};
  `}
/>

styled.div`
  color: blue;
  ${props.isPrimary && { fontSize: 15, fontWeight: bold }};
`;

styled.div({
   color: 'blue',
  props => props.isPrimary && { fontSize: 15, fontWeight: bold },
});
```

Which would become (simplified):

```css
.color-blue { color:  blue }
.font-size-15 { font-size: 15px }
.font-weight-bold { font-weight: bold }
```

```jsx
<div className={ax(['color-blue', props.isPrimary && 'font-size-15 font-weight-bold'])} />
```

**Css prop array**

```jsx
const red = { color: 'red', textSize: 12 }; 
<div  css={[isDanger && red, { background: 'blue' }]} />;
```

```css
.color-red { color: red }
.background-blue { background: blue }
```

```jsx
<div className={ax([isDanger && 'color-red text-size-12', 'background-blue'])} />
```

Same principle would apply to the class names API `css` prop.

## Testing notes

- Logical expression single (`isPrimary && css`)
- Logical expression multiple (`isPrimary && selected && css`)
- Conditional expression (`isPrimary ? css : otherCss`)
- Array with expressions (`css([isPrimary ? css : otherCss])`)
- All above with CSS prop
- All above with Styled
- All above with ClassNames

madou:
Closes #278 

**Todo**

- [x] spread conditional
- [x] object conditional
- [x] nested conditional
- [x] multi conditional
- [x] array conditional

**Out of scope**

We'll have follow up tickets for these.

- Ternary styles `isPrimary ? { color: 'red' } : { color: 'blue' }`
- Styled component support
- Template literal support

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,6 +20,7 @@ module.exports = {
     '@typescript-eslint/no-namespace': 'off',
     '@typescript-eslint/explicit-function-return-type': 'off',
     '@typescript-eslint/no-explicit-any': 'off',
+    '@typescript-eslint/no-use-before-define': 'off',
     '@typescript-eslint/no-empty-function': 'off',
     '@typescript-eslint/explicit-member-accessibility': 'off',
     '@typescript-eslint/no-non-null-assertion': 'off',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,7 +20,6 @@ module.exports = {
     '@typescript-eslint/no-namespace': 'off',
     '@typescript-eslint/explicit-function-return-type': 'off',
     '@typescript-eslint/no-explicit-any': 'off',
-    '@typescript-eslint/no-use-before-define': 'off',
     '@typescript-eslint/no-empty-function': 'off',
     '@typescript-eslint/explicit-member-accessibility': 'off',
     '@typescript-eslint/no-non-null-assertion': 'off',
--- a/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
+++ b/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
@@ -391,7 +391,7 @@ describe('css prop behaviour', () => {
     expect(actual).toInclude(':hover{color:red}');
   });
 
-  xit('should apply conditional object spread styles', () => {
+  it('should apply conditional object spread styles', () => {
     const actual = transform(`
     import '@compiled/react';
     import React from 'react';
@@ -405,6 +405,6 @@ describe('css prop behaviour', () => {
     );
   `);
 
-    expect(actual).toInclude('{color: blue}');
+    expect(actual).toInclude('className={ax([props.isPrimary&&"_syaz13q2"])}');
   });
 });
--- a/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
+++ b/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
@@ -391,7 +391,7 @@ describe('css prop behaviour', () => {
     expect(actual).toInclude(':hover{color:red}');
   });
 
-  it('should apply conditional object spread styles', () => {
+  it('should apply conditional logical expression object spread styles', () => {
     const actual = transform(`
     import '@compiled/react';
     import React from 'react';
@@ -400,6 +400,25 @@ describe('css prop behaviour', () => {
       <div css={{
         ...props.isPrimary && {
           color: 'blue',
+        }
+      }}>hello world</div>
+    );
+  `);
+
+    expect(actual).toInclude('className={ax([props.isPrimary&&"_syaz13q2"])}');
+  });
+
+  xit('should apply conditional ternary expression object spread styles', () => {
+    const actual = transform(`
+    import '@compiled/react';
+    import React from 'react';
+
+    const Component = props => (
+      <div css={{
+        ...props.isPrimary ? {
+          color: 'blue',
+        } : {
+          color: 'red',
         }
       }}>hello world</div>
     );
--- a/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
+++ b/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
@@ -390,4 +390,21 @@ describe('css prop behaviour', () => {
 
     expect(actual).toInclude(':hover{color:red}');
   });
+
+  xit('should apply conditional object spread styles', () => {
+    const actual = transform(`
+    import '@compiled/react';
+    import React from 'react';
+
+    const Component = props => (
+      <div css={{
+        ...props.isPrimary && {
+          color: 'blue',
+        }
+      }}>hello world</div>
+    );
+  `);
+
+    expect(actual).toInclude('{color: blue}');
+  });
 });
--- a/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
+++ b/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
@@ -466,4 +466,25 @@ describe('css prop behaviour', () => {
 
     expect(actual).toInclude('ax([(props.isPrimary||props.isMaybe)&&"_syaz13q2 _1wybgktf"])');
   });
+
+  it('should apply array conditional css', () => {
+    const actual = transform(`
+      import '@compiled/react';
+      import React from 'react';
+
+      const Component = props => (
+        <div css={[
+          { fontSize: 40, },
+          (props.isPrimary || props.isMaybe) && {
+            color: 'blue',
+            fontSize: 20,
+          },
+        ]}>hello world</div>
+      );
+    `);
+
+    expect(actual).toInclude(
+      'ax(["_1wyb1ylp",(props.isPrimary||props.isMaybe)&&"_syaz13q2 _1wybgktf"])'
+    );
+  });
 });
--- a/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
+++ b/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
@@ -528,4 +528,22 @@ describe('css prop behaviour', () => {
       'ax(["_1wybo7ao",props.isPrimary&&"_syaz13q2 _1wybgktf","_19itjoc5"])'
     );
   });
+
+  it('should use destructured props in conditional css rule', () => {
+    const actual = transform(`
+      import '@compiled/react';
+      import React from 'react';
+
+      const Component = ({ isPrimary }) => (
+        <div css={{
+          ...isPrimary && {
+            color: 'blue',
+            fontSize: 20,
+          },
+        }}>hello world</div>
+      );
+    `);
+
+    expect(actual).toInclude('ax([isPrimary&&"_syaz13q2 _1wybgktf"])');
+  });
 });
--- a/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
+++ b/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
@@ -487,4 +487,45 @@ describe('css prop behaviour', () => {
       'ax(["_1wyb1ylp",(props.isPrimary||props.isMaybe)&&"_syaz13q2 _1wybgktf"])'
     );
   });
+
+  it('should apply partial  conditional css rule', () => {
+    const actual = transform(`
+      import '@compiled/react';
+      import React from 'react';
+
+      const Component = props => (
+        <div css={{
+          ...props.isPrimary && {
+            color: 'blue',
+            fontSize: 20,
+          },
+          border: '1px solid black',
+        }}>hello world</div>
+      );
+    `);
+
+    expect(actual).toInclude('ax([props.isPrimary&&"_syaz13q2 _1wybgktf","_19itjoc5"])');
+  });
+
+  it('should apply unconditional before and after a conditional css rule', () => {
+    const actual = transform(`
+      import '@compiled/react';
+      import React from 'react';
+
+      const Component = props => (
+        <div css={{
+          fontSize: 15,
+          ...props.isPrimary && {
+            color: 'blue',
+            fontSize: 20,
+          },
+          border: '1px solid black',
+        }}>hello world</div>
+      );
+    `);
+
+    expect(actual).toInclude(
+      'ax(["_1wybo7ao",props.isPrimary&&"_syaz13q2 _1wybgktf","_19itjoc5"])'
+    );
+  });
 });
--- a/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
+++ b/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
@@ -393,37 +393,77 @@ describe('css prop behaviour', () => {
 
   it('should apply conditional logical expression object spread styles', () => {
     const actual = transform(`
-    import '@compiled/react';
-    import React from 'react';
+      import '@compiled/react';
+      import React from 'react';
 
-    const Component = props => (
-      <div css={{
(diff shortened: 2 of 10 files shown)


.eslintrc.js
```diff
@@ -19,6 +19,7 @@ module.exports = {
     '@typescript-eslint/no-namespace': 'off',
     '@typescript-eslint/explicit-function-return-type': 'off',
     '@typescript-eslint/no-explicit-any': 'off',
+    '@typescript-eslint/no-use-before-define': 'off',
```
pgmanutd, line 22: Thank god!! We are getting rid of this rule 😅. Can we remove all the comments related to this rule?

madou: yeah - i'll put another PR up to do that

madou:
@pgmanutd if ur feeling keen have a look at this :) still WIP

the architecture i think needs to change a little to support ternary statements.. but im also open to worrying about that later.

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -28,15 +41,59 @@ const normalizeContentValue = (value: string) => {
...
+ *
+ * Output:
+ *
+ * ```
+ * [{ type: 'unconditional', css: 'color: blue; font-size: 20px' }]
+ * ```
+ *
+ * @param arr
+ */
+const reduceUnconditionalCssItems = (arr: Array<CssItem>): Array<CssItem> => {
```
madou, line 61: i feel this is the worst part of the implementation ATM - but it is what it is... 🤔 any thoughts?

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -121,22 +175,25 @@ const extractObjectExpression = (node: t.ObjectExpression, meta: Metadata): CSSO
...
-      return css + q.value.raw + interpolation.value;
+      return acc + q.value.raw + interpolation.value;
     }
 
     if (t.isObjectExpression(interpolation)) {
       // We found an object like: css`${{ red: 'blue' }}`.
-      const result = extractObjectExpression(interpolation, updatedMeta);
-      variables = variables.concat(result.variables);
-      return css + result.css;
+      const result = buildCss(interpolation, updatedMeta);
```
madou, line 193: you'll notice a lot of call sites have changed from an explitiy `extractXyz` to `buildCss`

generic win!

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -173,13 +230,15 @@ const extractTemplateLiteral = (node: t.TemplateLiteral, meta: Metadata): CSSOut
       nextQuasis.value.raw = after.css; // Removes any suffixes from the next quasis.
       variables.push({ name: variableName, expression });
 
-      return before.css + `var(${variableName})`;
+      return acc + before.css + `var(${variableName})`;
```
madou, line 233: interestingly this was a bug fix.... how did this not fail before!

packages/jest/src/matchers.tsx
```diff
@@ -22,14 +22,24 @@ const findMediaRules = (
...
   for (const rule of allRules) {
-    if (!rule) return;
+    if (!rule) {
+      continue;
+    }
+
     if ('media' in rule) {
-      if (removeSpaces(rule.media) === removeSpaces(media) && 'rules' in rule) return rule.rules;
-      if ('rules' in rule) return findMediaRules(rule.rules, media);
+      if (removeSpaces(rule.media) === removeSpaces(media) && 'rules' in rule && rule.rules) {
```
madou, line 33: bug fix

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
```diff
@@ -390,4 +390,101 @@ describe('css prop behaviour', () => {
...
+  });
+
+  it('should apply multi conditional logical expression', () => {
+    const actual = transform(`
+      import '@compiled/react';
+      import React from 'react';
+
+      const Component = props => (
+        <div css={{
+          ...(props.isPrimary || props.isMaybe) && {
```
pgmanutd, line 459: Noice ❤️

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
```diff
@@ -390,4 +390,101 @@ describe('css prop behaviour', () => {
...
+
+  it('should apply array conditional css', () => {
+    const actual = transform(`
+      import '@compiled/react';
+      import React from 'react';
+
+      const Component = props => (
+        <div css={[
+          { fontSize: 40, },
+          (props.isPrimary || props.isMaybe) && {
```
pgmanutd, line 478: nit: Can we have example where `isPrimary` and `isMaybe` are coming from destructing props?
`const Component =({ isPrimary, isMaybe }) => (`

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
```diff
@@ -390,4 +390,101 @@ describe('css prop behaviour', () => {
...
+        <div css={{
+          ...props.isPrimary && {
+            color: 'blue',
+            fontSize: 20,
+          }
+        }}>hello world</div>
+      );
+    `);
+
+    expect(actual).toInclude('className={ax([props.isPrimary&&"_syaz13q2 _1wybgktf"])}');
```
pgmanutd, line 409: This is the place where compilation can't help much because `isPrimary` won't be known until runtime. Love to compact this `[props.isPrimary&&"_syaz13q2 _1wybgktf"]` into `[]` if `props.isPrimary` evaluates to `false`. But its not possible. Though bundlers can do some magic if `isPrimary` is statically defined like `<Component  isPrimary />`. Not sure.

packages/babel-plugin/src/utils/css-builders.tsx
pgmanutd: nit: description says merge and function name starts with reduce. Kinda confusing.

pgmanutd: > From what I understand from the implementation.

Might be a performance penalty but will clean the code:

- partition into two arrays using `filter` with check: `item.type === 'unconditional'`
- reduce `unconditional` array into an object `{ type: 'unconditional', css: 'color: blue; font-size: 20px' }` by concatenating `css`
- finally push this object into the other partitioned array
```ts
arrayWithoutUnconditionaItems.push({ type: 'unconditional', css: 'color: blue; font-size: 20px' })
```

Well not worse in terms of performance:
O(n + m) in first step
O(n) in second step
O(1) in last step

won't take much space either.

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -28,15 +41,59 @@ const normalizeContentValue = (value: string) => {
...
+ */
+const reduceUnconditionalCssItems = (arr: Array<CssItem>): Array<CssItem> => {
+  let unconditionalItem: UnconditionalCssItem | undefined;
+
+  return arr.reduce<CssItem[]>((acc, item, index) => {
+    if (!unconditionalItem && item.type === 'unconditional') {
+      unconditionalItem = item;
+    }
+
+    if (index === 0 || item.type !== 'unconditional' || !unconditionalItem) {
```
pgmanutd, line 69: This might fail in below scenario when `conditional` type comes first.
```ts
[{ type: 'conditional', css: 'color: blue' }, { type: 'unconditional', css: 'font-size: 20px' }]
```

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -322,6 +322,31 @@ export const getPropValue = (
...
+
+  cssOutput.css.forEach((item) => {
+    const css = transformCss(getItemCss(item));
+    const className = css.classNames.join(' ');
+
+    sheets.push(...css.sheets);
+
+    if (item.type === 'logical') {
+      classNames.push(t.logicalExpression('&&', item.expression, t.stringLiteral(className)));
+    } else if (item.type === 'unconditional') {
```
pgmanutd, line 342: nit: `unconditional` is fallback case. We can remove if check or probably have switch with last case as `unconditional` without break and logic falling to default case. Actually when I read this, I thought we are skipping other item types. Then I have to jump to the the file where item's type is defined 😅

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -7,8 +7,21 @@ import { Metadata } from '../types';
...
+  css: string;
+}
+
+interface LogicalCssItem {
+  type: 'logical';
+  expression: t.Expression;
   css: string;
+}
+
+type CssItem = UnconditionalCssItem | LogicalCssItem;
```
pgmanutd, line 21: nit: can be moved to types file.

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -55,17 +112,17 @@ const extractObjectExpression = (node: t.ObjectExpression, meta: Metadata): CSSO
...
+        const result = buildCss(propValue, updatedMeta);
+        css.push(...result.css.map((x) => ({ ...x, css: `${key} { ${getItemCss(x)} }` })));
+        variables.push(...result.variables);
         return;
       } else if (t.isTemplateLiteral(propValue)) {
         // We've found a template literal like: `fontSize: `${fontSize}px`
         const result = extractTemplateLiteral(propValue, updatedMeta);
-        value = result.css;
-        variables = variables.concat(result.variables);
+        css.push(...result.css.map((x) => ({ ...x, css: `${kebabCase(key)}: ${getItemCss(x)};` })));
```
pgmanutd, line 123: nit: lot of places now have `map` calls now along with `getItem` inside it. Can we extract it out if possible. No worries if its not possible atm. We can leave it.

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -55,17 +112,17 @@ const extractObjectExpression = (node: t.ObjectExpression, meta: Metadata): CSSO
...
         // We've found a numeric literal like: `fontSize: 12`
         value = addUnitIfNeeded(key, propValue.value);
-      } else if (t.isObjectExpression(propValue)) {
-        // We've found a nested object like: `':hover': { color: 'red' }`
-        const result = extractObjectExpression(propValue, updatedMeta);
-        css += `${key} { ${result.css} }`;
-        variables = variables.concat(result.variables);
+      } else if (t.isObjectExpression(propValue) || t.isLogicalExpression(propValue)) {
+        const result = buildCss(propValue, updatedMeta);
+        css.push(...result.css.map((x) => ({ ...x, css: `${key} { ${getItemCss(x)} }` })));
```
pgmanutd, line 117: nit: A comment above will be beneficial. Explains what type of scenario we are handling here. Extracting this into a function will be like cherry on top. hehe. But lets have a comment for now.

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -121,22 +175,25 @@ const extractObjectExpression = (node: t.ObjectExpression, meta: Metadata): CSSO
...
     }
 
     if (t.isObjectExpression(interpolation)) {
       // We found an object like: css`${{ red: 'blue' }}`.
-      const result = extractObjectExpression(interpolation, updatedMeta);
-      variables = variables.concat(result.variables);
-      return css + result.css;
+      const result = buildCss(interpolation, updatedMeta);
+      variables.push(...result.variables);
+      css.push(...result.css);
```
pgmanutd, line 195: nit: We have lots of `variables.push(...result.variables);` and `css.push(...result.css);` now. Can we extract them into a function which gives us final result? If its getting complex, we can skip this.

packages/babel-plugin/src/utils/css-builders.tsx
pgmanutd: Fug!

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
```diff
@@ -390,4 +390,101 @@ describe('css prop behaviour', () => {
...
+            fontSize: 20,
+          },
+        ]}>hello world</div>
+      );
+    `);
+
+    expect(actual).toInclude(
+      'ax(["_1wyb1ylp",(props.isPrimary||props.isMaybe)&&"_syaz13q2 _1wybgktf"])'
+    );
+  });
```
pgmanutd, line 489: This case is not working:

```ts
const Component = props => (
  <div css={{
    ...props.isPrimary && {
      color: 'blue',
      fontSize: 20,
    },
    border: '1px solid black'
  }}>hello world</div>
);
```

Getting this output:
```ts
/* File generated by @compiled/babel-plugin v0.0.0 */
    
import { ax, CC, CS } from "@compiled/react/runtime";
const _2 = "._1wybgktf{font-size:20px}";
const _ = "._syaz13q2{color:blue}";

const Component = props => <CC>
    <CS>{[_, _2]}</CS>
    {<div className={ax([props.isPrimary && "_syaz13q2 _1wybgktf"])}>hello world</div>}
  </CC>;
```

but if we put `border: '1px solid black'` before condition `props.isPrimary`, it works

pgmanutd (commented):
Nice work ❤️. Few nits and I think one bug. Check once.

packages/babel-plugin/src/utils/css-builders.tsx
madou: ah yeah, that would be a more readable wouldn't it. i'll have a play

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
madou: yes

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
madou: yeah thats a head scratcher - we could always put in extra code to see if any code resolves to false but it's questionable if it's worth the effort.

maybe folder for later?

packages/babel-plugin/src/utils/css-builders.tsx
madou: yeah, i'll rework the function to your suggestion

packages/babel-plugin/src/utils/css-builders.tsx
madou: the get item css was really only needed because i started investigating the ternary implementation and then needed that because the types were slightly different

this one is slightly different because this is a css declaration - whereas the other is a css rule 🤔 

we could make two functions.. let me have a play

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
pgmanutd: yeah later only. I was just discussing whether this is possible or not.

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
madou: nice catch

pgmanutd (approved):
lgtm

PR #381, opened 2020-11-24
fix: fixes ssr hydration warnings

madou:


packages/react/src/__tests__/server-side-hydrate.test.tsx
```diff
@@ -2,39 +2,52 @@ import React from 'react';
 import { renderToString } from 'react-dom/server';
 import ReactDOM from 'react-dom';
 import { styled } from '@compiled/react';
+import { isNodeEnvironment } from '@compiled/react/dist/runtime/is-node';
+import { useCache } from '@compiled/react/dist/runtime/provider';
+
+jest.mock('@compiled/react/dist/runtime/is-node');
```
pgmanutd, line 8: ahh alternative to `delete process.release.name`. Only thing looks weird is `'@compiled/react/dist/runtime/is-node'` hehe

packages/react/src/__tests__/server-side-hydrate.test.tsx
```diff
@@ -2,39 +2,52 @@ import React from 'react';
...
 
 describe('server side hydrate', () => {
-  it('should not log unexpected warnings when hydrating HTML', () => {
-    jest.spyOn(global.console, 'error').mockImplementation(() => {});
+  beforeEach(() => {
+    jest.spyOn(global.console, 'error');
+    flushEnvironment('node');
```
pgmanutd, line 13: `flushEnvironment` nice ❤️

packages/react/src/runtime/sheet.tsx
```diff
@@ -123,15 +123,28 @@ export const getStyleBucketName = (sheet: string): Bucket => {
...
  * Returns a style sheet object that is used to move styles to the head of the application
  * during runtime.
  *
  * @param opts StyleSheetOpts
  * @param inserted Singleton cache for tracking what styles have already been added to the head
  */
-export default function createStyleSheet(opts: StyleSheetOpts) {
-  return (css: string) => {
+export default function createStyleSheet() {
+  // Two step process for SSR.
```
pgmanutd, line 139: Nice. So we did move it during js execution phase. Even before layout phase.

packages/react/src/runtime/sheet.tsx
```diff
@@ -141,5 +154,18 @@ export default function createStyleSheet(opts: StyleSheetOpts) {
...
+      // as their client side counterparts have been applied above.
+      const ssrStyles = document.querySelectorAll('style[data-cmpld]');
+
+      for (let i = 0; i < ssrStyles.length; i++) {
+        const styleElement = ssrStyles[i];
+        styleElement.remove();
+      }
+
+      ssrStyleElementsCleaned = true;
+    }
```
pgmanutd, line 169: Bundlesize is going up. See if we can reduce it a bit. These steps are repeated:

```ts
const ssrStyles = document.querySelectorAll('style[data-cmpld]');

for (let i = 0; i < ssrStyles.length; i++) {
  const styleElement = ssrStyles[i];
}
```

pgmanutd: I think minifiers are smart enough to figure out such repeating, but still lets check once.

packages/react/src/runtime/style.tsx
```diff
@@ -47,23 +48,19 @@ function Style(props: StyleProps) {
       }
 
       return (
-        <style nonce={props.nonce}>
+        <style data-cmpld nonce={props.nonce}>
```
pgmanutd, line 51: Any short alternative to `data-cmpld`? `data-c`?

madou: i wanted something that wouldnt be used by somewhere else - we could go down to three? `data-cmp`

madou:
realized this needed some work - there will be a flash of unstyled elements briefly. let me fix.

madou:
@pgmanutd you can review this now

one current caveat is we don't clean up SSRd styles which means they appear duplicated (pull this branch locally at run `yarn start:ssr` to see

im not sure of the best way to handle it atm.

package.json
```diff
@@ -29,7 +30,8 @@
     "version": "bash scripts/version.sh && git add .",
     "version-nightly": "bash scripts/should-nightly-run.sh && npm version prerelease --no-git-tag-version --preid=$(git rev-parse HEAD)",
     "postversion": "bash scripts/push-if-git.sh",
-    "bundlesize": "yarn build && yarn build:dead-code-elimination && size-limit"
+    "bundlesize": "yarn build && yarn build:dead-code-elimination && size-limit",
+    "postinstall": "npx yarn-deduplicate"
```
pgmanutd, line 34: I was getting type errors when I ran this in the repo earlier. I thought it broke some of the packages.

packages/babel-plugin/src/index.tsx
```diff
@@ -122,7 +122,11 @@ export default declare<State>((api) => {
           }
 
           (['styled', 'ClassNames'] as const).forEach((apiName) => {
-            if (state.compiledImports && specifier.node?.imported.name === apiName) {
+            if (
+              state.compiledImports &&
+              t.isIdentifier(specifier.node?.imported) &&
```
pgmanutd, line 127: 🤔 Was it breaking types somewhere?

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -124,7 +124,7 @@ const extractTemplateLiteral = (node: t.TemplateLiteral, meta: Metadata): CSSOut
   let variables: CSSOutput['variables'] = [];
   // quasis are the string pieces of the template literal - the parts around the interpolations.
   const css = node.quasis.reduce((css, q, index) => {
-    const nodeExpression = node.expressions[index];
+    const nodeExpression: t.Expression = node.expressions[index] as t.Expression;
```
pgmanutd, line 127: 😅

packages/babel-plugin/src/index.tsx
madou: i think the types were improved haha

packages/react/src/runtime/style.tsx
pgmanutd: Not sure..lets stick to `data-cmpld` as of now.

packages/react/src/runtime/__tests__/style-ssr.test.tsx
```diff
@@ -1,18 +1,28 @@
...
   it('should render style as children on the server with nonce', () => {
     const result = renderToStaticMarkup(<Style nonce="1234">{[`.a { display: block; }`]}</Style>);
 
-    expect(result).toInclude('<style nonce="1234">.a { display: block; }</style>');
+    expect(result).toInclude(
+      '<style data-cmpld="true" nonce="1234">.a { display: block; }</style>'
+    );
+  });
+
+  it('should not output html comments', () => {
```
pgmanutd, line 20: Might be dumb to ask but where this test came from? 🤔

package.json
```diff
@@ -93,7 +96,7 @@
     },
     {
       "path": "./packages/react/dist/browser/runtime.js",
-      "limit": "451B",
+      "limit": "494B",
```
pgmanutd, line 99: 😕

package.json
```diff
@@ -10,8 +10,9 @@
     "benchmark:ssr": "cd benchmarks && yarn ssr",
     "benchmark:stylesheet": "cd benchmarks && yarn stylesheet",
     "benchmark:module-traversal": "cd packages/babel-plugin && yarn benchmark:module-traversal",
-    "clean": "rm -rf node_modules/.cache && rm -rf test/dead-code-elimination/dist && rm -f test/dead-code-elimination/tsconfig.tsbuildinfo && workspaces-run -- rm -rf dist -- rm -f tsconfig.tsbuildinfo -- rm -f tsconfig.browser.tsbuildinfo",
+    "clean": "rm -rf node_modules/.cache && rm -rf test/dead-code-elimination/dist && rm -f test/dead-code-elimination/tsconfig.tsbuildinfo && workspaces-run -- rm -rf dist -- rm -f tsconfig.tsbuildinfo -- rm -f tsconfig.browser.tsbuildinfo -- rm -rf build",
     "start": "yarn build:main && npx nodemon --exec \"start-storybook -p 6006 --ci\" --watch packages/babel-plugin/ -e tsx",
+    "start:ssr": "cd playground/ssr && yarn start",
```
pgmanutd, line 15: This requires `yarn build` otherwise it will throw error if someone runs `yarn clean` and then this command.

packages/react/src/runtime/provider/provider-browser.tsx
```diff
@@ -25,9 +25,28 @@ if (process.env.NODE_ENV === 'development' && isNodeEnvironment()) {
...
+ */
+const forEachSSRdStyleElement = (cb: (element: HTMLStyleElement) => void) => {
+  const ssrStyles = document.querySelectorAll<HTMLStyleElement>('style[data-cmpld]');
+  for (let i = 0; i < ssrStyles.length; i++) {
+    cb(ssrStyles[i]);
+  }
+};
+
+// Move all found server-side rendered style elements to the head before React hydration happens.
+forEachSSRdStyleElement((styleElement) => {
```
pgmanutd, line 40: Thanks for moving this here ❤️. This gives clear picture that this runs before React actually kicks in.

package.json
madou: yeah i fixed em

packages/react/src/runtime/__tests__/style-ssr.test.tsx
madou: when i was running ssr i noticed the html has html comments in it! because we were setting children as an array

i added this with the fix

package.json
madou: yeah not great

package.json
madou: cool i'll add it

packages/react/src/runtime/style.tsx
```diff
@@ -38,32 +35,27 @@ function Style(props: StyleProps) {
...
 
       if (!hasSheets) {
         return null;
       }
 
       return (
-        <style nonce={props.nonce}>
-          {styleBucketOrdering.map((bucket) => bucketedSheets[bucket])}
+        <style data-cmpld nonce={props.nonce}>
+          {styleBucketOrdering.map((bucket) => bucketedSheets[bucket]).join('')}
```
madou, line 47: @pgmanutd this is the SSR html comments fix part 2

packages/react/src/runtime/style.tsx
```diff
@@ -38,32 +35,27 @@ function Style(props: StyleProps) {
         }
 
         const bucketName = getStyleBucketName(sheet);
-        bucketedSheets[bucketName] = bucketedSheets[bucketName] || [];
-        bucketedSheets[bucketName]!.push(sheet);
+        bucketedSheets[bucketName] = (bucketedSheets[bucketName] || '') + sheet;
```
madou, line 38: @pgmanutd this is the SSR html comments fix part 1

pgmanutd:
> @pgmanutd you can review this now
> 
> one current caveat is we don't clean up SSRd styles which means they appear duplicated (pull this branch locally at run `yarn start:ssr` to see
> 
> image
> 
> im not sure of the best way to handle it atm.

OOPS missed this comment. Let me check.

pgmanutd:
> > @pgmanutd you can review this now
> > one current caveat is we don't clean up SSRd styles which means they appear duplicated (pull this branch locally at run `yarn start:ssr` to see
> > image
> > im not sure of the best way to handle it atm.
> 
> OOPS missed this comment. Let me check.

Thats too many duplication 😞. Can we clean them if possible? 🤔 But how? hmm May be just after render? `useLayoutEffect` or `useEffect`? I prefer `useLayoutEffect`. We already have `data-cmpld` attached to style tags.

Rough implementations can be as below. I tested it. Its working. Not sure if its the right approach.

```ts
const useIsomorphicLayoutEffect = isNodeEnvironment() ? () => {} : useLayoutEffect; // <=========== Here

interface StyleProps extends StyleSheetOpts {
  /**
   * CSS Rules.
   * Ensure each rule is a separate element in the array.
   */
  children: string[];
}

let cleanedupSSR = false; // <=========== Here

function Style(props: StyleProps) {
  const inserted = useCache();

  if (process.env.NODE_ENV === 'development') {
    props.children.forEach(analyzeCssInDev);
  }

  useIsomorphicLayoutEffect(() => { // <=========== Here
    if (!cleanedupSSR) {
      const ssrStyles = document.head.querySelectorAll<HTMLStyleElement>('style[data-cmpld]');
      
      for (let i = 0; i < ssrStyles.length; i++) {
        ssrStyles[i].remove();
      }

      cleanedupSSR = true;
    }
  }, []);

  if (props.children.length) { ...
```

madou:
> > > @pgmanutd you can review this now
> > > one current caveat is we don't clean up SSRd styles which means they appear duplicated (pull this branch locally at run `yarn start:ssr` to see
> > > image
> > > im not sure of the best way to handle it atm.
> > 
> > 
> > OOPS missed this comment. Let me check.
> 
> Thats too many duplication 😞. Can we clean them if possible? 🤔 But how? hmm May be just after render? `useLayoutEffect` or `useEffect`? I prefer `useLayoutEffect`. We already have `data-cmpld` attached to style tags.
> 
> Rough implementations can be as below. I tested it. Its working. Not sure if its the right approach.
> 
> ```ts
> const useIsomorphicLayoutEffect = isNodeEnvironment() ? () => {} : useLayoutEffect; // <=========== Here
> 
> interface StyleProps extends StyleSheetOpts {
>   /**
>    * CSS Rules.
>    * Ensure each rule is a separate element in the array.
>    */
>   children: string[];
> }
> 
> let cleanedupSSR = false; // <=========== Here
> 
> function Style(props: StyleProps) {
>   const inserted = useCache();
> 
>   if (process.env.NODE_ENV === 'development') {
>     props.children.forEach(analyzeCssInDev);
>   }
> 
>   useIsomorphicLayoutEffect(() => { // <=========== Here
>     if (!cleanedupSSR) {
>       const ssrStyles = document.head.querySelectorAll<HTMLStyleElement>('style[data-cmpld]');
>       
>       for (let i = 0; i < ssrStyles.length; i++) {
>         ssrStyles[i].remove();
>       }
> 
>       cleanedupSSR = true;
>     }
>   }, []);
> 
>   if (props.children.length) { ...
> ```

this one is tricky because if we did this there would be a chance of flash of unstyled elements (we could remove SSRd styles before we've added them back)

emotion gets around this by hydrating the styles rendered via SSR - we could potentially do the same thing (pumping it through `sheet.tsx`)

the only tricky thing is splitting up the CSS rules in a performant manner.

pgmanutd:
> > > > @pgmanutd you can review this now
> > > > one current caveat is we don't clean up SSRd styles which means they appear duplicated (pull this branch locally at run `yarn start:ssr` to see
> > > > image
> > > > im not sure of the best way to handle it atm.
> > > 
> > > 
> > > OOPS missed this comment. Let me check.
> > 
> > 
> > Thats too many duplication 😞. Can we clean them if possible? 🤔 But how? hmm May be just after render? `useLayoutEffect` or `useEffect`? I prefer `useLayoutEffect`. We already have `data-cmpld` attached to style tags.
> > Rough implementations can be as below. I tested it. Its working. Not sure if its the right approach.
> > ```ts
> > const useIsomorphicLayoutEffect = isNodeEnvironment() ? () => {} : useLayoutEffect; // <=========== Here
> > 
> > interface StyleProps extends StyleSheetOpts {
> >   /**
> >    * CSS Rules.
> >    * Ensure each rule is a separate element in the array.
> >    */
> >   children: string[];
> > }
> > 
> > let cleanedupSSR = false; // <=========== Here
> > 
> > function Style(props: StyleProps) {
> >   const inserted = useCache();
> > 
> >   if (process.env.NODE_ENV === 'development') {
> >     props.children.forEach(analyzeCssInDev);
> >   }
> > 
> >   useIsomorphicLayoutEffect(() => { // <=========== Here
> >     if (!cleanedupSSR) {
> >       const ssrStyles = document.head.querySelectorAll<HTMLStyleElement>('style[data-cmpld]');
> >       
> >       for (let i = 0; i < ssrStyles.length; i++) {
> >         ssrStyles[i].remove();
> >       }
> > 
> >       cleanedupSSR = true;
> >     }
> >   }, []);
> > 
> >   if (props.children.length) { ...
> > ```
> 
> this one is tricky because if we did this there would be a chance of flash of unstyled elements (we could remove SSRd styles before we've added them back)
> 
> emotion gets around this by hydrating the styles rendered via SSR - we could potentially do the same thing (pumping it through `sheet.tsx`)
> 
> the only tricky thing is splitting up the CSS rules in a performant manner.

Damn tricky. I totally forgot you removed step-2 before. hehe. mmm. Somehow if we can figure out last `Style` rendering on the page and then remove them. Other than that, removing each style from head when their client counterpart are moved to head will not be performant at all.
Last `Style` rendering on the page can do the trick. But we need some sort of count before. And thats difficult. Damn. So what should we do? Create a card for investigation or leave it as it is?

pgmanutd (approved):
lgtm. Lets have one card for that duplicate styles issue.

madou:
done https://github.com/atlassian-labs/compiled/issues/388

PR #383, opened 2020-11-24
Use cache directly without wrapping it into a ref

Andarist:


madou:
nice! remove that code

madou (approved):
lgtm

PR #385, opened 2020-12-02
Update cli run instructions

pgmanutd:
@madou: See if this PR makes sense. Make cli instructions more clear.

### DIFF
--- a/packages/babel-plugin/src/__tests__/index.test.tsx
+++ b/packages/babel-plugin/src/__tests__/index.test.tsx
@@ -148,7 +148,7 @@ describe('babel plugin', () => {
       import { CC as CompiledRoot, ax } from '@compiled/react/runtime';
       import '@compiled/react';
 
-      const classes = ax['1', '2'];
+      const classes = ax(['1', '2']);
 
       <CompiledRoot>
         <div css={{ display: 'block' }}  />
@@ -163,7 +163,7 @@ describe('babel plugin', () => {
       import * as React from 'react';
       import { CC as CompiledRoot, ax, CC, CS } from '@compiled/react/runtime';
       const _ = \\"._1e0c1ule{display:block}\\";
-      const classes = ax[('1', '2')];
+      const classes = ax(['1', '2']);
       <CompiledRoot>
               <CC>
           <CS>{[_]}</CS>
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -24,8 +24,10 @@
     "dist",
     "src"
   ],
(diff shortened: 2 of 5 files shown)


packages/babel-plugin/src/__tests__/index.test.tsx
```diff
@@ -148,7 +148,7 @@ describe('babel plugin', () => {
       import { CC as CompiledRoot, ax } from '@compiled/react/runtime';
       import '@compiled/react';
 
-      const classes = ax['1', '2'];
+      const classes = ax(['1', '2']);
```
madou, line 151: hehe, nice

packages/cli/package.json
```diff
@@ -24,8 +24,10 @@
     "dist",
     "src"
   ],
+  "peerDependencies": {
+    "@compiled/react": ">= 0.4.7"
```
madou, line 28: ohhh - was it using its own dependency before?

packages/react/src/codemods/emotion-to-compiled/README.md
```diff
@@ -59,4 +59,6 @@ const Component = (props) => (
 
 ## Updating Instructions
 
-When a new version of `@compiled/react` is released, Please update `@compiled/react` and `@compiled/cli` to the same new version simultaneously.
+- Bump up `@compiled/react` to x.x.x version.
```
madou, line 62: Can we give an example instead, e,g:

```
When wanting to update to a later version make sure both `@compiled/cli` is being ran with the same version -
watch out for it being cached!

For example when upgrading `@compiled/react` to `v0.6.0` where you've already used the CLI,
on your next run explicitly set the version number:

npx @compiled/cli@0.6.0 --preset codemods
```

packages/cli/package.json
pgmanutd: Not exactly. It was not used. I just added it. It was anyways taking from cwd and that will be the directory where we are running this cli. So made it peer so that it ll warn if user tries to install cli in the repo. Otherwise if they are using `npx`, they can go through the Readme file for the instructions.

packages/react/src/codemods/emotion-to-compiled/README.md
pgmanutd: Not sure if it should go in `@compiled/cli`. Thats just generic cli with different presets. I added this in codemod preset:
```md
Please ensure you have installed `@compiled/react` in the directory where you are running this preset.
```

We can have on website 😃

madou (approved):
Lgtm

PR #386, opened 2020-12-03
fix(classnames): check if style is coming from inner scope only

### LINKED ISSUE #378: Class names style replacement pass is targetting anything called style
Relates to: #374 

```tsx
const EmphasisText = ({ className, children, style }) => (
  <ClassNames>
    {({ css }) => (
      <span
        style={style} // <--- should not be transformed!
        className={
          css({
            color: '#00b8d9',
            textTransform: 'uppercase',
            fontWeight: 700,
          }) +
          ' ' +
          className
        }>
        {children}
      </span>
    )}
  </ClassNames>
);
```

pgmanutd:
This closes #378 

@madou: Is this that simple? We are now just checking inner scope. And that will be children function expression.

### DIFF
--- a/packages/babel-plugin/src/class-names/__tests__/behaviour.test.tsx
+++ b/packages/babel-plugin/src/class-names/__tests__/behaviour.test.tsx
@@ -278,4 +278,34 @@ describe('class names behaviour', () => {
         </CC>;"
     `);
   });
+
+  it('should not transform style identifier when its coming from outer scope', () => {
+    const actual = transform(`
+      import { ClassNames } from '@compiled/react';
+
+      const EmphasisText = ({ className, children, style }) => (
+        <ClassNames>
+          {({ css }) => (
+            <span
+              style={style}
+              className={
+                css({
+                  color: '#00b8d9',
+                  textTransform: 'uppercase',
+                  fontWeight: 700,
+                }) +
+                ' ' +
+                className
+              }>
+              {children}
+            </span>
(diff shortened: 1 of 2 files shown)


packages/babel-plugin/src/class-names/__tests__/behaviour.test.tsx
```diff
@@ -278,4 +278,34 @@ describe('class names behaviour', () => {
...
+                className
+              }>
+              {children}
+            </span>
+          )}
+        </ClassNames>
+      );
+  `);
+
+    console.log(actual);
```
madou, line 307: can remove

packages/babel-plugin/src/class-names/index.tsx
```diff
@@ -108,7 +108,11 @@ export const visitClassNamesPath = (path: NodePath<t.JSXElement>, meta: Metadata
   // Second pass to replace all usages of `style`.
   path.traverse({
     Identifier(path) {
-      if (path.node.name !== 'style' || path.parentPath.isProperty()) {
+      if (
+        path.node.name !== 'style' ||
+        path.parentPath.isProperty() ||
+        !path.scope.hasOwnBinding('style')
```
madou, line 114: nice

madou:
looks good! probably is this simple hehe. just remove the console log and we're good to go

madou (approved):
Lgtm

Issue #390, opened 2020-12-07
Conditional CSS rules for styled


madou:
We've implemented conditional logical expressions in #379. We now need to add support to `styled` apis.

```js
styled.div`
  color: 'blue';
  ${props => props.isPrimary && ({ color: 'red' })};
`;

styled.div(
  { color: 'blue' },
  props => props.isPrimary && ({ color: 'red' }),
);

styled.span`
  color: ${(props) => (props.primary ? '#00B8D9' : '#36B37E')};
`;
```

Conditional template literal CSS is blocked by #313 - atm the CSS builder code assumes if we are returning a template literal we are returning a CSS declaration, not a CSS rule.

Kelin2025:
Double thank you for someone who's doing this. The lack of this feature is the reason why I didn't switch to Linaria and stay with styled-components  
Question - will it put inline expression into a component or it guesses what the function will return? 
I mean, will it also support something like this:
```ts
const Button = styled.button`
  ${props => ({
    normal: { padding: "20px", fontSize: "16px" },
    large: { padding: "40px", fontSize: "32px" }
  })[props.$size]}
`
```
Or just ternary operators only?

PR #392, opened 2020-12-07
feat: CSS multi args

### LINKED ISSUE #352: Styled args support
Array is already supported.

```tsx
const Button = styled.button(
  { color: 'darkorchid' },
  { fontSize: 12 }
)
```

Should also work with `css()` apis from ClassNames.

## Links

- https://github.com/atlassian-labs/compiled/issues/222#issuecomment-723670605

madou:
This PR enables support for multiple args in the styled and ClassNames apis:

```js
styled.div(
  { color: 'red' },
  ` font-size: 20 `,
  [{ backgroundColor: 'blue', padding: 10 }]
);

<ClassNames>
  {({ css }) => (
    <div
      className={css({ fontSize: 20 }, `color: blue;`, [{ padding: 20 }], {
        backgroundColor: 'red',
      })}>
      hello world
    </div>
  )}
</ClassNames>
```

Needed to update types and a little bit of the CSS builder, but for the most part no major architecture changes were needed.

Closes #352

**TODO**

- [x] styled api
- [x] ClassNames api

### DIFF
--- a/packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
+++ b/packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
@@ -272,8 +272,9 @@ describe('styled component behaviour', () => {
     ]);
   });
 
-  it('should add missing semicolon', () => {
-    const actual = transform(`
+  it('should not throw when template literal CSS has no terminating semi colon', () => {
+    expect(() => {
+      transform(`
       import { styled } from '@compiled/react';
 
       const ListItem = styled.div(
@@ -281,11 +282,6 @@ describe('styled component behaviour', () => {
         { fontSize: 20 }
       );
     `);
-
-    expect(actual).toIncludeMultiple([
-      '{color:red}',
-      '{font-size:20px}',
-      'ax(["_syaz5scu _1wybgktf",props.className])',
-    ]);
+    }).not.toThrow();
   });
 });
--- a/packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
+++ b/packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
@@ -254,4 +254,21 @@ describe('styled component behaviour', () => {
     expect(actual).toInclude('({as:C="div",style,textSize,...props},ref)');
     expect(actual).toInclude('"--_1j0t240":"super"+((()=>{return textSize;})()||"")+"big"');
   });
+
+  it('should collect args as styles', () => {
+    const actual = transform(`
+      import { styled } from '@compiled/react';
+
+      const ListItem = styled.div(
+        { color: 'darkorchid' },
+        { fontSize: 12 },
+      );
+    `);
+
+    expect(actual).toIncludeMultiple([
+      '{color:darkorchid}',
+      '{font-size:12px}',
+      'ax(["_syaz1paq _1wyb1fwx",props.className])',
+    ]);
+  });
 });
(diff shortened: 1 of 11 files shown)


ben-rogerson:
From the hospital? Hope it's nothing too serious 😬

This looks pretty sweet though, I haven't styled comma-separated values without an array before!

Usually I'll chuck them into an array like this:

```js
styled.div([
  { color: 'red' },
  `font-size: 20`,
  [{ backgroundColor: 'blue', padding: 10 }]
]);
```

Then either push in the conditionals like this:

```js
styled.div(({ myProp }) => [
  myProp && { color: 'red' },
]);
```

or per line, like this:

```js
styled.div([
  ({ myProp }) => myProp && { color: 'red' }
]);
```

But if you've already got support for array styling AND you've just invented a new way to style here then hats off!

madou:
hi! nah all good my missus just went for a test and im waiting for her 😅 

oh! well maybe we don't need this API then - yeah array is already supported. hmm 🤔 i wonder if emotion supports this, let me check.

the conditional styled api will be coming in with #390

ben-rogerson:
Oh that's great to know, perhaps I'll switch over to paren-styling™ - now there's two less characters to type 👍 
Looks like next release is really shaping up here, amazing work 👏👏👏

packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
```diff
@@ -254,4 +254,38 @@ describe('styled component behaviour', () => {
...
+    `);
+
+    expect(actual).toIncludeMultiple([
+      '{color:darkorchid}',
+      '{font-size:12px}',
+      'ax(["_syaz1paq _1wyb1fwx",props.className])',
+    ]);
+  });
+
+  it('should add missing semicolon', () => {
```
pgmanutd, line 275: Looks like test statement doesn't seems to match what we are expecting 🤔. Did we miss something? 🤔

```ts
  '{color:red}',
  '{font-size:20px}',
  'ax(["_syaz5scu _1wybgktf",props.className])',
]);
```

packages/babel-plugin/src/styled/index.tsx
```diff
@@ -4,14 +4,22 @@ import { buildStyledComponent } from '../utils/ast-builders';
 import { buildCss } from '../utils/css-builders';
 import { Metadata, Tag } from '../types';
 
+interface StyledData {
+  tag: {
```
pgmanutd, line 8: This is not type `Tag` now which is coming from `../types`? Got confused 😅

madou: Oh I just extracted the types from the return types. I'll see if they are the same 🤔

packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
madou: Ah it's because it gets into atomic CSS so it's a bit cryptic. You can see the string is missing a semi colon - this test makes sure that scenario still works. 

We could change to a should not throw? 🤔

pgmanutd: hehe. Test is right. What I was saying that test statement and expectation are not matching. Semicolon is not present in expectation.

madou: Hehe ok, maybe can just call it "should not throw when template literal CSS has not terminating semi colon"

pgmanutd: Cool. And expect should not throw error. Otherwise it should if we remove the fix.

madou:
re-review ready @pgmanutd 👏

pgmanutd (approved):
lgtm

PR #395, opened 2020-12-09
fix(codemod): restore comments attached to transformed node paths

pgmanutd:
The [recipe](https://github.com/facebook/jscodeshift/blob/master/recipes/retain-first-comment.md) mentioned in jscodeshift docs was only working for top level node paths.

This fix ensures they (comments) will get restored even when transformed node paths lies in between the statements.

### DIFF
--- a/packages/react/src/codemods/emotion-to-compiled/__tests__/emotion-to-compiled.test.tsx
+++ b/packages/react/src/codemods/emotion-to-compiled/__tests__/emotion-to-compiled.test.tsx
@@ -380,6 +380,54 @@ describe('emotion-to-compiled transformer', () => {
   defineInlineTest(
     { default: transformer, parser: 'tsx' },
     {},
+    `
+    // @top-level comment
+
+    /** @jsx jsx */
+    import { ClassNames, CSSObject, css as c, jsx } from '@emotion/core';
+    // comment 1
+    import * as React from 'react';
+    `,
+    `
+    /* TODO: (from codemod) "ClassNames" is not exported from "@compiled/react" at the moment. Please find an alternative for it. */
+    /* TODO: (from codemod) "CSSObject" is not exported from "@compiled/react" at the moment. Please find an alternative for it. */
+    // @top-level comment
+
+    import '@compiled/react';
+
+    // comment 1
+    import * as React from 'react';
+    `,
+    'it should not remove top level comments when transformed'
+  );
+
+  defineInlineTest(
+    { default: transformer, parser: 'tsx' },
+    {},
+    `
+    // @top-level comment
+
+    /** @jsx jsx */
+    import * as React from 'react';
+    // comment 1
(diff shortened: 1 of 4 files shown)


pgmanutd:
@madou: Strange. My branch is up to date with master and still chromatic is showing 2 new stories 🤔

madou:
Oh I think we didn't approve the latest change to master. Can you approve
them and then rebase?

packages/react/src/codemods/emotion-to-compiled/__tests__/emotion-to-compiled.test.tsx
```diff
@@ -377,6 +377,54 @@ describe('emotion-to-compiled transformer', () => {
...
+
+    /** @jsx jsx */
+    import * as React from 'react';
+    // comment 1
+    import { ClassNames, CSSObject, css as c, jsx } from '@emotion/core';
+    `,
+    `
+    /* TODO: (from codemod) "ClassNames" is not exported from "@compiled/react" at the moment. Please find an alternative for it. */
+    /* TODO: (from codemod) "CSSObject" is not exported from "@compiled/react" at the moment. Please find an alternative for it. */
+    // @top-level comment
```
madou, line 418: the top level comment should stay as the top level comment - here its been moved down a few places

packages/react/src/codemods/emotion-to-compiled/__tests__/emotion-to-compiled.test.tsx
```diff
@@ -377,6 +377,54 @@ describe('emotion-to-compiled transformer', () => {
...
+
+    /** @jsx jsx */
+    import { ClassNames, CSSObject, css as c, jsx } from '@emotion/core';
+    // comment 1
+    import * as React from 'react';
+    `,
+    `
+    /* TODO: (from codemod) "ClassNames" is not exported from "@compiled/react" at the moment. Please find an alternative for it. */
+    /* TODO: (from codemod) "CSSObject" is not exported from "@compiled/react" at the moment. Please find an alternative for it. */
+    // @top-level comment
```
madou, line 394: same here

packages/react/src/codemods/emotion-to-compiled/__tests__/emotion-to-compiled.test.tsx
pgmanutd: hmm..its todo from codemod. User has to remove them as they are breaking changes. Problem is that its difficult to differentiate between top-level comment and comment just before imports:

```ts
// @top-level

// first comment 1
// first comment 2
import * as React2 from 'react2';
```

in above case its difficult to differentiate between `// @top-level` and `// first comment 1` 😞

madou: Hmm yeah ok, lgtm

madou (approved):

PR #396, opened 2020-12-09
feat: handle classnames props edge cases

### LINKED ISSUE #374: Class names props access edge cases
- Object access props props.style
- Object access props invalid (dontexist.css, dontexist.style)
- Destructured props - changed name ({ css: cs, style: stl })
- https://github.com/atlassian-labs/compiled/pull/361#discussion_r526758130 `const { css } = arg;`

pgmanutd:
This closes #374


packages/babel-plugin/src/class-names/index.tsx
```diff
@@ -108,21 +168,36 @@ export const visitClassNamesPath = (path: NodePath<t.JSXElement>, meta: Metadata
...
   path.traverse({
-    Identifier(path) {
-      if (
-        path.node.name !== 'style' ||
-        path.parentPath.isProperty() ||
-        !path.scope.hasOwnBinding('style')
-      ) {
-        // Nothing to do - skip.
-        return;
+    Expression(path) {
```
pgmanutd, line 171: We need to traverse `style` only after `css` traversal is done otherwise it's breaking in case when `style` is not `undefined`.

Check test `should replace style identifier with css variable object` from `packages/babel-plugin/src/class-names/__tests__/behaviour.test.tsx` file.

packages/babel-plugin/src/class-names/index.tsx
```diff
@@ -7,6 +7,43 @@ import { buildCss, getItemCss } from '../utils/css-builders';
...
+  path.replaceWith(styleValue);
+};
+
+/**
+ * Will recursively checks if type (`css` or `style`) coming from destructuring.
+ *
+ * @param type `css` or `style` type
+ * @param node Any Expression node
+ */
+const isTypeComingFromDestructuring = (
```
madou, line 31: we probably want to put this into `ast.tsx`

packages/babel-plugin/src/class-names/index.tsx
```diff
@@ -26,14 +63,37 @@ const extractStyles = (path: NodePath<t.Expression>): t.Expression[] | t.Express
...
+      const styles = path.node.arguments as t.Expression[];
+      return styles;
+    }
+  }
+
+  if (t.isCallExpression(path.node) && t.isMemberExpression(path.node.callee)) {
+    // filter out invalid calls like dontexist.css({})
+    if (
+      t.isIdentifier(path.node.callee.object) &&
+      !path.scope.hasOwnBinding(path.node.callee.object.name)
```
madou, line 83: do we need to handle `doesexist.css`? (except its just another variable from somewhere else)

madou:
looks great! few small q's 😄

packages/babel-plugin/src/class-names/index.tsx
pgmanutd: oooo cool

packages/babel-plugin/src/class-names/index.tsx
pgmanutd: let me check once if its not needed

pgmanutd: How about this case? It will break if we don't handle this case. If `css` is coming from outside of scope.

```ts
import { ClassNames } from '@compiled/react';

const ListItem = () => (
  <ClassNames>
    {(props) => (<div style={dontexist.style} className={dontexist.css({ fontSize: '20px' })}>hello, world!</div>)}
  </ClassNames>
);
```

madou: arent we already handling this 🤔 or ur saying dontexist.css should be shouldexist.css?

i reckon let's not add too much if we don't need it. this is pretty hypothetical atm 😄

packages/babel-plugin/src/class-names/index.tsx
madou: thanks!

madou (dismissed):
lgtm

madou (approved):
lgtm

love removing code!

PR #403, opened 2020-12-11
Destructuring in styled interpolation functions

### LINKED ISSUE #397: Destructuring in styled interpolation functions aren't handled
**Describe the bug**
```js
import colors from 'colors';

export const BadgeSkeleton = styled.span`
    background-color: ${({ loading }) => (loading ? colors.N20 : colors.N40)};
    background-color: ${(propzzz) => (propzzz.loading ? colors.N20 : colors.N40)};
`;
```

All blows up.

pgmanutd:
This closes #397


examples/styled-interpolation-function-behavior.tsx
```diff
@@ -14,24 +14,24 @@ interface Props {
...
 
 const FunctionStyledTemplateLiteral = styled.div<Props>`
-  color: ${(props) => props.color};
+  color: ${({ color }) => color};
   font-size: ${(props) => props.textSize}px;
   background-color: ${(props) => {
     return props.bgColor;
   }};
-  border: 5px ${(props) => props.borderStyle} black;
+  border: 5px ${({ borderStyle }) => borderStyle} black;
```
madou, line 34: can we add these tests as well

```js
({ borderStyle: bs }) => bs
propz => propz.borderStyle
```

pgmanutd: Sure sure

pgmanutd: Looks little bit tricky. Damn. We have hardcoded `props` check as of now. `propz` is now dynamic 😅

madou: yeah 😅 

make it work, make it right, make it fast? hehe

pgmanutd: Was lil bit tricky. Damn! Finally 😅

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -145,6 +145,109 @@ const traverseStyledBinaryExpression = (node: t.BinaryExpression, nestedVisitor:
   return node;
 };
 
+/**
+ * Handles cases like:
+ * 1. `propz.loading` in `border-color: \${(propz) => (propz.loading ? colors.N100 : colors.N200)};`
+ * Outcome: It will replace `propz.loading` with `props.loading`.
+ *
+ * 2. `prop.notValidProp` in `border-color: \${(props) => (props.notValidProp ? colors.N100 : colors.N200)};`
```
madou, line 153: what is the edge case here? that the arg is `props` but we're using `prop`?

in this case should we just do nothing?

packages/babel-plugin/src/styled/__tests__/object-literal.test.tsx
```diff
@@ -373,4 +373,26 @@ describe('styled component object literal', () => {
...
+        color: ({ loading: l }) => l ? colors.N50 : colors.N10,
+        borderColor: (propz) => propz.loading ? colors.N100 : colors.N200,
+      });
+    `
+    );
+
+    expect(actual).toIncludeMultiple([
+      '{as:C="span",style,isLoading,loading,...props}',
+      'isLoading?colors.N20:colors.N40',
+      'loading?colors.N50:colors.N10',
```
madou, line 394: hmm i think this is smart (instead of using `l` you use `loading`) - because it gets around the potential of naming collisions

can we have one test that asserts that so the knowledge is passed onto the next developer?

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -145,6 +145,109 @@ const traverseStyledBinaryExpression = (node: t.BinaryExpression, nestedVisitor:
   return node;
 };
 
+/**
+ * Handles cases like:
+ * 1. `propz.loading` in `border-color: \${(propz) => (propz.loading ? colors.N100 : colors.N200)};`
```
madou, line 150: nit: could we handle this in an `enter()` visitor instead? so the logic of "rename propz to props" only needs to function once - instead of needing to traverse up the parent

🤔 thinking about it i see we check for arrow function or binary expression on L273/L277. perhaps this is fine?

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -145,6 +145,109 @@ const traverseStyledBinaryExpression = (node: t.BinaryExpression, nestedVisitor:
...
+ * Outcome: It will replace `propz.loading` with `props.loading`.
+ *
+ * 2. `prop.notValidProp` in `border-color: \${(props) => (props.notValidProp ? colors.N100 : colors.N200)};`
+ * Outcome: It will move `notValidProp` under `propsToDestructure` and replaces `props.notValidProp` with `notValidProp`.
+ *
+ * @param path MemberExpression path
+ * @param propsToDestructure Array which gets mutated with props which are not valid eg. props.notValidProp
+ */
+const handleMemberExpressionInStyledInterpolations = (
```
madou, line 159: nit: 

```diff
-handleMemberExpressionInStyledInterpolations
+handleMemberExpressionInStyledInterpolation
```

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -145,6 +145,109 @@ const traverseStyledBinaryExpression = (node: t.BinaryExpression, nestedVisitor:
...
+ * 1. `isLoading` in `background-color: \${({ isLoading }) => (isLoading ? colors.N20 : colors.N40)};`
+ * Outcome: It will move `isLoading` under `propsToDestructure`.
+ *
+ * 2. `l` in `color: \${({ loading: l }) => (l ? colors.N50 : colors.N10)};`
+ * Outcome: It will move `loading` under `propsToDestructure` and replaces `l` with `loading`.
+ *
+ * @param path Identifier path
+ * @param propsToDestructure Array which gets mutated with props which extracted from destructuring
+ */
+const handleDestructuringInStyledInterpolations = (
```
madou, line 216: nit:

```diff
-handleDestructuringInStyledInterpolations
+handleDestructuringInStyledInterpolation
```

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -145,6 +145,109 @@ const traverseStyledBinaryExpression = (node: t.BinaryExpression, nestedVisitor:
...
+  const memberExpressionKey = path.node.object;
+
+  if (!t.isIdentifier(memberExpressionKey)) {
+    return;
+  }
+
+  const traversedUpFunctionPath = path.find((parentPath) => parentPath.isFunction());
+  const memberExpressionKeyName = memberExpressionKey.name;
+
+  const isMemberExpressionKeyNameEqualsFunctionParams =
```
madou, line 172: nit: `isMemberExpressionNameTheSameAsFunctionFirstParam`?

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -145,6 +145,109 @@ const traverseStyledBinaryExpression = (node: t.BinaryExpression, nestedVisitor:
...
+      if (memberExpressionKeyName !== 'props') {
+        path.replaceWith(
+          t.memberExpression(t.identifier('props'), t.identifier(memberExpressionValueName))
+        );
+      }
+
+      return;
+    }
+
+    if (!propsToDestructure.includes(memberExpressionValueName)) {
```
madou, line 197: can we add some comments here explaining what we're doing

(if valid html attribute let it through - else destructure to prevent)

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -145,6 +145,109 @@ const traverseStyledBinaryExpression = (node: t.BinaryExpression, nestedVisitor:
...
+
+    if (!t.isIdentifier(memberExpressionValue)) {
+      return;
+    }
+
+    const memberExpressionValueName = memberExpressionValue.name;
+
+    if (isPropValid(memberExpressionValueName)) {
+      // Convert cases like propz.color to props.color
+      if (memberExpressionKeyName !== 'props') {
```
madou, line 188: can we move `'props'` to module scope and reuse it? or move to `constants.tsx`? 🤔

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -145,6 +145,109 @@ const traverseStyledBinaryExpression = (node: t.BinaryExpression, nestedVisitor:
...
+ * Outcome: It will move `loading` under `propsToDestructure` and replaces `l` with `loading`.
+ *
+ * @param path Identifier path
+ * @param propsToDestructure Array which gets mutated with props which extracted from destructuring
+ */
+const handleDestructuringInStyledInterpolations = (
+  path: NodePath<t.Identifier>,
+  propsToDestructure: string[]
+) => {
+  // We are not interested in parent object property like `({ loading: l }) => l`
```
madou, line 220: is this comment wrong?

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -145,6 +145,109 @@ const traverseStyledBinaryExpression = (node: t.BinaryExpression, nestedVisitor:
...
+ *
+ * 2. `l` in `color: \${({ loading: l }) => (l ? colors.N50 : colors.N10)};`
+ * Outcome: It will move `loading` under `propsToDestructure` and replaces `l` with `loading`.
+ *
+ * @param path Identifier path
+ * @param propsToDestructure Array which gets mutated with props which extracted from destructuring
+ */
+const handleDestructuringInStyledInterpolations = (
+  path: NodePath<t.Identifier>,
+  propsToDestructure: string[]
```
madou, line 218: instead of leaking props to destructure can we instead return any identifiers we've found that we want to destructure?

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -145,6 +145,109 @@ const traverseStyledBinaryExpression = (node: t.BinaryExpression, nestedVisitor:
...
+ *
+ * 2. `prop.notValidProp` in `border-color: \${(props) => (props.notValidProp ? colors.N100 : colors.N200)};`
+ * Outcome: It will move `notValidProp` under `propsToDestructure` and replaces `props.notValidProp` with `notValidProp`.
+ *
+ * @param path MemberExpression path
+ * @param propsToDestructure Array which gets mutated with props which are not valid eg. props.notValidProp
+ */
+const handleMemberExpressionInStyledInterpolations = (
+  path: NodePath<t.MemberExpression>,
+  propsToDestructure: string[]
```
madou, line 161: instead of leaking props to destructure can we instead return any identifiers we've found that we want to destructure?

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -330,6 +330,49 @@ const findNamedExportModuleNode = (
...
+ * @param resolveFor Either resolve destructuring key or its value
+ */
+export const resolveIdentifierComingFromDestructuring = ({
+  name,
+  node,
+  resolveFor = 'key',
+}: {
+  name: string;
+  node: t.Expression | undefined;
+  resolveFor?: 'key' | 'value';
```
madou, line 349: what if we just returned both?

madou (dismissed):
hey! left some comments have a look :) 

looking pretty great! biggest thoughts are:

1. reworking the need to pass through `propsToDestructure`, and
2. seeing if we can do something about needing to traverse back up the AST to find the parent function

for 2) its more of a thought than anything, if it makes the code shittier then let's not bother

:fingerguns:

packages/babel-plugin/src/utils/ast-builders.tsx
pgmanutd: oops..it should be `props.notValidProp`. Sorry 😅

packages/babel-plugin/src/styled/__tests__/object-literal.test.tsx
pgmanutd: 🎉 yups sure

packages/babel-plugin/src/utils/ast-builders.tsx
pgmanutd: So we are already traversing down. Here we are making sure `propz` is coming from function params. We picked member expression property, traversed to make sure it is coming from function params, and renamed it. Earlier we hardcoded the check that value should be `props`.

packages/babel-plugin/src/utils/ast-builders.tsx
pgmanutd: So in this statement, `({ loading: l }) => l` when we are checking for `Identifier`, both `: l` and `=> l` are identifier. And function is parent for both. We are not interested in modifying `: l`. We just need to modify `=> l` to `=> loading`. If we don't skip, second didn't get modify because we have modified `: l` earlier and second identifier one is nowhere to be found inside function params.
Hope this makes sense 😅

packages/babel-plugin/src/utils/ast.tsx
pgmanutd: Yups we can 😃

pgmanutd:
> hey! left some comments have a look :)
> 
> looking pretty great! biggest thoughts are:
> 
> 1. reworking the need to pass through `propsToDestructure`, and
> 2. seeing if we can do something about needing to traverse back up the AST to find the parent function
> 
> for 2) its more of a thought than anything, if it makes the code shittier then let's not bother
> 
> :fingerguns:

First is fine. Second one hmm 🤔. Problem is how I can find where it is coming from.  Looks like `scope` won't work because `parent` for `loading ? color1: color2` will be `ConditionalExpression` and not the function. I need destructuring key in cases like `{ loading: l }`.

packages/babel-plugin/src/utils/ast.tsx
pgmanutd: Another thought. Here we are traversing until we hit dead end. We either need to search for `key` or `value`. I mean where we will break the recursion. When either `key` is found or `value` is found?

madou: no worries then :)

packages/babel-plugin/src/utils/ast-builders.tsx
madou: ahhh 🤔 

so why don't we add `loading` to props to de-structure here?

alternatively now that i think about it - should we be adding destructured props anyway? currently the behaviour should be

1. am i a valid HTML element? you will be passed down to the raw HTML element
2. am i not valid? you will be destructured

do we need to rework anything to support this behaviour? 🤔

madou (approved):
lgtm

PR #408, opened 2020-12-15
fix: fixes important being including in suffix

### LINKED ISSUE #398: !important isn't kept with interpolations
Something I noticed when testing in Jira. Might need to double check what the problem is.

```js

styled.div`
  font-size: ${props => props.size}px !important;
`;
```

madou:
Closes #398


DarkPurple141 (approved):
Can't see anything glaring.

Issue #346, opened 2020-11-05
Interpolations with a prefix/suffix that are undefined result in the prefix/suffix being added to the inline style


madou:
```jsx
export const HorizontalStack = styled.div`
  margin-top: ${(props) => props.spacing}rem;
`;
```

Results in:

```css
--_1g0t409: rem;
```

Instead it should just not be defined.

This could potentially be fixed in conjunction with #333

---

We can introduce another util that will handle this logic. Behaviour required:

1. Handle adding suffix/prefix only if the variable is defined
2. If the variable is undefined return an empty variable

```ts
function interpolation(variable?: string | number, suffix?: string, prefix?: string): string;
```

PR #412, opened 2020-12-16
fix: reverts fallback dynamic css decl value

### LINKED ISSUE #402: Dynamic CSS declarations with 0 value don't work
```js
styled.div`
  max-height: ${(props) => (props.open ? props.inherentHeight || 0 : 0)}px;
`
```

Need to double check that this is generating valid CSS.

madou:
This introduces a regression which means `undefined` can show up. We need a more holistic approach when it comes to getting rid of dynamic values that have a suffix/prefix.

For now - this means `undefinedpx` will appear. But at least `0` values can work again. I'll create a follow up ticket to look at fixing this properly.

Closes #402.

Follow up: https://github.com/atlassian-labs/compiled/issues/346

### DIFF
--- a/examples/default-css-variables.tsx
+++ b/examples/default-css-variables.tsx
@@ -1,0 +1,19 @@
+import * as React from 'react';
+import { styled } from '@compiled/react';
+
+const Container = styled.div<{ open: boolean; inherentHeight: number }>`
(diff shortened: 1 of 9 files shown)


pgmanutd:
Lets handle this in follow up ticket. Looks nasty though.

pgmanutd (approved):
lgtm

PR #414, opened 2020-12-16
Classnames codemod

pgmanutd:
@madou: Just go through the test case once. Added TODO's for `ClassNames` props so that consumers can easily replace them. Also added logic for adding TODO's for missing `styled-components` api's. Did some cleanup too 😅

### DIFF
--- a/packages/react/src/codemods/emotion-to-compiled/__tests__/emotion-to-compiled.test.tsx
+++ b/packages/react/src/codemods/emotion-to-compiled/__tests__/emotion-to-compiled.test.tsx
@@ -25,12 +25,24 @@ describe('emotion-to-compiled transformer', () => {
     { default: transformer, parser: 'tsx' },
     {},
     `
+    import { ClassNames } from '@emotion/core';
+    `,
+    `
+    import { ClassNames } from '@compiled/react';
+    `,
+    'it transforms ClassNames named @emotion/core import'
+  );
+
+  defineInlineTest(
+    { default: transformer, parser: 'tsx' },
+    {},
+    `
     /** @jsx jsx */
-    import { css, jsx } from '@emotion/core';
+    import { css, jsx, ClassNames } from '@emotion/core';
     `,
     `
     import * as React from 'react';
-    import '@compiled/react';
+    import { ClassNames } from '@compiled/react';
     `,
     'it transforms all named @emotion/core imports'
   );
@@ -40,11 +52,11 @@ describe('emotion-to-compiled transformer', () => {
     {},
     `
     /** @jsx jsx */
-    import { css as c, jsx } from '@emotion/core';
+    import { css as c, jsx, ClassNames as CN } from '@emotion/core';
     `,
     `
     import * as React from 'react';
-    import '@compiled/react';
+    import { ClassNames as CN } from '@compiled/react';
     `,
     'it transforms all named @emotion/core imports with different imported name'
   );
@@ -54,12 +66,12 @@ describe('emotion-to-compiled transformer', () => {
     {},
     `
     /** @jsx jsx */
-    import { css, jsx } from '@emotion/core';
+    import { css, jsx, ClassNames } from '@emotion/core';
     import styled from '@emotion/styled';
     `,
     `
     import * as React from 'react';
-    import { styled } from '@compiled/react';
+    import { ClassNames, styled } from '@compiled/react';
     `,
     'it transforms all named @emotion/core and default @emotion/styled imports'
   );
@@ -69,12 +81,12 @@ describe('emotion-to-compiled transformer', () => {
     {},
     `
     /** @jsx jsx */
-    import { css as c, jsx } from '@emotion/core';
+    import { css as c, jsx, ClassNames as CN } from '@emotion/core';
     import sc from '@emotion/styled';
     `,
     `
     import * as React from 'react';
-    import { styled as sc } from '@compiled/react';
+    import { ClassNames as CN, styled as sc } from '@compiled/react';
     `,
     'it transforms all named @emotion/core with different imported name and default with different name than "styled" @emotion/styled imports'
   );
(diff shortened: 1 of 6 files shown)


packages/react/src/codemods/emotion-to-compiled/__tests__/emotion-to-compiled.test.tsx
```diff
@@ -336,7 +362,7 @@ describe('emotion-to-compiled transformer', () => {
 
     const Component = () => (
       <ClassNames>
-        {({ css, cx }) => (
+        {({ css }) => (
           <SomeComponent
             wrapperClassName={css({ color: 'green' })}
             css={c\`background-color: green \`}
```
madou, line 368: we need to remove `c` from here

packages/react/src/codemods/emotion-to-compiled/__tests__/emotion-to-compiled.test.tsx
```diff
@@ -413,18 +444,77 @@ describe('emotion-to-compiled transformer', () => {
...
+    import * as React from 'react';
+    import { ClassNames as CN } from '@compiled/react';
+
+    const Component = () => (
+      <CN>
+        {({
+          css,
+
+          /* TODO(@compiled/react codemod): Please replace "cx" with "ax" from "@compiled/react/runtime".
+          Usage: import { ax } from '@compiled/react/runtime';
```
madou, line 489: love we're giving a lot of context, this is great 😄

packages/react/src/codemods/emotion-to-compiled/emotion-to-compiled.tsx
```diff
@@ -116,76 +83,120 @@ const replaceEmotionCoreCSSTaggedTemplateExpression = (
...
-        }
+    objectPattern.properties.push(styleObjectProperty);
 
-        return false;
-      });
+    addCommentBefore({
+      j,
+      collection: j(styleObjectProperty),
+      message: `We have exported "${styleIdentifierName}" from "${imports.emotionCoreImportNames.ClassNames}" props.
+      If you are using dynamic declarations, make sure to set the "${styleIdentifierName}"
```
madou, line 139: would be good to link to the website docs here https://compiledcssinjs.com/docs/class-names#dynamic-declarations

packages/react/src/codemods/emotion-to-compiled/__tests__/emotion-to-compiled.test.tsx
pgmanutd: 😕 `css as c` is coming from import. And its the input. Its removed from output. I am confused now 😅

packages/react/src/codemods/emotion-to-compiled/emotion-to-compiled.tsx
pgmanutd: I wanted to. But then I thought, its old website.

pgmanutd: Ahh..you took url from preview one.

packages/react/src/codemods/styled-components-to-compiled/__tests__/styled-components-to-compiled.test.tsx
```diff
@@ -79,6 +79,25 @@ describe('styled-components-to-compiled transformer', () => {
...
+    { default: transformer, parser: 'tsx' },
+    {},
+    `
+    import styled, { css, keyframes, createGlobalStyle, ThemeProvider, withTheme } from 'styled-components';
+    import * as React from 'react';
+    `,
+    `
+    /* TODO(@compiled/react codemod): "css" is not exported from "@compiled/react" at the moment. Please find an alternative for it. */
+    /* TODO(@compiled/react codemod): "keyframes" is not exported from "@compiled/react" at the moment. Please find an alternative for it. */
+    /* TODO(@compiled/react codemod): "createGlobalStyle" is not exported from "@compiled/react" at the moment. Please find an alternative for it. */
```
madou, line 92: only the things that will break usage should be commented i think

so `css` and `keyframes` - everything else wouldn't break style usage so we could remove them for now.

madou (changes requested):
looks great! few comments i've had 😄

packages/react/src/codemods/emotion-to-compiled/emotion-to-compiled.tsx
madou: omg you're up late!!

packages/react/src/codemods/emotion-to-compiled/__tests__/emotion-to-compiled.test.tsx
madou: oh, im a dumb dumb 😄

packages/react/src/codemods/emotion-to-compiled/emotion-to-compiled.tsx
pgmanutd: Going out soon. hehe. Thought of wrapping up this PR.

packages/react/src/codemods/styled-components-to-compiled/__tests__/styled-components-to-compiled.test.tsx
pgmanutd: 🤔 What about other api's which are exported from `styled-components`? They will definitely break the code right? These comments ensure that codemod is smart enough to figure out all the exports which are now unresolved since we have converted import path from `styled-component` to `@compiled/react`. And it will be easier for user to figure out api's which are removed from `styled-components`.

madou: hmmm true okay 😄

madou (approved):
lgtm

PR #417, opened 2020-12-19
chore: remove commonjs bundle split points

### LINKED ISSUE #405: Commonjs bundle split points don't work in strict bundlers
We currently use commons js `require` to split the browser vs main bundles: https://github.com/atlassian-labs/compiled/blob/5306a52fa57897159daf2032d89ecbce471073de/packages/react/src/runtime/provider/index.tsx#L4

This breaks in strict bundlers, like snowpack https://github.com/atlassian-labs/compiled/issues/387#issuecomment-743943745

We need to inline everything into one module and split inside a single module instead.

madou:
This PR merges the two browser/main bundles into one to remove the `require` call in the code - enabling strict mode modules to function correctly.

Interestingly this also reduces the bundle size slightly.

Closes #405


package.json
```diff
@@ -96,7 +96,7 @@
     },
     {
       "path": "./packages/react/dist/browser/runtime.js",
-      "limit": "489B",
+      "limit": "445B",
```
pgmanutd, line 99: Noice ❤️

packages/react/src/__tests__/server-side-hydrate.test.tsx
```diff
@@ -36,16 +26,20 @@ describe('server side hydrate', () => {
   };
 
   it('should not log any warnings when hydrating HTML', () => {
-    const StyledDiv = styled.div`
-      font-size: 12px;
-      color: blue;
-      border: 1px solid black;
-    `;
+    // It's notoriously hard to do both server and client rendering in this test.
```
pgmanutd, line 29: 👍

packages/react/src/runtime/style-cache.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * On the server this ensures the minimal amount of styles will be rendered
+ * safely using React Context.
+ *
+ * On the browser this turns into a fragment with no React Context.
+ */
+const StyleCacheProvider: ProviderComponent = (props) => {
+  if (isNodeEnvironment()) {
+    // This code path isn't conditionally called at build time - safe to ignore.
```
pgmanutd, line 46: nit: 🤔 It will work we know. Just wondering won't it confuse developers who are so familiar with rules of hooks? Is there a way we can split it into two components like we did before and do something like:

```ts
const StyleCacheProvider = isNodeEnvironment() ? StyleServerCacheProvider: StyleClientCacheProvider;
```

or maybe 

```ts
const StyleCacheProvider: ProviderComponent = (props) => {
  if (isNodeEnvironment()) {
    return <StyleServerCacheProvider {...props} />
  }
  
  return <StyleClientCacheProvider {...props} />
}
```

Its still difficult to do as above for `useCache` though 😞

madou: this keeps the bundlesize down and keeps the react tree as small as possible - i wouldn't want to make more TBH

pros & cons - in this instance having good amounts of comments should work - i'll update them a bit though

pgmanutd: No worries..hehe 😃

pgmanutd (approved):
lgtm

Issue #413, opened 2020-12-16
Re-write prefix/suffix template literal interpolation handler


madou:
We've had a few bugs with the current implementation - it's honestly a pain:

- https://github.com/atlassian-labs/compiled/issues/345
- https://github.com/atlassian-labs/compiled/issues/346
- #400
- #398

Current idea is to strip out the current code and then write a Post CSS plugin (using the `postcss-values-parser` library to parse the values) to remove prefix/suffixes around CSS variables and then callback with them. Will spike it out soon.

**Input**

```css
content: "var(--hello-world)";
width: var(--hello-world)px;
```

**Output**

```css
content: var(--hello-world);
width: var(--hello-world);
```

```js
{
  name: '--hello-world',
  prefix: '"',
  suffix: '"',
}

{
  name: '--hello-world',
  suffix: 'px',
}
```

madou:
@DarkPurple141 sensible chuckle

Issue #345, opened 2020-11-05
Referencing the same interpolation breaks inline variables


madou:
**Describe the bug**

```jsx
export const HorizontalStack = styled.div<{ spacing?: number; gap?: number }>`
  margin-top: ${(props) => props.spacing}rem;
  margin-bottom: ${(props) => props.spacing}rem;

  margin-right: ${(props) => props.gap}rem;

  @media only screen and (min-width: 900px) {
    && > * {
      margin-right: ${(props) => props.gap}rem;

      &:last-child {
        margin-right: 0;
      }
    }
  }
`;
```

Results in

```css
--_1g0t409: rem;
--_4cpxcj: 4;
```

Removing the first `margin-right` declaration results in:

```css
--_1g0t409: rem;
--_4cpxcj: 4rem;
```

Looks like it's getting caught with the suffix handling.

PR #418, opened 2020-12-19
fix: simplify interpolation handling

madou:
This PR simplifies the interpolation handling by making more assumptions, and handling only things that would actually happen. Obviously invalid suffix/prefixes and the like are now unsupported.

Closes https://github.com/atlassian-labs/compiled/issues/413 
Closes https://github.com/atlassian-labs/compiled/issues/345

### DIFF
--- a/packages/css/src/utils/string-interpolations.tsx
+++ b/packages/css/src/utils/string-interpolations.tsx
@@ -11,9 +11,13 @@ export interface BeforeInterpolation {
 }
 
 /**
- * Will remove any suffix out of the CSS and return them both.
+ * Will remove any valid suffix out of the CSS and return them both.
+ * Handles both terminated and un-terminated CSS.
  *
- * E.g. `'px;font-size: 20px;'` would return `"px"` as the suffix and `";font-size: 20px;"` as the CSS.
+ * Some examples:
+ * - `'px;font-size: 20px;'` would return `"px"` as the suffix and `";font-size: 20px;"` as the CSS.
+ * - `'"'` would return `'"'` as the suffix and `''` as the CSS.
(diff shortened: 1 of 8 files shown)


packages/css/src/utils/string-interpolations.tsx
```diff
@@ -15,43 +11,25 @@ export interface BeforeInterpolation {
 }
 
 /**
- * Will return any suffix out of the CSS and return them both.
+ * Will remove any suffix out of the CSS and return them both.
  *
  * E.g. `'px;font-size: 20px;'` would return `"px"` as the suffix and `";font-size: 20px;"` as the CSS.
```
pgmanutd, line 16: Should we add examples for more cases? Like when regex matched and when not. Might be helpful in future.

packages/css/src/utils/string-interpolations.tsx
```diff
@@ -62,39 +40,17 @@ export const cssAfterInterpolation = (css: string): AfterInterpolation => {
  * @param css all the CSS before the interpolation
```
pgmanutd, line 40: Can we add some examples here too? hehe

pgmanutd:
Looks good. Love that we have cleaned messy code hehe. If possible just add few examples in jsdoc which will be helpful for future developers.

packages/css/src/utils/string-interpolations.tsx
madou: good idea

packages/css/src/utils/string-interpolations.tsx
madou: done!

madou:
all done

pgmanutd (approved):
lgtm

PR #419, opened 2020-12-19
chore: adds cjs dist for react pkg

### LINKED ISSUE #387: ES modules can't be used in next.js
I've been following the project for a while and I'm really excited about the release coming up. I mainly use next.js for most of my projects and would love to play around with this too! Unfortunately, the packages are only bundled as es modules, and while sometimes it works to use [next-transpile-modules](https://github.com/martpie/next-transpile-modules), it doesn't work with this set of packages.

We use [tsdx](https://github.com/formium/tsdx) for a lot of our packages, it could be an easy win to use that here, too!

Thanks so much

madou:
Closes #387


pgmanutd:
Looks good. Should we test this with any boilerplate repo before merging?

madou:
i have a nextjs install locally - i'll cut a nightly and test there

pgmanutd (approved):
lgtm

PR #420, opened 2020-12-20
chore: upgrades minor and patch deps

madou:
This actually shows the type errors we saw when doing your interpolation changes @pgmanutd.

This is a pre-req for the `runtime: automatic` jsx fix.


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "typescript.tsdk": "node_modules/typescript/lib"
```
pgmanutd, line 2: Nice, we configured it here. I had similar one in my user settings.

package.json
```diff
@@ -31,14 +31,14 @@
     "version-nightly": "bash scripts/should-nightly-run.sh && npm version prerelease --no-git-tag-version --preid=$(git rev-parse HEAD)",
     "postversion": "bash scripts/push-if-git.sh",
     "bundlesize": "yarn build && yarn build:dead-code-elimination && size-limit",
-    "postinstall": "npx yarn-deduplicate"
+    "postinstall": "npx yarn-deduplicate && yarn --ignore-scripts"
```
pgmanutd, line 34: Are we ignoring postinstall script for dependencies? Any specific reason? 🤔

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -46,12 +46,16 @@ const hoistSheet = (sheet: string, meta: Metadata): t.Identifier => {
...
 
   const sheetIdentifier = meta.parentPath.scope.generateUidIdentifier('');
-  const parent = meta.parentPath.findParent((path) => path.isProgram()).get('body') as NodePath[];
-  const path = parent.filter((path) => !path.isImportDeclaration())[0];
-
-  path.insertBefore(
-    t.variableDeclaration('const', [t.variableDeclarator(sheetIdentifier, t.stringLiteral(sheet))])
-  );
+  const parent = meta.parentPath.findParent((path) => path.isProgram());
+  const parentBody = parent && (parent.get('body') as NodePath[]);
```
pgmanutd, line 50: So all of these started breaking just after patch/minor bump ups? 😨

packages/react/src/codemods/codemods-helpers.tsx
```diff
@@ -21,11 +23,14 @@ export const getImportDeclarationCollection = ({
   j: JSCodeshift;
   collection: Collection<any>;
   importPath: string;
-}) =>
-  collection
+}): Collection<ImportDeclaration> => {
```
pgmanutd, line 26: Love explicit typings over implicit one. TS compiler will be fast if we specify return types.

packages/babel-plugin/src/utils/ast-builders.tsx
madou: yeah

package.json
madou: only for postinstall - this is to stabalize the yarn lock after de-duping (and so it doesnt result in an infinite loop)

unfortunately there can be instances where running yarn after yarn de-dupe will result in the yarn lock file updating again

packages/react/src/codemods/codemods-helpers.tsx
madou: we could add an eslint rule for that too.

pgmanutd: Maybe later. Should we add card for this? Because it might require changes too.

madou: added https://github.com/atlassian-labs/compiled/issues/422

PR #421, opened 2020-12-20
fix: fixes automatic jsx transform not working

### LINKED ISSUE #349: Test with new jsx automatic transform
Currently unsure if we need to do anything to make it work.

madou:
Makes some changes to ensure jsx automatic mode works. Don't merge until #420 is merged.

Currently the automatic behavior is behind a babel plugin option instead of trying to infer - inferring was more trouble than its worth TBH.

**Question**

- What should we call the option? The _behaviour_ is to just add `React` if its missing. Currently it's called `importReact`.

Closes #349


examples/class-names-dynamic.tsx
```diff
@@ -55,7 +54,7 @@ const ClassNamesArray = ({ weight }: { weight: number }) => (
   </ClassNames>
 );
 
-export const Object = () => <ClassNamesObject color="pink" />;
```
madou, line 58: interestingly webpack adds an `Object` check now which made this stack overflow

package.json
```diff
@@ -34,12 +34,13 @@
     "postinstall": "npx yarn-deduplicate"
   },
   "devDependencies": {
+    "@babel/plugin-transform-react-jsx-self": "^7.12.1",
```
madou, line 37: had to add these for some reason - wasn't found otherwise.

package.json
```diff
@@ -34,12 +34,13 @@
...
+    "@babel/plugin-transform-react-jsx-source": "^7.12.1",
     "@babel/preset-env": "^7.11.0",
     "@babel/preset-react": "^7.10.4",
     "@babel/preset-typescript": "^7.10.4",
     "@compiled/jest": "*",
     "@compiled/react": "*",
-    "@emotion/core": "^10.1.1",
```
madou, line 42: removed the emotion test because i'd need to upgrade it to v11 for the new automatic stuff. rather just remove.

packages/babel-plugin/src/__tests__/index.test.tsx
```diff
@@ -80,10 +80,11 @@ describe('babel plugin', () => {
     expect(output?.code).toMatchInlineSnapshot(`
       "/* File generated by @compiled/babel-plugin v0.0.0 */
 
+      import { forwardRef } from 'react';
```
madou, line 83: forward ref now imported as a named import.

packages/babel-plugin/src/index.tsx
```diff
@@ -90,11 +90,17 @@ export default declare<State>((api) => {
...
-          if (!path.scope.getBinding('React')) {
+          const shouldImportReact = state.opts.importReact === undefined || state.opts.importReact;
+          if (shouldImportReact && !path.scope.getBinding('React')) {
             // React is missing - add it in at the last moment!
             path.unshiftContainer('body', template.ast(`import * as React from 'react'`));
           }
 
+          if (state.compiledImports.styled && !path.scope.getBinding('forwardRef')) {
+            // forwardRef is missing - add it in at the last moment!
+            path.unshiftContainer('body', template.ast(`import { forwardRef } from 'react'`));
```
madou, line 101: foward ref imported here. its bleeding a bit because this ideally would be in the styled visitor - but that doesnt have any knowledge of the program and i didn't really want to do more work.

we could move it in though.

packages/babel-plugin/src/types.tsx
```diff
@@ -1,20 +1,29 @@
...
+   * Will import the React namespace if it is missing.
+   * When using the `'automatic'` jsx runtime set this to `false`.
+   *
+   * Defaults to `true`.
+   */
+  importReact?: boolean;
 }
 
-export interface State {
+export interface State extends PluginPass {
```
madou, line 26: realized plugin pass is the type for this.

packages/babel-plugin/src/index.tsx
pgmanutd: Yes later please

pgmanutd:
@madou: Heading out. Will review it once m back. Seems like you just reviewed your own PR 😆

pgmanutd:
> @madou: Heading out. Will review it once m back. Seems like you just reviewed your own PR 😆

Sorry was bit busy today. Will review this in morning 😃

packages/babel-plugin/src/index.tsx
madou: created this https://github.com/atlassian-labs/compiled/issues/424

DarkPurple141 (dismissed):
LGTM

DarkPurple141 (approved):
get good

PR #423, opened 2020-12-21
Refactor dynamic interpolations to use a util

### LINKED ISSUE #333: Fix leaking CSS variables
**Describe the bug**
Non-inherited CSS properties are being inherited thanks to the shared CSS variable declaration (they have the same name).

This is OK behaviour for properties that inherit - not ok for ones that shouldn't.

**To Reproduce**

```jsx
import React from 'react';

const Box = styled.div<{ border?: string }>`
  border: ${(props) => props.border || ''};
`;

export const NestedWithDynamicValues = () => (
  <Box border="2px solid red">
    Should have border
    <Box>Should not have border</Box>
  </Box>
);
```

**Expected behavior**

Nested `Box` should not have a border.

**Additional context**

We can fix this by unsetting the CSS variable when it resolves to undefined.

```css
--value: ;
```

Currently we resolve to an empty string if it has a prefix/suffix - to fix this we will probably have to always do it for everything.

**Testing notes**

- Css prop interpolations
- Styled component interpolations

madou:
Closes https://github.com/atlassian-labs/compiled/issues/346
Closes #333

This PR introduces an `ix` function which handles CSS custom property values, with an optional  suffix/prefix.

Currently only the suffix and prefix handling is done for template literal CSS, so for example:

```js
styled.div`
  content: "${props => props.content}";
`;
```

Would result in this `ix` call:

```js
ix(props.content, '"', '"');
```

If we did this for object CSS instead you would need to handle this yourself:

```js
styled.div({
  content: props => ('"' + props.content + '"'),
});
```

Which ends up with the `ix` call:

```js
ix('"' + props.content + '"'); // no suffix or prefix argument
```


packages/react/src/runtime/css-custom-property.tsx
```diff
@@ -0,0 +1,31 @@
+/**
+ * Returns a CSS custom property value with an optional suffix & prefix.
+ * Prefix will only be added if there is a suffix.
+ * If the value is undefined a fallback value will be returned to prevent children inheriting parent values.
+ *
+ * @param value
+ * @param suffix
+ * @param prefix
+ */
+export default function cssCustomPropertyValue(
```
pgmanutd, line 10: Do we need to do some perf testing here since we introduce this during runtime?

pgmanutd: Nevermind. It won't be that slow..hehe. Few if checks only.

pgmanutd:
Can we add bundlesize check for `ix`?

packages/react/src/runtime/css-custom-property.tsx
```diff
@@ -0,0 +1,31 @@
...
+  value: string | number | number,
```
pgmanutd, line 11:
```suggestion
  value: string | number | undefined | null,
```

packages/react/src/runtime/css-custom-property.tsx
madou: 😄

packages/react/src/runtime/css-custom-property.tsx
madou: done

packages/react/src/styled/index.tsx
```diff
@@ -31,15 +31,15 @@ export type Interpolations<TProps extends Record<string, unknown>> = (
  * props from `StyledProps`.
  */
 export interface StyledFunctionFromTag<TTag extends keyof JSX.IntrinsicElements> {
-  <TProps extends Record<string, unknown>>(
+  <TProps extends unknown>(
```
madou, line 34: fixing types that i introduced a regression to in the TS change before, oops

pgmanutd (approved):
lgtm

PR #425, opened 2020-12-22
chore: bumps typescript

madou:
When we did #349 we didn't upgrade typescript - so the examples were type erroring (but didn't break builds because it didn't typecheck during storybook build).

This upgrades TS to 4.1 to add in the new jsx behaviour.


pgmanutd (approved):
lgtm

PR #426, opened 2020-12-22
fix: upgrade jest to fix a security vuln

madou:
Fixing https://github.com/atlassian-labs/compiled/security/dependabot/yarn.lock/node-notifier/open


pgmanutd (approved):
lgtm

PR #427, opened 2020-12-23
fix: ensures CSS values, selectors, at rules are normalized

### LINKED ISSUE #318: Ensure CSS property values are consistently formatted
We want to sure we reduce the possibility for duplicate declarations - for that we need to ensure that all values can be the same.

## Same case

E.g:

**Before**

```css
color: currentColor;
```

**After**

```css
color: currentcolor;
```

There are some case-sensitive things we want to avoid:

- CSS variables
- Classes/ID references
- Content property
- Identifiers (animation name)

Probably make this as a new PostCSS plugin.

## Consistent "empty"

Change everything to `0`, over `none` and the like.

**Before**

```
width: 0px;
border: none;
```

**After**

```
width: 0;
border: 0;
```

### LINKED ISSUE #326: Normalize selectors in atomic plugin
**Describe the bug**
Currently in the atomic plugin two semantically same selectors can generate different group hashes:

```
> :hover {}
>:hover {}
:before {}
::before {}
```

Results in two different hashes. We should normalize the selector in the atomic plugin so this can't happen.

madou:
Closes #326
Closes #318

This turns on some Post CSS plugins to normalize both selectors and at rules during `development` builds, and then more Post CSS plugins to normalize declaration values during `production` builds.

The overarching point of this PR is to reduce the total amount of CSS generated from equivalent but slightly different CSS declarations.


packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
```diff
@@ -522,7 +522,7 @@ describe('css prop behaviour', () => {
       );
     `);
 
-    expect(actual).toInclude('ax([props.isPrimary&&"_syaz13q2 _1wybgktf","_19itjoc5"])');
+    expect(actual).toInclude('ax([props.isPrimary&&"_syaz13q2 _1wybgktf","_19it97hw"])');
```
madou, line 525: border order changes so new value hash here

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
```diff
@@ -543,7 +543,7 @@ describe('css prop behaviour', () => {
     `);
 
     expect(actual).toInclude(
-      'ax(["_1wybo7ao",props.isPrimary&&"_syaz13q2 _1wybgktf","_19itjoc5"])'
+      'ax(["_1wybo7ao",props.isPrimary&&"_syaz13q2 _1wybgktf","_19it97hw"])'
```
madou, line 546: border order changes so new value hash here

packages/css/src/plugins/expand-shorthands/__tests__/background.test.tsx
```diff
@@ -33,16 +33,4 @@ describe('background property expander', () => {
           "
     `);
   });
-
-  it('should order value nodes alphabetically greater than one', () => {
```
madou, line 37: we can delete the ordering plugin because Post CSS has one.

packages/css/src/plugins/expand-shorthands/background.tsx
```diff
@@ -10,10 +10,5 @@ export const background: ConversionFunction = (value) => {
     return [{ prop: 'background-color', value: value.nodes[0].toString() }];
   }
 
-  const orderedValues = value.nodes
-    .map((val) => val.toString())
-    .sort()
-    .join(' ');
-
-  return [{ value: orderedValues }];
+  return [{ value: value.nodes.join(' ') }];
```
madou, line 13: as above, we deleted the ordering plugin so we can get rid of this

packages/css/src/plugins/normalize-css.tsx
```diff
@@ -0,0 +1,72 @@
...
+];
+
+/**
+ * This plugin runs a subset of the cssnao plugins to normalize CSS during build.
+ * During a production build it will run more plugins.
+ */
+export const normalizeCSS = (): Plugin<never>[] => {
+  const preset = cssnano();
+  // We exclude async because we need this to run synchronously as ts transformers aren't async!
+  const extraPlugins = process.env.NODE_ENV === 'production' ? PROD_PLUGINS : [];
```
madou, line 54: current logic: we don't want people to "opt in" to this behaviour. thats why we force it during production builds.

packages/css/src/utils/__tests__/css-transform.test.tsx
```diff
@@ -55,7 +143,7 @@ describe('#css-transform', () => {
       );
 
       expect(actual.join('\n')).toMatchInlineSnapshot(
-        `"._1sfm1q9v:hover div, ._f8pj1q9v:focus{color:hotpink}"`
+        `"._f8pj1q9v:focus, ._1sfm1q9v:hover div{color:hotpink}"`
```
madou, line 146: this is now ordered but its equivalent so all good 👍

pgmanutd (approved):
lgtm

PR #462, opened 2020-12-24
[WIP] Global component

madou:
Will be tinkering with this over the holidays. Chime in if you're interested.


pgmanutd:
> Will be tinkering with this over the holidays. Chime in if you're interested.

Noice..btw renovate PR's almost crashed my inbox..damn so many mails 😅

packages/react/src/global/index.tsx
```diff
@@ -0,0 +1,10 @@
+import { CssFunction } from '../types';
+import { createSetupError } from '../utils/error';
+
+interface GlobalProps {
+  style: CssFunction | CssFunction[];
```
pgmanutd, line 5: 🤔 Should we use `style` or something else? It might confuse consumers because react in built components accept `style` props which translates to inline styles.

Emotion also uses `styles` instead of `style`. Should we use that or something else 🤔?

Not sure 😅

madou:
Oh yeah definitely use styles 😅

packages/react/src/global/index.tsx
madou: Ah definitely 😄 should have been `styles`

PR #476, opened 2021-01-05
feat: functions with arguments mixin support

### LINKED ISSUE #341: Functions with arguments mixin support
Previously we've introduced function mixin support here #275 - however arguments aren't supported. Let's make it work with arguments. 

**Things to consider:**

- Naming of arguments - we need to make sure they align

```jsx
const mixin = (mode) => ({
  color: mode
});


const Component = (props) => <div css={[mixin(props.color)]} />
```

Becomes

```jsx
<div style={{ '--propscolor': props.color }} className="_asd1232s" />
```

```css
._asd1232s { color: var(--propscolor) }
```

- Static arguments should resolve to static declarations

```jsx
const mixin = (mode) => ({
  color: mode
});

<div css={[mixin('blue')]} />
// or
const blue = 'blue';
<div css={[mixin(blue)]} />
```

Becomes

```css
color: blue;
```

**Testing notes/edge cases**

- Simple args (string, number)
- Complex args (objects, arrays)
- Destructured args `({ color })`
- Destructured renamed args `({ color: clr })`

**Things to ignore:**

- Mutable things outside the function. We should only allow the passed arguments to "be dynamic" - everything else should resolve to something static with the already implemented traversal behaviour.

pgmanutd:
This closes #341 

@madou: Did something 😅 Its working though 🎉

### DIFF
--- a/packages/babel-plugin/src/class-names/__tests__/object-literal.test.tsx
+++ b/packages/babel-plugin/src/class-names/__tests__/object-literal.test.tsx
@@ -128,4 +128,96 @@ describe('class names object literal', () => {
 
     expect(actual).toIncludeMultiple([':hover{color:red}', '{font-size:20px}']);
   });
+
+  it('should transform object with argument arrow function variable', () => {
+    const actual = transform(`
+        import { ClassNames } from '@compiled/react';
+
+        const color1 = 'black';
+        const mixin = ({ color1, color2: c }, color3, radius) => ({
+          color: color1,
+          backgroundColor: c,
+          borderColor: color3 ,
+          borderRadius: radius,
+        });
+
+        const color = { red: 'red' };
+        const greenColor = 'green';
+
+        const Component = (props) => {
+          const color2 = 'black';
+
+          return (
+            <ClassNames>
+              {({ css }) => <div className={css({ ...mixin({ color1: color.red, color2: 'blue' }, greenColor, 10) })} />}
+            </ClassNames>
+          );
+        };
+    `);
+
+    expect(actual).toIncludeMultiple([
+      '{color:red}',
+      '{background-color:blue}',
+      '{border-color:green}',
+      '{border-radius:10px}',
+    ]);
+  });
+
+  it('should transform object with unresolved argument arrow function variable', () => {
+    const actual = transform(`
+        import { ClassNames } from '@compiled/react';
+
+        const radius = 10;
+        const mixin = (color1, radius, size, weight) => ({
+          color: color1,
+          borderRadius: radius,
+          fontSize: size,
+          fontWeight: weight
+        });
+
+        const Component = (props) => (
+          <ClassNames>
+            {({ css, style }) => <div style={style} className={css({ ...mixin(props.color1, radius) })} />}
+          </ClassNames>
+        );
+      `);
+
+    expect(actual).toIncludeMultiple([
+      '{color:var(--_zo7lop)}',
+      '"--_zo7lop":ix(props.color1)',
+      '{border-radius:10px}',
+      '{font-weight:var(--_u6vle4)}',
+      '"--_u6vle4":ix()',
+      '{font-size:var(--_kre2x8)}',
+      '"--_kre2x8":ix()',
+    ]);
+  });
+
+  it('should transform object with argument arrow function variable inside member expression', () => {
+    const actual = transform(`
+        import { ClassNames } from '@compiled/react';
+
+        const mixin = {
+          value: (color1, r, color2) => ({
+            color: color1,
+            borderRadius: r,
+            borderColor: color2,
+          })
+        }
+
+        const radius = 10;
+
+        const Component = (props) => (
+          <ClassNames>
+            {({ css, style }) => <div style={style} className={css({ ...mixin.value(props.color1, radius, 'red') })} />}
+          </ClassNames>
+        );
+      `);
+
+    expect(actual).toIncludeMultiple([
+      '"--_zo7lop":ix(props.color1)',
+      '{border-radius:10px}',
+      '{border-color:red}',
+    ]);
+  });
 });
--- a/packages/babel-plugin/src/class-names/__tests__/string-literal.test.tsx
+++ b/packages/babel-plugin/src/class-names/__tests__/string-literal.test.tsx
@@ -127,4 +127,96 @@ describe('class names string literal', () => {
 
     expect(actual).toInclude(`{color:blue}`);
   });
+
+  it('should transform template string with argument arrow function variable', () => {
+    const actual = transform(`
+        import { ClassNames } from '@compiled/react';
+
+        const color1 = 'black';
+        const mixin = ({ color1, color2: c }, color3, radius) => ({
+          color: color1,
+          backgroundColor: c,
+          borderColor: color3 ,
+          borderRadius: radius,
+        });
+
+        const color = { red: 'red' };
+        const greenColor = 'green';
+
+        const Component = (props) => {
+          const color2 = 'black';
+
+          return (
+            <ClassNames>
+              {({ css }) => <div className={css\`\${mixin({ color1: color.red, color2: 'blue' }, greenColor, 10)}\`} />}
+            </ClassNames>
+          );
+        };
+    `);
+
+    expect(actual).toIncludeMultiple([
+      '{color:red}',
+      '{background-color:blue}',
+      '{border-color:green}',
+      '{border-radius:10px}',
+    ]);
+  });
+
+  it('should transform template string with unresolved argument arrow function variable', () => {
+    const actual = transform(`
+        import { ClassNames } from '@compiled/react';
+
+        const radius = 10;
+        const mixin = (color1, radius, size, weight) => ({
+          color: color1,
+          borderRadius: radius,
+          fontSize: size,
+          fontWeight: weight
+        });
+
+        const Component = (props) => (
+          <ClassNames>
+            {({ css, style }) => <div style={style} className={css\`\${mixin(props.color1, radius)}\`} />}
+          </ClassNames>
+        );
+      `);
+
+    expect(actual).toIncludeMultiple([
+      '{color:var(--_zo7lop)}',
+      '"--_zo7lop":ix(props.color1)',
+      '{border-radius:10px}',
+      '{font-weight:var(--_u6vle4)}',
+      '"--_u6vle4":ix()',
+      '{font-size:var(--_kre2x8)}',
+      '"--_kre2x8":ix()',
+    ]);
+  });
+
+  it('should transform template string with argument arrow function variable inside member expression', () => {
+    const actual = transform(`
+        import { ClassNames } from '@compiled/react';
+
+        const mixin = {
+          value: (color1, r, color2) => ({
+            color: color1,
+            borderRadius: r,
+            borderColor: color2
+          })
+        }
+
+        const radius = 10;
+
+        const Component = (props) => (
+          <ClassNames>
+            {({ css, style }) => <div style={style} className={css\`\${mixin.value(props.color1, radius, 'red')}\`} />}
+          </ClassNames>
+        );
+      `);
+
+    expect(actual).toIncludeMultiple([
+      '"--_zo7lop":ix(props.color1)',
+      '{border-radius:10px}',
+      '{border-color:red}',
+    ]);
+  });
 });
--- a/packages/babel-plugin/src/styled/__tests__/string-literal.test.tsx
+++ b/packages/babel-plugin/src/styled/__tests__/string-literal.test.tsx
@@ -217,8 +217,7 @@ describe('styled component string literal', () => {
         \`;
       `);
 
-    expect(actual).toInclude('{color:var(--_16ywsic)}');
-    expect(actual).toInclude(`"--_16ywsic":ix(em('blue'))`);
+    expect(actual).toInclude('{color:blue}');
   });
 
   it('should transform template string with no argument arrow function variable', () => {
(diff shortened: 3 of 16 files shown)


packages/babel-plugin/src/class-names/__tests__/object-literal.test.tsx
```diff
@@ -128,4 +128,96 @@ describe('class names object literal', () => {
...
+          </ClassNames>
+        );
+      `);
+
+    expect(actual).toIncludeMultiple([
+      '{color:var(--_zo7lop)}',
+      '"--_zo7lop":ix(props.color1)',
+      '{border-radius:10px}',
+      '{font-weight:var(--_u6vle4)}',
+      '"--_u6vle4":ix()',
```
madou, line 190: this doesn't look right

packages/babel-plugin/src/class-names/__tests__/object-literal.test.tsx
```diff
@@ -128,4 +128,96 @@ describe('class names object literal', () => {
...
+      '{font-size:var(--_kre2x8)}',
+      '"--_kre2x8":ix()',
```
madou, line 192: this doesn't look right

packages/babel-plugin/src/class-names/__tests__/string-literal.test.tsx
```diff
@@ -127,4 +127,96 @@ describe('class names string literal', () => {
...
+          </ClassNames>
+        );
+      `);
+
+    expect(actual).toIncludeMultiple([
+      '{color:var(--_zo7lop)}',
+      '"--_zo7lop":ix(props.color1)',
+      '{border-radius:10px}',
+      '{font-weight:var(--_u6vle4)}',
+      '"--_u6vle4":ix()',
```
madou, line 189: this doesn't look right

packages/babel-plugin/src/class-names/__tests__/string-literal.test.tsx
```diff
@@ -127,4 +127,96 @@ describe('class names string literal', () => {
...
+      '{font-size:var(--_kre2x8)}',
+      '"--_kre2x8":ix()',
```
madou, line 191: this doesn't look right

packages/babel-plugin/src/styled/__tests__/string-literal.test.tsx
```diff
@@ -217,8 +217,7 @@ describe('styled component string literal', () => {
         \`;
       `);
 
-    expect(actual).toInclude('{color:var(--_16ywsic)}');
-    expect(actual).toInclude(`"--_16ywsic":ix(em('blue'))`);
+    expect(actual).toInclude('{color:blue}');
```
madou, line 220: dope

packages/babel-plugin/src/types.tsx
```diff
@@ -75,6 +75,11 @@ export interface Metadata {
    * Path of a parent node.
    */
   parentPath: NodePath<any>;
+
+  /**
+   * Path of the own node.
+   */
+  ownPath?: NodePath<any>;
```
madou, line 82: nit: `ownPath` -> `path`?

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -96,6 +96,39 @@ const toCSSDeclaration = (key: string, result: CSSOutput) => {
   return result.css.map((x) => ({ ...x, css: `${kebabCase(key)}: ${getItemCss(x)};` }));
 };
 
+/**
+ * Will return the variable declarator value node from meta's own path.
+ * It will compare the variable declarator node id's name with passed node name and
+ * returns its stringified value along with the node.
+ *
+ * Eg. If passed node is an Identifier with name `abcd` and somewhere in own path
+ * we have any variable declarator `const abcd = obj.value`, it will return
```
madou, line 105: nit: can we break this out into input outputs to make it read a little easier?

e.g.

```
Input: 
`const abcd = obj.value`

Output:
`{ variableName: '???', expression: '???'  }`
```

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -96,6 +96,39 @@ const toCSSDeclaration = (key: string, result: CSSOutput) => {
...
+ * stringified `obj.value` along with node `obj.value`.
+ * If variable declarator's value is undefined `const abcd = undefined`, it will
+ * return stringified `abcd` along with undefined.
+ *
+ * @param node
+ * @param meta Meta data used during the transformation.
+ */
+const getVariableDeclaratorValueForOwnPath = (node: t.Expression, meta: Metadata) => {
+  let variableName = generate(node).code;
```
madou, line 114: can you add a comment for what this will produce? for example does this make the stringified "obj.value" i guess?

(as a nit: any way we can avoid needing to generate the code?)

packages/babel-plugin/src/utils/evaluate-expression.tsx
```diff
@@ -205,6 +206,90 @@ const traverseFunction = (expression: t.Function, meta: Metadata) => {
...
+/**
+ * Will find the function node for the call expression and wrap an IIFE around it (to avoid name collision)
+ * and move all the parameters mapped to passed arguments in the IIFE's scope (own scope in this case).
+ * It will also set own scope path so that when we recursively evaluate any node,
+ * we will look for its binding in own scope first, then parent scope.
+ *
+ * @param expression Expression we want to interrogate.
+ * @param meta Meta data used during the transformation.
+ */
+const traverseCallExpression = (expression: t.CallExpression, meta: Metadata) => {
```
madou, line 218: todo - review this later 😄

packages/babel-plugin/src/class-names/__tests__/object-literal.test.tsx
pgmanutd: 🤔 So this is the case when passed args and params count doesn't matched. When we passed less args than params. 🤔 Should we remove this altogether?

packages/babel-plugin/src/utils/evaluate-expression.tsx
pgmanutd: No rush 😄

packages/babel-plugin/src/class-names/__tests__/object-literal.test.tsx
madou: ah yeah i see now.

hmm, i think it could go both ways. what do you think

pgmanutd: 🤔 Let me see if its possible to remove `'"--_u6vle4":ix()'` completely if not param is not resolved.

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -450,7 +483,11 @@ export const resolveBindingNode = (
   referenceName: string,
   meta: Metadata
 ): PartialBindingWithMeta | undefined => {
-  const binding = meta.parentPath.scope.getBinding(referenceName);
+  // Check binding in own scope first so that manually created scopes can be
+  // evaluated first then parent scopes or scopes coming from different module.
+  const binding =
+    meta.ownPath?.scope.getOwnBinding(referenceName) ||
```
madou, line 489: when we say manually created scopes what are we talking about? us adding new variables into the scope?

are we missing the part where we add the binding to the scope? could we add it at a previous step?

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -436,6 +436,39 @@ const resolveObjectPatternValueNode = (
...
+ * equal to reference name for which we have to resolve binding.
+ *
+ * Eg. If we have something like `const { key: value } = { key: 'something' }`, and
+ * reference name is `value`, it will return `key` so that it can be resolved to
+ * 'something' otherwise it won't get resolved.
+ *
+ * @param node Object pattern node which we have to investigate
+ * @param referenceName Reference name for which `binding` to be resolved
+ */
+const getObjectPatternKeyWhenNotSameAsValue = (
```
madou, line 450: is there a less wordy name we could give this funtion?

`getDestructuredObjectPatternKey`? 🤔

packages/babel-plugin/src/utils/evaluate-expression.tsx
```diff
@@ -205,6 +206,90 @@ const traverseFunction = (expression: t.Function, meta: Metadata) => {
...
+  const callee = expression.callee;
+  let value: t.Node | undefined | null = undefined;
+  let updatedMeta: Metadata = meta;
+
+  if (t.isExpression(callee)) {
```
madou, line 223: can we give some examples throughout this function of what each if statements callee/node would look like (so we don't forget later) please

packages/babel-plugin/src/utils/evaluate-expression.tsx
```diff
@@ -205,6 +206,90 @@ const traverseFunction = (expression: t.Function, meta: Metadata) => {
...
+    let functionNode = null;
+
+    if (t.isIdentifier(callee)) {
```
madou, line 226: nit: could we simplify the start by pumping everything through `evaluateExpression`?

packages/babel-plugin/src/utils/evaluate-expression.tsx
```diff
@@ -205,6 +206,90 @@ const traverseFunction = (expression: t.Function, meta: Metadata) => {
...
+        (argument) => evaluateExpression(argument as t.Expression, updatedMeta).value
+      );
+
+      updatedMeta.parentPath.traverse({
+        CallExpression(callExpressionPath) {
+          // Only look for expression for which we are evaluating
+          if (callExpressionPath.node === expression) {
+            const callExpressionWrappingNode = wrapNodeInIIFE(callExpressionPath.node);
+
+            const [wrappingNodePath] = callExpressionPath.replaceWith(callExpressionWrappingNode);
```
madou, line 259: what are we replacing here? (comment please?)

packages/babel-plugin/src/utils/evaluate-expression.tsx
```diff
@@ -205,6 +206,90 @@ const traverseFunction = (expression: t.Function, meta: Metadata) => {
...
+      const { params } = functionNode;
+
+      const evaluatedArguments = expression.arguments.map(
+        (argument) => evaluateExpression(argument as t.Expression, updatedMeta).value
+      );
+
+      updatedMeta.parentPath.traverse({
+        CallExpression(callExpressionPath) {
+          // Only look for expression for which we are evaluating
+          if (callExpressionPath.node === expression) {
```
madou, line 256: are we doing this just to get ahold of the expression path? could use use `getPathForExpression` instead?

packages/babel-plugin/src/utils/evaluate-expression.tsx
```diff
@@ -205,6 +206,90 @@ const traverseFunction = (expression: t.Function, meta: Metadata) => {
...
+            const [wrappingNodePath] = callExpressionPath.replaceWith(callExpressionWrappingNode);
+
+            wrappingNodePath.traverse({
+              ArrowFunctionExpression(arrowFunctionExpressionPath) {
+                params
+                  .filter((param) => t.isIdentifier(param) || t.isObjectPattern(param))
+                  .forEach((param, index) => {
+                    const evaluatedArgument = evaluatedArguments[index];
+
+                    arrowFunctionExpressionPath.scope.push({
```
madou, line 268: so we are doing this purely to add things to the scope? is this related to the ownPath hasOwnBinding logic?

packages/babel-plugin/src/utils/evaluate-expression.tsx
```diff
@@ -205,6 +206,90 @@ const traverseFunction = (expression: t.Function, meta: Metadata) => {
...
+      updatedMeta.parentPath.traverse({
+        CallExpression(callExpressionPath) {
+          // Only look for expression for which we are evaluating
+          if (callExpressionPath.node === expression) {
+            const callExpressionWrappingNode = wrapNodeInIIFE(callExpressionPath.node);
+
+            const [wrappingNodePath] = callExpressionPath.replaceWith(callExpressionWrappingNode);
+
+            wrappingNodePath.traverse({
+              ArrowFunctionExpression(arrowFunctionExpressionPath) {
```
madou, line 262: nit: can we do this without traversing?

madou (changes requested):
left some comments :) looking great!

packages/babel-plugin/src/utils/ast.tsx
pgmanutd: So the flow is:

- We pick the parameters
- Then evaluate the args by traversing
- Loop through params, and map the args with params
- Pull the params and create an IIFE around the calling function `CallExpression` which creates a new scope around calling function and isolates everything (We add bindings at this step)
- Add that IIFE node path to `ownPath`, and check only for own binding so that only isolated params are evaluated
- If at any step, `ownPath` is not set (module traversal case or any other case), it will pick things from `parentPath`

Since we are recursively traversing, its difficult to add bindings in `resolveBindingNode` because thats done under `traverseCallExpression` and `ownPath` is only variable which guarantees that recursively traversed things evaluates properly. 

Hope this makes sense 😅

packages/babel-plugin/src/utils/evaluate-expression.tsx
pgmanutd: Oh..we cant actually because `evaluateExpression` will evaluate things and we just need the `functionNode` reference.
For eg. 

```ts
function abcd(color) {
  return {
    color
  }
}

abcd('blue');
```

will return `ObjectExpression` `{ color }` if we use `evaluateExpression`, but we only wanted that function node so that we can isolate the parameters and map them to args.

packages/babel-plugin/src/utils/evaluate-expression.tsx
pgmanutd: Sure

packages/babel-plugin/src/utils/evaluate-expression.tsx
pgmanutd: Ah! here were are just checking the reference. There can be many `CallExpression` under that `expression`, and we are only interested in nodes when `CallExpression` is same as passed `expression`

packages/babel-plugin/src/utils/evaluate-expression.tsx
pgmanutd: yups..I know its weird. So it creates an isolated scope and `ownPath` is only variable we use for getting the own binding because we are evaluating things recursively without any state in between.

packages/babel-plugin/src/utils/evaluate-expression.tsx
pgmanutd: 🤔 Let me see. I wanted `ArrowFunctionExpression`'s scope actually to push variables in that isolated scope we created using IIFE

pgmanutd:
> left some comments :) looking great!
> 
> might also be good to zoom later to go through this more in depth

hehe. I know this PR is bit hard to understand. I did my best to explain the flow. I pinged you. But you went away 😛

packages/babel-plugin/src/utils/evaluate-expression.tsx
madou: ah ok

madou:
you'll need to rebase to get the new GH actions config

packages/babel-plugin/src/class-names/__tests__/object-literal.test.tsx
pgmanutd: Tried. Need lot of code change for this 😞 Can we create a card for this cleanup?

packages/babel-plugin/src/types.tsx
pgmanutd: I have kept it as `ownPath` for now. Added detailed description for it.

packages/babel-plugin/src/class-names/__tests__/object-literal.test.tsx
madou: i don't think its the end of the world, let's keep it for now.

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -436,6 +436,44 @@ const resolveObjectPatternValueNode = (
...
+ * equal to reference name for which we have to resolve binding.
+ *
+ * Eg. If we have something like `const { key: value } = { key: 'something' }`, and
+ * reference name is `value`, it will return `key` so that it can be resolved to
+ * 'something' otherwise it won't get resolved.
+ * Input: `node: const { key: value } = { key: 'something' }, referenceName: 'value'`
+ * Output: `'key'`
+ *
+ * Input: `node: const { key } = { key: 'something' }, referenceName: 'value'`
+ * Output: `'value'`
```
madou, line 450: why does this return value? should this not return `key`

madou (approved):
Just one question, LGTM

packages/babel-plugin/src/utils/ast.tsx
pgmanutd: ah because we are saving it in result `let result = referenceName;` and if `keyName !== keyValue && keyValue === referenceName` condition fails, it will return the result with same value.

PR #491, opened 2021-01-11
fix: Add boolean to CSS Prop type

### LINKED ISSUE #489: Conditional array CSS prop type errors
**Describe the bug**

The following code got TypeScript error.
https://compiledcssinjs.com/docs/css-prop#conditional-rules

```
TS2322: Type 'false | { border: string; color: string; }' is not assignable to type 'string | number | void | CSSProps | AnyKeyCssProps<void>'.   Type 'boolean' is not assignable to type 'string | number | void | CSSProps | AnyKeyCssProps<void>'.
```

**Expected behavior**
No errors.

**Screenshots**
error

**Desktop (please complete the following information):**

- OS: macOS BigSir 11.1
- Browser Google Chrome
- Version 87.0.4280.141（Official Build） （x86_64）

**Develop Environment**

- TypeScript: 4.1.3
- @compiled/react: 0.5.2

sonatard:
fix #489

This is my first contribution to JavaScript library 😄 

Please give me advice about tests and Storybook. I can't find CSS Prop tests.

### DIFF
--- a/examples/static-evaluation.tsx
+++ b/examples/static-evaluation.tsx
@@ -1,4 +1,5 @@
 import { styled } from '@compiled/react';
+import '@compiled/react';
 
 export default {
   title: 'ast/static evaluation',
--- a/examples/static-evaluation.tsx
+++ b/examples/static-evaluation.tsx
@@ -1,5 +1,4 @@
 import { styled } from '@compiled/react';
-import '@compiled/react';
 
 export default {
   title: 'ast/static evaluation',
--- a/packages/react/src/jsx/__tests__/index.test.tsx
+++ b/packages/react/src/jsx/__tests__/index.test.tsx
@@ -129,4 +129,31 @@ describe('css prop', () => {
     expect(getByText('hello world')).toHaveCompiledCss('font-size', '20px');
     expect(getByText('hello world')).not.toHaveCompiledCss('font-size', '12px');
   });
+
+  it('should create css from object reference with true condition', () => {
+    const condition = true;
+    const base = { fontSize: 12 };
+    const { getByText } = render(<div css={condition && base}>hello world</div>);
+
+    expect(getByText('hello world')).toHaveCompiledCss('font-size', '12px');
+  });
+
+  it('should create css from object reference with false condition', () => {
+    const condition = false;
+    const base = { fontSize: 12 };
+    const { getByText } = render(<div css={condition && base}>hello world</div>);
+
+    expect(getByText('hello world')).not.toHaveCompiledCss('font-size', '12px');
+  });
+
+  it('should create css from array of object reference with condition', () => {
+    const condition = true;
+    const base = { fontSize: 12 };
+    const base2 = { fontSize: 20 };
+    const { getByText } = render(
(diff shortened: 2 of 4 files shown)


madou:
Cool! Good job

You can add a unit test here https://github.com/atlassian-labs/compiled/blob/master/packages/react/src/jsx/__tests__/index.test.tsx

You can do something similar to this test https://github.com/atlassian-labs/compiled/blob/master/packages/react/src/jsx/__tests__/index.test.tsx#L39

pgmanutd:
Congrats @sonatard for your first contribution 🎉. See if you can add one boolean example using css array prop here: https://github.com/atlassian-labs/compiled/blob/master/examples/conditional-rules.tsx

sonatard:
Thanks!! I'll try it.

madou:
No worries - let us know how you go! Were you able to run storybook and tests locally? Anything that could be documented better in CONTRIBUTING.md?

sonatard:
>  Were you able to run storybook and tests locally? 

Yes, No Problem!
I added tests and Storybook. Please review it.

And It ’s a little off topic, I use WebStorm which prints `Attribute css is not allowed here` warning in test files.
Test files import `import '@compiled/react';`. I can't check type error in test  file and Storybook.

スクリーンショット 2021-01-12 10 27 10

madou:
Thanks! Will review soon. 

For the webstorm error - from looking at the error I think you need to just add it to webstorm as a custom HTML attribute? It must have its own linting 🤔

packages/react/src/jsx/__tests__/index.test.tsx
```diff
@@ -129,4 +129,31 @@ describe('css prop', () => {
...
+
+    expect(getByText('hello world')).toHaveCompiledCss('font-size', '12px');
+  });
+
+  it('should create css from object reference with false condition', () => {
+    const condition = false;
+    const base = { fontSize: 12 };
+    const { getByText } = render(<div css={condition && base}>hello world</div>);
+
+    expect(getByText('hello world')).not.toHaveCompiledCss('font-size', '12px');
```
sonatard, line 146: I want to use an empty expect value. But I don't know how to use it.

madou: hmmm that doesn't exist ATM

I think your assertion is okay though - it _is_ asserting the expected behaviour 😄

examples/static-evaluation.tsx
```diff
@@ -1,4 +1,5 @@
 import { styled } from '@compiled/react';
+import '@compiled/react';
```
madou, line 2: nit: this isn't needed - css prop is enabled when the module is found so the above `@compiled/react` has enabled it 👍

sonatard: I don't know why I added this. 🤣

madou (approved):
LGTM! I'll just pull your fork locally to make sure the VR tests look good - unfortunately they aren't being ran in forked pull requests 😭 (made an issue here #497)

sonatard:
Thanks for reviewing!! Is there anything else I need to do?

madou:
Not at the moment, thanks!

PR #495, opened 2021-01-11
fix: fixes cli not finding codemods

### LINKED ISSUE #490: No codemods available right now
When running `npx @compiled/cli@latest` or `npx @compiled/cli --preset codemods` I am getting the following error:

> No codemods available right now.

I have already followed the installation steps here: https://compiledcssinjs.com/docs/installation

madou:
Closes #490

### DIFF
--- a/packages/react/src/codemods/constants.tsx
+++ b/packages/react/src/codemods/constants.tsx
@@ -1,5 +1,18 @@
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const pkgJson = require('../../../package.json');
+/**
+ * Because we don't want to include package json in the build artefacts
+ * AND because the location of package json will be different depending
+ * on using source or built assets - we do this. Not great but for now
+ * it's better than re-writing the build setup.
+ */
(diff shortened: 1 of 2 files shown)


packages/react/src/codemods/constants.tsx
```diff
@@ -1,5 +1,18 @@
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const pkgJson = require('../../package.json');
+/**
+ * Because we don't want to include package json in the build artefacts
+ * AND because the location of package json will be different depending
+ * on using source or built assets - we do this. Not great but for now
+ * it's better than re-writing the build setup.
+ */
+const tryGetPkgJson = () => {
```
madou, line 7: this is shit but it works - without including package json into the built assets (changing the TS build to have ./ as base path instead of ./src i don't know if we can stabilize this.

my preference is to make this fix asap, and maybe re-evaluate build setup later.

alexreardon (approved):

PR #502, opened 2021-01-13
conditional export from the main package for cjs and esm

### LINKED ISSUE #482: React package babel plugin entry point has main and module in package.json reversed
**Describe the bug**
https://github.com/atlassian-labs/compiled/blob/8908a402de570f98054e45b0233a7c9233af1fbc/packages/react/babel-plugin/package.json#L2-L3

If I understand correctly, `main` should be common js (cjs) and `module` should be ES module (esm). Is it just a mistake?

### LINKED ISSUE #501: Sometimes doesn't work with node version more than 12.19.x
It looks like that compiled _might_ have a hard dependency on node version. In my current setup, which uses next.js and atlaskit components, the build in CI (github actions if important) started failing after I added `@atlaskit/lozenge` to the mix. Babel plugin is installed.

failed-build

It was mitigated by reducing node version to the version `12.16.x` from just `12.x`. Not sure how to reproduce this setup in isolation without spending days on it, hope the description and the screen will help :)

UPD:

12.15.x node version - everything is good
12.19 - fails (so starts failing somewhere in between, probably when they enabled modules support by default)
14.15.1 (latest of 14.x) - still failing
15.5.1 (latest of 14.x) - still failing

UPD:
PR with the fix: https://github.com/atlassian-labs/compiled/pull/502

adevnadia:
Closes #501 
Closes #482 

Based on this: https://nodejs.org/api/packages.html#packages_conditional_exports

Exporting different versions based on the import type of the package. Fixes this issue: https://github.com/atlassian-labs/compiled/issues/501

### DIFF
--- a/packages/react/babel-plugin/package.json
+++ b/packages/react/babel-plugin/package.json
@@ -1,5 +1,5 @@
 {
-  "main": "../dist/esm/babel-plugin.js",
-  "module": "../dist/cjs/babel-plugin.js",
+  "main": "../dist/cjs/babel-plugin.js",
+  "module": "../dist/esm/babel-plugin.js",
   "types": "../dist/esm/babel-plugin.d.ts"
 }
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -24,9 +24,18 @@
     "typescript"
   ],
   "exports": {
-    ".": "./dist/esm/index.js",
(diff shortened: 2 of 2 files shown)


pgmanutd:
Also fixes #482

madou (approved):
LGTM - I've added a test for this and other node versions in this PR https://github.com/atlassian-labs/compiled/pull/503

PR #504, opened 2021-01-14
BREAKING CHANGE: removes babel plugin from react pkg

### LINKED ISSUE #484: postcss-values-parser@4.0.0 license issue
Latest version of Switcher has `@atlaskit/lozenge` bumped to `10.1.0` - this pulls in `@compiled/react` which itself pulls in `postcss-values-parser@4.0.0`. The problem with this is the license for that library is incompatible (MPL-2.0) and its breaking Confluence FE pipeline.

@madou: If you are around. Do take a look at this. I am investigating how `postcss-values-parser@4.0.0` end up in the deps.

### LINKED ISSUE #493: Separate babel plugin from react package
Currently the babel plugin is re-exported as a separate entrypoint in the react package to simplify consumption. Unfortunately it means that people are being forced to install the dependencies even if they're not writing styles with Compiled - which is causing some friction.

This change would be removing the re-exported entrypoint, updating documentation (readme + website) to mention the need for installing the babel plugin package, and cutting a new minor release.

**Before**

```
npm i @compiled/react
```

**After**

```
npm i @compiled/react @compiled/babel-plugin
```

madou:
**BREAKING CHANGE**

Removes babel plugin entrypoint from the react package. This will be a minor version bump (in our case, breaking because we are `v0.x.x`.

Closes #493 
Closes #484

**TODO**

- [ ] Update website with new consumption info

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,14 +31,14 @@ const StyledButton = styled.button`
 Install the React package:
 
 ```bash
-npm install @compiled/react
+npm install @compiled/react @compiled/babel-plugin
 ```
 
 Configure [Babel](https://babeljs.io/docs/en/config-files):
 
 ```json
 {
-  "plugins": ["@compiled/react/babel-plugin"]
+  "plugins": ["@compiled/babel-plugin"]
 }
 ```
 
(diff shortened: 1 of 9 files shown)


pgmanutd (approved):
lgtm

pgmanutd:
We might have to update website docs too https://compiledcssinjs.com/docs/installation#babel-setup

PR #512, opened 2021-01-18
Parcel transformer spike

### LINKED ISSUE #344: Modifying values outside of a module doesn't update the component styles
**Describe the bug**

When developing (using Parcel) in a component and modifying some properties outside of the module that are imported - don't end up recompiling the component - thus they don't get updated.

```jsx
// colors.tsx
export const heroBackground = 'linear-gradient(120deg,blue 0%, #8fd3f4 100%)';

// hero.tsx
import React from 'react';
import '@compiled/core';
import { heroBackground } from '../utils/colors';

export const Hero = ({ children }: { children: React.ReactNode }) => {
  return (
    <div
      css={{
        background: heroBackground,
      }}>
      {children}
    </div>
  );
};
```

What's worse - sometimes you get into a state where it just won't update. In this case is it our caching behaviour, or is it the bundler? Both?

**To Reproduce**

1. Use code above
1. Run with your favourite bundler
1. Update hero background
1. Notice that the component doesn't have the updated color
1. Now modify hero component
1. Notice that is now is updated as expected

**Expected behavior**

It should update the component color without needing the modify the component file.

**Additional context**

Not sure what we can do here. Is it possible to fix this if we create a bundler plugin to make sure any depending module gets re-transformed if something changes?

**Questions**

- Does this also affect Webpack
- How can we force a bundler to recompile a consuming module (so it runs the Babel plugin over it, again)

madou:
Spiking out a Parcel plugin to fix #344.

**TODO**
- [x] Update parcel engine version to `^2.0.0 || ^2.0.0-beta.1`
- [x] Add transform and parse step (transformer should be standalone, not require Babel to follow)
- [x] Remove hardcoded code
- [x] Included files rebuild importing files
- [x] Test with TypeScript
- [x] Add parcel to readme
- [x] Pass through opts
- [x] Website documentation https://github.com/compiled/website/pull/72
- [x] Source maps

**Later**

- [ ] Fix https://github.com/parcel-bundler/parcel/issues/5927
- [ ] Use `babel` type for AST so it can be re-used

**Links**

- Included files discussion https://github.com/parcel-bundler/parcel/discussions/5677


packages/parcel-transformer/src/index.tsx
```diff
@@ -0,0 +1,30 @@
+import { Transformer } from '@parcel/plugin';
+import semver from 'semver';
+
+/**
+ * Compiled Transformer
+ *
+ * Should run before any other Babel transformers.
+ */
+export default new Transformer({
```
pgmanutd, line 9: Not much code needed for transformer? 🤔 If it is this small, then noice 😄

madou: Would be a bit more involved to pick out the dependencies that we need to "include" - but yeah

asvny: Would it be better if we follow a pattern like `*.compiled.ts` and then glob those files?

madou: Out of the box we won't recommend that (because of CSS prop) - but we'll definitely document that as a performance optimization your build tooling could opt into.

madou: Also for context when I say dependencies it's talking about files that are statically evaluated and included in the "importing" component - which #344 goes into a bit of detail around why the DX is a bit painful ATM.

madou:
image

Some extra info!

packages/parcel-transformer/src/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+    includedFiles.forEach((file) => {
+      asset.addIncludedFile(file);
+    });
+
+    if (result?.ast) {
+      asset.setAST({
+        // TODO: Currently if we set this as `'babel'` the babel transformer blows up.
+        // Let's figure out what we can do to reuse it.
+        type: 'compiled',
+        version: '0.0.0',
```
mischnic, line 71: This should be 
```
      type: 'babel',
      version: '7.0.0',
```

madou: Thanks! I had it like that before but it was blowing up. I'll try again after I upgrade the nightly and get everything working.

examples/packages/parcel/.babelrc
```diff
@@ -0,0 +1,7 @@
+{
```
mischnic, line 1: This file will never be read because this is monorepo ([Babel's rule](https://babeljs.io/docs/en/config-files#monorepos), not Parcel-specific)

packages/parcel-transformer/src/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+    const ast = await parseAsync(code, {
+      filename: asset.filePath,
+      caller: { name: 'compiled' },
+    });
+
+    return ast;
+  },
+
+  async transform({ asset, ast }) {
```
mischnic, line 31: The AST isn't passed in here, you need to call `const ast = await asset.getAST();`

madou: according to https://v2.parceljs.org/plugin-system/transformer/ it is? 🤔

packages/parcel-transformer/src/index.tsx
madou: @mischnic i get this error when updating this to babel - does the babel transformer not accept a standard babel AST?

```
$ parcel serve ./src/index.html --open
ℹ️ Server running at http://localhost:1234
🚨 Build failed.
Error: this[statement.type] is not a function
TypeError: this[statement.type] is not a function
    at Object.Program (/Users/madou/projects/compiled/node_modules/astring/dist/astring.js:228:27)
    at Object.Program (/Users/madou/projects/compiled/node_modules/@parcel/babel-ast-utils/lib/generator.js:54:34)
    at Object.generator.<computed> [as Program] (/Users/madou/projects/compiled/node_modules/@parcel/babel-ast-utils/lib/generator.js:317:10)
    at generate (/Users/madou/projects/compiled/node_modules/astring/dist/astring.js:1149:29)
    at generateAST (/Users/madou/projects/compiled/node_modules/@parcel/babel-ast-utils/lib/index.js:123:43)
    at generate (/Users/madou/projects/compiled/node_modules/@parcel/babel-ast-utils/lib/index.js:163:10)
    at async /Users/madou/projects/compiled/node_modules/@parcel/core/lib/Transformation.js:456:24
    at async Promise.all (index 0)
    at async Transformation.runPipeline (/Users/madou/projects/compiled/node_modules/@parcel/core/lib/Transformation.js:452:7)
    at async Transformation.runPipelines (/Users/madou/projects/compiled/node_modules/@parcel/core/lib/Transformation.js:368:40)
✨  Done in 36.16s.
➜  compiled git:(parcel-plugin) ✗
```

packages/parcel-transformer/src/index.tsx
mischnic: I'll update that. The source of truth is this section: https://v2.parceljs.org/plugin-system/transformer/#Transformer

madou: No worries thanks, will update the code.

(As a side - is there any TypeScript types lying around for the `Transformer` class?)

madou:
Going to merge the work in progress stuff for external testing. It works for now. Will follow up with improvements later

PR #530, opened 2021-02-01
Webpack loader

### LINKED ISSUE #344: Modifying values outside of a module doesn't update the component styles

madou:
Fix #344 for Webpack users.

**TODO**

- [x] Creation of `transform` export in babel plugin
- [x] Async webpack loader
- [x] Single pass caching
- [x] Webpack dev server rebuilds and shows changes after modifying statically evaluated module
- [x] Move babel plugin to its own file and re-export through `index.tsx`
- [x] Error handling
- [x] Static evaluation not working
- [x] Included files callback tested and working
- [x] Loader tests
- [x] Documentation
- [x] Configuration pass through

**Outstanding thoughts**

- Can we hook into a single babel pass instead of doing another with our `transform` function? Parcel has this capability.

**Related**

- Docs change here https://github.com/compiled/website/pull/54


madou:
@pgmanutd want to do an early review?

To run locally:

```bash
git checkout webpack-plugin-spike
yarn
yarn start:webpack
```

packages/babel-plugin/src/babel-plugin.tsx
```diff
@@ -0,0 +1,195 @@
+import { declare } from '@babel/helper-plugin-utils';
```
madou, line 1: this file hasn't changed - its just been renamed.

packages/babel-plugin/src/index.tsx
```diff
@@ -1,185 +1,2 @@
...
-          return;
-        }
-
-        visitCssPropPath(path, { state, parentPath: path });
-      },
-    },
-  };
-});
+export { default } from './babel-plugin';
+export { transform } from './transform';
```
madou, line 2: so we can transform directly if needed. currently using this in webpack loader - however it might not be needed in parcel. maybe.

packages/babel-plugin/src/types.tsx
```diff
@@ -11,8 +11,12 @@ export interface PluginOptions {
 
   /**
    * Whether to use the cache or not. Will make subsequent builds faster.
+   *
+   * - `true` caches for the duration of the node run (useful for single builds)
+   * - `"file-pass"` caches per file pass (useful for watch mode)
+   * - `false` turns caching off
    */
-  cache?: boolean;
+  cache?: boolean | 'file-pass';
```
madou, line 19: not too sure if this is needed anymore. keeping cache as `true` doesn't seem to affect the run (things still change as expected).

pgmanutd:
> @pgmanutd want to do an early review?
> 
> To run locally:
> 
> ```shell
> git checkout webpack-plugin-spike
> yarn
> yarn start:webpack
> ```

Noice. I ll have a look 😃. I looked it once few days back. I think lot has changed since then.

madou:
@pgmanutd I'm going to merge this to use in the parcel spike (want to push this out anyway 😄) - if you can have a review and we can change things in a follow up PR if needed 👍

packages/babel-plugin/src/transform.tsx
```diff
@@ -0,0 +1,33 @@
...
+ * @param code Code to transform
+ * @param opts Userland options
+ */
+export async function transformAsync(code: string, opts: TransformOpts): Promise<TransformResult> {
+  const includedFiles: string[] = [];
+
+  const result = await babelTransformAsync(code, {
+    babelrc: false,
+    configFile: false,
+    filename: opts.filename,
```
madou, line 23: will break TS & Flow I think. Need to infer some config from userland.

pgmanutd:
> @pgmanutd I'm going to merge this to use in the parcel spike (want to push this out anyway 😄) - if you can have a review and we can change things in a follow up PR if needed 👍

No worries. Later it is 😄

PR #561, opened 2021-02-14
Extract to CSS spike

madou:
Nothing to see here, yet.


pgmanutd:
> Nothing to see here, yet.

🤔 hehe now? Sorry not in touch with compiled lately. Are we removing all the imports, `CC`, `CS` wrappers, local atomic variables, `ax` & `ix` runtimes and pulling children up? This way they we will have only `classnames` left right? And what about extracted `sheets` 🤔? Are we thinking of creating fake random stylesheet import and putting these sheets into that?

madou:
> Are we removing all the imports, `CC`, `CS` wrappers, local atomic variables, `ax` & `ix` runtimes and pulling children up?

Hi! Almost - see https://github.com/atlassian-labs/compiled/pull/561/files#diff-371fb3e15b7968f4a468c26cdfc3f951134bf4caf6263ef77a8a99ad666da7b5R47

`ax` and `ix` still remain for now.

> And what about extracted `sheets` 🤔? Are we thinking of creating fake random stylesheet import and putting these sheets into that?

Still thinking about this, but yeah something like this 🤔

pgmanutd:
> > Are we removing all the imports, `CC`, `CS` wrappers, local atomic variables, `ax` & `ix` runtimes and pulling children up?
> 
> Hi! Almost - see https://github.com/atlassian-labs/compiled/pull/561/files#diff-371fb3e15b7968f4a468c26cdfc3f951134bf4caf6263ef77a8a99ad666da7b5R47
> 
> `ax` and `ix` still remain for now.
> 
> > And what about extracted `sheets` 🤔? Are we thinking of creating fake random stylesheet import and putting these sheets into that?
> 
> Still thinking about this, but yeah something like this 🤔

Noice. Instead of writing to file system & importing, faking imports & resolving our extracted sheets might work. Not sure. I did similar for express in past. Serving stylesheets on fly with caching based on its content and it worked.

madou:
cool its working

**not extracted**

image

image

**extracted**

image

image

packages/babel-plugin-extract/src/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+      if (t.isObjectProperty(prop) && t.isExpression(prop.value)) {
+        children.push(prop.value);
+      }
+    });
+  }
+
+  return children;
+};
+
+const removeStyleDeclarations = (node: t.Node, parentPath: NodePath<any>, pass: PluginPass) => {
```
DarkPurple141, line 38:
```suggestion
const removeStyleDeclarations = <T extends {} = any>(node: t.Node, parentPath: NodePath<T>, pass: PluginPass) => {
```
Dunno, may get better type inference from consumers?

packages/babel-plugin-extract/src/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+    const [styles] = node.children;
+    if (t.isJSXExpressionContainer(styles) && t.isArrayExpression(styles.expression)) {
+      styles.expression.elements.forEach((value) => {
+        if (!t.isIdentifier(value)) {
+          return;
+        }
+
+        const binding = parentPath.scope.getBinding(value.name);
+        if (binding) {
+          const value = ((binding?.path?.node as t.VariableDeclarator)?.init as t.StringLiteral)
```
DarkPurple141, line 91: eesh

DarkPurple141 (commented):
Dunno if I've got enough context.

madou:
Closing to start again.

PR #626, opened 2021-03-11
chore: resolves eslint warnings

madou:
Warnings got to over 100 so I figured it's time to fix them (and make sure they don't come up again!)

### DIFF
--- a/examples/packages/benchmarks/src/ssr/playground.tsx
+++ b/examples/packages/benchmarks/src/ssr/playground.tsx
@@ -2,15 +2,17 @@ import React, { useContext, createContext } from 'react';
 
 const Cache = createContext<Record<string, true> | null>(null);
 
-export const useCache = () => {
+export const useCache = (): Record<string, true> => {
   return useContext(Cache) || {};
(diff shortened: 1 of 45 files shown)


examples/packages/benchmarks/src/ssr/playground.tsx
```diff
@@ -2,15 +2,17 @@ import React, { useContext, createContext } from 'react';
 
 const Cache = createContext<Record<string, true> | null>(null);
 
-export const useCache = () => {
+export const useCache = (): Record<string, true> => {
```
DarkPurple141, line 5: this return type seems sus, also wouldn't it better be `typeof Cache` ?

DarkPurple141 (approved):
nothing really to call it that I could see.

examples/packages/benchmarks/src/ssr/playground.tsx
madou: potato potato

its smelly but its the same type after all. also this is in benchmarks who cares hehe

Issue #388, opened 2020-12-07
Styles are duplicated when SSRd


madou:
**Describe the bug**
Styles are duplicated when server-side rendered.

**To Reproduce**
See: https://github.com/atlassian-labs/compiled/pull/381#issuecomment-736077147

**Expected behavior**
Styles should not be duplicated.

PR #649, opened 2021-03-17
fix: style duplication issue #388

MatthewCharlton:
To solve the issue I have added a script tag as a sibling to the style tag when SSR'd. The script tag contains the stringifyed inserted cache data and is added to the client cache object when the bundle is executed.

This solution does have a few downsides: 
1) Adding an extra DOM node next to the style tag with the inserted classnames cache effectively doubles the DOM nodes needed to operate.
2) When the size of the application grows and there are a large number of SSR'd components there is potential that the parsing of both the style and script tags will block execution long enough to be noticeable. It might be prudent to add a batching function. e.g. raf or settimeout.

### DIFF
--- a/examples/packages/ssr/src/app.js
+++ b/examples/packages/ssr/src/app.js
@@ -11,6 +11,9 @@ const Header = () => (
       height: 56,
       display: 'flex',
       alignItems: 'center',
+      '@media only screen and (max-width:1100px)': {
+        background: 'green',
+      },
     }}>
     server side rendering
   </header>
--- a/examples/packages/ssr/src/app.js
+++ b/examples/packages/ssr/src/app.js
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import '@compiled/react';
+import { SSRCacheComponent } from '@compiled/react/runtime';
 
 const Footer = () => <footer css={{ background: 'purple', padding: 8 * 4 }}>footer</footer>;
 
@@ -19,16 +20,25 @@ const Header = () => (
   </header>
 );
 
+const Button = ({ count, setCount }) => (
+  <button css={{ background: 'white', borderRadius: 3 }} onClick={() => setCount((count += 1))}>
+    Count {count}
+  </button>
+);
+
 const Content = () => (
   <main css={{ padding: 8 * 4, height: '200vh', background: 'blue' }}>content</main>
 );
 
 export default function Home() {
+  const [count, setCount] = React.useState(0);
   return (
     <div css={{ fontSize: '100%' }}>
       <Header />
+      <Button count={count} setCount={setCount} />
       <Content />
       <Footer />
+      <SSRCacheComponent />
     </div>
   );
 }
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     },
     {
       "path": "./packages/react/dist/browser/runtime.js",
-      "limit": "445B",
+      "limit": "599B",
       "import": "{ CC, CS }",
       "ignore": [
         "react"
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     },
     {
       "path": "./packages/react/dist/browser/runtime.js",
-      "limit": "585B",
+      "limit": "538B",
       "import": "{ CC, CS }",
       "ignore": [
         "react"
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     },
     {
       "path": "./packages/react/dist/browser/runtime.js",
-      "limit": "599B",
+      "limit": "585B",
       "import": "{ CC, CS }",
       "ignore": [
         "react"
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     },
     {
       "path": "./packages/react/dist/browser/runtime.js",
-      "limit": "538B",
+      "limit": "729B",
       "import": "{ CC, CS }",
       "ignore": [
         "react"
--- a/packages/react/src/runtime/sheet.tsx
+++ b/packages/react/src/runtime/sheet.tsx
@@ -75,6 +75,8 @@ function lazyAddStyleBucketToHead(bucketName: Bucket, opts: StyleSheetOpts): HTM
     }
 
     const tag = document.createElement('style');
+    // set data-cmpld attribute to "h" for hydrated
+    tag.setAttribute('data-cmpld', 'h');
     opts.nonce && tag.setAttribute('nonce', opts.nonce);
     tag.appendChild(document.createTextNode(''));
     styleBucketsInHead[bucketName] = tag;
--- a/packages/react/src/runtime/style-cache.tsx
+++ b/packages/react/src/runtime/style-cache.tsx
@@ -7,8 +7,7 @@ import { ProviderComponent, UseCacheHook } from './types';
  * Cache to hold already used styles.
  * React Context on the server - singleton object on the client.
  */
-let ClientCache = {};
-const Cache: any = createContext<Record<string, true> | null>(null) || ClientCache;
+const Cache: any = isNodeEnvironment() ? createContext<Record<string, true> | null>(null) : {};
 
 if (!isNodeEnvironment()) {
   /**
@@ -17,20 +16,38 @@ if (!isNodeEnvironment()) {
   const ssrStyles = document.querySelectorAll<HTMLStyleElement>('style[data-cmpld]');
   for (let i = 0; i < ssrStyles.length; i++) {
     // Create the client cache for all SSR'd classes
-    const inserted: Record<string, true> = {};
     const stylesheet = ssrStyles[i];
     const rulesText = stylesheet.innerText;
 
     if (rulesText) {
-      rulesText.split('}').forEach((rule: string) => {
-        const sheet = `${rule}}`;
-        inserted[sheet] = true;
-      });
+      // We split the stylesheet innerText by the closing } then loop over the array
+      const arr = rulesText.split('}');
+
+      // Nested rule string
+      let nestedRule = '';
+
+      for (let i = 0; i < arr.length; i++) {
+        const myRule = `${arr[i]}}`;
+        // check if rule starts with @
+        if (myRule.charCodeAt(0) === 64) {
+          // begin building nested rule string
+          nestedRule = myRule;
+          // if the nested rule ends with }} then it's considered complete
+        } else if (nestedRule.substr(-2) === '}}') {
+          // add nested rule to Cache and reset nestedRule string
+          Cache[nestedRule] = true;
+          nestedRule = '';
+          // if we have started building a nested rule but it is not complete then append an }
+        } else if (nestedRule) {
+          nestedRule += '}';
+        } else {
+          // if not an @ rule then just add to Cache
+          Cache[myRule] = true;
+        }
+      }
     }
 
-    ClientCache = { ...ClientCache, ...inserted };
     // Move all found server-side rendered style elements to the head before React hydration happens.
-
     document.head.appendChild(stylesheet);
   }
 }
@@ -47,7 +64,7 @@ export const useCache: UseCacheHook = () => {
   }
 
   // On the client we use the object singleton.
-  return ClientCache;
+  return Cache;
 };
 
 /**
--- a/packages/react/src/runtime/style-cache.tsx
+++ b/packages/react/src/runtime/style-cache.tsx
@@ -7,13 +7,23 @@ import { ProviderComponent, UseCacheHook } from './types';
  * Cache to hold already used styles.
  * React Context on the server - singleton object on the client.
  */
-const Cache: any = isNodeEnvironment() ? createContext<Record<string, true> | null>(null) : {};
+let ClientCache: { [key: string]: any } = {};
+const Cache: any = createContext<Record<string, true> | null>(null) || ClientCache;
 
 if (!isNodeEnvironment()) {
   /**
+   * Iterates through all found script elements with styles already inserted when server side rendering.
+   */
+  const compiledCaches = document.querySelectorAll<HTMLScriptElement>('script[data-cmpld-cache]');
+  for (let i = 0; i < compiledCaches.length; i++) {
+    let cacheData = {};
+    try {
+      cacheData = JSON.parse(compiledCaches[i].innerText);
+    } catch {}
+    ClientCache = { ...ClientCache, ...cacheData };
+  }
+  /**
    * Iterates through all found style elements generated when server side rendering.
-   *
-   * @param cb
    */
   const ssrStyles = document.querySelectorAll<HTMLStyleElement>('style[data-cmpld]');
   for (let i = 0; i < ssrStyles.length; i++) {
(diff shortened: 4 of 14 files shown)


madou:
cool! thanks matt. will dig in soon.

MatthewCharlton:
I am unable to run the benchmarks or the storybook performance examples to see how much of a negative impact the extra DOM nodes and processing has on performance. I have another idea similar to this not using a script tag but I don't know performance-wise how that will perform vs this solution.

madou:
Yeah! It's tricky. I think we don't want to move too quickly here -- currently the duplicated styles are quite _benign_. They look weird from a developers perspective, but they won't affect the styles on the page.

Because of that, we want to avoid increasing bundle size and SSR output where possible.

I'm hesitant to move forward with this implementation purely because of that TBH - what was your other idea? 😄

MatthewCharlton:
Yeh the other idea is the same concept but instead of using the cache object directly you can parse the style innerText for the classes then create the cache object. 
This means no extra DOM nodes and you can do the text parsing operation in the same loop when moving the style tags to the head.

madou:
Ok, put up a spike and let's talk through it!

MatthewCharlton:
Ok so I have update the branch with the other implementation. Let me know what you think

packages/react/src/runtime/style-cache.tsx
```diff
@@ -7,18 +7,31 @@ import { ProviderComponent, UseCacheHook } from './types';
  * Cache to hold already used styles.
  * React Context on the server - singleton object on the client.
  */
-const Cache: any = isNodeEnvironment() ? createContext<Record<string, true> | null>(null) : {};
+let ClientCache = {};
+const Cache: any = createContext<Record<string, true> | null>(null) || ClientCache;
```
madou, line 11: this will always use the `createContext` flow - can we revert to what it was previously? it's behind a node environment check so the `create context` flow never makes its way to the client

```suggestion
const Cache: any = isNodeEnvironment() ? createContext<Record<string, true> | null>(null) : {};
```

packages/react/src/runtime/style-cache.tsx
```diff
@@ -7,18 +7,31 @@ import { ProviderComponent, UseCacheHook } from './types';
...
    */
   const ssrStyles = document.querySelectorAll<HTMLStyleElement>('style[data-cmpld]');
   for (let i = 0; i < ssrStyles.length; i++) {
+    // Create the client cache for all SSR'd classes
+    const inserted: Record<string, true> = {};
+    const stylesheet = ssrStyles[i];
+    const rulesText = stylesheet.innerText;
+
+    if (rulesText) {
+      rulesText.split('}').forEach((rule: string) => {
```
madou, line 25: can we use a for loop here? it's faster

an edge case with this implementation will be nested rules, how can we accommodate this? 🤔 

```
@media screen {
  .rule { display: block }
}
```

packages/react/src/runtime/style-cache.tsx
```diff
@@ -7,18 +7,31 @@ import { ProviderComponent, UseCacheHook } from './types';
...
+        const sheet = `${rule}}`;
+        inserted[sheet] = true;
+      });
+    }
+
+    ClientCache = { ...ClientCache, ...inserted };
```
madou, line 31:
```suggestion
```

packages/react/src/runtime/style-cache.tsx
```diff
@@ -7,18 +7,31 @@ import { ProviderComponent, UseCacheHook } from './types';
...
 if (!isNodeEnvironment()) {
   /**
    * Iterates through all found style elements generated when server side rendering.
-   *
-   * @param cb
    */
   const ssrStyles = document.querySelectorAll<HTMLStyleElement>('style[data-cmpld]');
   for (let i = 0; i < ssrStyles.length; i++) {
+    // Create the client cache for all SSR'd classes
+    const inserted: Record<string, true> = {};
```
madou, line 20:
```suggestion
```

packages/react/src/runtime/style-cache.tsx
```diff
@@ -7,18 +7,31 @@ import { ProviderComponent, UseCacheHook } from './types';
...
+    const stylesheet = ssrStyles[i];
+    const rulesText = stylesheet.innerText;
+
+    if (rulesText) {
+      rulesText.split('}').forEach((rule: string) => {
+        const sheet = `${rule}}`;
+        inserted[sheet] = true;
```
madou, line 27:
```suggestion
+       Cache[sheet] = true;
```

packages/react/src/runtime/style-cache.tsx
```diff
@@ -7,18 +7,31 @@ import { ProviderComponent, UseCacheHook } from './types';
  * Cache to hold already used styles.
  * React Context on the server - singleton object on the client.
  */
-const Cache: any = isNodeEnvironment() ? createContext<Record<string, true> | null>(null) : {};
+let ClientCache = {};
```
madou, line 10:
```suggestion
```

packages/react/src/runtime/style-cache.tsx
```diff
@@ -27,14 +40,14 @@ if (!isNodeEnvironment()) {
...
-    // On the server we use React Context to we don't leak the cache between SSR calls.
+    // On the server we use React Context so we don't leak the cache between SSR calls.
     // During runtime this hook isn't conditionally called - it is at build time that the flow gets decided.
     // eslint-disable-next-line react-hooks/rules-of-hooks
     return useContext(Cache) || {};
   }
 
   // On the client we use the object singleton.
-  return Cache;
+  return ClientCache;
```
madou, line 50:
```suggestion
  return Cache;
```

madou (changes requested):
hey mate, had a look over. good direction! few things to iron out

packages/react/src/runtime/style-cache.tsx
MatthewCharlton: I didn't use a for loop initially because it uses more bytes. Happy to change it. For future reference I assume performance takes precedence over size?

MatthewCharlton:
BTW TypeScript blows up when I try to run the benchmarks. Is this a known issue? I feel like we should we fix the benchmarks first so we can validate these changes. Want me to create an issue to resolve it?

madou:
> BTW TypeScript blows up when I try to run the benchmarks. Is this a known issue? I feel like we should we fix the benchmarks first so we can validate these changes. Want me to create an issue to resolve it?

The benchmarks were created in an ad-hoc way TBH, sure let's make an issue to fix them up (and also run them in CI).

packages/react/src/runtime/style-cache.tsx
```diff
@@ -12,13 +12,43 @@ const Cache: any = isNodeEnvironment() ? createContext<Record<string, true> | nu
...
+        } else {
+          // if not an @ rule then just add to Cache
+          Cache[myRule] = true;
+        }
+      }
+    }
+
     // Move all found server-side rendered style elements to the head before React hydration happens.
-    document.head.appendChild(ssrStyles[i]);
+    document.head.appendChild(stylesheet);
```
madou, line 51: hmm there's a high chance the styles we add won't be ordered, which could create some problems. as a fix we can use `insertRule`:

```
import insertRule from './sheet';

insertRule(myRule);
```

packages/react/src/runtime/style-cache.tsx
```diff
@@ -12,13 +12,43 @@ const Cache: any = isNodeEnvironment() ? createContext<Record<string, true> | nu
 if (!isNodeEnvironment()) {
   /**
    * Iterates through all found style elements generated when server side rendering.
-   *
-   * @param cb
    */
   const ssrStyles = document.querySelectorAll<HTMLStyleElement>('style[data-cmpld]');
   for (let i = 0; i < ssrStyles.length; i++) {
+    // Create the client cache for all SSR'd classes
+    const stylesheet = ssrStyles[i];
```
madou, line 19: can you see if we can write some unit tests? looking at the tests there aren't any written, so it might just be a bit tricky to test this flow. let me know how you go 😄 

would be interested in these test cases:

1. hydrate single style
2. hydrate nested styles
3. hydrate ordered styles (bunch of pseudo classes, media queries, and regular CSS rules)

examples/packages/ssr/src/app.js
```diff
@@ -11,21 +12,33 @@ const Header = () => (
...
       alignItems: 'center',
+      '@media only screen and (max-width:1100px)': {
+        background: 'green',
+      },
     }}>
     server side rendering
   </header>
 );
 
+const Button = ({ count, setCount }) => (
```
MatthewCharlton, line 23: I added this to test the style consolidation of possibly 1000's of style elements was non-blocking

package.json
```diff
@@ -112,7 +112,7 @@
     },
     {
       "path": "./packages/react/dist/browser/runtime.js",
-      "limit": "445B",
+      "limit": "729B",
```
MatthewCharlton, line 115: I know :(
However if the changes proposed in this PR are considered the way to go then perhaps there are ways to reduce this.

packages/react/src/runtime/sheet.tsx
```diff
@@ -75,6 +75,8 @@ function lazyAddStyleBucketToHead(bucketName: Bucket, opts: StyleSheetOpts): HTM
     }
 
     const tag = document.createElement('style');
+    // set data-cmpld attribute to "h" for hydrated
+    tag.setAttribute('data-cmpld', 'h');
```
MatthewCharlton, line 79: The idea here is mainly to tell which style tags were created by the library. Can revert this if needed.

packages/react/src/runtime/style-cache.tsx
```diff
@@ -1,24 +1,70 @@
...
+              break;
+          }
+
+          // Add the rule to a new style tag in the head
+          insertRule(rule, { nonce });
+        }
+      }
+      // remove SSR'd style tag
+      stylesheet.remove();
+    }, 20);
```
MatthewCharlton, line 67: Time value is arbitrary, can tweak to see what gives the best results.

packages/react/src/runtime/style.tsx
```diff
@@ -43,7 +43,8 @@ export default function Style(props: StyleProps): JSX.Element | null {
       }
 
       return (
-        <style data-cmpld nonce={props.nonce}>
+        // set data-cmpld attribute to "s" for SSR
+        <style data-cmpld="s" nonce={props.nonce}>
```
MatthewCharlton, line 47: This obviously adds a few bytes here but when there are hundreds of SSR'd style tags on the DOM it's actually fewer bytes because React turns this into ```data-cmpld="true"``` so it's actually 3 less bytes per style tag.

packages/react/src/runtime/style-cache.tsx
```diff
@@ -1,24 +1,70 @@
...
 
 /**
  * Cache to hold already used styles.
  * React Context on the server - singleton object on the client.
  */
-const Cache: any = isNodeEnvironment() ? createContext<Record<string, true> | null>(null) : {};
+let Cache: any = isNodeEnvironment() ? createContext<Record<string, true> | null>(null) : {};
 
 if (!isNodeEnvironment()) {
+  const ssrCache = document.querySelector('script[data-cmpld="c"]');
```
MatthewCharlton, line 14: I ran a bunch of performance tests on the other method and it started to get quite bad when there were ~1000 SSR'd style tags on the DOM. This creates the cache immediately. Then we can move all the SSR'd style tags to the head where we can asynchronously begin to consolidate them using ```insertRule```.

packages/react/src/runtime/style-cache.tsx
```diff
@@ -1,24 +1,70 @@
...
    */
-  const ssrStyles = document.querySelectorAll<HTMLStyleElement>('style[data-cmpld]');
+  const ssrStyles = document.querySelectorAll<HTMLStyleElement>('style[data-cmpld="s"]');
   for (let i = 0; i < ssrStyles.length; i++) {
-    // Move all found server-side rendered style elements to the head before React hydration happens.
-    document.head.appendChild(ssrStyles[i]);
+    // Create the client cache for all SSR'd classes
+    const stylesheet = ssrStyles[i];
+    // move all SSR style tags to the head straight away
+    document.head.appendChild(stylesheet);
```
MatthewCharlton, line 32: We move these to the head so the styles don't disappear. Then we go through and consolidate them.

examples/packages/ssr/src/app.js
```diff
@@ -11,21 +12,33 @@ const Header = () => (
...
 
 export default function Home() {
+  const [count, setCount] = React.useState(0);
   return (
     <div css={{ fontSize: '100%' }}>
       <Header />
+      <Button count={count} setCount={setCount} />
       <Content />
       <Footer />
+      <SSRCacheComponent />
```
MatthewCharlton, line 41: Not sure how we want to handle this but I put this here for testing but thinking about it I think this should go in the ```server.js``` file and wrapped in the ```CacheProvider```.

Like: 
```js
... snip ...
import { SSRCacheComponent, CC as CacheProvider } from '@compiled/react/runtime'; // perhaps we can re-export CC as CacheProvider so you can just import CacheProvider directly?
import App from './app';

const assets = require(process.env.RAZZLE_ASSETS_MANIFEST);

const server = express();
server
  .disable('x-powered-by')
  .use(express.static(process.env.RAZZLE_PUBLIC_DIR))
  .get('/*', (req, res) => {
    const context = {};
    const markup = renderToString(
      <CacheProvider>
        <App />
        <SSRCacheComponent />
      </CacheProvider>
    );
... snip ...

**Comment by madou** [comment]
Hey! Will have a review when I can. Had a shower thought though since we're writing a lot of code to handle this -- what if we just... didn't fix this? It would be benign, it doesn't affect anything, and we keep bundle size and runtime as fast as it can be.

It would also have the same behaviour when we do critical CSS with extracted styles (#671). Maybe that's okay?

File: packages/react/src/runtime/style-cache.tsx [hunk 18]
```diff
@@ -1,24 +1,70 @@
...
    * Iterates through all found style elements generated when server side rendering.
-   *
-   * @param cb
    */
-  const ssrStyles = document.querySelectorAll<HTMLStyleElement>('style[data-cmpld]');
+  const ssrStyles = document.querySelectorAll<HTMLStyleElement>('style[data-cmpld="s"]');
   for (let i = 0; i < ssrStyles.length; i++) {
-    // Move all found server-side rendered style elements to the head before React hydration happens.
-    document.head.appendChild(ssrStyles[i]);
+    // Create the client cache for all SSR'd classes
```
MatthewCharlton, line 29: - [ ] remove this comment. It is outdated

packages/react/src/runtime/style-cache.tsx
MatthewCharlton: Have added tests. Will add more to cover the rest of the cases above

MatthewCharlton:
@madou Hey, yeh I had the same thought, however the current implementation simply moves the SSR'd styles to the head instead of consolidating them using ```insertRule``` which was your concern [here](https://github.com/atlassian-labs/compiled/pull/649#discussion_r606720163), it also pollutes the head with many style tags.
For me this proposal became less about the caching and more about the styles consolidation the more I worked on it. 
I think the caching problem is solved here with a pretty simple solution, and requires very little user effort. It's the conversion from SSR'd styles to ```insertRule``` that I am also trying to solve. Perhaps the scope has increased too much?
Let me know what you want to do. Feel free to decline the PR and I can work on something else.

madou:
Closing for now, the current trade off isn't worth while (benign duplicates vs. introducing more runtime to handle it).

PR #658, opened 2021-03-21
Extract CSS for webpack4

madou:
I've struggled to be able to test this with multiple versions of webpack existing in the repo. Even being able to test that it's working is problematic (to keep both webpack 4 + 5 working).

I'll probably need to opt out of workspaces to make it work. Maybe.

**TODO**

- [x] Cleanup code
- [ ] Try to test the code


danieldelcore (approved):
LGTM 👍 !

PR #695, opened 2021-04-02
Parcel extraction

madou:
**Local playground**

```bash
yarn
yarn start:parcel
```

Example in `examples/packages/parcel`

**New packages**

- `parcel-transformer-css` - converts `css:xyz?style={rule}` import query params into CSS content, equivalent to the css loader we wrote for webpack
- `parcel-optimizer` - sorts a CSS bundle
- `parcel-config` - holds all config for compiled as a single entrypoint, convince wrapper

**`parcel-transformer` changes**

- Add extract option
- Extract from node modules
- Add `css:` import to file pass for every style declaration found

**Questions**

- Should we add the imports via the AST?
- How can we share the AST with the babel transformer? It throws an error
- How can we move styles from one bundle to another? (E.g. hoisting unstable CSS to the primary CSS bundle)
- Else - Is it possible to disable code splitting CSS? `asset.isSplittable = false` doesn't work
- Is it possible to "delete" the `css:xyz` imports after transformation?

### DIFF
--- a/examples/packages/parcel/.compiledcssrc
+++ b/examples/packages/parcel/.compiledcssrc
@@ -1,3 +1,4 @@
 {
-  "importReact": false
+  "importReact": false,
+  "extract": true
 }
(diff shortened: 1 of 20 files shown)


madou:
Closing, this can be used as a reference when picking it up later.

PR #724, opened 2021-04-11
Code split CSS chunks

### LINKED ISSUE #669: NextJS CSS extraction support
NextJS support is desirable, but currently it doesn't work well. There are two schools of problems:

1. Style sheet sorting doesn't work (starting a conversation here https://github.com/vercel/next.js/issues/8582#issuecomment-804678546)
2. SSR build doesn't like the CSS imports we place

We need to investigate these two areas.

We want to create an example that has no extraction during dev, extraction during prod, and sorting & SSR work as expected.

- When extracted SSR should not interweave style tags (because the runtime is stripped).
- Style sheet should be sorted

**Additional context**
We currently collate CSS into a single style sheet using webpack cache groups - nextJS seemingly disables this. We might need an alternative implementation.

### LINKED ISSUE #722: [Webpack] Code split CSS
Currently CSS extraction forces all styles to be inside one style sheet to simplify the first implementation. Ideally, we want to be able to only load the CSS that is used instead of downloading it all, however there are some constraints:

1. Unsafe atomic styles need to be hoisted to the top entry point
2. All style sheets need to be sorted

When we say unsafe atomic styles, we're primarily talking about pseudo styles defined here https://compiledcssinjs.com/docs/atomic-css#automatic-pseudo-class-ordering.

**Why do they need to be hoisted?**

If we don't the order of applied styles won't be consistent, resulting in different experiences depending what chunks were loaded and when. We want to avoid that at all costs.

madou:
Closes #722

Seems to work, need to think about edge cases. As well as what happens with multiple entry points. This might also closes #669 as we remove the need for placing everything into a single style sheet via cache groups.

@jantimon how's this look?

**Changes**

- Compiled now supports async CSS chunks
- To enable this Compiled now takes control over the entire CSS pipeline
- Any CSS declared inside or outside of Compiled will be sorted and potentially hoisted if unsafe for atomic styles

**TODO**

- [x] Investigate potential edge cases & side effects
- [ ] Update docs to talk through how Compiled will take control of the entire CSS pipeline

### DIFF
--- a/.changeset/fresh-tools-refuse.md
+++ b/.changeset/fresh-tools-refuse.md
@@ -1,0 +1,5 @@
+---
+'@compiled/webpack-loader': minor
+---
+
+**BREAKING** Compiled now takes control of the entire CSS pipeline. This means any CSS declared inside or outside of Compiled will be sorted and potentially hoisted if unsafe.
--- a/.changeset/large-pans-retire.md
+++ b/.changeset/large-pans-retire.md
@@ -1,0 +1,5 @@
+---
+'@compiled/css': patch
+---
+
+Sort now can remove unstable atomic rules with the `removeUnstableRules` option set to `true`.
--- a/.changeset/silent-planets-invent.md
+++ b/.changeset/silent-planets-invent.md
@@ -1,0 +1,5 @@
+---
+'@compiled/webpack-loader': patch
+---
+
+Compiled now supports async chunked CSS. When components are code split and have unique styles only used in that chunk, its styles will be in their own style sheet.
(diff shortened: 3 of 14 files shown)


cloudkite:
Nice work!! Does this enable multiple entry points? Not sure how common this is in the wild, but our app uses multiple entry points to build bundles for independent apps (website, chrome extension, embed etc) within a monorepo. Being able to only include styles used for a particular entry point would be great!

madou:
Hi! Yeah exactly, need to investigate what happens and probably tweak it 🙂

madou:
Closing, this can be used as a reference when picking it up later.

PR #759, opened 2021-05-14
Catch unhandled exception on inserting rules with prefixed selectors

skoob13:
Usage of `::placeholder` pseudo element throws the error in production when it gets prefixed:

```
DOMException: Failed to execute 'insertRule' on 'CSSStyleSheet': Failed to parse the rule '._1idr12ax::-moz-placeholder{color:red}'.
```

After some research I found that Chrome specifically throws an exception on `insertRule` with prefixed pseudo elements. The easiest way to fix is to add try/catch block. For example, [Emotion](https://github.com/emotion-js/emotion/blob/f3c51e8aca287170e43110bee2e4527eacfcada4/packages/sheet/src/index.js#L137) handles the exception similarly. They even have a comment for `::placeholder` pseudo element and `:read-write` selector.

Reproduction: [Codesandbox](https://codesandbox.io/s/competent-bush-dcnoi?file=/pages/index.js) (development).
Demo: https://csb-dcnoi-ei3wqlflp-georgiyridgebox.vercel.app/ (production env where `insertRule` is used).

One thing that I can't figure out is how to make a test🙂 It requires a browser environment.

### DIFF
--- a/.changeset/stale-badgers-exercise.md
+++ b/.changeset/stale-badgers-exercise.md
@@ -1,0 +1,5 @@
+---
+'@compiled/react': patch
+---
+
+Catch unhandled exception on inserting rules with prefixed selectors.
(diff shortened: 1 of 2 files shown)


madou:
Thanks! All good on the test front.

madou (approved):

PR #769, opened 2021-05-23
Consider adding babel-plugin-macros to bundle transformers

iiwii:
Adds macro support to be able to use macros such as [`twin.macro`](https://github.com/ben-rogerson/twin.macro) with the loader/transformer. Previously, taking Twin as an example, I could only get it working using `@compiled/babel-plugin`, and so Webpack CSS extraction wasn't possible.


madou:
Hi thanks for raising this PR. Currently there isn't a desire to hard code babel macros into the webpack plugin so this will need some extra thinking, could you raise an issue and we can discuss potential ways forward please? Cheers

PR #776, opened 2021-07-10
feat(css function): add support of additional parameters

Dwlad90:
Add support of array additional parameters to css function.

The bug: 

```
 const colors = { blue: () => 'blue' };
  return (
    <div
      css={css`
      /*Expected 1 arguments, but got 2.*/
        background-color: ${colors.blue()};
        color: #00f;      `}>
      I&m blue
    </div>
  );
```
[Link](https://codesandbox.io/s/distracted-morning-ihd3h?file=/pages/index.tsx) to sandbox with the problem

[Proof to the solution on StackOverflow](https://stackoverflow.com/questions/49442210/template-interpolation-in-function-argument-results-in-ts2554expected-1-argumen)

### DIFF
--- a/.changeset/red-kids-agree.md
+++ b/.changeset/red-kids-agree.md
@@ -1,0 +1,5 @@
+---
+'@compiled/react': patch
+---
+
+add support of additional parameters to css function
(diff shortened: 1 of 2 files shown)


madou:
Thanks for the type fixed 😄

madou (approved):

PR #790, opened 2021-07-29
Codemod plugin system

JakeLane:
Simple plugin system that allows extension to the JSCodemod with a defined interface. Currently, only two methods supported but will be extended in future PRs.

This PR also pulls out the codemod into a new package, `@compiled/codemods`. @madou anything special need to happen here?


madou:
Hey @JakeLane thanks for the ping. Looking good! Main thing to point out was [project references](https://www.typescriptlang.org/docs/handbook/project-references.html) are setup correctly, which it looks like you've done!

The other is to make sure you create a changeset for the new package, run `yarn changeset` and follow the prompts, and then check it in. Any change we put up should include a changeset describing what the change is, and which packages are affect 😄.

As a side to the changeset stuff, make sure to check what renovate is updating as they don't have changesets unfortunately, something we might want to automate later.

JakeLane:
Thanks for the heaps up @madou!

packages/codemods/src/emotion-to-compiled/__tests__/emotion-to-compiled.test.tsx
```diff
@@ -3,7 +3,7 @@ jest.disableAutomock();
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const defineInlineTest = require('jscodeshift/dist/testUtils').defineInlineTest;
 
-import transformer from '../emotion-to-compiled';
+import { transformer } from '../emotion-to-compiled';
```
MonicaOlejniczak, line 6: I'm not a huge fan of exporting the `transformer` just for testing. Is there anything wrong with using the public API through the default import here? It seems like you just need to rename `pluginModule` to `plugin`, and handle passing in the `pluginModule` internally as well as paths.

packages/codemods/src/codemods-helpers.tsx
```diff
@@ -10,11 +11,39 @@ import {
...
+const getPlugin = async (pluginPath: string | undefined) => {
+  if (pluginPath) {
+    try {
+      const pluginModule = await import(pluginPath);
+      return pluginModule.default;
+    } catch (err) {
+      console.error(
+        chalk.red(`${chalk.bold(`Plugin at path '${pluginPath}' was not loaded`)}\n${err}`)
+      );
+      process.exit(1);
```
MonicaOlejniczak, line 34: I would just throw the error here, if you really want the exit code then specify `process.exitCode = 1;`

packages/codemods/src/plugins/contract.ts
```diff
@@ -0,0 +1,62 @@
...
+// We want to ensure the config contract is correct so devs can get type safety
+type ValidateConfig<T, Struct> = T extends Struct
+  ? Exclude<keyof T, keyof Struct> extends never
+    ? T
+    : never
+  : never;
+
+type BaseConfig = { j: JSCodeshift };
+
+export interface CodemodPlugin {
```
MonicaOlejniczak, line 12: Later on this interface might not scale, as it's specific to changing away from one library to compiled (i.e. the codemods currently available). It should be ok for now though.

packages/codemods/src/plugins/contract.ts
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Insert AST nodes after the compiled import
+   *
+   * @param config The configuration object
+   * @param config.j The JSCodeshift object
+   * @param config.newImport The new import node replaced in `buildImport`
+   *
+   * @returns Nodes to insert after the import
+   */
+  insertAfterImport?<T>(
```
MonicaOlejniczak, line 59: How do you see this working with your initial feature flagging attempt? In Jira's case at least, you want to insert before or after the `styled` or `css` declaration, not the import.

packages/codemods/src/emotion-to-compiled/__tests__/emotion-to-compiled.test.tsx
JakeLane: Yeah, this is a mild workaround as testUtils in jscodeshift does not handle async transformations, which breaks testability through those helper libraries

packages/cli/src/presets/codemods/codemods.tsx
```diff
@@ -33,6 +33,11 @@ const codemodChoice: Array<Choice<keyof CodemodOptions>> = [
     name: 'ignorePattern',
     message: '--ignore-pattern',
   },
+  {
+    name: 'plugin',
+    message: '--plugin (path to source)',
```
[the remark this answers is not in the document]
JakeLane: - [ ] update

packages/codemods/src/plugins/contract.ts
```diff
@@ -0,0 +1,62 @@
+import { ImportDeclaration, Collection, JSCodeshift } from 'jscodeshift';
```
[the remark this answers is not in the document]
JakeLane: - [ ] rename

packages/codemods/src/plugins/contract.ts
```diff
@@ -0,0 +1,62 @@
...
+   *
+   * @param config The configuration object
+   * @param config.j The JSCodeshift object
+   * @param config.newImport The new import node replaced in `buildImport`
+   *
+   * @returns Nodes to insert before the import
+   */
+  insertBeforeImport?<T>(
+    config: ValidateConfig<T, BaseConfig & { newImport: Collection<ImportDeclaration> }>
+  ): any;
```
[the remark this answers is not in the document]
JakeLane: I matched the type on the jscodeshift method here - I can possibly be more specific if I read the source

- [ ] check source

packages/codemods/src/plugins/contract.ts
JakeLane: I can rename this to be specific and export it non-default to make the transition later easier

packages/cli/src/presets/codemods/codemods.tsx
```diff
@@ -33,6 +33,11 @@ const codemodChoice: Array<Choice<keyof CodemodOptions>> = [
     name: 'ignorePattern',
     message: '--ignore-pattern',
   },
+  {
+    name: 'plugin',
```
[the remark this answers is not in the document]
JakeLane: I wasn't expecting to support multiple plugins but should be simple enough to implement. My main concern is it getting confusing if the same methods are implemented multiple times. We could simply override behaviours based on the call order when a plugin is specified.

For example, if both plugins implemented `buildImport` in `--plugin /path/to/plugin/1 --plugin /path/to/plugin/2` - plugin 2 would win

packages/codemods/src/plugins/contract.ts
JakeLane: This interface doesn't include all the methods required to implement the Jira plugin yet. I was intending to make a follow-up PR with more functionality to reduce the scope of the PR.

My plan would likely be to add hooks before and after the processing of the css and styled blocks, as those will need to be transformed by the codemod to handle the features not implemented.

packages/cli/src/presets/codemods/codemods.tsx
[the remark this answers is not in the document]
MonicaOlejniczak: Yeah, so multiple plugins becomes iffy when you have both a cli, and the default codemod preset. If a single interface is developed for all plugins to implement, it should be ok. But if not, then you would want to move towards something more like babel or eslint, which enforces you to have a configuration file.

packages/codemods/src/index.ts
```diff
@@ -0,0 +1 @@
+export { CodemodPlugin } from './plugins/contract';
```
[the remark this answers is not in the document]
JakeLane: TIL

packages/codemods/src/README.md
```diff
@@ -0,0 +1,35 @@
+# codemods
```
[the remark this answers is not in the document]
JakeLane: I think I'm going to delete this README and move the usage to the main one - as I see no need for both at the moment

packages/codemods/src/transforms/styled-components-to-compiled/styled-components-to-compiled.tsx
```diff
@@ -4,16 +4,23 @@ import {
   hasImportDeclaration,
   convertDefaultImportToNamedImport,
   addCommentForUnresolvedImportSpecifiers,
-} from '../codemods-helpers';
+  withPlugin,
+} from '../../codemods-helpers';
+import { CodemodPlugin } from '../../plugins/types';
```
MonicaOlejniczak, line 9: /nit `import type`

packages/codemods/src/transforms/emotion-to-compiled/emotion-to-compiled.tsx
```diff
@@ -20,7 +19,9 @@ import {
   replaceImportDeclaration,
   mergeImportSpecifiersAlongWithTheirComments,
   addCommentBefore,
-} from '../codemods-helpers';
+  withPlugin,
+} from '../../codemods-helpers';
+import { CodemodPlugin } from '../../plugins/types';
```
MonicaOlejniczak, line 24: /nit `import type`

packages/codemods/src/plugins/types.ts
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * Insert AST nodes before the compiled import
+   *
+   * @param config The configuration object
+   * @param config.j The JSCodeshift object
+   * @param config.newImport The new import node replaced in `buildImport`
+   *
+   * @returns Nodes to insert before the import
+   */
+  insertBeforeImport?<T>(
```
MonicaOlejniczak, line 54: I still think `insertBeforeImport` and `insertAfterImport` could just be replaced with having access to the program. The babel plugin in compiled uses `visitor: { Program: { exit(path, state) {} } }` to do final insertions, including inserting at particular imports.

packages/cli/src/presets/codemods/codemods.tsx
```diff
@@ -33,6 +33,19 @@ const codemodChoice: Array<Choice<keyof CodemodOptions>> = [
     name: 'ignorePattern',
     message: '--ignore-pattern',
   },
+  {
+    name: 'plugins',
+    message: 'number of plugins',
```
MonicaOlejniczak, line 38: It's a bit weird you have to specify the number of plugins here, is there no way of specifying the plugin paths in a more straightforward way?

packages/codemods/src/codemods-helpers.tsx
```diff
@@ -129,26 +258,32 @@ export const convertDefaultImportToNamedImport = ({
...
 
-      const newNode = importDeclarationPath.node;
+      j(importDeclarationPath).insertAfter(
+        applyAfterImport({ j, plugins, originalImport: importDeclarationPath.node, newImport })
+      );
 
-      if (newNode !== oldNode) {
-        newNode.comments = comments;
-      }
+      // const insertBeforeNodes = (
```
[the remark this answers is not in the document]
JakeLane: agh I pushed the wrong commit, I'll fix this now

packages/cli/src/presets/codemods/codemods.tsx
JakeLane: yeah, I had a lot of trouble coming up with a nice UX for this with the enquirer library. I may be able to do this with newlines - I'll have another go at it

JakeLane: Ended up going to a following prompt with the `List` import

packages/codemods/src/plugins/types.ts
JakeLane: Done - just added `program` as the only entrypoint for now

packages/codemods/src/codemods-helpers.tsx
```diff
@@ -10,11 +11,59 @@ import {
...
+  if (!pluginPathsInput) return [];
+  const pluginPaths = Array.isArray(pluginPathsInput) ? pluginPathsInput : [pluginPathsInput];
+
+  return Promise.all(
+    pluginPaths.map(async (path) => {
+      try {
+        const pluginModule = await import(path);
+
+        const pluginName = pluginModule?.default?.metadata?.name ?? null;
+        if (pluginName === null) {
```
MonicaOlejniczak, line 37: Can you just have `!pluginName` to also check for `undefined` and empty strings, then omit `?? null`?

packages/codemods/src/codemods-helpers.tsx
```diff
@@ -104,13 +153,44 @@ export const findImportSpecifierName = ({
...
+  j: JSCodeshift;
+  plugins: Array<CodemodPlugin>;
+  originalNode: ImportDeclaration;
+  defaultSpecifierName: string;
+  namedImport: string;
+}) =>
+  // Run default plugin first and apply plugins in order
+  [DefaultPlugin, ...plugins].reduce((currentNode, plugin, i, array) => {
+    const buildImportImpl = plugin.migrationTransform?.buildImport ?? null;
+    if (buildImportImpl === null) return currentNode;
```
MonicaOlejniczak, line 172: Same here

MonicaOlejniczak (approved):

Issue #799, opened 2021-08-06
Unconditional rules don't override values in a CSS mixin if defined within 2 mixins


pancaspe87:
**Describe the bug**
When trying to override a property declared in a CSS mixin with a rule that is placed within 2 mixins, the property defined in the mixin is still rendered. Hence override doesn't work

**To Reproduce**

1. Go to 'examples/stories/styled-string.tsx'
2. Add a `font-size: 30px;` between `${primary};` and `${secondary}; `in DoubleUse

**Expected behavior**
Font-size should be 30px instead of 32px.

**Screenshots**
![image](https://user-images.githubusercontent.com/14813563/128450684-3d7dac55-ee9e-48c7-b5a4-66748aa33b34.png)

![image](https://user-images.githubusercontent.com/14813563/128450741-02b2fb68-9db0-4dc2-8b67-948fd6d5177b.png)

pancaspe87:
This should be fixed by the same fix for #798

Issue #798, opened 2021-08-06
Classes are not generated for rules defined before a mixin


pancaspe87:
**Describe the bug**
When placing a CSS declaration before a CSS mixin, only the declarations within the mixin are compiled.

**To Reproduce**

1. Go to 'examples/stories/styled-string.tsx'
2. Add a CSS declaration before `${primary};` in `SingleUse`
3. Inspect element in the browser
4. There are no classes for padding

**Expected behavior**
Element should includes classes also for padding

**Screenshots**
NB:Screenshot was taken when working a different file.
<img width="1541" alt="Screen Shot 2021-08-04 at 11 23 51 am" src="https://user-images.githubusercontent.com/14813563/128471949-2312cd9b-279c-471b-8227-efba7280700d.png">

PR #805, opened 2021-08-10
fix: css is generated for rules defined before a mixins, css mixins can be overriden

pancaspe87:
This PR addresses the issues identified in #798 and #799.

I was able to replicate the issues for each API - namely CSS Prop, ClassNames and Styled - but only for template literals.

When having CSS declarations before a mixin, quasis at the same index of a node expression were not evaluated. Hence classes were not generated.

For example, only the node expression was included in the css output. The quasis with value `'\n  border: 1px solid black;\n` wasn't evaluated. 
Screen Shot 2021-08-10 at 4 57 34 pm


packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -274,6 +274,14 @@ const extractTemplateLiteral = (node: t.TemplateLiteral, meta: Metadata): CSSOut
 
     const { value: interpolation, meta: updatedMeta } = evaluateExpression(nodeExpression, meta);
 
+    // If there is a node Expression ( eg a mixin), the quasi at the same index won't be evaluated
+    // unless it contains a valid CSS declaration with no interpolation.
+    // This is needed to evaluate CSS declarations defined before a mixin
+    if (nodeExpression && templateElementContainsValidStaticCss(q, interpolation)) {
```
MonicaOlejniczak: `nodeExpression` is redundant here, since it is checked for above

packages/babel-plugin/src/class-names/__tests__/string-literal.test.tsx
```diff
@@ -148,17 +148,63 @@ describe('class names string literal', () => {
...
+      '{padding-bottom:15px}',
       '{color:red}',
       '{background-color:blue}',
       '{border-color:green}',
       '{border-radius:10px}',
+      '{font-size:20px}',
+    ]);
+  });
+
+  it('should be able to override properties in a mixin', () => {
```
pancaspe87, line 173: This test case is for #799, the rest of tests refer to #798

packages/babel-plugin/src/css-prop/__tests__/string-literal.test.tsx
```diff
@@ -207,17 +207,61 @@ describe('css prop string literal', () => {
     expect(actual).toInclude('{font-size:30px}');
   });
 
+  it('should be able to override properties in a mixin', () => {
```
pancaspe87, line 210: This test case is for #799, the rest of tests refer to #798

packages/babel-plugin/src/styled/__tests__/string-literal.test.tsx
```diff
@@ -57,6 +57,37 @@ describe('styled component string literal', () => {
     ]);
   });
 
+  it('should be able to override properties in a mixin', () => {
```
pancaspe87, line 60: This test case is for #799, the rest of tests refer to #798

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -286,6 +286,11 @@ const extractTemplateLiteral = (node: t.TemplateLiteral, meta: Metadata): CSSOut
       const result = buildCss(interpolation, updatedMeta);
       variables.push(...result.variables);
       css.push(...result.css);
+
```
pancaspe87, line 289: This is all we need for both issues.
At first I was having a different logic (https://github.com/atlassian-labs/compiled/pull/805/commits/65633a0f0c25802299a1a3821549b90cbc774454), then realised that could be easily simplified in here :)

MonicaOlejniczak: Yep, this is much better :)

MonicaOlejniczak (dismissed):

packages/babel-plugin/src/utils/css-builders.tsx
MonicaOlejniczak: Yep, this is much better! :)

MonicaOlejniczak (dismissed):

MonicaOlejniczak (approved):

PR #808, opened 2021-08-11
Update package.json and tsconfig.json files

MonicaOlejniczak:
## Background
The `package.json` files and `tsconfig.json` files are hard to scan when they are not sorted alphabetically. This set of changes sorts both sets of files, removes a redundant tsconfig option (`noImplicitAny`), and adds in a `prettier` script so that prettier can be run directly outside of commits and IDE setups.

## Changes
* Sort package.json with sort-package-json
* Sort tsconfig.json alphabetically
* Add prettier script
* Remove noImplicitAny in favour of strict

### DIFF
--- a/packages/babel-plugin/package.json
+++ b/packages/babel-plugin/package.json
@@ -2,8 +2,6 @@
   "name": "@compiled/babel-plugin",
   "version": "0.6.14",
   "description": "A familiar and performant compile time CSS-in-JS library for React.",
-  "author": "Michael Dougall",
-  "license": "Apache-2.0",
   "homepage": "https://compiledcssinjs.com",
   "bugs": "https://github.com/atlassian-labs/compiled/issues/new?assignees=&labels=bug&template=bug_report.md",
   "repository": {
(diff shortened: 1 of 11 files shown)


packages/babel-plugin/package.json
```diff
@@ -2,15 +2,15 @@
...
-  "author": "Michael Dougall",
-  "license": "Apache-2.0",
   "homepage": "https://compiledcssinjs.com",
   "bugs": "https://github.com/atlassian-labs/compiled/issues/new?assignees=&labels=bug&template=bug_report.md",
   "repository": {
     "type": "git",
     "url": "https://github.com/atlassian-labs/compiled.git",
     "directory": "packages/babel-plugin"
   },
+  "license": "Apache-2.0",
```
pancaspe87, line 12: Hey @MonicaOlejniczak , out of curiosity, what's the reason to place these fields below?

pancaspe87 (approved):

packages/babel-plugin/package.json
MonicaOlejniczak: I used [sort-package-json](https://github.com/keithamus/sort-package-json) which defines the sort order, and it is loosely based off the order specified in https://docs.npmjs.com/cli/v7/configuring-npm/package-json

Issue #731, opened 2021-04-14
getting an error when import { css } from '@compiled/react


cameronbraid:
```
SyntaxError: /app/src/pages/Compiled.tsx: CallExpression isn't a supported CSS type - try using an object or string. (This is an error on an internal node. Probably an internal
error. Location has been estimated.)
  4 | import { css } from '@compiled/react';
  5 |
> 6 | const buttonCss = css`
    |                   ^^^
  7 |   border: 1px solid blue;
  8 |   border-radius: 3px;
  9 |   font-size: 10px;
    at File.buildCodeFrameError (/app/node_modules/@babel/core/lib/transformation/file/file.js:240:12)
    at NodePath.buildCodeFrameError (/app/node_modules/@babel/traverse/lib/path/index.js:138:21)
    at Object.buildCodeFrameError (/app/node_modules/@compiled/babel-plugin/src/utils/ast.tsx:77:42)
    at Object.buildCss (/app/node_modules/@compiled/babel-plugin/src/utils/css-builders.tsx:434:9)
    at Object.buildCss (/app/node_modules/@compiled/babel-plugin/src/utils/css-builders.tsx:360:20)
    at Object.visitCssPropPath (/app/node_modules/@compiled/babel-plugin/src/css-prop/index.tsx:42:21)
    at PluginPass.JSXOpeningElement (/app/node_modules/@compiled/babel-plugin/src/babel-plugin.tsx:216:9)
    at newFn (/app/node_modules/@babel/core/node_modules/@babel/traverse/lib/visitors.js:175:21)
    at NodePath._call (/app/node_modules/@babel/core/node_modules/@babel/traverse/lib/path/context.js:55:20)
    at NodePath.call (/app/node_modules/@babel/core/node_modules/@babel/traverse/lib/path/context.js:42:17)
    at NodePath.visit (/app/node_modules/@babel/core/node_modules/@babel/traverse/lib/path/context.js:92:31)
    at TraversalContext.visitQueue (/app/node_modules/@babel/core/node_modules/@babel/traverse/lib/context.js:116:16)
    at TraversalContext.visitSingle (/app/node_modules/@babel/core/node_modules/@babel/traverse/lib/context.js:85:19)
    at TraversalContext.visit (/app/node_modules/@babel/core/node_modules/@babel/traverse/lib/context.js:144:19)
    at Function.traverse.node (/app/node_modules/@babel/core/node_modules/@babel/traverse/lib/index.js:82:17)
[...]

PR #809, opened 2021-08-11
Override the default storybook babel configuration

MonicaOlejniczak:
## Background
Currently, storybook does not allow a `css` style to be defined as a variable:

```tsx

export default {
  title: 'test',
};

const foo = css`
  color: blue;
`;

export const Foo = (): JSX.Element => (
  <div css={foo}>hello world</div>
);
```

```sh
SyntaxError: /compiled/examples/stories/foo.tsx: CallExpression isn't a supported CSS type - try using an object or string. (This is an error on an internal node. Probably an internal error. Location has been estimated.)
   5 | };
   6 |
>  7 | const foo = css`
     |             ^^^
   8 |   color: blue;
   9 | `;
  10 |
```

This seems to occur because the css tagged template expression is transpiled into a call expression, and this syntax is not supported. Instead, the babel plugin should run on the source code.

## Changes
* Add a `.babelrc.json` file to `.storybook`, which runs `@compiled/babel-plugin` first
* Add a new story for a tagged template defined by a variable

### DIFF
--- a/.storybook/.babelrc.json
+++ b/.storybook/.babelrc.json
@@ -1,0 +1,4 @@
+{
+  "plugins": [["@compiled/babel-plugin", { "importReact": false }]],
+  "presets": ["@babel/preset-typescript", ["@babel/preset-react", { "runtime": "automatic" }]]
+}
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,7 +1,7 @@
 module.exports = {
   addons: ['storybook-addon-performance/register', 'storybook-addon-pseudo-states'],
-  stories: ['../examples/stories/*.tsx'],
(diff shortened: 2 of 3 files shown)


pancaspe87 (approved):
Hey @MonicaOlejniczak, I think this should also fix https://github.com/atlassian-labs/compiled/issues/731

If you look at the [last comment](https://github.com/atlassian-labs/compiled/issues/731#issuecomment-819516262), it sounds exactly like the issue that we encountered

MonicaOlejniczak:
@pancaspe87 The issue you linked was something that happened in a development app. While it may have the same underlying cause, the changes here will only fix the storybook setup.

PR #810, opened 2021-08-12
Fix identifiers shadowed by a function argument

MonicaOlejniczak:
## Changes
* Patch `@compiled/babel-plugin`
* Update the logic in `evaluateExpression`, so that the `ownPath` is set on a new meta object. This ensures that the correct expression is generated in the style prop, when an identifier is shadowed by a function argument.
* Create generic `transform` test utility, that will later be used for the `keyframes` API
* Add test case to `css-builders`
* Add `prettier` as a dev dependency, for the test utility

### DIFF
--- a/packages/babel-plugin/src/__tests__/css-builder.test.tsx
+++ b/packages/babel-plugin/src/__tests__/css-builder.test.tsx
@@ -1,13 +1,4 @@
-import { transformSync } from '@babel/core';
-import babelPlugin from '../index';
-
-const transform = (code: string) => {
-  return transformSync(code, {
-    configFile: false,
-    babelrc: false,
-    plugins: [babelPlugin],
-  })?.code;
-};
+import { transform } from './test-utils';
 
 describe('css builder', () => {
   it('should keep nested unconditional css together', () => {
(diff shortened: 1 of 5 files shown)


packages/babel-plugin/src/__tests__/css-builder.test.tsx
```diff
@@ -19,4 +10,43 @@ describe('css builder', () => {
...
+      const color = baseColor;
+
+      <div css={{
+        backgroundColor: getBackgroundColor(customBackgroundColor),
+        color
+      }} />
+    `);
+
+    // Make sure color is used over customBackgroundColor
+    expect(actual).toIncludeMultiple(['{color:var(--_1ylxx6h)}', '"--_1ylxx6h": ix(color)']);
```
pancaspe87, line 28: Hey @MonicaOlejniczak , just for my understanding, without the fix in `evaluateExpression`, we would have the same colour for both `backgroundColor` and `color`? 
For example, if `customBackgroundColor` is yellow, we would have yellow applied on both properties?

pancaspe87 (approved):

pancaspe87 (approved):

Issue #14, opened 2019-12-18
RFC: Keyframes api 


madou:
<details>
  <summary>Previous thoughts</summary>

## Static

Transforms:

```jsx
const fadeIn = keyframes({
  from: { opacity: 0 },
  to: { opacity: 1 },
});
```

To:

```jsx
export const fadeIn = {
  name: 'keyframes-a',
  keyframes: `
    from: { opacity: 0; }
    to: { opacity: 1; }
  `,
  variables: {},
};
```

And then consuming:

Transforms:

```jsx
import { fadeIn } from './fade-in';

<div css={{ animation: fadeIn }}>
  hello world
</div>
```

To:

```jsx
import { fadeIn } from './fade-in';

<>
  <style>
    {`
      @keyframes ${fadeIn.name} {
        ${fadeIn.keyframes}
      }

      .a { animation: ${fadeIn.name} }
    `}
  </style>
  <div className="a">
    hello world
  </div>
</>
```

## Dynamic

Transforms:

```jsx
export const fadeIn = (props) => keyframes({
  from: { opacity: 0 },
  '50%': { color: props.color },
  to: { opacity: 1 },
});
```

To:

```jsx
export const fadeIn = (props) => ({
  name: 'keyframes-a',
  keyframes: `
    from: { opacity: 0; }
    50%: { color: var(--color-a); }
    to: { opacity: 1; }
  `,
  variables: { '--color-a': props.color },
});
```

And then consuming it, transforms:

```jsx
import { fadeIn } from './fade-in';

const Component = props =>
<div css={{ animation: fadeIn({ color: props.color }) }}>
  hello world
</div>
```

To:

```jsx
import { fadeIn } from './fade-in';

const Component = props => {
  const keyFramesA = fadeIn({ color: props.color });

  return <>
    <style>
      {`
        @keyframes ${keyFramesA.name} {
          ${keyFramesA.keyframes}
        }

        .a { animation: ${keyFramesA.name} }
      `}
    </style>
    <div style={keyFramesA.variables} className="a">
      hello world
    </div>
  </>
};
```
</details>

First pass will be hashing animation keyframes name and then only being able to use them locally.

```js
<div
  css={`
    @keyframes fade-in {
      from { opacity: 0 }
      to { opacity: 1 }
    }

    animation: fade-in;
  `}
/>
```

To:

```css
@keyframes cc-abc123 {
  from { opacity: 0 }
  to { opacity: 1 }
}

.cc-abc123 {
[...]

PR #812, opened 2021-08-13
Add support for keyframes

MonicaOlejniczak:
These changes add `keyframes` to `@compiled/react`, which closes https://github.com/atlassian-labs/compiled/issues/14 and allows keyframes to be created with either a call expression or tagged template expression:

```jsx
import { css, keyframes } from '@compiled/react';

// Object call expression
const fadeOut = keyframes({
  from: {
    opacity: 1,
  },
  to: {
    opacity: 0,
  },
});

<div css={{ animation: `${fadeOut} 2s ease-in-out` }} />
```

```jsx
import { css, keyframes } from '@compiled/react';

// Tagged template expression
const fadeOut = keyframes`
  from {
    opacity: 1;
  }
  to {
    opacity: 0;
  }
`;

<div css={{ animation: `${fadeOut} 2s ease-in-out` }} />
```

## Changes
* Add `keyframes` runtime API to `@compiled/react` and associated tests
* Handle `keyframes` usages in `@compiled/babel-plugin` by:
  * Detecting `keyframes` imports
  * Cleaning up `keyframes` usages
  * Simplifying `evaluateExpression` for the `css` tagged template expression (i.e. check `t.isTaggedTemplateExpression(value)` over `value && isCompiledCSSTemplateLiteral(value, updatedMeta)`), and handling the keyframes call expression via `value && isCompiledKeyframesCallExpression(value, updatedMeta))`
  * Handling keyframe usages in the css builder by using the `isCompiledKeyframesTaggedTemplateExpression` and `isCompiledKeyframesCallExpression` helper utilities from `ast.tsx`
  * Fixing a bug that grabs an incorrect imported node in `ast.tsx`
  * Introducing the concept of a `sheet` on `CSSItem`, so that its styles are not within a CSS declaration
  * Introducing the concept of a `context` on `Metadata`, so that common runtime variables can be generated for a keyframe
* Add inline keyframes tests and examples for css prop and styled
* Add keyframes test in `ClassNames`
* Add keyframes examples and tests
* Add option to disable the autoprefixer by setting `process.env.AUTOPREFIXER` to `off` in `@compiled/css`, which is primarily used for tests but may prove useful for others
* Make `meta` JSDoc comments consistent and correctly named


packages/babel-plugin/src/keyframes/__fixtures__/index.tsx
```diff
@@ -0,0 +1,47 @@
+export const longhandCssPropObjectCallExpression = `
```
pancaspe87, line 1: Do we also need to provide examples for `ClassNames`?

packages/react/src/keyframes/__tests__/index.test.tsx
```diff
@@ -0,0 +1,324 @@
...
+        );
+      });
+
+      it('given a template literal call expression argument', () => {
+        const fadeOut = keyframes(`from { opacity: 1; } to { opacity: 0; }`);
+        const { getByText } = render(<div css={{ animationName: fadeOut }}>hello world</div>);
+
+        expect(getByText('hello world')).toHaveCompiledCss('animation-name', 'klmf72q');
+        expect(getKeyframe('klmf72q')).toMatchInlineSnapshot(
+          `"@keyframes klmf72q{0%{opacity:1}to{opacity:0}}"`
```
pancaspe87, line 44: Hey @MonicaOlejniczak, most likely I am overthinking, but I would expect same keyframes to generate the same output. All these tests have the same properties `{0%{opacity:1}to{opacity:0}}`, but generate different names.

Maybe we could check how other libraries deal with keyframes

packages/babel-plugin/src/keyframes/__tests__/call-expression.test.tsx
```diff
@@ -0,0 +1,1157 @@
...
+        it('identifier that references a number literal', () => {
+          const actual = transform(`
+            import { keyframes } from '@compiled/react';
+
+            const fromFontSize = 14;
+            const toFontSize = 18;
+
+            const enlargeFont = keyframes({
+              from: {
+                fontSize: \`\${fromFontSize}px\`,
```
pancaspe87, line 589: I haven't tried yet, but it seems that `addUnitIfNeeded` should deal with missing units. Maybe we could omit `px ` from one of these declarations and see what happens :)

packages/babel-plugin/src/keyframes/__tests__/call-expression.test.tsx
```diff
@@ -0,0 +1,1157 @@
...
+
+    describe('referenced through a css prop object call expression', () => {
+      describe('by inlining static values', () => {
+        it('used with the longhand syntax', () => {
+          const actual = transform(createSmokeTest(longhandCssPropObjectCallExpression));
+
+          expect(actual).toMatchInlineSnapshot(`
+            "const _4 = \\"._1pgl1ytf{animation-timing-function:ease-in-out}\\";
+            const _3 = \\"._j7hq18s7{animation-name:k6qtb99}\\";
+            const _2 = \\"@keyframes k6qtb99{25%,0%{opacity:1}50%{opacity:0.5}to{opacity:0}}\\";
```
pancaspe87, line 49: nit not sure if we have control on it, but should we use `100%` instead of `to`?

packages/babel-plugin/src/keyframes/__tests__/tagged-template.test.tsx
```diff
@@ -0,0 +1,840 @@
+import { transform } from '../../__tests__/test-utils';
```
pancaspe87, line 1: I haven't reviewed all the test cases yet, so forgive me if you already addressed these cases...just wondering what happens if
- we have a keyframe defined multiple times ( eg https://developer.mozilla.org/en-US/docs/Web/CSS/@keyframes#when_a_keyframe_is_defined_multiple_times)
- we have multiple animations ( eg https://css-tricks.com/almanac/properties/a/animation/#multiple-animations)

packages/babel-plugin/src/keyframes/__tests__/call-expression.test.tsx
MonicaOlejniczak: postcss will convert any input I provide to the shortest form. So in this case, `from` becomes `0%` and `to` stays as `to`

packages/react/src/keyframes/__tests__/index.test.tsx
MonicaOlejniczak: I did initially as well, but it's because I am generating the name from the raw input before processing the css. So a keyframe of `keyframes({})` with the same values, will be different to `keyframes('')`. This should be fine in practice, as applications should use a consistent syntax.

Emotion and styled components both generate the name by hashing the resulting CSS, which could be done as an optimisation later but there is a bit of trickiness to it.

packages/babel-plugin/src/keyframes/__tests__/call-expression.test.tsx
MonicaOlejniczak: Yes, it does deal with this case. Omitting it just use the value `fromFontSize`. I won't add tests for these under keyframes, since the `addUnitIfNeeded` logic should be tested more thoroughly elsewhere.

packages/babel-plugin/src/keyframes/__fixtures__/index.tsx
MonicaOlejniczak: Added some tests

packages/babel-plugin/src/keyframes/__tests__/tagged-template.test.tsx
MonicaOlejniczak: Added tests for these cases

JakeLane (approved):

Issue #800, opened 2021-08-06
Storybook doesn't work without importing css even though mixins are imported


pancaspe87:
**Describe the bug**
When creating a story that imports CSS mixins from an external file, story is not created unless we also import `css`

**To Reproduce**

1. Run project locally
2. Go to 'examples/stories/styled-string.tsx'
3. Remove css`
    text-transform: uppercase;
  ` from the import
4. Remove css from ArrayUse
5. Story doesn't work anymore

**Expected behavior**
Story should be created

**Screenshots**
![image](https://user-images.githubusercontent.com/14813563/128451074-1baf11bf-f560-49af-9784-bbe8696286ed.png)

![image](https://user-images.githubusercontent.com/14813563/128451124-1439be17-48e4-4b69-90bf-90d535b0371d.png)

PR #823, opened 2021-09-10
Fix missing key prop on generated React elements

### LINKED ISSUE #802: Compiled is not adding the key attribute
**Describe the bug**
Consider this component:
```
import '@compiled/react'

const MyComponent = () => {
  return ['foo', 'bar'].map((string) => {
    return (
      <div key={string} css={{ backgroundColor: 'blue' }}>
        {string}
      </div>
    )
  })
}
```

Rendering this component results in the following warning: `Warning: Each child in a list should have a unique "key" prop.`

It seems that the key prop is not added to the element that the library is adding.

**To Reproduce**
See example above

Note: I'm using the following versions:
```
"@compiled/react": "^0.6.13",
"@compiled/babel-plugin": "^0.6.14"
```

JakeLane:
Fixes #802 

Simple change to add a key prop to generated React elements. This will suppress the React warning and remove the performance bug where React will always trigger re-rendering for children components ([per react docs](https://reactjs.org/docs/lists-and-keys.html)).

### DIFF
--- a/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
+++ b/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
@@ -615,4 +615,19 @@ describe('css prop behaviour', () => {
 
     expect(actual).toInclude('ax([isPrimary&&"_syaz13q2 _1wybgktf"])');
   });
+
+  it('should retain keys for mapped react components', () => {
+    const actual = transform(`
+      import '@compiled/react';
+      import React from 'react';
+
+      ['foo', 'bar'].map((str) => (
+        <div key={str} css={{ backgroundColor: 'blue' }}>
+          {str}
+        </div>
+      ));
+    `);
+
+    expect(actual).toInclude('<CC key={str}>');
+  });
 });
--- a/packages/babel-plugin/src/css-prop/__tests__/object-literal.test.tsx
+++ b/packages/babel-plugin/src/css-prop/__tests__/object-literal.test.tsx
@@ -729,4 +729,20 @@ describe('css prop object literal', () => {
 
     expect(actual).toInclude('{color:red}');
   });
+
+  it('should retain keys for mapped react components', () => {
+    const actual = transform(`
+        import '@compiled/react';
+
+        ['foo', 'bar'].map((str) => (
+          <div key={str} css={{ backgroundColor: 'blue' }}>
+            {str}
+          </div>
+        ));
+      `);
+
+    console.log(actual);
+    expect(actual).toMatch(/<CC key={.+}>/);
+    expect(actual).toMatch(/<CS key={.+}>/);
+  });
 });
--- a/packages/babel-plugin/src/css-prop/__tests__/object-literal.test.tsx
+++ b/packages/babel-plugin/src/css-prop/__tests__/object-literal.test.tsx
@@ -729,18 +729,4 @@ describe('css prop object literal', () => {
 
     expect(actual).toInclude('{color:red}');
   });
-
-  it('should retain keys for mapped react components', () => {
-    const actual = transform(`
-        import '@compiled/react';
-
-        ['foo', 'bar'].map((str) => (
-          <div key={str} css={{ backgroundColor: 'blue' }}>
-            {str}
-          </div>
-        ));
-      `);
-
-    expect(actual).toInclude('<CC key={str}>');
-  });
 });
--- a/packages/babel-plugin/src/css-prop/__tests__/object-literal.test.tsx
+++ b/packages/babel-plugin/src/css-prop/__tests__/object-literal.test.tsx
@@ -741,8 +741,6 @@ describe('css prop object literal', () => {
         ));
       `);
 
-    console.log(actual);
-    expect(actual).toMatch(/<CC key={.+}>/);
-    expect(actual).toMatch(/<CS key={.+}>/);
+    expect(actual).toInclude('<CC key={str}>');
   });
 });
--- a/packages/babel-plugin/src/utils/ast-builders.tsx
+++ b/packages/babel-plugin/src/utils/ast-builders.tsx
@@ -369,6 +369,34 @@ const styledTemplate = (opts: StyledTemplateOpts, meta: Metadata): t.Node => {
   }) as t.Node;
 };
 
+const getKeyAttributeExpression = (node: t.Expression): t.Expression | null => {
+  if (!t.isJSXElement(node)) {
+    return null;
+  }
+
+  const keyAttribute = node.openingElement.attributes.find(
+    (attribute) => t.isJSXAttribute(attribute) && attribute.name.name === 'key'
+  ) as t.JSXAttribute | undefined;
+
+  if (
+    !keyAttribute ||
+    !t.isJSXExpressionContainer(keyAttribute.value) ||
+    !t.isExpression(keyAttribute.value.expression)
+  ) {
+    return null;
+  }
+
+  return keyAttribute.value.expression;
+};
+
+const buildKeyAttribute = (expression: t.Expression, label: string) =>
+  generate(
+    t.jsxAttribute(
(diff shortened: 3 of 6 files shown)


packages/babel-plugin/src/css-prop/__tests__/object-literal.test.tsx
```diff
@@ -729,4 +729,20 @@ describe('css prop object literal', () => {
...
+        import '@compiled/react';
+
+        ['foo', 'bar'].map((str) => (
+          <div key={str} css={{ backgroundColor: 'blue' }}>
+            {str}
+          </div>
+        ));
+      `);
+
+    console.log(actual);
```
MonicaOlejniczak, line 744: You shouldn't need to log the actual here, since jest will output it for you when there is a mismatch. I presume this is lingering code though

packages/babel-plugin/src/css-prop/__tests__/object-literal.test.tsx
```diff
@@ -729,4 +729,20 @@ describe('css prop object literal', () => {
...
+    expect(actual).toMatch(/<CC key={.+}>/);
```
MonicaOlejniczak, line 745: You could be more explicit and replace this with:

```javascript
  '<CC key="foo">',
  '<CS key="foo">',
  '<CC key="bar">',
  '<CS key="foo">',
]);
```

Or if you want to be a bit more programmatic:

```javascript
expect(actual).toIncludeMultiple(['foo', 'bar'].flatMap((name) => [
  `<CC key="${name}">`,
  `<CS key="${name}">`,
]));
```

Though if you remove the key on `CS` (refer to another comment), then it can be simplified to just `map`.

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -380,10 +408,14 @@ const styledTemplate = (opts: StyledTemplateOpts, meta: Metadata): t.Node => {
...
   const nonceAttribute = meta.state.opts.nonce ? `nonce={${meta.state.opts.nonce}}` : '';
 
+  const keyAttributeExpression = getKeyAttributeExpression(node);
+
   return template(
     `
-  <CC>
-    <CS ${nonceAttribute}>{%%cssNode%%}</CS>
+  <CC ${keyAttributeExpression ? buildKeyAttribute(keyAttributeExpression, '--cc') : ''}>
+    <CS ${
```
MonicaOlejniczak, line 416: The key should only be needed on the first node in the list (i.e. `CC`) — you should be able to remove this

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -369,6 +369,34 @@ const styledTemplate = (opts: StyledTemplateOpts, meta: Metadata): t.Node => {
...
+  const keyAttribute = node.openingElement.attributes.find(
+    (attribute) => t.isJSXAttribute(attribute) && attribute.name.name === 'key'
+  ) as t.JSXAttribute | undefined;
+
+  if (
+    !keyAttribute ||
+    !t.isJSXExpressionContainer(keyAttribute.value) ||
+    !t.isExpression(keyAttribute.value.expression)
+  ) {
+    return null;
```
MonicaOlejniczak, line 386: `return;` and `return undefined;` is used more frequently in the codebase. Either is generally more preferential to `null`, see: https://basarat.gitbook.io/typescript/recap/null-undefined#final-thoughts

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -369,6 +369,34 @@ const styledTemplate = (opts: StyledTemplateOpts, meta: Metadata): t.Node => {
   }) as t.Node;
 };
 
+const getKeyAttributeExpression = (node: t.Expression): t.Expression | null => {
+  if (!t.isJSXElement(node)) {
+    return null;
+  }
+
+  const keyAttribute = node.openingElement.attributes.find(
```
MonicaOlejniczak, line 377: You should prefer a type guard over casting here:

```typescript
const isKeyAttribute = (attribute: t.JSXAttribute | t.JSXSpreadAttribute): attribute is t.JSXAttribute =>
    t.isJSXAttribute(attribute) && attribute.name.name === 'key';

const keyAttribute = node.openingElement.attributes.find(isKeyAttribute);
```

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -380,10 +408,14 @@ const styledTemplate = (opts: StyledTemplateOpts, meta: Metadata): t.Node => {
 export const compiledTemplate = (node: t.Expression, sheets: string[], meta: Metadata): t.Node => {
   const nonceAttribute = meta.state.opts.nonce ? `nonce={${meta.state.opts.nonce}}` : '';
 
+  const keyAttributeExpression = getKeyAttributeExpression(node);
+
   return template(
     `
-  <CC>
-    <CS ${nonceAttribute}>{%%cssNode%%}</CS>
+  <CC ${keyAttributeExpression ? buildKeyAttribute(keyAttributeExpression, '--cc') : ''}>
```
MonicaOlejniczak, line 415: You should be able to use the original `key` expression here, without appending `--cc`

packages/babel-plugin/src/css-prop/__tests__/object-literal.test.tsx
JakeLane: yeah forgot we don't have an eslint rule to disallow this :s

packages/babel-plugin/src/utils/ast-builders.tsx
JakeLane: I assumed the same but React was complaining about duplicate keys. I suspect storybook may about been mounting the story twice in the same tree for some reason - I'll double check this

packages/babel-plugin/src/utils/ast-builders.tsx
JakeLane: Same concern as https://github.com/atlassian-labs/compiled/pull/823#discussion_r707842373

packages/babel-plugin/src/utils/ast-builders.tsx
JakeLane: Yep, that is the case - I could only remove these if I did a `yarn clean`. This could be related to https://github.com/atlassian-labs/compiled/issues/800 . I'll look into this as this is breaking the dev loop.

MonicaOlejniczak: Let me take a look at this as well, since I thought the new babel setup fixed it and I forgot to close out that issue

packages/babel-plugin/src/css-prop/__tests__/object-literal.test.tsx
```diff
@@ -729,4 +729,18 @@ describe('css prop object literal', () => {
 
     expect(actual).toInclude('{color:red}');
   });
+
+  it('should retain keys for mapped react components', () => {
```
MonicaOlejniczak, line 733: You should add one test for styled components like you have for the css prop here

JakeLane: There's no key added to the generated react components because it's not needed

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
```diff
@@ -615,4 +615,19 @@ describe('css prop behaviour', () => {
 
     expect(actual).toInclude('ax([isPrimary&&"_syaz13q2 _1wybgktf"])');
   });
+
+  it('should retain keys for mapped react components', () => {
+    const actual = transform(`
+        import '@compiled/react';
```
MonicaOlejniczak, line 621: /nit should be de-indented by 1

MonicaOlejniczak (dismissed):

MonicaOlejniczak (approved):

Issue #301, opened 2020-09-23
Update to PostCSS 8


madou:
See: https://github.com/postcss/postcss/releases?after=8.0.2

PR #837, opened 2021-09-28
Update to postcss v8

andparsons:
I've done some initial work updating the depandancies, plugins and ensuring the tests now pass.

Aims to address #301


packages/css/src/__tests__/transform.test.tsx
```diff
@@ -25,7 +25,7 @@ describe('#css-transform', () => {
       const { sheets: actual } = transformCss(`
         margin-left: initial;
         content: 'hello';
-        color: hotpink;
+        color: rebeccapurple;
```
andparsons, line 28: Updated colormin didn't convert hotpink to #ff69b4 as they are the same length, a longer color name makes for a better test

packages/css/src/plugins/expand-shorthands/index.tsx
```diff
@@ -58,31 +58,36 @@ const shorthands: Record<string, ConversionFunction> = {
...
+export const expandShorthands = (): Plugin => {
   const filter = new RegExp(Object.keys(shorthands).join('|'));
 
-  return (root) => {
-    root.walkDecls(filter, (decl) => {
-      const valueNode = parse(decl.value);
-      const expand = shorthands[decl.prop];
+  return {
+    postcssPlugin: 'expand-shorthands',
+    Once(root) {
```
MonicaOlejniczak, line 66: This can be replaced with:

```tsx
Declaration(declaration) {
  if (!shorthands[declaration.prop]) {
    return;
  }

  // remaining logic here...
}
```

or this, which should be the fastest:

```tsx
const onDeclaration = (declaration: Declaration) => {
  // remaining logic here...
};

// ...
Declaration: Object.keys(shorthands).reduce((keys, shorthand) => {
  keys[shorthand] = onDeclaration;
  return keys;
}, {})
```

Refer to https://github.com/postcss/postcss/blob/main/docs/guidelines/plugin.md#23-use-fast-nodes-scanning

packages/react/src/__tests__/ssr.test.tsx
```diff
@@ -134,7 +134,7 @@ describe('SSR', () => {
 
     expect(result.split('</style>').join('</style>\n')).toMatchInlineSnapshot(`
       "<style data-cmpld=\\"true\\" nonce=\\"k0Mp1lEd\\">._1e0c1txw{display:flex}._1wyb12am{font-size:50px}._syaz1cnh{color:purple}._ysv75scu:link{color:red}._105332ev:visited{color:pink}._f8pjbf54:focus{color:green}._30l31gy6:hover{color:yellow}._9h8h13q2:active{color:blue}@supports (display:grid){._1df61gy6:focus{color:yellow}._7okp11x8:active{color:black}}@media (max-width:800px){._1o8z1gy6:focus{color:yellow}._1cld11x8:active{color:black}}</style>
-      <a href=\\"https://atlassian.design\\" class=\\"_1e0c1txw _1wyb12am _syaz1cnh _30l31gy6 _9h8h13q2 _ysv75scu _7okp11x8 _1df61gy6 _f8pjbf54 _105332ev _1cld11x8 _1o8z1gy6\\">Atlassian Design System</a>"
+      <a href=\\"https://atlassian.design\\" class=\\"_1e0c1txw _1wyb12am _syaz1cnh _30l31gy6 _9h8h13q2 _ysv75scu _1df61gy6 _7okp11x8 _f8pjbf54 _105332ev _1o8z1gy6 _1cld11x8\\">Atlassian Design System</a>"
```
MonicaOlejniczak, line 137: Why did the class order change here?

packages/css/src/plugins/merge-duplicate-at-rules.tsx
```diff
@@ -18,33 +17,38 @@ import type { ChildNode, AtRule } from 'postcss';
...
-          node: atRule,
-          children: {},
-        };
-      }
-
-      atRule.each((node) => {
-        const stringifiedNode = node.toString();
-        if (!atRuleStore[name].children[stringifiedNode]) {
-          atRuleStore[name].children[stringifiedNode] = node;
+      root.walkAtRules((atRule) => {
```
MonicaOlejniczak, line 26: `walkAtRules` can be replaced with a top level `AtRule` instead of `Once`, and I believe `root` can instead have `atRule.root()`

packages/css/src/plugins/parent-orphaned-pseudos.tsx
```diff
@@ -16,24 +16,29 @@ const prependNestingTypeToSelector = (selector: selectorParser.Node) => {
...
  */
-export const parentOrphanedPseudos = plugin('parent-orphened-pseudos', () => {
-  return (root) => {
-    root.walkRules((rule) => {
-      const { selectors } = rule;
+export const parentOrphanedPseudos = (): Plugin => {
+  return {
+    postcssPlugin: 'parent-orphened-pseudos',
+    Once(root) {
+      root.walkRules((rule) => {
```
MonicaOlejniczak, line 23: `walkRules` can be replaced with a top-level `Rule` instead of `Once`

MonicaOlejniczak (changes requested):
Thanks for the contribution! Overall, the changes look good. I have mostly left feedback on how some of the more trivial plugins can be updated to use the new API :)

packages/react/src/__tests__/browser.test.tsx
```diff
@@ -37,7 +37,7 @@ describe('browser', () => {
     render(
       <>
         <StyledDiv>hello world</StyledDiv>
-        <StyledDiv>hello world</StyledDiv>
+        <StyledDiv>Hello World!</StyledDiv>
```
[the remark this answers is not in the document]
andparsons: I was having the tests in this file continue to fail unless I made any sort of change to the file. Coming back to this now I can see that it would be a Jest cache issue.
Running `npx jest --clearCache` now allows the test to pass locally with the original, so I'll undo this silly change.

package.json
```diff
@@ -53,6 +53,8 @@
   "resolutions": {
     "@babel/core": "^7.15.5",
     "jest": "^26.6.3",
+    "postcss": "^8.0.0",
```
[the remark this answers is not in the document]
andparsons: The package `css-modules-loader-core` is abandoned and has a `postcss v6` dependancy, and `@parcel/transformer-postcss` requires it. Not doing this causes types to fail pior to the tests running. I'm not sure how else to resolve this as opposed to this resolution approach.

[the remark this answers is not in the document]
madou: Can we target the specific package in the resolution? Something like:

```
"@parcel/transformer-postcss/**/postcss": "^8.0.0"
```

Would need to make sure it gets the correct packages 😄 https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/#toc-how-to-use-it

packages/css/src/plugins/atomicify-rules.tsx
```diff
@@ -200,33 +221,38 @@ const atomicifyAtRule = (node: AtRule, opts: AtomicifyOpts): AtRule => {
...
+          default:
+            break;
+        }
+      });
+    },
   };
-});
+};
+
+module.exports.postcss = true;
```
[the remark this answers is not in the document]
andparsons: This is great advice, I'm seeing the tests run fine with this update

packages/parcel-transformer/types/json-source-map.d.ts
```diff
@@ -0,0 +1,14 @@
+declare module 'json-source-map' {
```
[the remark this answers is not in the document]
andparsons: I was getting type errors before the tests would run. Removing these now seems to still run fine, so I've updated the branch as such.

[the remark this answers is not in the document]
andparsons: Turns out these were needed after all during `yarn run build`
I'm not sure how they were okay not existing before, as parcel v2 has never had those types available by the look of it.
Added some comments of where the types are from.

packages/css/src/plugins/discard-duplicates.tsx
```diff
@@ -1,15 +1,14 @@
...
 
 /**
  * Discards top level duplicate declarations.
  */
-export const discardDuplicates = plugin<{ callback: (sheet: string) => void }>(
-  'discard-duplicates',
-  () => {
-    const decls: Record<string, Declaration[]> = {};
+export const discardDuplicates = (): Plugin => {
+  const decls: Record<string, Declaration[]> = {};
```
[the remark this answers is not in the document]
andparsons: I've now inlined as such

packages/css/src/plugins/normalize-current-color.tsx
```diff
@@ -1,15 +1,20 @@
...
-    root.walkDecls((rule) => {
-      const lowerValue = rule.value.toLowerCase();
-      if (lowerValue === 'currentcolor' || lowerValue === 'current-color') {
-        rule.value = 'currentColor';
-      }
-    });
+export const normalizeCurrentColor = (): Plugin => {
+  return {
+    postcssPlugin: 'normalize-current-color',
+    OnceExit(root) {
```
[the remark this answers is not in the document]
andparsons: I've now updated as such

packages/css/src/plugins/expand-shorthands/index.tsx
andparsons: I think the faster method lacks the typescript support, have gone with the first suggestion which is working fine.

packages/react/src/__tests__/ssr.test.tsx
andparsons: I'm not too sure as nothing around ordering classes has been changed.
I don't think it's a big issue though as it's just the class attribute, the other tests around the lvfffha style order are still valid.

packages/css/src/plugins/merge-duplicate-at-rules.tsx
andparsons: Due to the fact the atRule gets removed and later appended to root, I've split the work into an AtRule call and a OnceExit call that should run at the end. The tests stay valid with this approach.

packages/css/src/plugins/parent-orphaned-pseudos.tsx
andparsons: I'm not sure if it's due to having a child call to walkPseudos, but I am unable to get the tests to pass after a few attempts.

andparsons:
There may be some further optimisations to changing calls to `Once` to be more granular for those still using it, but I am unable to do this myself without finding the tests then show something breaks.

Hopefully the rest of the changes look good to you, thanks for the speedy feedback on this!

packages/parcel-transformer/types/json-source-map.d.ts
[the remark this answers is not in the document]
MonicaOlejniczak: I believe your install resolved the nightly `2.0.0-nightly.655` version in master to `2.0.0-rc.0`. Using the RC candidate is better, and between the versions the types were then needed.

packages/css/src/plugins/parent-orphaned-pseudos.tsx
MonicaOlejniczak: Okay, no problem! Thanks for trying :)

packages/css/src/plugins/expand-shorthands/index.tsx
MonicaOlejniczak: Awesome! Though I don't see the first change here, did you forget to push it up?

packages/css/src/plugins/atomicify-rules.tsx
```diff
@@ -156,9 +164,20 @@ const atomicifyRule = (node: Rule, opts: AtomicifyOpts): Rule[] => {
  * @param node
  * @param opts
  */
-const atomicifyAtRule = (node: AtRule, opts: AtomicifyOpts): AtRule => {
-  const children: Node[] = [];
-  const newNode = node.clone({ nodes: children });
+const atomicifyAtRule = (node: AtRule, opts: AtomicifyOpts, postcss: Postcss): AtRule => {
+  const children: ChildNode[] = [];
+  const newNode = postcss.atRule({
```
MonicaOlejniczak, line 169: How come the original `node.clone({ nodes: children })` isn't used here? It should be part of the `AtRule` API

packages/css/src/plugins/atomicify-rules.tsx
```diff
@@ -171,24 +190,26 @@ const atomicifyAtRule = (node: AtRule, opts: AtomicifyOpts): AtRule => {
...
-        children.push(atomicifyDecl(childNode, atRuleOpts));
+        children.push(atomicifyDecl(childNode, atRuleOpts, postcss));
         break;
 
       default:
         break;
     }
   });
 
+  newNode.nodes = children;
```
MonicaOlejniczak, line 211: Is this needed when it's initialised above?

packages/css/src/plugins/atomicify-rules.tsx
```diff
@@ -1,5 +1,4 @@
-import type { Node, Declaration, Container, Rule, AtRule } from 'postcss';
-import { plugin, decl, rule } from 'postcss';
```
MonicaOlejniczak, line 2: These should still be valid exports, meaning you don't have to pass around `postcss: Postcss` everywhere

packages/css/src/plugins/atomicify-rules.tsx
```diff
@@ -102,19 +101,24 @@ const buildAtomicSelector = (node: Declaration, opts: AtomicifyOpts) => {
...
+  const newRule = postcss.rule({
+    raws: { before: '', after: '', between: '', selector: { raw: '', value: '' } },
+    nodes: [newDecl],
+    selector,
+  });
 
   // We need to link the new node to a parent else autoprefixer blows up.
   newDecl.parent = newRule;
-  newDecl.raws.before = '';
   newRule.parent = opts.parentNode!;
```
MonicaOlejniczak, line 121: Just to be a bit cleaner, you should be able to place this inside `newRule` as a key. It seems to be overridden from https://github.com/postcss/postcss/blob/main/lib/node.js#L52 which `rule` extends from

packages/css/src/plugins/atomicify-rules.tsx
andparsons: https://evilmartians.com/chronicles/postcss-8-plugin-migration#step-4-remove-postcss-imports says it can be kept s is for a while, so I'll add them back in.

packages/css/src/plugins/atomicify-rules.tsx
andparsons: It wasn't working without reassignment, perhaps due to some change in how the clone behaviour deals with references.
I've updated this to push new node items directly to newNode.nodes, and this resolves the need for the reassignment

packages/css/src/plugins/atomicify-rules.tsx
andparsons: parent doesn't seem to exist on the RuleProps type, just the Rule type after creation

packages/css/src/plugins/atomicify-rules.tsx
andparsons: I was initially having issues after the update, coming back to this now the use of clone seems to work as it should, have now updated as such

andparsons:
@MonicaOlejniczak it looks like the parcel plugins have updated due to a change in the lockfile, which is breaking it's usage in parcel-transformer.
Should I fix the version of parcel to 2.0.0-nightly.657 and open another pull request to update it to 2.0.0-rc.0, or should I update the parcel-transformer in this pull-request?

packages/css/src/plugins/expand-shorthands/index.tsx
andparsons: I sure did, all pushed now

MonicaOlejniczak:
@andparsons I didn't realise I had to enable the build for contributors, but it's now linked up and you can see the failure in the pipeline. Given that the update breaks the transformer, and because this pull request is specific to postcss, I would revert the lockfile to use the nightly version. If you have free time, then it would be greatly appreciated if you could fix the transformer in another pull request! I'm not expecting it though, so don't feel pressured. We'll probably have to fix it sometime soon anyway, as it should break internal builds in an Atlassian product once they upgrade.

andparsons:
@MonicaOlejniczak I've temporarily fixed the parcel version, and have opened #842 to address the update to parcel@2.0.0-rc.0

madou:
Hi @andparsons looks like quite a bit of conflicts - but we'd love this contribution if you want to pick it up again!

madou:
Superseded by https://github.com/atlassian-labs/compiled/pull/946

PR #838, opened 2021-09-28
Make strip plugin work with esbuild/rollup bundled code

youzi:
Currently the strip plugin checks for exact matches of React/CC/CS to replace/remove. Tools like rollup and esbuild often rename imports by adding a number (eg `import React from 'react'` becomes `import React2 from 'react'`). This makes it that the plugin does not work with input from these tools.

Example input:
```js
import React4 from "react";
import { ax as ax2, ix as ix2, CC as CC2, CS as CS2 } from "@compiled/react/runtime";
const _2 = "._syaz13q2{color:blue";
const _ = "._1wyb1fwx{font-size:12px";

const Component = () => /*#__PURE__*/React4.createElement(CC2, null, /*#__PURE__*/React4.createElement(CS2, null, [_, _2]), /*#__PURE__*/React4.createElement("div", {
  className: ax2(["_1wyb1fwx _syaz13q2"])
}, "hello world"));
```

This PR makes the `isComponent` check less strict by checking if a node name starts with `"React"` vs checking if it equals `"React"`. It also now stores the imports that it removed and will remove any calls that match one of those imports.

This makes the plugin fully functional in the cases that I have tested (stripping runtime from code bundled with esbuild)

### DIFF
--- a/packages/babel-plugin-strip-runtime/src/__tests__/first-party-strip.test.tsx
+++ b/packages/babel-plugin-strip-runtime/src/__tests__/first-party-strip.test.tsx
@@ -115,7 +115,7 @@ describe('first party strip runtime', () => {
 
         const Component = () => <div css={{ fontSize: 12, color: 'blue' }}>hello world</div>
       `;
-      console.log(baked);
+
       const actual = transform({ runtime: 'classic', run: 'extract' })(baked);
 
       expect(actual).toMatchInlineSnapshot(`
--- a/packages/babel-plugin-strip-runtime/src/__tests__/first-party-strip.test.tsx
+++ b/packages/babel-plugin-strip-runtime/src/__tests__/first-party-strip.test.tsx
@@ -193,7 +193,7 @@ describe('first party strip runtime', () => {
         });
       `;
 
-      const actual = transform({ runtime: 'classic', run: 'extract' })(baked);
+      const actual = transform({ runtime: 'automatic', run: 'extract' })(baked);
 
       expect(actual).toMatchInlineSnapshot(`
         "import { ax as ax2, ix as ix2 } from \\"@compiled/react/runtime\\";
--- a/packages/babel-plugin-strip-runtime/src/__tests__/first-party-strip.test.tsx
+++ b/packages/babel-plugin-strip-runtime/src/__tests__/first-party-strip.test.tsx
@@ -152,7 +152,7 @@ describe('first party strip runtime', () => {
       `);
     });
 
-    it('should remove CSS prop runtime when js is bundled by other tools (eg. esbuild/rollup)', () => {
+    it('should remove CSS prop runtime when js is bundled by other tools (eg. esbuild/rollup) classic', () => {
       const baked = `
         import React4 from "react";
         import { ax as ax2, ix as ix2, CC as CC2, CS as CS2 } from "@compiled/react/runtime";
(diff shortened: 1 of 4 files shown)


packages/babel-plugin-strip-runtime/src/index.tsx
```diff
@@ -63,7 +65,8 @@ export default declare<PluginPass>((api) => {
           // We've found something that looks like React.createElement(...)
           // Now we want to check if it's from the Compiled Runtime and if it is - replace with its children.
           const component = path.node.arguments[0];
-          if (!isCCComponent(component)) {
+          // @ts-ignore
```
madou, line 68: What was the TypeScript error you're suppressing? Could we see if we can fix that instead? 😄

packages/babel-plugin-strip-runtime/src/__tests__/first-party-strip.test.tsx
```diff
@@ -152,6 +152,29 @@ describe('first party strip runtime', () => {
...
+        import React4 from "react";
+        import { ax as ax2, ix as ix2, CC as CC2, CS as CS2 } from "@compiled/react/runtime";
+        const _2 = "._syaz13q2{color:blue";
+        const _ = "._1wyb1fwx{font-size:12px";
+
+        const Component = () => /*#__PURE__*/React4.createElement(CC2, null, /*#__PURE__*/React4.createElement(CS2, null, [_, _2]), /*#__PURE__*/React4.createElement("div", {
+          className: ax2(["_1wyb1fwx _syaz13q2"])
+        }, "hello world"));`;
+
+      const actual = transform({ runtime: 'classic', run: 'extract' })(baked);
```
madou, line 166: Let's test the automatic runtime as well.

packages/babel-plugin-strip-runtime/src/utils/ast.tsx
```diff
@@ -13,7 +13,7 @@ export const isCreateElement = (node: t.Node): node is t.CallExpression => {
   return (
     t.isMemberExpression(node) &&
     t.isIdentifier(node.object) &&
-    node.object.name === 'React' &&
+    node.object.name.startsWith('React') &&
```
madou, line 16: Can we leave a comment here talking through why we do "starts with" (because of different bundler outputs).

madou:
Thanks for the contribution! I've left some comments.

packages/babel-plugin-strip-runtime/src/index.tsx
youzi: Fixed

packages/babel-plugin-strip-runtime/src/__tests__/first-party-strip.test.tsx
youzi: Done

packages/babel-plugin-strip-runtime/src/utils/ast.tsx
youzi: Done

youzi:
I've updated the PR, now covering automatic runtime as well.

There is one more exact check that I believe needs work, but I'm not familiar enough with the project to know if it should be adjusted. https://github.com/atlassian-labs/compiled/blob/9413477040feb1da8a5299738cc0945f69b803f2/packages/babel-plugin-strip-runtime/src/utils/ast.tsx#L176

Perhaps someone can pick that up.

JakeLane (approved):

JakeLane:
Thanks for the contribution @youzi! Looks good to me but I'll wait for @madou to have a look.

Can you also please add a changeset? There are instructions [here](https://github.com/atlassian/changesets/blob/main/docs/adding-a-changeset.md) and that will let us cut a release with your changes.

packages/babel-plugin-strip-runtime/src/utils/ast.tsx
```diff
@@ -74,13 +78,17 @@ export const getJsxRuntimeChildren = (node: t.CallExpression): Array<t.Expressio
  * @param node
  * @returns
  */
-export const isCCComponent = (node: t.Node): boolean => {
-  if (t.isIdentifier(node) && node.name === 'CC') {
-    return true;
+export const isCCComponent = (node: t.Node, removed: PluginPass['removed']): boolean => {
+  if (t.isIdentifier(node)) {
+    if (node.name === 'CC' || removed.has(node.name)) {
```
madou, line 83: One issue I'm seeing with this is `removed` will have both the identifier names of `CC` and `CS` components. That would mean that this function would return true if its either a `CC` or `CS`, no? 

🤔 

If we go back to the drawing board the main thing we want to do is understand what the identifiers are for these components, right?

Instead of using `removed` should we either:

1. Do something similar to the React change and do a "starts with" check
2. Or, potentially more resilient, store the identifiers in the pass object.

```
this.identifiers.React = path.node.local.name;
this.identifiers.CC = ...
this.identifiers.CS = ...
```

And so on. What's your thoughts?

MonicaOlejniczak: +1

youzi: Option 2 would be most robust: you can't just go stripping everything that starts with `CC` or `CS` :). 

For `CC` and `CS` it should also check the source of the import: users might have other imports carrying these names. You don't want to strip those out.

madou:
Hey @youzi we'd be keen for this contribution - do you need any help? There are a few conflicts now.

madou:
Hey @youzi we'd be keen for this contribution - do you need any help? There are a few conflicts now.

youzi:
@madou Unfortunately I'm not able to work on this at the moment. If someone could implement the requested changes that would be great.

itsdouges:
Closing a stale, we're more than happy to take a look if anyone in the community wants to pick this up again!

PR #839, opened 2021-09-30
Add Flow types support

JakeLane:
A simple implementation for flow types with an aim to reduce dev loop impact.

This change takes advantage of [flowgen](https://github.com/joarwilk/flowgen) which is impressive but doesn't have a perfect implementation. A shell script is used to extract any override files (none existing yet), generate flow types with flowgen, and patch the flow types.

The patch process is implemented with simple shell transform commands (`sed` and `awk`) to keep complexity down.

Jest types are provided but since global augmentation is not supported in flow, do not function by default.


scripts/generate-flow-types.sh
```diff
@@ -0,0 +1,63 @@
...
+  if [ ! -f "${file%.*.*}.js.flow" ]; then
+    echo "Type '${file%.*.*}.js.flow' was not generated"
+    exit 1
+  fi
+done
+
+echo "Patching types"
+find "${PACKAGES[@]}" -type f -path '*/dist/*' -name '*.js.flow' -print0 | while read -rd $'\0' file; do
+  # Change import to import type (bug in flowgen)
+  sed -i '' 's/import {/import type {/g' "$file"
```
JakeLane, line 36: Would like your opinion here @MonicaOlejniczak (and @madou if you're interested).

I put this together as a way to patch together the types as I corrected the errors. Do we think this is manageable enough for the dev loop in its current form? I'm not married to this idea but I liked how simple and readable it ended up so I kept the approach. 

I think plain transformation without parsing is sufficient but I could re-write this into TS.

examples/packages/flow-types/src/index.js
```diff
@@ -0,0 +1,56 @@
+// @flow strict-local
```
MonicaOlejniczak, line 1: What is this file and package for exactly? Is it to showcase how to apply flow types, or to test the types? If the latter, then tests should live in the package they're defined in (i.e. `@compiled/react`). You can do something similar to what I setup for the RLL repository, so that you have test files like so: https://github.com/atlassian-labs/react-loosely-lazy/blob/master/packages/core/react-loosely-lazy/src/lazy/__tests__/lazy.flow.test.js

yarn.lock
```diff
@@ -8381,11 +8381,30 @@ flatten@^1.0.2:
   resolved "https://registry.yarnpkg.com/flatten/-/flatten-1.0.3.tgz#c1283ac9f27b368abc1e36d1ff7b04501a30356b"
   integrity sha512-dVsPA/UwQ8+2uoFe5GHtiBMu48dWLTdsuEd7CKGlZlD78r1TTWBvDuFaFGKCo/ZfEr95Uk56vZoX86OsHkUeIg==
 
+flow-bin@^0.160.1:
+  version "0.160.1"
+  resolved "https://packages.atlassian.com/api/npm/npm-remote/flow-bin/-/flow-bin-0.160.1.tgz#ae24ad6def1f05ae37789b6cc332b58975d29de0"
```
MonicaOlejniczak, line 8386: Can you make this resolve from `yarnpkg` instead of atlassian?

scripts/generate-flow-types.sh
MonicaOlejniczak: I think it's fine, since you should always be importing the type

In general, I prefer writing complex scripts with node / typescript, but it's fairly readable atm.

.github/workflows/build.yml
```diff
@@ -35,6 +35,9 @@ jobs:
       - name: Build source
         run: yarn build
 
+      - name: Build flow types
+        run: yarn build:types-flow
```
MonicaOlejniczak, line 39: Normally the generated types are still committed to the repository, so that you can visually inspect the generated output or manually modify them. I think that would be good to have, in addition to having a build step validate that the current types are up to date with the generated types. This might complicate the process too much, but figured I would mention it anyway.

MonicaOlejniczak (changes requested):

examples/packages/flow-types/src/index.js
JakeLane: The main intention was to break builds if the type imports failed when importing the package. Moving to regular unit test files sounds good to me, I don't think there's a huge value to importing the package retrospectively - things are gonna break elsewhere first.

yarn.lock
JakeLane: ooh nice catch

examples/packages/flow-types/src/index.js
JakeLane: - [ ] Move to test files in package

yarn.lock
JakeLane: - [ ] Fix lock file

.github/workflows/build.yml
JakeLane: sounds good to me. I'll see if I can chuck it in the pre-commit hook as well.

.flowconfig
```diff
@@ -0,0 +1,19 @@
...
+
+[declarations]
+<PROJECT_ROOT>/flow_typed/.*
+.*/node_modules/.*
+
+[lints]
+
+[options]
+
+[strict]
```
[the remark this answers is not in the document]
JakeLane: Done 👍

.github/workflows/build.yml
JakeLane: Didn't end up putting it in the pre-commit hook yet so the scope doesn't blow out too much (type generation is slow so I'd need to make sure the staged changes are only applied) but will consider doing this later in a future PR if the dev loop suffers too much

PR #842, opened 2021-10-01
Update to parcel 2.0.0-rc.0

andparsons:
Updates the parcel-transformer to match the changes in parcel@2.0.0-rc-0

### DIFF
--- a/packages/parcel-transformer/src/transformer.tsx
+++ b/packages/parcel-transformer/src/transformer.tsx
@@ -24,12 +24,12 @@ export default new Transformer<UserlandOpts>({
     const contents: UserlandOpts = {};
 
     if (conf) {
-      config.shouldInvalidateOnStartup();
+      config.invalidateOnStartup();
       Object.assign(contents, conf.contents);
     }
 
     // We always set a result so it's only evaluated once.
-    config.setResult(contents);
+    return contents;
   },
 
   canReuseAST() {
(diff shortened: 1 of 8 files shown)


packages/parcel-transformer/src/transformer.tsx
```diff
@@ -49,13 +48,18 @@ export default new Transformer<UserlandOpts>({
...
       filename: asset.filePath,
       caller: { name: 'compiled' },
       rootMode: 'upward-optional',
     });
 
-    return ast;
+    return {
+      type: 'babel',
+      version: '7.0.0',
+      code,
```
MonicaOlejniczak, line 60: You should remove `code`, as it's not used elsewhere in the transformer and shouldn't be needed in general

packages/parcel-transformer/src/transformer.tsx
```diff
@@ -111,13 +115,14 @@ export default new Transformer<UserlandOpts>({
...
-  generate({ ast, asset }) {
+  async generate({ ast, asset }) {
     // TODO: We're using babels standard generator. Internally parcel does some hacks in
     // the official Babel transformer to make it faster - using ASTree directly.
     // Perhaps we should do the same thing.
-    const { code, map } = generate(ast.program, {
+    const { code } = generate(ast.program, {
       filename: asset.filePath,
     });
+    const map = await asset.getMap();
```
MonicaOlejniczak, line 125: This will create inaccurate source maps, as it will use an old one. You should be able to update the line before to:

```
const { code, map } = generate(ast.program, {
  filename: asset.filePath,
  sourceMaps: true,
});
```

MonicaOlejniczak (changes requested):

madou:
Hi @andparsons - considering Parcel 2 is officially released would you be interested in contributing this further?

JakeLane:
Superseded by $794

PR #863, opened 2021-10-20
Keep `css` import in SC-to-Compiled codemod

flakeparadigm:
With the work being done in #857, we can now support keeping the `css` tagged template literals from Styled Components. Additionally, this implementation should make it easy to whitelist any additional Styled Components named imports we may eventually add support for.


JakeLane:
I think breaking changes for the codemod package are fine, as long as we follow semver correctly. We're expecting to change the API occasionally to support new features in our codemods.

PR #873, opened 2021-10-27
Fix ternary condition applying all styles when taking truthy path

### LINKED ISSUE #871: Falsy path styles always applied in conditional expressions
**Describe the bug**
When conditional ternary expressions evaluate to true, styles from both truthy and falsy paths get applied.

**To Reproduce**
With the following condition, if `ABTest()` returns true all styles will get applied:
```
${() => ABTest() ? 'font-weight: bold' : css`font-size: 20px; color: red;`};
```
Problem does not exist if both paths modify the same style as falsy style is overridden with the truthy style. For example:
```
${() => ABTest() ? 'color: blue' : 'color: red'};
```

**Expected behavior**
Only styles from one side of the ternary should be applied.

at-nathan:
Resolves #871 

This PR changes the behaviour of ternary conditionals to not apply falsy path styles by default, instead making both paths into logical CSS items. This behaviour only works for situations like:
```
${() => ABTest() ? 'color: blue' : 'color: red'};
```
As the style property is the same on both paths can be overridden with the truthy path style. However when styles on both paths don't match up. For example:
```
${() => ABTest() ? 'font-weight: bold' : css`font-size: 20px; color: red;`};
```
Then we get unwanted styles for the falsy path.

### DIFF
--- a/.changeset/cold-toes-perform.md
+++ b/.changeset/cold-toes-perform.md
@@ -1,0 +1,5 @@
+---
+'@compiled/babel-plugin': minor
+---
+
+**Breaking change:** Ternary conditionals will no longer add falsy path styles when expression evaluates truthy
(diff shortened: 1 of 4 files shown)


MonicaOlejniczak (approved):
LGTM

JakeLane (approved):

PR #901, opened 2021-11-09
Add disable directives for css prop

JakeLane:
Usages like the following do not work:

```
import { css as cssSC } from "styled-components";
import { css as cssCompiled } from "@compiled/react";

const stylesSC = cssSC`
  color: red;
`;

const stylesCompiled = cssCompiled`
  color: blue;
`;

<>
  <div css={stylesSC}>Red text</div>
  <div css={stylesCompiled}>Blue text</div>
</>;
```

This is because Compiled cannot tell that `stylesSC` is defined with styled-components. To allow usages to opt-out, we should add a comment directive to allow the Compiled babel plugin to ignore certain props. e.g.

```
{isCompiledEnabled()
  ? <div css={{ color: 'red' }} />
  // @compiled-disable-next-line transform-css-prop
  : <div css={{ color: 'blue' }} />
}
```

This PR introduces `// @compiled-disable-next-line transform-css-prop` and `// @compiled-disable transform-css-prop` as comment directives in the babel plugin. When these directives are placed on either the CSS prop or the JSX opening element, Compiled will ignore the AST node and not traverse.

### DIFF
--- a/examples/stories/css-prop-static-object.tsx
+++ b/examples/stories/css-prop-static-object.tsx
@@ -82,3 +82,19 @@ export const ObjectLiteralMapWithKeys = (): JSX.Element => (
     ))}
   </div>
 );
+
+export const ObjectLiteralDisabledSameLine = (): JSX.Element => (
+  <h1
+    css={{ color: 'red' }} // @compiled-disable transform-css-prop
+  >
+    Black text
+  </h1>
+);
+
+export const ObjectLiteralDisabledNextLine = (): JSX.Element => (
+  <h1
+    // @compiled-disable-next-line transform-css-prop
+    css={{ color: 'red' }}>
+    Black text
+  </h1>
+);
--- a/examples/stories/css-prop-static-object.tsx
+++ b/examples/stories/css-prop-static-object.tsx
@@ -83,15 +83,15 @@ export const ObjectLiteralMapWithKeys = (): JSX.Element => (
   </div>
 );
 
-export const ObjectLiteralDisabledSameLine = (): JSX.Element => (
+export const ObjectExpressionDisabledSameLine = (): JSX.Element => (
   <h1
-    css={{ color: 'red' }} // @compiled-disable transform-css-prop
+    css={{ color: 'red' }} // @compiled-disable-line transform-css-prop
   >
     Black text
   </h1>
 );
 
-export const ObjectLiteralDisabledNextLine = (): JSX.Element => (
+export const ObjectExpressionDisabledNextLine = (): JSX.Element => (
   <h1
     // @compiled-disable-next-line transform-css-prop
     css={{ color: 'red' }}>
--- a/packages/babel-plugin/src/constants.tsx
+++ b/packages/babel-plugin/src/constants.tsx
@@ -1,1 +1,3 @@
 export const PROPS_IDENTIFIER_NAME = 'props';
+export const COMPILED_DIRECTIVE_DISABLE_SAME_LINE = '@compiled-disable';
+export const COMPILED_DIRECTIVE_DISABLE_NEXT_LINE = '@compiled-disable-next-line';
--- a/packages/babel-plugin/src/constants.tsx
+++ b/packages/babel-plugin/src/constants.tsx
@@ -1,3 +1,5 @@
 export const PROPS_IDENTIFIER_NAME = 'props';
-export const COMPILED_DIRECTIVE_DISABLE_SAME_LINE = '@compiled-disable';
+
+export const COMPILED_DIRECTIVE_DISABLE_LINE = '@compiled-disable-line';
 export const COMPILED_DIRECTIVE_DISABLE_NEXT_LINE = '@compiled-disable-next-line';
+export const COMPILED_DIRECTIVE_TRANSFORM_CSS_PROP = 'transform-css-prop';
--- a/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
+++ b/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
@@ -633,29 +633,29 @@ describe('css prop behaviour', () => {
 
   it('should not transform css prop with comment directive', () => {
     const actual = transform(`
-      import "@compiled/react";
+      import { css } from '@compiled/react';
 
       // @compiled-disable-next-line transform-css-prop
-      const RedDiv = <div css={{ color: "red" }} />;
+      const RedDiv = <div css={{ color: 'red' }} />;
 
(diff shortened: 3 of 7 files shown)


examples/stories/css-prop-static-object.tsx
```diff
@@ -82,3 +82,19 @@ export const ObjectLiteralMapWithKeys = (): JSX.Element => (
     ))}
   </div>
 );
+
+export const ObjectLiteralDisabledSameLine = (): JSX.Element => (
```
MonicaOlejniczak, line 86: `ObjectLiteralDisabledSameLine` > `ObjectExpressionDisabledSameLine`

examples/stories/css-prop-static-object.tsx
```diff
@@ -82,3 +82,19 @@ export const ObjectLiteralMapWithKeys = (): JSX.Element => (
...
+  <h1
+    css={{ color: 'red' }} // @compiled-disable transform-css-prop
+  >
+    Black text
+  </h1>
+);
+
+export const ObjectLiteralDisabledNextLine = (): JSX.Element => (
```
MonicaOlejniczak, line 94: `ObjectLiteralDisabledNextLine` > ObjectExpressionDisabledNextLine

packages/babel-plugin/src/constants.tsx
```diff
@@ -1 +1,3 @@
 export const PROPS_IDENTIFIER_NAME = 'props';
+export const COMPILED_DIRECTIVE_DISABLE_SAME_LINE = '@compiled-disable';
```
MonicaOlejniczak, line 2: `@compiled-disable` > `@compiled-disable-line` makes this in-line with eslint

Then rename `COMPILED_DIRECTIVE_DISABLE_SAME_LINE` to `COMPILED_DIRECTIVE_DISABLE_LINE`

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.tsx
```diff
@@ -630,4 +630,32 @@ describe('css prop behaviour', () => {
...
+  it('should not transform css prop with comment directive', () => {
+    const actual = transform(`
+      import "@compiled/react";
+
+      // @compiled-disable-next-line transform-css-prop
+      const RedDiv = <div css={{ color: "red" }} />;
+
+      const GreenDiv = () => (
+        <div
+          css={{ color: "green" }} // @compiled-disable transform-css-prop
```
MonicaOlejniczak, line 643: Maybe add a different usage here, like:

```
css`color: green`
```

packages/babel-plugin/src/css-prop/index.tsx
```diff
@@ -16,6 +23,28 @@ const getJsxAttributeExpression = (node: t.JSXAttribute) => {
   throw new Error('Value of JSX attribute was unexpected.');
 };
 
+const isCssPropDisabled = (path: NodePath<t.Node>, meta: Metadata): boolean => {
+  const { before, same } = getNodeDirectiveComments(path, meta);
```
MonicaOlejniczak, line 27: `same` > `current`

packages/babel-plugin/src/css-prop/index.tsx
```diff
@@ -2,8 +2,15 @@ import * as t from '@babel/types';
...
 import { buildCompiledComponent } from '../utils/ast-builders';
 import { buildCss } from '../utils/css-builders';
+import {
+  COMPILED_DIRECTIVE_DISABLE_NEXT_LINE,
+  COMPILED_DIRECTIVE_DISABLE_SAME_LINE,
+} from '../constants';
+import { getNodeDirectiveComments } from '../utils/ast';
 import type { Metadata } from '../types';
 
+const COMPILED_DIRECTIVE_TRANSFORM_CSS_PROP = 'transform-css-prop';
```
MonicaOlejniczak, line 12: Should this go in `constants`?

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -740,3 +741,40 @@ const tryWrappingBlockStatementInIIFE = (node: t.BlockStatement | t.Expression)
...
+ * Get comments for `path` in both the line before and on the same line.
+ *
+ * e.g.
+ * `<div css={{color: 'green'}} /> // @compiled-disable` will output `{before: [], same: [...]}
+ *
+ * @param path {NodePath<t.Node>}
+ * @param meta {Metadata} Context for the transform
+ * @returns {before: t.CommentLine[], same: t.CommentLine[]} Comments before and on the same line as the input path
+ */
+export const getNodeDirectiveComments = (
```
MonicaOlejniczak, line 755: Can this go in its own util file? The large utils are unbearable to read. Maybe something like `utils/directives.ts` though I'm confused why `directive` is needed in the name since it's meant to be generically retrieving comments? Maybe just `utils/comments.ts` and `getComments`?

MonicaOlejniczak (changes requested):

packages/babel-plugin/src/utils/ast.tsx
JakeLane: Yeah makes sense to me. I put directive as the function only retrieves the above and same line, which is close to the directive use-case.

JakeLane: I went with `getNodeComments` as I want to keep the context that the comments are relevant to the AST node specified

MonicaOlejniczak (approved):

packages/babel-plugin/src/css-prop/index.tsx
```diff
@@ -26,20 +52,25 @@ const getJsxAttributeExpression = (node: t.JSXAttribute) => {
  */
 export const visitCssPropPath = (path: NodePath<t.JSXOpeningElement>, meta: Metadata): void => {
   let cssPropIndex = -1;
-  const cssProp = path.node.attributes.find((attr, index): attr is t.JSXAttribute => {
-    if (t.isJSXAttribute(attr) && attr.name.name === 'css') {
+  const cssProp = path.get('attributes').find((attr, index): attr is NodePath<t.JSXAttribute> => {
```
at-nathan, line 55: Oh nice, does `.get` allow you to traverse the `NodePath` Object to find a property?

at-nathan (approved):

packages/babel-plugin/src/css-prop/index.tsx
JakeLane: yeah, I think this is something we're doing wrong throughout the repo. Once you unbox the path into a node, there's no way to get it back, so we should probably pass around paths instead.

PR #903, opened 2021-11-09
feat: styled-components attrs codemod

blackheart-atlassian:
Work in progress PR, just wanna to start a discussion and hear your thoughts. I know it might be a lot to review, but want to hear some feedback as soon as possible.

we transform from:
```
export const Input = styled.input.attrs(props => ({
  style: (props) => ({
    left: props.left,
    top: props.top,
  }),
  type: 'text',
}))`
  position: absolute;
`;
```

to: 
```
const CompiledInput = styled.div`
  position: absolute;
  left: ${({ left }) => left};
  top: ${({ top }) => top};
`;

export const Input = (props) => <CompiledInput type="text" {...props} />;
```


packages/codemods/src/transforms/styled-components-to-compiled/__tests__/styled-components-attributes-normalize.test.tsx
```diff
@@ -0,0 +1,413 @@
...
+        }),
+      }))\`
+        position: absolute;
+      \`;
+      `,
+    `
+      import { styled } from '@compiled/react';
+
+      export const Input = styled.input\`
+left: $\{props => props.left\};
```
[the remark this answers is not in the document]
blackheart-atlassian: that will require adding extra indentations in the transformer's code. do we wanna do it?

packages/codemods/src/helpers/styled-components-attributes.ts
```diff
@@ -0,0 +1,278 @@
...
+  plugins: CodemodPluginInstance[];
+  templateExpressions: Collection<TaggedTemplateExpression>;
+}): void => {
+  templateExpressions.forEach((templateExpression) => {
+    let newComponentDeclaration = null;
+    let extraContent = null;
+
+    const componentDeclarator = templateExpression.parentPath;
+    const componentDeclaration = componentDeclarator.parentPath.parentPath;
+    const componentName = componentDeclarator.value.id.name;
```
JakeLane, line 209: nit: I think this would be easier to read if the variables were moved closer to their first usage

packages/codemods/src/plugins/default.ts
```diff
@@ -14,6 +14,34 @@ const defaultCodemodPlugin: CodemodPlugin = {
         return newImport;
       },
 
+      buildAttributes({ originalNode, transformedNode, composedNode }) {
```
TimeRaider, line 17: other ones use `currentNode`, should we stick to similar naming maybe?

packages/codemods/src/plugins/default.ts
```diff
@@ -14,6 +14,34 @@ const defaultCodemodPlugin: CodemodPlugin = {
...
+          originalNode.value.type === 'TaggedTemplateExpression'
+        ) {
+          const newDeclaration = j(originalNode).replaceWith(transformedNode).get();
+
+          if (composedNode) {
+            let candidateForInsertion = newDeclaration;
+
+            while (
+              candidateForInsertion.parentPath &&
+              candidateForInsertion.parentPath.name !== 'body'
```
TimeRaider, line 29:
```suggestion
              candidateForInsertion.parentPath?.name !== 'body'
```
are you looping to find the first node that has `parentPath === 'body'`?

packages/codemods/src/plugins/default.ts
```diff
@@ -14,6 +14,34 @@ const defaultCodemodPlugin: CodemodPlugin = {
...
+              candidateForInsertion = candidateForInsertion.parentPath;
+            }
+
+            j(candidateForInsertion).insertBefore(composedNode);
+          }
+
+          return newDeclaration;
+        }
+
+        throw new Error(
```
TimeRaider, line 40: not sure if I understand this. You have the following condition:
```js
if (
  originalNode.value.type === 'VariableDeclaration' ||
  originalNode.value.type === 'TaggedTemplateExpression'
)
``` 
but here you mention `ExportNamedDeclaration` and `ExportDefaultDeclaration`

also, should we even `throw` at all or should we just skip?

packages/codemods/src/plugins/default.ts
TimeRaider: also shouldn't the signature match `BuildAttributesContext`?

packages/codemods/src/plugins/types.ts
```diff
@@ -40,6 +44,20 @@ export type BuildRefAttributesContext<T> = ValidateConfig<
...
+  {
+    // The original component declaration
+    originalNode: ASTPath<VariableDeclaration | TaggedTemplateExpression>;
+    // The existing component declaration
+    currentNode: ASTPath<VariableDeclaration | TaggedTemplateExpression>;
+    // The original node after transforms
+    transformedNode: VariableDeclaration | TaggedTemplateExpression | ArrowFunctionExpression;
+    // The composed node that's been created during transformation
+    composedNode: VariableDeclaration | TaggedTemplateExpression | null;
+  }
```
TimeRaider, line 58: maybe it's me but these comments don't help :) Not sure what the difference is between `original` and `exiting` nodes. Also each property has `Node` suffix but 2 of them are `ASTPath`s :) Again, maybe it's just me

packages/codemods/src/transforms/styled-components-to-compiled/README.md
```diff
@@ -35,3 +35,20 @@ on your next run explicitly set the version number:
...
 ```
+
+## Gotchas
+
+`styled.div.attrs` spread properties are not supported.
+
+_Example_
+
+```
+styled.div.attrs({
```
TimeRaider, line 46: is this an example of what NOT to do?

packages/codemods/src/transforms/styled-components-to-compiled/styled-components-to-compiled.tsx
```diff
@@ -53,6 +54,21 @@ const transformer = (fileInfo: FileInfo, api: API, options: Options): string =>
...
   });
 
+  const taggedTemplateExpressions = collection.find(
+    j.TaggedTemplateExpression,
+    ({ tag: expression }) =>
+      expression.callee.object.object.name === 'styled' &&
+      expression.callee.property.name === 'attrs'
+  );
+
+  if (taggedTemplateExpressions.length) {
```
TimeRaider, line 64: can be inlined .... :D j/k

packages/codemods/src/utils/styled-components-attributes.ts
```diff
@@ -0,0 +1,367 @@
...
+const getDeclarationData = (expression: ASTPath) => {
+  const parentDeclaration = expression.parentPath;
+
+  if (
+    parentDeclaration.value.type === 'VariableDeclarator' &&
+    parentDeclaration.value.id.type === 'Identifier'
+  ) {
+    return {
+      name: parentDeclaration.value.id.name as string,
+      declaration: findParent(expression, 'VariableDeclaration') as StyledAttributesDeclarationNode,
```
TimeRaider, line 211:
```suggestion
      declaration: findParent(expression, 'VariableDeclaration') as StyledAttributesDeclarationNode,
      isAnonymous: false,
```

packages/codemods/src/plugins/default.ts
blackheart-atlassian: yep, just need to find a place where we can add a new component declaration

packages/codemods/src/transforms/styled-components-to-compiled/README.md
blackheart-atlassian: yeeeah, not sure what to put in the description here and thought maybe example would be helpful. But it feels like it's rather confusing, isn't it?

packages/codemods/src/plugins/types.ts
blackheart-atlassian: so, `original` node is the node without any transformations and the `current` node is the node after applying plugins consecutively. You also have an access to a transformed node after initial transformations but without any plugins applied to it yet.

packages/codemods/src/plugins/default.ts
blackheart-atlassian: true, I can't think of any scenarios in the current implementation where it'll be something else. I'll change it to a warning

JakeLane (dismissed):

packages/codemods/src/utils/styled-components-attributes.ts
```diff
@@ -0,0 +1,367 @@
...
+  name,
+}: {
+  j: JSCodeshift;
+  expression: TaggedTemplateExpression;
+  name: string;
+}) => j.variableDeclaration('const', [j.variableDeclarator(j.identifier(name), expression)]);
+
+/**
+ * Returns declaration data
+ * @returns data.name - component name if it's a variable diclaration or generic component name
```
at-nathan, line 198: typo `diclaration`. declaration?

packages/codemods/src/utils/styled-components-attributes.ts
```diff
@@ -0,0 +1,367 @@
...
+
+  if (
+    parentDeclaration.value.type === 'VariableDeclarator' &&
+    parentDeclaration.value.id.type === 'Identifier'
+  ) {
+    return {
+      name: parentDeclaration.value.id.name as string,
+      declaration: findParent(expression, 'VariableDeclaration') as StyledAttributesDeclarationNode,
+    };
+  } else {
```
at-nathan, line 213: nit: Redundant else

packages/codemods/src/plugins/default.ts
at-nathan: 👍

packages/codemods/src/transforms/styled-components-to-compiled/__tests__/styled-components-attributes-normalize.test.tsx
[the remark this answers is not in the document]
at-nathan: @MonicaOlejniczak I had a go at getting some simple formatting going here using `pretty-print` functionality. However it made no different because it's a tagged string literal and the printer lib has no context on how `styled-components` should be formatted.

I had a go at formatting it myself but ran into complexity dealing with expressions.

JakeLane (approved):

PR #904, opened 2021-11-09
Support nested ternary operators

flakeparadigm:


packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
```diff
@@ -417,14 +417,41 @@ describe('styled component behaviour', () => {
...
+      '._19bvftgi{padding-left:8px}',
+      '._n3tdftgi{padding-bottom:8px}',
+      '._u5f3ftgi{padding-right:8px}',
+      '._ca0qftgi{padding-top:8px}',
+      '._19itlf8h{border:2px solid blue}',
+      '._1wyb1ul9{font-size:30px}',
+      '._syaz1xo0{color:var(--_13kr4bd)}',
+    ]);
+
+    expect(actual).toInclude(
```
at-nathan, line 442: This `toInclude` can go inside `toIncludeMultiple`. I think we're trying to keep everything in one assertion if possible.

packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
```diff
@@ -441,6 +468,28 @@ describe('styled component behaviour', () => {
...
+    `);
+
+    expect(actual).toIncludeMultiple([
+      '._bfhk1x77{background-color:white}',
+      '._syaz11x8{color:black}',
+      '._syaz13q2{color:blue}',
+      '._syaz5scu{color:red}',
+    ]);
+
+    expect(actual).toInclude(
```
at-nathan, line 488: Ditto here

packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
```diff
@@ -463,19 +512,49 @@ describe('styled component behaviour', () => {
     );
   });
 
+  it('should not allow a logical statement with a conditional right-hand side', () => {
```
at-nathan, line 515: Nice!

packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
```diff
@@ -463,19 +512,49 @@ describe('styled component behaviour', () => {
...
+    `);
+
+    expect(actual).toIncludeMultiple([
+      '._syazbf54{color:green}',
+      '._syaz5scu{color:red}',
+      'className={ax(["",props.isPrimary?"_syazbf54":"_syaz5scu",props.className])}',
+    ]);
+  });
+
+  it('should apply conditional CSS when using "key: value; key: value; ..." in string form', () => {
```
at-nathan, line 543: hmm this is cool, previously consumer needed to use `css()` to do multi style declarations. I wonder if styled-components supports this?

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -473,14 +473,13 @@ export const buildStyledComponent = (tag: Tag, cssOutput: CSSOutput, meta: Metad
   // Rely on transformCss to remove duplicates and return only the last unconditional CSS for each property
   const uniqueUnconditionalCssOutput = transformCss(unconditionalCss.join(''));
 
-  // Rely on transformItemCss to build logicalExpressions for logical CSS
-  const logicalCssOutput = transformItemCss({ css: logicalCss, variables: cssOutput.variables });
-
-  const sheets = [...uniqueUnconditionalCssOutput.sheets, ...logicalCssOutput.sheets];
+  // Rely on transformItemCss to build expressions for conditional & logical CSS
+  const conditionalCssOutput = transformItemCss(conditionalCss);
```
at-nathan, line 477: Oh so `transformItemCss` never did anything with `variables: cssOutput.variables`?

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -263,15 +307,33 @@ const extractConditionalExpression = (node: t.ConditionalExpression, meta: Metad
...
+    });
+  } else if (consequentCss?.length) {
+    // convert single-sided conditional into logical statements
+    css.push(...getLogicalItemFromConditionalExpression(consequentCss, node, 'consequent'));
+  } else if (alternateCss?.length) {
+    css.push(...getLogicalItemFromConditionalExpression(alternateCss, node, 'alternate'));
   }
 
-  return { css: mergeSubsequentUnconditionalCssItems(css), variables };
+  return { css: css, variables };
```
at-nathan, line 336: nit: `{ css, variables }`

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -91,7 +91,9 @@ const mergeSubsequentUnconditionalCssItems = (arr: Array<CssItem>): Array<CssIte
  * @param item
  */
 export const getItemCss = (item: CssItem): string => {
-  return item.css;
+  return item.type === 'conditional'
+    ? [...item.consequent, ...item.alternate].map(getItemCss).join('')
```
at-nathan, line 95: ❤️

at-nathan (commented):
Looking good! Thoughts on adding a storybook example to `conditional-rules-styled`?

packages/babel-plugin/src/utils/types.tsx
```diff
@@ -19,7 +26,7 @@ export interface SheetCssItem {
   css: string;
 }
 
-export type CssItem = UnconditionalCssItem | LogicalCssItem | SheetCssItem;
+export type CssItem = UnconditionalCssItem | ConditionalCssItem | LogicalCssItem | SheetCssItem;
```
at-nathan, line 29: I think we will need to generate the flow types too.

packages/babel-plugin/src/utils/ast-builders.tsx
flakeparadigm: Apparently not. I double checked several times to make sure I wasn't going crazy 😂 haha

packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
flakeparadigm: Ah yeah, I didn't do anything particular to support this. It already worked before my changes, it would just be pulled out into logical statements like...
```
props.isPrimary&&"_syazbf54 _1wyb1fwx", !props.isPrimary&&"_syaz5scu _1wybexct"
```

flakeparadigm: Looks like this would work in Styled Components as well, based on the little demo they have:
Screen Shot 2021-11-11 at 11 42 25 am

flakeparadigm: Yep, using the demo on their site, I can confirm this also works in styled-components.

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -543,7 +605,7 @@ const extractTemplateLiteral = (node: t.TemplateLiteral, meta: Metadata): CSSOut
 
   css.push({ type: 'unconditional', css: literalResult });
 
-  // Deals with Conditional CSS Rules
+  // Deals with Conditional CSS Rules from Logical Expressions
   node.expressions.forEach((prop) => {
```
MonicaOlejniczak, line 609: Can this logic be moved into the main looping logic above? I can't quite remember the reasoning for having it here

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -238,10 +282,10 @@ const callbackIfFileIncluded = (meta: Metadata, next: Metadata) => {
...
   const conditionalPaths: ['consequent', 'alternate'] = ['consequent', 'alternate'];
-  const css = [];
-  const variables = [];
+  const css: CSSOutput['css'] = [];
+  const variables: CSSOutput['variables'] = [];
 
-  for (const path of conditionalPaths) {
+  const [consequentCss, alternateCss] = conditionalPaths.map((path) => {
     const pathNode = node[path];
     let cssOutput;
```
MonicaOlejniczak, line 291: Do we still need all the checks before `t.isIdentifier` or can we dumb it down to always try and build the css if it's not an identifier or conditional?

packages/babel-plugin/src/utils/ast-builders.tsx
```diff
@@ -526,26 +525,42 @@ export const getPropValue = (
...
+  cssItems.forEach((item) => {
+    switch (item.type) {
+      case 'conditional':
+        const consequent = transformItemCss(item.consequent);
+        const alternate = transformItemCss(item.alternate);
 
-    sheets.push(...css.sheets);
+        sheets.push(...consequent.sheets, ...alternate.sheets);
+        classNames.push(
+          t.conditionalExpression(item.test, consequent.classNames[0], alternate.classNames[0])
```
MonicaOlejniczak, line 542: Is it possible for there to be multiple `classNames`? If so, should they be joined together somehow? Or should the types change to not be an array?

packages/babel-plugin/src/utils/css-builders.tsx
flakeparadigm: I'm honestly not very certain about why it isn't in the main loop, but there is already a logic path in the main loop that would conflict with this one related to support for keyframes.

I updated this comment because the use of the word "Conditional" here was throwing me off every time I searched for it in the code. Could be worth investigating further, but I think it would go beyond the scope of this work.

packages/babel-plugin/src/utils/css-builders.tsx
flakeparadigm: Sadly simplifying it here doesn’t play nicely with the logic & assumptions in `buildCss` - it expects things/expressions that we know to contain whole CSS declarations. However, Compiled supports using Conditional Expressions that either return a whole declaration, or just the value side of a declaration.

So we have to do the filtering here to make sure we only try to `buildCss` with whole declarations or more.

packages/babel-plugin/src/utils/ast-builders.tsx
flakeparadigm: It shouldn't be possible to have multiple entries in `classNames` here (which is a confusing as the `classNames` array actually represents actual JS Expressions that could contain multiple CSS class names).

I've re-worked the logic here to make that expectation clear up-front when parsing the JS in the first place.

flakeparadigm:
@at-nathan Thanks for the suggestion, I've added some storybook examples.

packages/babel-plugin/src/utils/types.tsx
flakeparadigm: Looks like we only generate flowtypes for the `jest` and `react` packages, so these changes don't affect those types.

JakeLane (approved):

MonicaOlejniczak (approved):

PR #906, opened 2021-11-11
Support namespace imports & export specifiers

### LINKED ISSUE #630: Importing through star syntax does not statically evaluate
**Describe the bug**
Importing through star syntax does not statically evaluate.

**To Reproduce**

```js
import * as colors from './imports/colors';

export const Blue = styled.span`
  color: ${colors.blueviolet};
`;
```

```css
._syaz57xd{color:var(--_103kgnr)}
```

Changing to extracting the values individually works.

```js
import { blueviolet }  from './imports/colors';

export const Blue = styled.span`
  color: ${blueviolet};
`;
```

```css
._syazmu8g{color:blueviolet}
```

**Expected behavior**
Values are statically evaluated.

at-nathan:
Resolves #630

Allows styles to be statically evaluated when using:
- Namespace imports - `import * as theme from 'theme';`
- Export specifiers - `export { colors };`

Does not address current issues where:
- Member expressions breaks early when identifiers and call expressions are used mid chain
- Member expressions assume parent path is same throughout the chain

Future work will be done to support:
- Nested namespace imports
- Deconstruction from namespace imports

### DIFF
--- a/.changeset/small-swans-approve.md
+++ b/.changeset/small-swans-approve.md
@@ -1,0 +1,5 @@
+---
+'@compiled/babel-plugin': minor
+---
+
+Add support for namespace imports and export specifiers
--- a/packages/babel-plugin/src/__tests__/module-traversal.test.tsx
+++ b/packages/babel-plugin/src/__tests__/module-traversal.test.tsx
@@ -536,9 +536,9 @@ describe('module traversal', () => {
 
       expect(result).toIncludeMultiple([
         '{color:red}',
-        // Cannot be resolved statically because of bug where
-        // we assume parent path of member expression is the same throughout
-        // the entire chain
+        // We're not handling statically evaluating param values used from a namespace chain,
+        // as the long term strategy is to only statically evaluate expressions scoped locally
+        // to the current file.
         '{background-color:var(--_1w9snyi)}',
       ]);
     });
--- a/packages/babel-plugin/src/__tests__/module-traversal.test.tsx
+++ b/packages/babel-plugin/src/__tests__/module-traversal.test.tsx
@@ -36,7 +36,7 @@ describe('module traversal', () => {
     expect(result).toInclude('{color:blue}');
   });
 
-  it('should replace an identifier referencing a default import specificer string literal', () => {
+  it('should replace an identifier referencing a default import specifier string literal', () => {
     const result = transform(
       `
       import '@compiled/react';
@@ -50,7 +50,7 @@ describe('module traversal', () => {
     expect(result).toInclude('{color:red}');
   });
 
-  it('should replace an identifier referencing a default import specificer string literal', () => {
+  it('should replace an identifier referencing a default import specifier string literal', () => {
     const result = transform(
       `
       import '@compiled/react';
@@ -64,7 +64,7 @@ describe('module traversal', () => {
     expect(result).toInclude('{color:red}');
   });
 
-  it('should replace an identifier referencing a named import specifier object', () => {
+  it('should replace an identifier referencing a named import specifier object from a variable declaration export', () => {
     const result = transform(
       `
       import '@compiled/react';
@@ -76,6 +76,20 @@ describe('module traversal', () => {
     );
 
     expect(result).toInclude('{color:red}');
+  });
+
+  it('should replace an identifier referencing a named import specifier object from an export specifier', () => {
+    const result = transform(
+      `
+      import '@compiled/react';
+      import React from 'react';
+      import { danger } from '../__fixtures__/mixins/objects';
+
+      <div css={{ color: danger }} />
+    `
+    );
+
+    expect(result).toInclude('{color:blue}');
   });
 
   it('should replace an identifier referencing a node modules named import specifier object', () => {
@@ -106,7 +120,7 @@ describe('module traversal', () => {
     expect(result).toInclude('{font-size:12px}');
   });
 
-  it('should inline css from a object spread referencing a named import object', () => {
+  it('should inline css from an object spread referencing a named import object', () => {
     const result = transform(
       `
       import '@compiled/react';
@@ -121,7 +135,7 @@ describe('module traversal', () => {
     expect(result).toInclude('{font-size:12px}');
   });
 
-  it('should inline css from a object with multiple identifiers referenced from a named import', () => {
+  it('should inline css from an object with multiple identifiers referenced from a named import', () => {
     const result = transform(
       `
       import '@compiled/react';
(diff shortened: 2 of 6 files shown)


packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -199,10 +200,6 @@ export const getMemberExpressionMeta = (
   let bindingIdentifier: t.Identifier | null = null;
   let originalBindingType: t.Expression['type'] = 'Identifier';
 
-  if (t.isIdentifier(expression.property)) {
```
at-nathan, line 202: Removed this as it duplicated the last item in the access path. Making it hard to tell if the member path had been fully traversed. We already do this on line: 219.

packages/babel-plugin/src/__tests__/module-traversal.test.tsx
```diff
@@ -36,7 +36,7 @@ describe('module traversal', () => {
     expect(result).toInclude('{color:blue}');
   });
 
-  it('should replace an identifier referencing a default import specificer string literal', () => {
+  it('should replace an identifier referencing a default import specifier string literal', () => {
```
JakeLane, line 39: 🏆

packages/babel-plugin/src/__tests__/module-traversal.test.tsx
```diff
@@ -435,4 +449,109 @@ describe('module traversal', () => {
...
+        const color = { blue: 'blue' };
+
+        const Component = (props) => {
+          return <div css={{ ...objects.colorMixin2(color.blue) }} />
+        };
+    `);
+
+      expect(result).toIncludeMultiple([
+        '{color:red}',
+        // Cannot be resolved statically because of bug where
```
JakeLane, line 539: We should probably make sure to link each workaround in the repo to a github issue in the comment so we can track what needs to be updated

packages/babel-plugin/src/utils/export-traversers.tsx
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * E.g: `export default color;`
+ * Will return the identifier node for `color`.
+ *
+ * @param ast File we want to traverse.
+ */
+export const getDefaultExport = (ast: t.File): Result<t.ExportDefaultDeclaration> | undefined => {
+  let result;
+
+  traverse(ast, {
```
JakeLane, line 24: Not really a problem for the PR as this is the status quo but we'll probably have scale issues unless we remove this stuff per https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md#avoid-traversing-the-ast-as-much-as-possible

JakeLane (dismissed):

packages/babel-plugin/src/utils/evaluate-expression.tsx
```diff
@@ -133,6 +134,24 @@ const evaluateIdentifierBindingMemberExpression = (
...
 
+const evaluateNamespaceImportExpression = (accessPath: t.Identifier[], meta: Metadata) => {
+  const exportName = accessPath[0].name;
+  const updatedMeta: Metadata = { ...meta };
+  const { file } = meta.state;
+  const result =
+    exportName === 'default' ? getDefaultExport(file) : getNamedExport(file, exportName);
+
+  if (result) {
+    updatedMeta.parentPath = result.path.parentPath;
```
at-nathan, line 145: I think I'm suppose to set `ownPath` instead of `parentPath`. I tried it and it didn't make a difference we hit identities from other paths. I might be doing something wrong here. Will look into it more.

packages/babel-plugin/src/utils/export-traversers.tsx
at-nathan: 👍 Maybe I should do a `replaceWith` and build an object expression? Is traversing an object expression a lot cheaper? I would also have to add all the exports as properties even if they are not used?

packages/babel-plugin/src/__tests__/module-traversal.test.tsx
```diff
@@ -435,4 +449,109 @@ describe('module traversal', () => {
...
+        import React from 'react';
+        import * as objects from '../__fixtures__/mixins/objects';
+
+        <div css={{ color: objects.danger }} />
+      `);
+
+      expect(result).toInclude('{color:blue}');
+    });
+
+    it('should inline css from an object spread', () => {
```
MonicaOlejniczak, line 490: Should this include `referencing ...` like the other examples?

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -387,19 +384,11 @@ const findDefaultExportModuleNode = (
...
-  let foundNode: t.Node | undefined = undefined;
-  let foundParentPath: NodePath | undefined = undefined;
-
-  traverse(ast, {
-    ExportDefaultDeclaration(path) {
-      foundParentPath = path as NodePath;
-      foundNode = path.node.declaration as t.Node;
-    },
-  });
+  const result = getDefaultExport(ast);
```
MonicaOlejniczak, line 387: Since this function `findDefaultExportModuleNode ` and `findNamedExportModuleNode` don't do anything anymore, can you remove them and just inline the logic at the call sites?

packages/babel-plugin/src/utils/evaluate-expression.tsx
MonicaOlejniczak: Yeah, I'd expect both to be set since they're used to resolve bindings

packages/babel-plugin/src/utils/export-traversers.tsx
MonicaOlejniczak: `export default` should always be top-level. You be able to manually `find` it in the `program` nodes quite easily instead of `traverse`. Same applies to `getNamedExport`

packages/babel-plugin/src/utils/evaluate-expression.tsx
MonicaOlejniczak: Yeah, I'd expect both to be set since they're used to resolve bindings

packages/babel-plugin/src/__tests__/module-traversal.test.tsx
at-nathan: I left it out of the test name as I wanted to just test spreading still works with namespaces. The previous 2 tests assert that export specifiers and variable declarations are correctly chained to the namespace. But I can add it to be more specific.

packages/babel-plugin/src/utils/export-traversers.tsx
at-nathan: @MonicaOlejniczak I started implementing the loop and remembered we need to get the path for metadata :( Not sure if there is a way to get the path of a node. Should do a `traverse` with enter then do the loop?

MonicaOlejniczak (approved):

packages/babel-plugin/src/utils/export-traversers.tsx
MonicaOlejniczak: I'd just leave it as is for now

JakeLane (approved):

.changeset/small-swans-approve.md
```diff
@@ -0,0 +1,5 @@
+---
+'@compiled/babel-plugin': minor
```
madou, line 2: @MonicaOlejniczak @at-nathan @JakeLane Hey gang, one thing we need to be careful about (at least when we go for wider adoption) is currently since the majority of Compiled's packages are pre 1.0 - each minor version as far as semver is concerned around de-duping, is a major version.

So if we don't have any API change - most changes should flow through patch changes.

PR #918, opened 2021-11-16
Fix: Perform Node/SSR check without process

flakeparadigm:
Change the Node/SSR check to be based on the presence of `document` instead of looking for Node Processes.


flakeparadigm:
The failing builds are due to the bundle size increase - caused by removing the build-time cleanup of SSR related code in the Browser build.

Are we ok with those increases, or should I put the transformation back into the build?

MonicaOlejniczak:
@flakeparadigm I suspect the increase is just from adding the import. We could keep the transform, but just inline the new value. Or, remove the transform like in these changes, and instead remove `isNodeEnvironment` and inline the check. If that doesn't decrease it very much, I'll take a look at the output more.

flakeparadigm:
@MonicaOlejniczak Ah, just did some experimenting and it looks like the increase comes from two things:
- The fact that all the code guarded by `isNodeEnvrionment()` checks cannot be optimised out by a compiler (previously, it was optimised out of bundles because `if (false) { ... }` gets completely removed.
- The new reset functions for caches push it just barely over the current limit (even when the `isNodeEnvrionment` is replaced with `false`.

Changing the transform to inline the `typeof document ...` actually ends up bigger (by 1 byte) than just importing the `isNodeEnvrionment`.

Ultimately, it looks like our options are either... keep the forced explicit `false` in the `browser` output, or accept the bundle size increase.

MonicaOlejniczak:
@flakeparadigm Right, I just had a look at the code and there is much more than I anticipated in the server path. I think given what it's stripping out, it makes sense to keep the transform in for the `browser` build tbh.

packages/react/src/runtime/sheet.tsx
```diff
@@ -28,7 +28,15 @@ export const styleBucketOrdering: Bucket[] = [
 /**
  * Holds all style buckets in memory that have been added to the head.
  */
-const styleBucketsInHead: Partial<Record<Bucket, HTMLStyleElement>> = {};
+let styleBucketsInHead: Partial<Record<Bucket, HTMLStyleElement>>;
```
MonicaOlejniczak, line 31: Can we remove the `Partial` type?

flakeparadigm: We cannot, when assigning the default value later we get this error:
```
Type '{}' is missing the following properties from type 'Record<Bucket, HTMLStyleElement>': "", l, v, w, and 5 more.ts(2740)
```

packages/react/src/keyframes/__tests__/index.test.tsx
```diff
@@ -2,20 +2,28 @@ import { keyframes, styled } from '@compiled/react';
...
+    document.head.querySelectorAll('style'),
     (style) => style.innerHTML
   ).join('\n');
 
   return styles.substring(styles.indexOf(`@keyframes ${name}`));
 };
 
 describe('keyframes', () => {
+  afterEach(() => {
+    resetStyleBuckets();
```
MonicaOlejniczak, line 22: Can't these helpers be removed now that the `browser` build is mostly reverted?

package.json
```diff
@@ -128,7 +128,7 @@
     },
     {
       "path": "./packages/react/dist/browser/runtime.js",
-      "limit": "445B",
+      "limit": "475B",
```
MonicaOlejniczak, line 131: Why did it increase, is it from the helpers?

flakeparadigm: Yeah, it was from the helper functions.

packages/react/src/keyframes/__tests__/index.test.tsx
flakeparadigm: No, the helpers are necessary because these tests now believe they're running in the browser. Previously, all tests ran in SSR mode unless the `isNodeEnvrionment` helper was explicitly mocked to `false`  (because `process` exists even in JSDOM environments - just doesn't try to hide it in any way).

I can switch this test to run in SSR mode again by changing the Jest environment for it, however it seems like these tests are meant to test browser behaviour since there's no mention of SSR in here.

MonicaOlejniczak: Okay, let me have a look to see if there's a better way of tackling this

MonicaOlejniczak: Using environment variables seems better here, since it can be stripped out in the builds...

In the `ts-transform-define`, add:

```
"replace": {
  "isNodeEnvironment()": "process.env.IS_NODE_EXPRESSION",
  "process.env.NODE_ENV === 'test'": "false"
}
```

Then make these changes:

```
const Cache: any = isNodeEnvironment() ? createContext<Record<string, true> | null>(null) : {};

export const useCache: UseCacheHook = () => {
  if (process.env.NODE_ENV === 'test' && process.env.CACHE === 'false') {
    return {};
  }

  // ...
};
```

```
const styleBucketsInHead: Partial<Record<Bucket, HTMLStyleElement>> = {};

function lazyAddStyleBucketToHead(bucketName: Bucket, opts: StyleSheetOpts): HTMLStyleElement {
  if (!styleBucketsInHead[bucketName] || (process.env.NODE_ENV === 'test' && process.env.CACHE === 'false')) {
    // ...
 }

 return styleBucketsInHead[bucketName]!;
};
```

In the test:

```
beforeAll(() => {
  process.env.CACHE = 'false';
});

afterAll(() => {
  delete process.env.CACHE;
});
```
I tried a bunch of other ways with jest mocks, isolated modules, separate test utils entrypoint, but to no avail.

flakeparadigm: > It might be better putting the check in its own util, as that could be stripped out more easily. Not sure if it impacts the bundlesize much though.

I can't actually get `ts-transform-define` to work with just the check just written out inline, so looks like I have to go with the function anyway. Looks like in the end, it only adds 1B to the bundle size that we check for.

MonicaOlejniczak: Awesome, 1B is fine! If we move to a proper bundler later on, we can tree-shake it properly.

packages/react/src/runtime/disable-cache.tsx
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * Any code within this check will be removed in the output bundles:
+ *
+ * ```js
+ * if (isCacheDisabled()) {
+ *   // This code will be removed from the output bundles
+ * }
+ * ```
+ */
+export const isCacheDisabled = (): boolean => {
```
MonicaOlejniczak, line 12: I would name the file `is-cache-disabled` or `cache`

MonicaOlejniczak (dismissed):

packages/react/src/runtime/disable-cache.tsx
flakeparadigm: I'll go with `cache` - I was so torn on which name to pick 😖 haha

JakeLane (dismissed):

packages/react/src/runtime/is-node.tsx
```diff
@@ -11,6 +11,5 @@
  * ```
  */
 export const isNodeEnvironment = (): boolean => {
-  // https://nodejs.org/api/process.html#process_process_release
-  return typeof process !== 'undefined' && process?.release?.name === 'node';
+  return typeof document === 'undefined';
```
madou, line 14: Is there a difference between checking for window vs. document?

packages/tsconfig.options.json
```diff
@@ -13,6 +13,14 @@
     "noImplicitThis": true,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
+    "plugins": [
```
madou, line 16: Could this be placed in the React package since it's only used there?

packages/react/src/runtime/disable-cache.tsx
madou: It might be good to mark this explicitly as used only in tests? Both the file name and function being called something obvious like `INTERNAL_TESTING_ONLY_isStyleCacheDisabled()`. Love the jsdoc mentioned code is removed during bundling! It might also be good to add extra context why it was needed to be used?

Is what we're doing only needed to test the sheet insertion logic? Could consumers need this capability as well (is there a bigger story we're missing?)

packages/react/src/runtime/is-node.tsx
```diff
@@ -11,6 +11,5 @@
  * ```
  */
 export const isNodeEnvironment = (): boolean => {
```
madou, line 13: Did you have any thoughts on this vs. renaming `isBrowserEnvironment`? More change I suppose?

madou:
Oh you're getting an interesting error in CI`error compiled@0.0.0: The engine "node" is incompatible with this module. Expected version "^v14.16.1". Got "16.13.0"`

🤔

packages/react/src/runtime/is-node.tsx
JakeLane: This was part of the spec in the PIR process but looking at the implementation of our SSR babel plugin, `document` and `window` are interchangeable. Let's go with `window` to be more friendly with bundlers by default.

packages/react/src/runtime/is-node.tsx
JakeLane: I think it makes sense to keep the interface the same unless we have a major reason to refactor.

packages/react/src/runtime/is-node.tsx
flakeparadigm: Ah, I missed this quick question - we chose `document` here because Emotion uses it for similar checks as well.

packages/react/tsconfig.json
```diff
@@ -3,7 +3,15 @@
...
-    "tsBuildInfoFile": "tsconfig.esm.tsbuildinfo"
+    "tsBuildInfoFile": "tsconfig.esm.tsbuildinfo",
+    "plugins": [
+      {
+        "transform": "ts-transform-define",
+        "replace": {
+          "isCacheDisabled()": "false"
+        }
+      }
+    ]
```
flakeparadigm, line 14: This will only apply to the `esm` build - the other builds for React use `"extends": "../tsconfig.options.json",` as well, so they won't pick up changes.

Check with @MonicaOlejniczak what might be the best way forward, as I believe she was working on these files recently.

packages/react/src/runtime/is-node.tsx
flakeparadigm: (`window` is probably just fine too. That's just the only particular reason we chose `document` over `window`)

packages/react/tsconfig.json
JakeLane: I ended up creating a common config so we don't duplicate code here

MonicaOlejniczak (approved):

PR #919, opened 2021-11-17
Restructure babel-plugin-strip-runtime package

MonicaOlejniczak:
## Changes
* Add `prettier` dev dependency for tests
* Refactor and restructure existing tests to increase readability
* Remove `README.md` from `files`, since it's always included by npm
* Remove duplicate logic in `removeStyleDeclarations` with `processElement`
* Split out `utils` into individual files to better manage readability
* Support `.ts` test files


JakeLane (approved):

jest.config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   testEnvironment: 'jsdom',
-  testMatch: ['**/src/**/*.test.tsx'],
+  testMatch: ['**/src/**/*.test.(ts|tsx)'],
```
madou, line 3: nit: You'll find only `tsx` files used in the entire repo atm outside of typedefs. This was deliberate to just have only one way to make a file :-).

MonicaOlejniczak: Ah I didn't realise it was deliberate. I personally prefer the distinction, which is why I added them, as then you will know by looking at the file name if it contains any JSX in it or it's pure TS.

PR #922, opened 2021-11-17
feat: adds prettier check to ci

madou:
This pull request replaces the `prettier` command with two new ones: `prettier:check` and `prettier:fix` - and then runs `prettier:fix` in CI.


JakeLane (approved):

.prettierignore
```diff
@@ -0,0 +1,4 @@
+*.js
```
MonicaOlejniczak, line 1: /nit Can these be alphabetised / in the order they'd appear in the IDE? I mostly just want `coverage` before `dist`

```
coverage
dist
CHANGELOG.md
*.js
```

MonicaOlejniczak (dismissed):

.prettierignore
madou: Yep!

JakeLane (approved):

MonicaOlejniczak (approved):

at-nathan (approved):

PR #923, opened 2021-11-17
chore: adds typecheck ci

madou:
This pull request adds `typecheck` as a command and implements basic typechecking over the codebase. Still tinkering with it, if you have an idea how to typecheck "everything" better please let me know.

I kind of hate the state of how we need to build the project before all the types flow through.

Should be merged after #920

**TODO**
- [ ] Fix type errors 
- [ ] Sacrifice a goat


madou:
This one is tricky, will come back to it later.

PR #924, opened 2021-11-17
Improve local dev loop to run straight from source

madou:
This pull request improves the local dev loop and removes the need to build source before running tests and storybook. It also fixes types in tests so all matchers now appear as expected.

**TODO**
- [x] Fix type errors
- [x] Fix builds
- [x] Add changesets
- [x] Rebase from master
- [x] Exclude tests and perf folders from dist

**MY ASK TO YOU**

Please check this branch out and play with this locally. Does it work well for you?

**FOLLOW UP**

Things that would be good to do as a follow up. ATM I want to jump back into the local JSX namespace stuff (which is what I was _actually_ working on for innovation week 😄).

- Investigate better solution for tsconfig paths
- Investigate if we can get rid of all the entrypoint `js` files


types/jest-matchers/index.d.ts
```diff
@@ -0,0 +1,2 @@
+import '../../packages/jest/src/index';
+import '../../node_modules/jest-extended';
```
MonicaOlejniczak, line 2: How come you can't import `jest-extended` directly?

types/jest-matchers/index.d.ts
```diff
@@ -0,0 +1,2 @@
+import '../../packages/jest/src/index';
```
MonicaOlejniczak, line 1: Is this to avoid building? Or does `@compiled/jest` not work?

madou: Ignore for now still tweaking!

packages/react/src/__tests__/jest-matcher.test.tsx
```diff
@@ -1,13 +1,6 @@
 import React from 'react';
 import { render } from '@testing-library/react';
 import { styled } from '@compiled/react';
-import { toHaveCompiledCss } from '../matchers';
```
madou, line 4: Had to move this into the react package has it was causing a cyclic dependency and TS project references weren't happy.

.github/workflows/build.yml
```diff
@@ -49,3 +46,6 @@ jobs:
 
       - name: Build parcel example
         run: yarn build:parcel
+
+      - name: Compare committed flow types to generated
```
madou, line 50: Moved this here so it's the last thing to run, as I wanted to get feedback for everything else first.

.github/workflows/test.yml
```diff
@@ -57,3 +48,12 @@ jobs:
           yarn bundlesize
           export GITHUB_RUN_NUMBER=GITHUB_RUN_NUMBER_TMP
           echo "$GITHUB_RUN_NUMBER"
+
+      - name: Build source for remainder tests
```
madou, line 52: Moved tests that are dependent on built source here. Might be worth splitting out to another workflow.

examples/packages/parcel-resolver-alias/tsconfig.json
```diff
@@ -6,5 +6,5 @@
     "module": "commonjs"
   },
   "exclude": ["dist"],
-  "references": []
+  "references": [{ "path": "../../../packages/jest" }]
```
madou, line 9: Because of type checking now being enabled in all projects pretty much every project needs to reference `jest` because of the global matcher.

jest.config.json
```diff
@@ -1,6 +1,7 @@
 {
   "moduleNameMapper": {
-    "^@compiled/react$": "<rootDir>/packages/react/src/index.tsx"
+    "^@compiled/react/runtime$": "<rootDir>/packages/react/src/runtime.tsx",
```
madou, line 3: Needed so Jest would pick up the internal modules (that aren't built).

package.json
```diff
@@ -22,26 +22,29 @@
     "build:webpack": "yarn build:browser && yarn build:examples-babel && ttsc --build examples/packages/webpack/tsconfig.json && yarn workspace @compiled/webpack-app build",
     "build:webpack:extract": "yarn build:browser && yarn build:examples-babel && ttsc --build examples/packages/webpack/tsconfig.json && cd examples/packages/webpack && EXTRACT_TO_CSS=true yarn build",
     "bundlesize": "yarn build && size-limit",
-    "clean": "rm -rf node_modules/.cache && find . -name *.tsbuildinfo -type f -delete",
+    "clean": "yarn clean:node-modules && yarn clean:ts-cache && yarn clean:dist",
```
madou, line 25: Separated clean into four commands, added back dist, fixes up the ts cache one as it needed quotations.

package.json
```diff
@@ -22,26 +22,29 @@
...
+    "clean:ts-cache": "find . -name \"*.tsbuildinfo\" -type f -delete",
     "flow-types": "scripts/flow-types.sh",
     "postinstall": "npx yarn-deduplicate && yarn --ignore-scripts",
     "lint": "eslint --config .eslintrc.js --ext flow,json,ts,tsx .",
     "lint:fix": "yarn lint -- --fix",
     "prettier:check": "prettier ./ --check",
     "prettier:fix": "prettier ./ --write",
     "release": "yarn clean && yarn build && yarn changeset publish",
-    "start": "npx nodemon --exec \"yarn build && start-storybook -p 6006 --ci\" --watch packages/babel-plugin/ -e tsx",
+    "start": "npx nodemon --exec \"start-storybook -p 6006 --ci\" --watch packages/babel-plugin/ -e tsx",
```
madou, line 36: Removed the need for building source! Happy days! Example running in storybook locally:

https://user-images.githubusercontent.com/6801309/142385801-f6aedf31-ed7f-40e5-8582-112bbe0712cd.mov

package.json
```diff
@@ -22,26 +22,29 @@
...
     "start:cli": "cd packages/cli && yarn start",
     "start:inspect": "npx nodemon --exec \"node --inspect-brk node_modules/.bin/start-storybook -p 6006 --ci\" --watch packages/babel-plugin/ -e tsx",
     "start:parcel": "yarn build:browser && yarn build:esm && ttsc --build examples/packages/parcel/tsconfig.json && yarn workspace @compiled/parcel-app start",
     "start:prod": "NODE_ENV=production yarn start",
     "start:ssr": "yarn build && cd examples/packages/ssr && yarn start",
     "start:webpack": "yarn build:browser && yarn build:examples-babel && ttsc --build examples/packages/webpack/tsconfig.json && yarn workspace @compiled/webpack-app start",
     "start:webpack:extract": "yarn build:browser && yarn build:examples-babel && ttsc --build examples/packages/webpack/tsconfig.json && cd examples/packages/webpack && EXTRACT_TO_CSS=true yarn start",
-    "test": "yarn build:cjs && yarn build:esm && jest --no-cache",
+    "test": "jest --no-cache",
```
madou, line 44: Removed the need for building when running tests! This means when you use any of Compiled's dependencies, anywhere in tests, you won't be stuck in that bizzaro world of needing to restart the test runner.

This doesn't help for the jest matcher or babel plugin though since the jest environment needs them to run - so we'll need to keep restarting them when developing them with the react package.

Example:

https://user-images.githubusercontent.com/6801309/142388009-7397c030-c563-47de-86ee-5d0cb394e4ac.mov

packages/babel-plugin-strip-runtime/index.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * This exists for local sessions so source files don't need to be built.
+ * This is an unfortunate hack if you're reading this and have a spare
+ * eng health week please find a better way to not have to do this.
+ */
+module.exports = require('./src');
```
madou, line 6: One of the worst parts of this fix. These are scattered around the repo, used for picking up entrypoints that aren't yet built. Each case of this has a pairing jsdoc with the explanation.

Would love to not need to do this if anyone has ideas!

packages/babel-plugin/index.js
```diff
@@ -0,0 +1,18 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const tsNode = require('ts-node');
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const tsPaths = require('tsconfig-paths');
+
+tsNode.register({
```
madou, line 6: Since the babel plugin is the first to be picked up its setting up tsnode + tsconfig paths. This is used both by storybook and by jest.

There could be a better way to set this up.

packages/babel-plugin/src/__tests__/index.test.tsx
```diff
@@ -17,7 +17,7 @@ describe('babel plugin', () => {
       babelOpts
     );
 
-    expect(output.code).toMatchInlineSnapshot(
+    expect(output?.code).toMatchInlineSnapshot(
```
madou, line 20: Because types are properly checking now I had to fix them up in a few tests.

packages/babel-plugin/src/utils/ast.tsx
```diff
@@ -594,6 +594,11 @@ export const resolveBindingNode = (
     }
 
     const moduleImportSource = binding.path.parentPath.node.source.value;
+    if (moduleImportSource.startsWith('@compiled')) {
```
madou, line 597: Tests were failing because Compiled was trying to traverse itself.

packages/cli/src/__tests__/main.test.tsx
```diff
@@ -62,9 +62,12 @@ describe('main', () => {
...
 
   it('should warn when invalid preset is provided using prompt', async () => {
-    castToJestMock(AutoComplete).mockImplementation(() => ({
-      run: () => Promise.resolve('invalid-using-prompt'),
-    }));
+    castToJestMock(AutoComplete).mockImplementation(
+      () =>
+        ({
+          run: () => Promise.resolve('invalid-using-prompt'),
+        } as any)
```
madou, line 69: These were implicit any - now explicit. If anyone has types for them LMK.

packages/cli/src/__tests__/main.test.tsx
```diff
@@ -110,11 +116,4 @@ describe('main', () => {
 
     expect(presets.mockedPreset).toHaveBeenCalledWith(cli);
   });
-
-  it('should find codemods in node modules', () => {
```
madou, line 114: No longer testing against dist!

packages/cli/src/presets/codemods/__tests__/codemods.test.tsx
```diff
@@ -21,7 +21,7 @@ const expectCodemodToHaveBeenRan = (name: string, runPath: string) => {
   expect(execAsync).toHaveBeenCalledWith(
     expect.stringMatching(
       new RegExp(
-        `.*--transform=.*node_modules\\/@compiled\\/codemods\\/dist\\/transforms\\/${name}\\/index.js ${regexPath}`
+        `.*--transform=.*node_modules\\/@compiled\\/codemods\\/src\\/transforms\\/${name}\\/index.tsx ${regexPath}`
```
madou, line 24: No longer building against dist.

packages/cli/src/presets/codemods/__tests__/codemods.test.tsx
```diff
@@ -37,22 +37,31 @@ const setupCliRunner = (opts: {
   codemodOpts?: CodemodOptions;
   pluginPaths?: string[];
 }) => {
-  castToJestMock(AutoComplete).mockImplementation(({ choices, result }) => ({
-    run: () => Promise.resolve(result(choices[opts.choice])),
-  }));
+  castToJestMock(AutoComplete).mockImplementation(
```
madou, line 40: Implicit anys again.

packages/webpack-loader/src/__tests__/compiled-loader.test.tsx
```diff
@@ -15,7 +15,7 @@ describe.each<'development' | 'production'>(['development', 'production'])(
         mode,
       });
 
-    it('does not transform files that do not contain @compiled/react', async () => {
+    it.concurrent('does not transform files that do not contain @compiled/react', async () => {
```
madou, line 18: Seemed to shave off some time overall so I added these in. It means it runs each test at the same time.

packages/webpack-loader/src/__tests__/extract-plugin.test.tsx
```diff
@@ -13,34 +13,44 @@ describe('CompiledExtractPlugin', () => {
...
+        disableExtractPlugin: true,
+      }).catch((err) => err);
+
+      expect(errors).toEqual([
+        expect.objectContaining({
+          message: expect.stringContaining("You forgot to add the 'CompiledExtractPlugin' plugin"),
+        }),
+      ]);
+    },
+    10000
```
madou, line 29: Had to bump the timeout as some tests were failing locally/in CI. It might be because webpack has to do more work now transpiling on the fly instead of going straight to source.

packages/webpack-loader/src/css-loader.tsx
```diff
@@ -14,10 +14,13 @@ export default function CSSLoader(this: LoaderContext<void>): string {
  * Moves CSSloader to the end of the loader queue so it runs first.
  */
 export function pitch(this: LoaderContext<void>): void {
-  if (this.loaders[0].path !== __filename) {
+  if (this.loaders[0].pitch !== pitch) {
```
madou, line 17: Now the entry point filename + path isn't always the same so I had to look for another solution - I leaned on referential equality to do it instead.

scripts/flow-types.sh
```diff
@@ -3,7 +3,7 @@ PACKAGES=({packages/react,packages/jest})
 
 generate() {
   echo "Generating flow types with flowgen"
-  find "${PACKAGES[@]}" -type f -path '*/dist/*' -name '*.d.ts' -not -path '*/cjs/*' -not -path '*/browser/*' -not -path '*/__fixtures__/*' -print0 | while read -rd $'\0' file; do
+  find "${PACKAGES[@]}" -type f -path '*/dist/*' -name '*.d.ts' -not -path '*/cjs/*' -not -path '*/browser/*' -not -path '*/__tests__/*' -not -path '*/__perf__/*' -not -path '*/__fixtures__/*' -print0 | while read -rd $'\0' file; do
```
madou, line 6: Excluded tests and perfs since they're now in the built dist (follow up task: remove them from node_modules).

tsconfig.paths.json
```diff
@@ -0,0 +1,15 @@
+{
+  "compilerOptions": {
+    "baseUrl": ".",
+    "paths": {
```
madou, line 4: These are needed to wire up all the packages so TypeScript can do a build.

package.json
```diff
@@ -10,7 +10,8 @@
   ],
   "scripts": {
     "benchmark": "yarn build:cjs && yarn build:esm && jest --config benchmark.config.json",
```
MonicaOlejniczak, line 12: Can you remove the `build` scripts from here too? You'll need to include `benchmark` in the `react` tsconfigs, and add some mappers in

packages/babel-plugin-strip-runtime/index.js
MonicaOlejniczak: I think you would need something like https://www.npmjs.com/package/babel-plugin-module-resolver to fix it

tsconfig.paths.json
MonicaOlejniczak: Do we need the `..` one's? Removing them still results in functional stories + tests for me

packages/codemods/src/transforms/styled-components-to-compiled/__tests__/styled-components-to-compiled.test.tsx
```diff
@@ -178,7 +178,7 @@ describe('styled-components-to-compiled transformer', () => {
         {
           create: (_: FileInfo) => ({
             transform: {
-              buildImport: ({ originalNode, currentNode }) => {
+              buildImport: ({ originalNode, currentNode }: any) => {
```
MonicaOlejniczak, line 181: Should just be `BuildImportContext`

packages/webpack-loader/tsconfig.json
```diff
@@ -10,6 +10,7 @@
     { "path": "../babel-plugin" },
     { "path": "../babel-plugin-strip-runtime" },
     { "path": "../css" },
-    { "path": "../utils" }
+    { "path": "../utils" },
+    { "path": "../react" }
```
MonicaOlejniczak, line 14: /nit react above utils

packages/tsconfig.options.json
```diff
@@ -11,12 +11,13 @@
     "moduleResolution": "node",
     "noImplicitReturns": true,
     "noImplicitThis": true,
+    "noImplicitAny": true,
```
MonicaOlejniczak, line 14: This should be enabled under strict, or is something else disabling it?

See https://www.typescriptlang.org/tsconfig#noImplicitAny

> Default:
true if strict,
false otherwise.

packages/webpack-loader/src/__tests__/extract-plugin.test.tsx
```diff
@@ -13,34 +13,44 @@ describe('CompiledExtractPlugin', () => {
...
+  );
+
+  it.concurrent(
+    'extracts local styles',
+    async () => {
+      const actual = await bundle(join(fixturesPath, 'local-styles.tsx'));
+
+      expect(actual[assetName]).toMatchInlineSnapshot(`
       "._1wybdlk8{font-size:14px}
```
MonicaOlejniczak, line 38: Can the indentation be fixed here?

packages/webpack-loader/index.js
```diff
@@ -0,0 +1,18 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
MonicaOlejniczak, line 1: You can remove this and the css entry if you add this below `resolve` in the `test-utils` file:

```
resolveLoader: {
  alias: {
    '@compiled/webpack-loader/css-loader': join(__dirname, '..', 'css-loader.tsx'),
    '@compiled/webpack-loader': join(__dirname, '..', 'index.tsx'),
  }
},
```

packages/babel-plugin-strip-runtime/index.js
madou: Yeah, wouldn't want to use it yet since TSC is used to build dist. Fodder for consolidating on a single build tool though.

tsconfig.paths.json
madou: Cool, will remove.

package.json
madou: Thanks will do that

packages/codemods/src/transforms/styled-components-to-compiled/__tests__/styled-components-to-compiled.test.tsx
madou: Thanks! What would the type be to pass in?

packages/tsconfig.options.json
madou: Nice catch! Removed.

packages/webpack-loader/index.js
madou: Did you have any luck trying this? I did it locally and couldn't get it working - I always assumed this was for aliasing during bundling not bootstrapping 🤔.

packages/codemods/src/transforms/styled-components-to-compiled/__tests__/styled-components-to-compiled.test.tsx
MonicaOlejniczak: I'm not actually sure, I have wanted to remove the `T` since it doesn't seem necessary to me anymore. Can you help out here @JakeLane since you initially wrote that `ValidateConfig` type?

madou:
@MonicaOlejniczak Made most of the changes :-) Let me know if there's anything that needs a fix up - else I'm happy to put up a follow PR too.

packages/webpack-loader/index.js
MonicaOlejniczak: Yeah it worked for me when I tried it out. Putting it in `resolve` didn't work, but `resolveLoader` was fine

MonicaOlejniczak (approved):
LGTM, any lingering issues can be addressed later

PR #925, opened 2021-11-18
Add json-files eslint plugin

MonicaOlejniczak:
This resolves the inconsistent sorting of `package.json` files by introducing a fixable `sort-package-json` lint rule.

## Changes
* Add `eslint-plugin-json-files` and corresponding lint rule
* Fix flow linting
* Refactor `.eslintrc.js` to use overrides for specific file types
* Update `lint` script to include more file types, and run it over the root directory


madou (approved):

package.json
```diff
@@ -25,7 +25,7 @@
     "clean": "rm -rf node_modules/.cache && find . -name *.tsbuildinfo -type f -delete",
     "flow-types": "scripts/flow-types.sh",
     "postinstall": "npx yarn-deduplicate && yarn --ignore-scripts",
-    "lint": "eslint --config .eslintrc.js --ext tsx,ts ./packages/**/src ./examples",
```
at-nathan, line 28: Are we removing `./packages/**/src ./examples` because it's already included as part of `.`?

at-nathan (approved):

package.json
MonicaOlejniczak: Yep, that's right. I added in some excludes so it doesn't pick up too much either.

PR #927, opened 2021-11-19
Split out ast-builders into multiple utils

MonicaOlejniczak:
## Changes
* Add `getJSXAttribute` util
* Create `Variable` type
* Extract `buildCompiledComponent`, `buildDisplayName`, `compiledTemplate`, and `hoistSheet` to separate utils
* Move `importSpecifier` to `append-runtime-imports.ts`
* Rename `buildCssVariablesProp` to `buildCssVariables` and extract it to a separate util
* Rename `transformItemCss` to `transformCssItems` and extract it to a separate util


packages/babel-plugin/src/styled/index.tsx
```diff
@@ -1,9 +1,11 @@
 import * as t from '@babel/types';
 import type { NodePath } from '@babel/core';
+
+import type { Metadata, Tag } from '../types';
```
madou, line 4: Can we turn a lint rule on for imports?

MonicaOlejniczak: For sorting and grouping? I was planning to, otherwise not sure what you're referring to

madou: Yeah exactly

madou (dismissed):
LGTM

Issue #836, opened 2021-09-27
Template string literals are broken unless the API is enabled in a different usage


JakeLane:
**Describe the bug**

**To Reproduce**
The following code _does not_ generate blue text:
```
import '@compiled/react';

export const TemplateLiteral = (): JSX.Element => {
  return (
    <div
      css={`
        display: flex;
        font-size: 30px;
        color: blue;
      `}>
      blue text
    </div>
  );
};
```

The following code _does_ generate blue text:
```
import { css } from '@compiled/react';

const temp = css`
  display: flex;
  font-size: 30pfdsafdsalfdas;x;
  color: blue;
`;
console.log(temp);

export const TemplateLiteral = (): JSX.Element => {
  return (
    <div
      css={`
        display: flex;
        font-size: 30px;
        color: blue;
      `}>
      blue text
    </div>
  );
};
```

**Expected behaviour**
In the first example, the import to `@compiled/react` should mean that template literal strings in the `css` prop will be converted to Compiled styles. This will only happen if the `css` API is not enabled in the same file by a different usage.

madeleineostoja:
Kinda random aside but why not just force usage of the css tagged literal in css props? That’s what Emotion does and it leads imo to more foolproof code

PR #928, opened 2021-11-19
Local jsx namespace

madou:
This pull request replaces the global JSX namespace augmentation with a local JSX namespace for the `css` prop. This is a breaking type change for TS users (if you aren't using TypeScript this change will not affect you) - however we would update our docs to use jsx pragma for consistency either way.

The benefits of this are:

- No longer polluting the global namespace
- Collisions in the global namespace are now impossible
- Having deeper more integrated types are now possible, for example it is now a type error when using CSS prop on a JSX element that doesn't take a className prop. We can extend this later on for tokens etc.

**Before**

```tsx
import '@compiled/react';

function Component(props: { children: string }) {
  return <div />
}

<Component css={{ color: 'red' }} /> // no error
```

**After (classic runtime)**

```tsx
/** @jsx jsx */
import { jsx } from '@compiled/react';

<Component css={{ color: 'red' }} /> 
           ^^^ error css prop doesn't exist
```

**After (automatic runtime)**

```tsx
/** @jsxImportSource @compiled/react */

<Component css={{ color: 'red' }} />
           ^^^ error css prop doesn't exist
```

**TODO**

- [x] Fix flow type errors
- [x] Add a changeset

**Follow up**

-  [x] Update lint rule to place JSX pragma (Done: https://github.com/atlassian-labs/compiled/pull/933 https://github.com/atlassian-labs/compiled/pull/932)
- [ ] Update codemod to place JSX pragma
- [ ] Fix type errors in examples (and see how we can typecheck them!)


.changeset/silver-dots-reply.md
```diff
@@ -0,0 +1,42 @@
+---
+'@compiled/react': minor
+---
+
+**Breaking change:** When using the `css` prop with [TypeScript](https://www.typescriptlang.org) you now need to declare a jsx pragma enabling types for that module. Previously when importing the `@compiled/react` package the global JSX namespace would be polluted as a side effect potentially causing collisions with other CSS-in-JS libraries. Now thanks to the use of [locally scoped JSX namespaces](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#locally-scoped-jsx-namespaces) the global JSX namespace will no longer be polluted.
```
DarkPurple141, line 5: jsx or JSX? Should we be consistent?

packages/babel-plugin-strip-runtime/src/__tests__/strip-runtime-source-code.test.ts
```diff
@@ -96,8 +96,8 @@ describe('babel-plugin-strip-runtime using source code', () => {
         expect(actual).toMatchInlineSnapshot(`
           "/* app.tsx generated by @compiled/babel-plugin v0.0.0 */
 
-          import * as React from 'react';
           import { ax, ix } from '@compiled/react/runtime';
+          import * as React from 'react';
```
DarkPurple141, line 100: any reason these are no longer top imports?

packages/babel-plugin/src/types.tsx
```diff
@@ -65,6 +65,14 @@ export interface State extends PluginPass {
     styled?: string;
   };
 
+  /**
+   * Details of pragmas that are currently enabled in the pass.
+   */
+  pragma: {
+    jsxImportSource?: boolean;
```
DarkPurple141, line 72: can both be enabled at the same time?

packages/babel-plugin/src/utils/append-runtime-imports.ts
```diff
@@ -13,50 +13,55 @@ const importSpecifier = (name: string, localName?: string): t.ImportSpecifier =>
...
     });
   } else {
     // Add the runtime entrypoint module
-    path.insertBefore(
+    path.unshiftContainer(
+      'body',
       t.importDeclaration(
-        runtimeImportNames.map((runtimeImportName) => importSpecifier(runtimeImportName)),
-        t.stringLiteral(runtimeImportModuleName)
+        COMPILED_RUNTIME_IMPORTS.map((runtimeImportName) => importSpecifier(runtimeImportName)),
```
DarkPurple141, line 63: nit: doesn't need an inline func, can just be `map(importSpecifier)`

DarkPurple141 (approved):

.changeset/silver-dots-reply.md
madou: JSX

packages/babel-plugin-strip-runtime/src/__tests__/strip-runtime-source-code.test.ts
madou: Oh, I can fix that. Sec.

packages/babel-plugin/src/types.tsx
madou: Technically yes, but Babel or any bundler would throw

packages/babel-plugin/src/babel-plugin.tsx
```diff
@@ -48,19 +51,45 @@ export default declare<State>((api) => {
...
+        enter(_, state) {
+          const { file } = state;
+
+          if (file.ast.comments) {
+            for (const comment of file.ast.comments) {
+              const jsxSourceMatches = JSX_SOURCE_ANNOTATION_REGEX.exec(comment.value);
+              const jsxMatches = JSX_ANNOTATION_REGEX.exec(comment.value);
+
+              if (jsxSourceMatches && jsxSourceMatches[1] === COMPILED_MODULE) {
+                // jsxImportSource pragma found - turn on CSS prop!
```
JakeLane, line 67: This may resolve https://github.com/atlassian-labs/compiled/issues/836

JakeLane (approved):
awesome work!

packages/webpack-loader/src/__fixtures__/compiled-error.tsx
```diff
@@ -1,5 +1,5 @@
-import { css } from '@compiled/react';
-import React from 'react';
+/** @jsx jsx */
+import { jsx, css } from '@compiled/react';
```
MonicaOlejniczak, line 2: The order doesn't get autofixed? Bad rule

packages/webpack-loader/src/__fixtures__/local-styles.tsx
```diff
@@ -1,13 +1,14 @@
...
+import { jsx, styled } from '@compiled/react';
+import { Fragment } from 'react';
 
 const Styled = styled.div({
   color: 'blue',
 });
 
 export const App = (): JSX.Element => (
-  <>
+  <Fragment>
```
MonicaOlejniczak, line 10: Does `<>` not work with the `jsx`?

tsconfig.paths.json
```diff
@@ -7,6 +7,8 @@
       "@compiled/jest": ["packages/jest/src/index.tsx"],
       "@compiled/react": ["packages/react/src/index.tsx"],
       "@compiled/react/runtime": ["packages/react/src/runtime.tsx"],
+      "@compiled/react/jsx-runtime": ["packages/react/src/jsx/jsx-runtime.tsx"],
```
MonicaOlejniczak, line 10: /nit these two should be above `react/runtime`

packages/babel-plugin/src/babel-plugin.tsx
MonicaOlejniczak: I don't think it will, but haven't dug deeply into it

packages/babel-plugin/src/css-prop/__tests__/jsx-pragma.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+  return transformSync(code, {
+    configFile: false,
+    babelrc: false,
+    plugins: [babelPlugin],
+  })?.code;
+};
+
+describe('local jsx namespace', () => {
+  it('should transform css prop using jsx import source', () => {
+    const result = transform(`
```
MonicaOlejniczak, line 15: `result` to `actual` to match other test files

packages/babel-plugin/src/css-prop/__tests__/jsx-pragma.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <CC>
+          <CS>{[_]}</CS>
+          {<div className={ax([\\"_syaz5scu\\"])}>hello</div>}
+        </CC>;"
+    `);
+  });
+
+  it('should transform css prop using jsx pragma', () => {
+    const result = transform(`
+      /** @jsxRuntime classic */
```
MonicaOlejniczak, line 37: What happens with edge cases, should you test those too? i.e.`jsxRuntime automatic` with `@jsx jsx` and `@jsxImportSource` with the runtime specified, etc

stories/jsx-pragma.tsx
```diff
@@ -0,0 +1,40 @@
+/** @jsxRuntime classic */
+/** @jsx jsx */
+// TypeScript thinks we're using the automatic runtime however we've forced Babel
```
MonicaOlejniczak, line 3: How does babel handle `@jsxRuntime` exactly? I don't see any code references to it, and am unsure of where the magic comes from

stories/jsx-import-source-pragma.tsx
```diff
@@ -0,0 +1,35 @@
+/** @jsxImportSource @compiled/react */
+
+export default {
```
MonicaOlejniczak, line 3: Should these be stories or type tests? There are type tests for flow, you could use a similar format for typescript.

packages/react/src/__tests__/jest-matcher.test.tsx
```diff
@@ -1,6 +1,7 @@
-import { styled } from '@compiled/react';
+/** @jsxRuntime classic */
+/** @jsx jsx */
+import { jsx, styled } from '@compiled/react';
```
MonicaOlejniczak, line 3: Is there a way of enabling local namespaces without importing `jsx` and just using the pragma?

at-nathan (approved):

packages/react/src/__tests__/jest-matcher.test.tsx
```diff
@@ -1,6 +1,7 @@
-import { styled } from '@compiled/react';
+/** @jsxRuntime classic */
```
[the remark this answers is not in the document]
madou: Because the runtime is actually automatic, this forces Babel to use the classic runtime :-)

packages/webpack-loader/src/__fixtures__/local-styles.tsx
madou: Yeah, exactly. Unless you specify the fragment pragma.

packages/babel-plugin/src/css-prop/__tests__/jsx-pragma.test.tsx
madou: Build tools (Babel) throw - you can't define both.

stories/jsx-pragma.tsx
madou: It forces the runtime to be whatever is declared (classic or runtime). It's a babel thing unfortunately not supported by TS. Useful for testing :)

stories/jsx-import-source-pragma.tsx
madou: Both - there are tests testing these flows. They are also stories so it makes sure the runtime is actually set (because it imports from `@compiled/react/jsx-runtime`).

packages/react/src/__tests__/jest-matcher.test.tsx
madou: Using `jsxImportSource` and the automatic runtime 👍

packages/babel-plugin/src/utils/append-runtime-imports.ts
```diff
@@ -13,50 +13,55 @@ const importSpecifier = (name: string, localName?: string): t.ImportSpecifier =>
...
-    .getAllPrevSiblings()
-    .concat(path.getAllNextSiblings())
-    .find(
-      (path) =>
-        t.isImportDeclaration(path.node) && path.node.source.value === runtimeImportModuleName
-    ) as NodePath<t.ImportDeclaration> | undefined;
+  const previouslyDeclaredRuntimeDeclaration = path
+    .get('body')
+    .find((childPath): childPath is NodePath<t.ImportDeclaration> => {
+      return (
```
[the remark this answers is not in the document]
madou: I don't think there is a set standard, opportunity for a lint rule 😄

packages/react/src/__tests__/jest-matcher.test.tsx
MonicaOlejniczak: Yeah, so can we do that for this test?

stories/jsx-pragma.tsx
MonicaOlejniczak: Cool, found it here: https://babeljs.io/docs/en/babel-plugin-transform-react-jsx

packages/babel-plugin/src/css-prop/__tests__/jsx-pragma.test.tsx
MonicaOlejniczak: So would you say that it's not worth having tests for these cases, because it's testing the external babel behaviour?

stories/jsx-pragma.tsx
madou: Oh! Sorry should have gotten it for you.

JakeLane (dismissed):

packages/babel-plugin/src/css-prop/__tests__/jsx-pragma.test.tsx
madou: It will throw when defining both - but still good to know what happens in a test (and if its behaviour changes in the future) - I'll add one an expect an exception.

packages/react/src/__tests__/jest-matcher.test.tsx
madou: Ah nice, sorted

packages/babel-plugin/src/css-prop/__tests__/jsx-pragma.test.tsx
madou: Ah, it will throw when transforming the JSX elements with that plugin, however it won't throw in ours. I don't believe it's worth it to double up, what do you think?

MonicaOlejniczak: Yeah, all good to leave it as is

MonicaOlejniczak (approved):

PR #932, opened 2021-11-20
Refactor emotion-to-compiled ESLint rule

madou:
This pull request:

- Renames `emotion-to-compiled` rule to `no-emotion-css`
- Adds a violation for using `jsxImportSource` pragma with Emotion
- Updates rule to keep `jsx` pragma if used previously with Emotion


DarkPurple141 (approved):
lgtm

PR #933, opened 2021-11-20
feat: adds jsx-pragma rule

madou:
This pull request adds a new ESLint rule `jsx-pragma`. It is related to the #928 change which introduces a local JSX namespce in TypeScript - forcing TypeScript consumers to use the JSX pragma to enable local types instead of global types which can cause collisions.

This rule also works for regular development so if you use a css prop and you forget to enable the pragma - it will cause a violation and tell you to do something about it.

The configuration is set to `runtime` and defaults to `automatic`.  The name was chosen as it seems to be common place among dev tooling (context: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html).


packages/eslint-plugin/src/rules/jsx-pragma/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+          },
+        },
+        additionalProperties: false,
+      },
+    ],
+  },
+  create(context) {
+    const options: Options = context.options[0] || { pragma: 'jsxImportSource' };
+    const source = context.getSourceCode();
+    if (source.text.indexOf('css={') === -1) {
```
madou, line 36: 🤔 Wanted to bail early if the source code didn't include a css prop. Could there be a better way to check this?

JakeLane (dismissed):

packages/eslint-plugin/src/index.tsx
```diff
@@ -1,7 +1,9 @@
+import jsxPragma from './rules/jsx-pragma';
 import noEmotionCSS from './rules/no-emotion-css';
 
 export const rules = {
   'no-emotion-css': noEmotionCSS,
+  'jsx-pragma': jsxPragma,
```
MonicaOlejniczak, line 6: /nit add this before `no-emotion-css`

.changeset/six-poets-love.md
```diff
@@ -0,0 +1,5 @@
+---
+'@compiled/eslint-plugin': patch
+---
+
+Adds `jsx-pragma` rule.
```
MonicaOlejniczak, line 5: /nit I prefer no `.` for short, standalone sentences

packages/eslint-plugin/src/rules/jsx-pragma/README.md
```diff
@@ -0,0 +1,48 @@
+# @compiled/eslint-plugin/jsx-pragma
+
+This rule ensures a jsx pragma is used when using CSS prop.
```
MonicaOlejniczak, line 3: `when using CSS prop.` > ``when using the `css` prop.``

packages/eslint-plugin/src/rules/jsx-pragma/README.md
```diff
@@ -0,0 +1,48 @@
...
+You can configure which jsx pragma to use via the `runtime` option (either classic or automatic).
```
MonicaOlejniczak, line 4: Can this be under an `## Options` heading, it's pretty typical in eslint rules. The format normally has each option under its own heading like this https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md

packages/eslint-plugin/src/rules/jsx-pragma/__tests__/rule.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+      `,
+      options: [{ runtime: 'classic' }],
+      errors: [
+        {
+          messageId: 'preferJsx',
+        },
+      ],
+    },
+    {
+      code: `
```
MonicaOlejniczak, line 75: Can this indentation be fixed?

packages/eslint-plugin/src/rules/jsx-pragma/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+type Options = {
+  runtime: 'classic' | 'automatic';
+};
+
+const rule: Rule.RuleModule = {
+  meta: {
+    fixable: 'code',
+    type: 'problem',
+    messages: {
+      missingPragma: 'To use CSS prop you must set the {{ pragma }} pragma.',
```
MonicaOlejniczak, line 15: `To use CSS prop you` to ``To use the `css` prop, you``

packages/eslint-plugin/src/rules/no-emotion-css/index.tsx
```diff
@@ -1,5 +1,7 @@
 import type { Rule } from 'eslint';
-import type { ImportDeclaration, ImportSpecifier } from 'estree';
+import type { ImportSpecifier, ImportDeclaration } from 'estree';
```
MonicaOlejniczak, line 2: Can you merge in latest changes which include the import order?

packages/eslint-plugin/src/utils/ast-string.tsx
```diff
@@ -0,0 +1,63 @@
+import type {
```
MonicaOlejniczak, line 1: Can you rename this from `ast-string` to `ast-to-string`? It does highlight that `buildImportDeclaration` doesn't quite belong in this file either given its input parameter

packages/eslint-plugin/src/rules/jsx-pragma/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+          },
+        },
+        additionalProperties: false,
+      },
+    ],
+  },
+  create(context) {
+    const options: Options = context.options[0] || { runtime: 'automatic' };
+    const source = context.getSourceCode();
+    const jsxPragma = source.getAllComments().find((n) => n.value.indexOf('@jsx jsx') > -1);
```
MonicaOlejniczak, line 37: Can we store comments in a variable? I assume it's a little bit expensive

packages/eslint-plugin/src/rules/no-emotion-css/index.tsx
madou: I did 😄 This was fixed by the rule.

MonicaOlejniczak: Oh really? Okay, sometimes the sorting behaves weirdly

packages/eslint-plugin/src/rules/jsx-pragma/index.tsx
DarkPurple141: Not sure what you mean?

DarkPurple141: Oh I see. Yes next line. @madou use a variable for comments.

DarkPurple141 (approved):
LGTM

packages/eslint-plugin/src/rules/jsx-pragma/README.md
madou: Thanks will look at the examples

packages/eslint-plugin/src/rules/no-emotion-css/index.tsx
madou: Done

.changeset/six-poets-love.md
```diff
@@ -0,0 +1,7 @@
+---
+'@compiled/eslint-plugin': patch
+---
+
+Adds `jsx-pragma` rule,
```
MonicaOlejniczak, line 5: Formatting is a bit weird here?

MonicaOlejniczak (approved):

.changeset/six-poets-love.md
madou: This is a personal preference 😄 - you can have sentences that end with a comma or full stop and start the next work on a new line and it stays within the paragraph. I like this because it makes diffs and reviewing easier than reading a really long sentence left to right

PR #935, opened 2021-11-22
Lint js and jsx files

MonicaOlejniczak:
## Changes
* Fix eslint violations
* Target `js` and `jsx` files in root `eslint` script
* Update eslint config to share common rules in the first override, and rely on override merging for applying additional rules


madou (approved):
LGTM

PR #939, opened 2021-11-22
Fix import order issues for changset script and update eslint excludes

JakeLane:
Fix red master

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,6 +6,7 @@ module.exports = {
     '*.d.ts',
     'babel-cjs.js',
     'babel-esm.js',
+    'storybook-static'
   ],
   overrides: [
     {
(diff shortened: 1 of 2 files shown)


.eslintrc.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
     '*.d.ts',
     'babel-cjs.js',
     'babel-esm.js',
+    'storybook-static'
```
madou, line 9: Oh, sick. Does this speed things up?

madou (approved):
LGTM

jackbrown (approved):

.eslintrc.js
JakeLane: ye a fair bit - also caused local failures for me

PR #944, opened 2021-11-23
Simplify browser and examples build

MonicaOlejniczak:
## Changes
* Add missing project references to `@compiled/parcel-app` and `@compiled/webpack-app`
* Add `tsconfig.json` to `@compiled/ssr-app`
* Reduce the `CC` and `CS` browser bundle size limit to the current size
* Update `@compiled/react` tsconfig files
  * Make each configuration standalone
  * Apply the transform to the `browser` configuration
  * Hardcode `isNodeEnvironment()` to false
  * Remove `IS_NODE_EXPRESSION='false'` from the `build:browser` script
* Update example app scripts to no longer depend on `build:browser`, and only build what is required
* Simplify `build:webpack:extract` and `start:webpack:extract` scripts


madou (approved):
LGTM. Nice job! Do the example apps run as expected? And now we can modify packages without building source? 😃

MonicaOlejniczak:
@madou They sure do! I built and ran them all from a clean root directory.

Yeah, you only build the packages that you need. The example apps never listened to changes in the compiled packages (such as the babel plugin), so hot reloading will only work if you modify the example app directly. Otherwise, you'd need to rerun the command. I think if we want to get that part working better, we'd need to decide on a bundler and integrate through that.

PR #947, opened 2021-11-24
Update pkg jsdoc/json/rule

madou:
This pull request:

- updates the jsdoc descriptions for the `@compiled/react` package 
- removes redundant files in `files` field in pkg json in most packages
- updates homepage in pkg json for most packages
- updates the type signature of the class names component style prop
- adds url to the lint rules

### DIFF
--- a/.changeset/pink-apples-appear.md
+++ b/.changeset/pink-apples-appear.md
@@ -12,4 +12,4 @@
 '@compiled/webpack-loader': patch
 ---
 
-Package jsons have had their homepage and other properties updated.
+Update `homepage` and other `package.json` properties
--- a/.changeset/pink-apples-appear.md
+++ b/.changeset/pink-apples-appear.md
@@ -1,0 +1,15 @@
+---
+'@compiled/babel-plugin': patch
+'@compiled/babel-plugin-strip-runtime': patch
+'@compiled/cli': patch
+'@compiled/codemods': patch
+'@compiled/css': patch
+'@compiled/eslint-plugin': patch
+'@compiled/jest': patch
+'@compiled/parcel-transformer': patch
+'@compiled/react': patch
+'@compiled/utils': patch
+'@compiled/webpack-loader': patch
+---
+
+Package jsons have had their homepage and other properties updated.
--- a/.changeset/selfish-hounds-do.md
+++ b/.changeset/selfish-hounds-do.md
@@ -2,4 +2,4 @@
 '@compiled/react': patch
 ---
 
-Class name component now has its style prop passed through children as function typed as `CSSProperties` instead of a string map.
+`ClassNames` component now has its style prop typed as `CSSProperties` instead of a string map
--- a/.changeset/selfish-hounds-do.md
+++ b/.changeset/selfish-hounds-do.md
@@ -1,0 +1,5 @@
+---
+'@compiled/react': patch
+---
+
+Class name component now has its style prop passed through children as function typed as `CSSProperties` instead of a string map.
--- a/packages/codemods/package.json
+++ b/packages/codemods/package.json
@@ -1,14 +1,14 @@
 {
   "name": "@compiled/codemods",
   "version": "0.3.2",
-  "description": "Codemods (with JSCodeshift) for Compiled CSS-in-JS.",
+  "description": "A familiar and performant compile time CSS-in-JS library for React.",
   "keywords": [
     "codemod",
     "compiled",
     "css-in-js",
     "typescript"
   ],
-  "homepage": "https://compiledcssinjs.com",
+  "homepage": "https://compiledcssinjs.com/docs/pkg-cli",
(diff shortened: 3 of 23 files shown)


.changeset/pink-apples-appear.md
```diff
@@ -0,0 +1,15 @@
...
+'@compiled/css': patch
+'@compiled/eslint-plugin': patch
+'@compiled/jest': patch
+'@compiled/parcel-transformer': patch
+'@compiled/react': patch
+'@compiled/utils': patch
+'@compiled/webpack-loader': patch
+---
+
+Package jsons have had their homepage and other properties updated.
```
MonicaOlejniczak, line 15: /nit ``Update `homepage` and other `package.json` properties``

JakeLane (dismissed):

packages/codemods/package.json
```diff
@@ -1,14 +1,14 @@
 {
   "name": "@compiled/codemods",
   "version": "0.3.2",
-  "description": "Codemods (with JSCodeshift) for Compiled CSS-in-JS.",
+  "description": "A familiar and performant compile time CSS-in-JS library for React.",
```
MonicaOlejniczak, line 4: Shouldn't the description be specific to the package?

packages/eslint-plugin/src/rules/no-emotion-css/README.md
```diff
@@ -1,11 +1,9 @@
 # @compiled/eslint-plugin/emotion-to-compiled
 
 Ensures usage of the `@compiled/react` library over `@emotion/core` or `@emotion/styled`.
-This rule acts as a form of codemod.
+The `--fix` option [on the command line] automatically fixes problems reported by this rule.
 
-## Examples
-
-👎 Example of **incorrect** code for this rule:
+## Fail
```
MonicaOlejniczak, line 6: How come the change of style here? I don't think there is any general recommendation though, but I normally see the thumbs up / down approach

.changeset/selfish-hounds-do.md
```diff
@@ -0,0 +1,5 @@
+---
+'@compiled/react': patch
+---
+
+Class name component now has its style prop passed through children as function typed as `CSSProperties` instead of a string map.
```
MonicaOlejniczak, line 5: I would just be more concise, and use the component name directly:

`` `ClassNames` component now has its style prop typed as `CSSProperties` instead of a string map``

packages/react/src/styled/index.tsx
```diff
@@ -58,17 +58,45 @@ export interface StyledComponentInstantiator extends StyledComponentMap {
 }
 
 /**
- * Create a component that ties styles to an element which comes with built-in behavior such as `ref` and `as` prop support.
+ * ## Styled component
```
MonicaOlejniczak, line 61: Looks way better 👍

packages/react/src/css/index.tsx
```diff
@@ -9,21 +9,36 @@ import type {
 import { createSetupError } from '../utils/error';
 
 /**
- * Create styles that can be re-used between components with a template literal.
+ * ## CSS
```
MonicaOlejniczak, line 12: We should probably move the jsdoc to be associated with the bottom export, as this function is specific to the tagged template usage. Same applies to keyframes. Though I'm not sure what it will look like in IDEs.

packages/react/src/class-names/index.tsx
```diff
@@ -8,36 +8,46 @@ export type Interpolations = (BasicTemplateInterpolations | CssFunction | CssFun
...
  * </ClassNames>
  * ```
- *
- * For more help, read the docs:
- * https://compiledcssinjs.com/docs/api-class-names
- *
- * @param props
  */
-export function ClassNames(_: ClassNamesProps): JSX.Element {
+export function ClassNames({ children }: ClassNamesProps): JSX.Element;
```
MonicaOlejniczak, line 49: Do we need both definitions?

packages/codemods/package.json
madou: Probably - but this is the outlier ATM. Every other package uses the same description.

packages/eslint-plugin/src/rules/no-emotion-css/README.md
madou: It's consistent with the other Compiled rule README now (previously they were different) - it also matches more closely to the official ESLint rule READMEs.

packages/react/src/styled/index.tsx
madou: Thanks!

packages/react/src/class-names/index.tsx
madou: This makes the tooltip a bit more friendly

image

packages/react/src/css/index.tsx
madou: Ah it looks like it needs to be on the very top one else IDE's don't pick them up.

MonicaOlejniczak (approved):

Issue #407, opened 2020-12-15
Userland destructuring should follow the same behaviour as member expressions


pgmanutd:
Outcome of #403 

Right now following works for `MemberExpression` (Implemented [here](https://github.com/atlassian-labs/compiled/blob/master/packages/babel-plugin/src/utils/ast-builders.tsx#L179))

```ts
styled.div`
  color: ${props => props.color}; // added to HTML element
  background-color: ${props => props.bgColor}; // NOT added to HTML element
`;
```

It doesn't work for `Identifier` as of now which is coming from userland:

```ts
styled.div`
  color: ${({color}) => color}; // should be added to HTML element
  background-color: ${({$color}) => $color}; // should NOT added to HTML element
`;
```

PR #954, opened 2021-11-29
RFC: Rethinking how Compiled handles AST traversal/transformation

madou:
This pull request adds an RFC for modifying how Compiled generates styles at build time, encouraging Compiled to do what it does best - compile styles! While leaving room for the runtime to put things back together.

### DIFF
--- a/rfcs/0000-compiled-runtime.md
+++ b/rfcs/0000-compiled-runtime.md
@@ -47,7 +47,7 @@ Today Compiled does the following:
 
 This is a lot of steps and prone to running into edge cases unfortunately. But it does work! And allowed the initial architecture to work with any unmarked style rules. If we were to enforce that all style rules **must** be declared using the `css` function things get a little more interesting.
 
-Taking the same example above except styles are now defined through `css` and return a data structure with compiled styles.
+If we take the example above, ensure all styles flow through a `css` call site, and then compile each call site in place to a data structure Compiled can work with, things get _really_ interesting.
 
 ```diff
 /** @jsxImportSource @compiled/react */
@@ -69,7 +69,7 @@ Compiled would:
 2. Generate class name and styles
 3. Lean on the runtime to do the rest
 
-You can see a basic implementation of the above here: https://github.com/atlassian-labs/compiled/compare/babel-experiment?expand=1#diff-7c3b8f72b34155cdf17c42f282ed0cecec4fd2863ba71912213f8932159cd01bR17
+You can see a basic implementation of the above on the [`babel-experiment` branch](https://github.com/atlassian-labs/compiled/compare/babel-experiment?expand=1#diff-7c3b8f72b34155cdf17c42f282ed0cecec4fd2863ba71912213f8932159cd01bR17). Note only tests work, Storybook is broken.
 
 ## Motivation
 
@@ -77,12 +77,12 @@ The primary motivation is for Compiled maintainers to maintain and write less co
 
 Other motivating factors:
 
-- Reduce AST traveral
-- Enable very dynamic conditional styles to work
+- Reduce AST traveral (improved build performance)
+- Enable very dynamic conditional styles to work (improved DX)
 
 ## Detail design
 
-This change should work with all Compiled APIs and ideally be backwards compatible if possible.
+This change should work with all Compiled APIs and ideally be backwards compatible if possible, at least for a while.
 
 ### CSS
 
@@ -126,7 +126,7 @@ function Comp() {
 }
 ```
 
-When referencing an expression Compiled would try to statically evaluate it, if it fails it would be added as the third item in the array for all dynamic items. This behaviour can be re-used for truly dynamic styles.
+When referencing an expression Compiled would try to statically evaluate it, if it fails it would be added as the third item in the array along with all other dynamic styles.
 
 ```jsx
 import { N800 } from '@atlaskit/theme/colors';
@@ -138,7 +138,7 @@ const styles = css({ color: N800 });
 const styles = i(['_syaz4rde', ['._syaz4rde{color: var(--_kmurgp)}'], [['--_kmurgp', N800]]]);
 ```
 
-When composing style rules together in an array they would naturally work without extra AST transformation.
+When composing style rules together in an array they would naturally work **without extra AST transformation**.
 
 ```jsx
 const colorStyles = css({
@@ -164,15 +164,16 @@ const styles = i([colorStyles, hoverStyles]);
 
 **Considerations**
 
-- What would the best data structure be for these styles, what would encourage the fastest HOT path?
+- Could there be a better way to mark sites for extraction?
+- What would the best data structure be for these styles, what would encourage the fastest HOT path? I've chosen arrays as they don't take much bundle size vs. objects.
 - How can we de-duplicate style declarations in the same module?
-- Composition of style rules via object rest wouldn't be possible without further AST transformation
+- Composition of style rules via object rest wouldn't be possible without further AST transformation, though perhaps this remains unsupported? Have only a single way to do composition.
 
 ### Keyframes
 
 Keyframes are a bit tricky as they can't easily lean on the same pattern the previous API leans on while still encouraging static compiled strings that can be easily extracted later in time, including module traversal intricacies.
 
-This is probably the weakest part of this RFC currently.
+This is probably the weakest part of this RFC currently and would love any thoughts of how best to handle them.
 
 ```jsx
 import { css, keyframes } from '@compiled/react';
@@ -189,7 +190,7 @@ const styles = css({
 // 👇 transforms to
 
 const styles = i([
-  '_5sagymdr _5sagymdr',
+  '_5sagymdr _j7hqa2t1',
   [
     '._j7hqa2t1{animation-name:kfwl3rt}',
     '@keyframes kfwl3rt{to{opacity:1}}',
@@ -197,6 +198,32 @@ const styles = i([
   ],
 ]);
 ```
+
+Note that it in essence still behaves the same as it does today. The keyframes call site is erased and merged into the style rule. I had some an idea for keyframes to build its own style rule declaration (below) but it broke down when wanting to define animation in pseudo elements/classes.
+
+```jsx
+import { css, keyframes } from '@compiled/react';
+
+const fadeIn = keyframes({
+  to: { opacity: 1 },
+});
+
+const styles = css({
+  animationName: fadeIn,
+  animationDuration: '2s',
+});
+
+// 👇 transforms to
+
+const fadeIn = i([
+  '_j7hqa2t1',
+  ['._j7hqa2t1{animation-name:kfwl3rt}', '@keyframes kfwl3rt{to{opacity:1}}'],
+]);
+
+const styles = i([fadeIn, ['_5sagymdr', ['._5sagymdr{animation-duration:2s}']]]);
+```
+
+Tough.
 
 **Considerations**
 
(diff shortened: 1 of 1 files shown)


rfcs/0000-compiled-runtime.md
```diff
@@ -0,0 +1,417 @@
...
+```jsx
+import { css, styled } from '@compiled/react';
+
+const Comp = styled.div`
+  color: ${(props) => props.color};
+  ${(props) =>
+    props.isHidden
+      ? css({ display: 'none' })
+      : [props.isLarge && css({ fontSize: 50 }), props.isRed && css({ color: 'red' })]}
+`;
```
at-nathan, line 274: I wonder if it's worth first converting this to something like:
`<div css={...css(stringTemplateContent) }>` 

Would this reduce code needed to handle the `styled` API?

rfcs/0000-compiled-runtime.md
```diff
@@ -0,0 +1,417 @@
...
+// 👇 transforms to
+
+const _1 = i(['_syaz11x8', ['._syaz11x8{color:black}']]);
+
+function Comp() {
+  return <div css={_1} />;
+}
+```
+
+When referencing an expression Compiled would try to statically evaluate it, if it fails it would be added as the third item in the array for all dynamic items. This behaviour can be re-used for truly dynamic styles.
```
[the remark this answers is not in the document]
madou: Definitely! It just doesn't have a good story.

rfcs/0000-compiled-runtime.md
madou: Not really, styled needs to internally forward ref and construct the JSX element that was wrapped (either native or composed) - all that to say it would be easier to just write the small abstraction to handle it.

In one way you can imagine every time a "compiled component" is transformed into the AST - that's just where the abstraction would be used instead. So while we currently don't really count them as apart of the bundle size, it might end up reducing it even further by creating the abstractions even though the library "looks" bigger.

rfcs/0000-compiled-runtime.md
```diff
@@ -0,0 +1,473 @@
...
+  '_5sagymdr _j7hqa2t1',
+  [
+    '._j7hqa2t1{animation-name:kfwl3rt}',
+    '@keyframes kfwl3rt{to{opacity:1}}',
+    '._5sagymdr{animation-duration:2s}',
+  ],
+]);
+```
+
+Note that it in essence still behaves the same as it does today. The keyframes call site is erased and merged into the style rule. I had some an idea for keyframes to build its own style rule declaration (below) but it broke down when wanting to define animation in pseudo elements/classes.
```
MonicaOlejniczak, line 202: Can you elaborate more on:

> but it broke down when wanting to define animation in pseudo elements/classes.

Would be interested in seeing the example in practice, as the one below looks fine to me?

madou:
Marking it ready for review! Dig in if you get a chance.

rfcs/0000-compiled-runtime.md
madou: It's hypothetical of course but the overarching goal is how can we architect things so it requires the least amount of AST traversal (and knowledge of other nodes). If I decided to apply an animation only on hover:

```
import { css, keyframes } from '@compiled/react';

// Has no knowledge of styles
const fadeIn = keyframes({
  to: { opacity: 1 },
});

// Has no knowledge of fadeIn animation
const styles = css({
  animationDuration: '2s',
  ':hover': {
      animationName: fadeIn,
  }
});
```

How could we make this work when they don't have knowledge of each other? 🤔

MonicaOlejniczak: Right, so I encountered a similar kind of problem in my experimental implementation. They have an inherent dependency on each other, so I don't think there is any easy way around this. I handled the presence of `animationName` specifically (which breaks down for the tagged template expression syntax), and generated:

```js
const _3 = \\"._1pgl1ytf{animation-timing-function:ease-in-out}\\";
const _2 = \\"._j7hq13il{animation-name:var(--_129ax8o)}\\";
const _ = \\"._5sagymdr{animation-duration:2s}\\";

const fadeOut = {
  name: "kug81sc",
  sheet: "@keyframes kug81sc{25%,0%{opacity:1}25%{opacity:0.75}50%{opacity:0.5}to{opacity:0}}",
};

<CC>
  <CS>{[_, _2, _3]}</CS>
  <div
    className={ax([\\"_5sagymdr _j7hq13il _1pgl1ytf\\"])}
    style={{
      "--_129ax8o\\": ix(fadeOut),
    }}
  />
</CC>
```

The above works a bit differently to yours in that `keyframes()` does not generate an `animation-name` property and value, but instead has `name`. I keep the `animationName` in the original css and utilise the style prop to get the animation name at runtime. But now I have the problem where the `CS` component does not contain the keyframe css, since it's this weird global thing. I think in my case I need to include code evaluation to avoid this problem, and can then remove the need for the `style` prop.

Your approach is more self-contained by having `animation-name` in the keyframe, but how are you omitting the original `animationName` from the output `styles` to get `i([fadeIn, ...])`. Are you also looking specifically for the presence of the animation key?

As for hovering specifically with your implementation, you need code evaluation to get the real animation name as `kfwl3rt` or do it dynamically like in my version above.

rfcs/0000-compiled-runtime.md
madou: That said thinking about this further realistically the component that would be backing the jsx pragma would forward ref too... so it probably just comes down value prop of every component flushing through css pragma vs. just styled, as the former means turning the jsx pragma on and that could be a bit... finicky? Could be worth fleshing this out further though!

rfcs/0000-compiled-runtime.md
madou: Full disclosure most of this RFC isn't implemented yet 😄. Yeah it would handle animation keys as a special case and do something differently.

Love your use of dynamic styles! That actually makes me wonder if we could just reference things by value and then have it be a special case that needs to be handled during style extraction.

```jsx
const fadeIn = keyframes({
  to: { opacity: 1 },
});

const styles = css({
  animationDuration: '2s',
  ':hover': {
      animationName: fadeIn,
  }
});

// transforms to

const fadeIn = i(['kfwl3rt', ['@keyframes kfwl3rt{to{opacity:1}}']]);

const styles = i(['_j7hqa2t1 _5sagymdr', [`._j7hqa2t1{animation-name:${fadeIn[0]}}`, '._5sagymdr{animation-duration:2s}']]);
```

It makes style extraction harder, especially if cross module boundaries which is something we'd like to eliminate I imagine... hence your point about code eval. 

Tough! Something for us to chew on TBH.

madou: Bit of late night thinking in bed hehe so I got up to write it down. I've had a thought where we could keep the separation but still get the result we want using CSS custom properties.

```js
const fadeIn = keyframes({
  to: { opacity: 1 },
});

const styles = css({
  animationDuration: '2s',
  ':hover': {
      animationName: fadeIn,
  }
});

// transforms to

const fadeIn = i(['_5qwe', ['_5qwe{--anim:kfwl3rt}', '@keyframes kfwl3rt{to{opacity:1}}']]);

const styles = i(['_j7hqa2t1 _5sagymdr', [`._j7hqa2t1:hover{animation-name:var(--anim)}}`, '._5sagymdr{animation-duration:2s}']]);
```

**The crux of the solution**
- Any use of animation or animation name will always get transformed to be a CSS variable (probably something hashed, but it should be the same).
- Keyframes generates some styles with the animation name applied to the anim CSS custom property

**Pitfalls**
- It would only work with one animation on an element (as the custom property is static) so it isn't a perfect solution. But maybe there's something here with further digging.

DarkPurple141 (commented):
I see you've noted the potential runtime implication; it would be good to measure/validate this as one of the big pros of compiled is meant to be the minimal runtime.

Overall though the principle of a slightly more constrained API leading to a far simpler AST traversal makes a lot of sense to me.

rfcs/0000-compiled-runtime.md
```diff
@@ -0,0 +1,533 @@
...
+    />
+  );
+}
+```
+
+Less edge cases and AST transformation are needed to strip the runtime. Win win!
+
+### Styled
+
+Styled components would build on the same foundation as both the `css` func & prop, still leaning on a runtime to do the application of class names and style insertion.
```
madou, line 294: It would still be advantageous to block any non-DOM attributes from being passed down to the DOM element at runtime (related: https://github.com/atlassian-labs/compiled/issues/407)

This could look like another argument:

```jsx
styled.div`
  color: ${props => props.color};
`;

// to

styled('div', props => [
    ['_asd12s', ['_.asd12s{color:var(--as1112)}'], ['--as1112', props.color]]
  ], 
  ['color']); // <---
```

rfcs/0000-compiled-runtime.md
```diff
@@ -0,0 +1,533 @@
...
+When referencing an expression Compiled would try to statically evaluate it else if it fails be added as the third item in the array along with all other dynamic styles.
+
+```jsx
+import { N800 } from '@atlaskit/theme/colors';
+
+const styles = css({ color: N800 });
+
+// 👇 transforms to
+
+const styles = i(['_syaz4rde', ['._syaz4rde{color: var(--_kmurgp)}'], [['--_kmurgp', N800]]]);
```
madou, line 140: How might styles with suffix/prefix' affect dynamic styles? Lift and shift should work fine.

```jsx
import image from '!url-loader!./image.png';

const styles = css({ background: `url(${image})` });
```

```jsx
const styles = i(['_1asd931s', ['._1asd931s{background: var(--_921awg)}'], [['--_921awg', `url(${image})`]]]);
```

rfcs/0000-compiled-runtime.md
```diff
@@ -0,0 +1,533 @@
...
+};
+```
+
+For the strip runtime it needs to be able to work with both transformation targets to ensure all sourced of Compiled transformed components will be extracted, with configuration to turn on/off legacy/experimental extraction.
+
+## Unresolved questions
+
+- What edge cases could arise with style extraction?
+- What performance characteristics (bundle size vs. runtime) are affected by this change?
+- This doesn't completely get rid of the need of module traversal (for style declaration static evaluation) - is it possible to?
```
MonicaOlejniczak, line 533: Why do you say this? I think it can, if you make assumptions about some usages (or guarantee it through types). The only other problem is stylesheet extraction, but it can leverage module traversal itself or de-optimise the output if necessary (leave anything imported un-extracted).

madou: Only for variable import use. The answer could be just throw and don't allow it? Problem solved. It's a similar aspect to Tom's point around completely changing APIs, for example if we were to get rid of the "0 config SSR" aspect would could in essence boil down the API to just.. css prop/class name. And styled for the short term conversion. With the obvious assumption the production if anyone wants SSR to work, well, they'd turn on config to extract anyway?

🤔

madou: Here's a spike https://github.com/atlassian-labs/compiled/compare/rfc-compiled-runtime-spike?expand=1

**Setup**

```
git checkout rfc-compiled-runtime-spike
yarn clean
yarn
```

**Tests**

```
yarn test packages/react/src/__tests__/experimental.test.tsx
yarn test packages/react/src/experimental.tsx
```

**Storybook**

```
yarn start
```

Removing the constraint of zero config (streaming) SSR really simplifies everything.

MonicaOlejniczak: Yeah, sounds good! I think I'd want a slightly different API in the end, but that's pretty easy to change.

madou:
Closed as stale

PR #955, opened 2021-11-30
chore: adds only changed to chromatic run

madou:
This should reduce the total amount of snapshots used monthly.


JakeLane (dismissed):

madou:
Will hold off merging until we have test runs again so we can make sure its working correctly.

MonicaOlejniczak (dismissed):

JakeLane:
We now use Loki so no need for this PR anymore

PR #960, opened 2021-12-03
Add visual regression test GitHub action with Percy

JakeLane:
Add BrowserStack Percy as new visual regression library as we use this for other projects at Atlassian.

Storybook support are as follows: https://docs.percy.io/docs/storybook


at-nathan (approved):

JakeLane:
Seems to work fine but we're running into limits already 😢 . Closing for now as we can revive this in a pinch.

PR #984, opened 2021-12-14
Add config to typecheck parcel example

JakeLane:
Parcel does not type check by default so we need to ensure typescript does this for us.

### DIFF
--- a/examples/parcel/src/ui/typescript.tsx
+++ b/examples/parcel/src/ui/typescript.tsx
@@ -1,4 +1,5 @@
-import '@compiled/react';
+/** @jsxImportSource @compiled/react */
+// @ts-expect-error
 import { primary } from '../constants';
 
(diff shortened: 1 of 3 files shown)


examples/parcel/src/ui/typescript.tsx
```diff
@@ -8,7 +9,7 @@ interface TypeScriptProps {
 export const TypeScript = ({ color }: TypeScriptProps): JSX.Element => (
   <div
     css={{
-      display: 'inline-block',
+      display2: 'inline-block',
```
MonicaOlejniczak, line 12: display2?

JakeLane: oops forgot to remove this when testing the types

MonicaOlejniczak (approved):

PR #987, opened 2021-12-16
Add support in babel plugin for file imports

### LINKED ISSUE #956: Loading imported 'images' using babel loader
**Describe the bug**
When adding an image in React, I receive the following error:

```
Property expression of ExpressionStatement expected node to be of a type ["Expression"] but instead got "ImportDefaultSpecifier"
```

**To Reproduce**
- Create React App with craco
- New dependencies
    "@compiled/babel-plugin": "^0.11.1",
    "@compiled/react": "^0.10.0",
- Import an image and try to create a style to use the image
 ```tsx
 import dropdownSvg from './Dropdown.svg'; 

export const Select = styled.select({
    background: `#fff no-repeat center right .5rem`,
    backgroundImage: `url(${dropdownSvg})`,
});

```

**Expected behavior**
dropdownSvg is a string with a path in the /static/media folder, backgroundImage should be assigned that path

Additionally, some way to include imported images should be documented, currently there's basically no reference to using images in the docs that I can find

### PR DESCRIPTION
**Pull request by JakeLane**
Resolves #956 

Two major changes in this PR to the logic in the babel plugin:

We now only traverse files with configured file extensions. This is configured as an option (with a default) to the babel plugin and the loaders.

We also now treat `url(...)` as a special case for variable interpolation in CSS. This is required as `url` differs from other CSS functions.

### CODE REVIEW

File: packages/css/src/utils/__tests__/string-iterpolations.test.tsx [hunk 1]
```diff
@@ -214,120 +220,162 @@ describe('template literal to css', () => {
...
 
-      const extract = cssBeforeInterpolation(simpleParts[0]);
+      const [before] = cssAffixInterpolation(simpleParts[0], simpleParts[1]);
+
+      expect(before.variablePrefix).toEqual('-');
+      expect(before.css).toEqual('0 ');
+    });
+  });
+
+  describe('interpolations with url', () => {
```
JakeLane, line 358: For readability, only below are new tests

packages/css/src/utils/string-interpolations.tsx
```diff
@@ -64,3 +54,37 @@ export const cssBeforeInterpolation = (css: string): BeforeInterpolation => {
...
+ *
+ *
+ * @param before all the CSS _before_ the interpolation
+ * @param after all the CSS _after_ the interpolation
+ */
+export const cssAffixInterpolation = (
+  before: string,
+  after: string
+): [BeforeInterpolation, AfterInterpolation] => {
+  if (before.endsWith('url(') && after.startsWith(')')) {
```
MonicaOlejniczak, line 81: Should these be trimmed?

stories/assets/carat-down.svg
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
MonicaOlejniczak, line 1: Can you run these svgs through svgo with prettified markup?

stories/styled-file-imports.tsx
```diff
@@ -0,0 +1,28 @@
+import { styled } from '@compiled/react';
+
+import caratDown from './assets/carat-down.svg';
+import caratUp from './assets/carat-up.svg';
+
+export default {
+  title: 'styled/file imports',
+};
+
+const BackgroundImage = styled.div({
```
MonicaOlejniczak, line 10: What about `css` and tagged template syntax?

stories/types.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.svg' {
+  const content: any;
```
MonicaOlejniczak, line 2: Why don't you type this as a string? Normally it would return either the url or the raw content

examples/webpack/src/app.js
```diff
@@ -3,6 +3,7 @@ import '@compiled/react';
 import { Suspense, lazy } from 'react';
 
 import { primary } from './common/constants';
+import { CustomFileExtension } from './ui/custom-file-extension.customjsx';
```
MonicaOlejniczak, line 6: The custom file extension should be for an imported value. Importing a component like this won't be transformed by compiled. Same for the parcel example.

examples/parcel/.parcelrc
```diff
@@ -1,9 +1,14 @@
 {
   "extends": "@parcel/config-default",
   "transformers": {
-    "*.{js,jsx,ts,tsx}": [
+    "*.{js,jsx,ts,tsx,customjsx}": [
       "@compiled/parcel-transformer",
       "..."
+    ],
+    "*.customjsx": [
```
MonicaOlejniczak, line 8: Is this one needed?

examples/parcel/.compiledcssrc
```diff
@@ -1,3 +1,10 @@
 {
-  "importReact": false
+  "importReact": false,
+  "extensions": [
```
MonicaOlejniczak, line 3: Just an idea, but... instead of providing extensions in the config here, you could be a bit more clever in parcel and webpack. What you really want to know is which extensions are used with the `@compiled/parcel-transformer` or the `@compiled/webpack-loader`. I'm not sure how easy it would be to codify that though.

MonicaOlejniczak (changes requested):

examples/parcel/tsconfig.json
```diff
@@ -1,9 +1,9 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "rootDir": "src",
```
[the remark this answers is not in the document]
JakeLane: I removed it as the configuration started complaining as we import from `lib/`

examples/parcel/.compiledcssrc
JakeLane: yeah, I had a quick look into that before implementing the config and ended up going this route because it got complex very quick, as I'd have to programmatically get the matcher/regex.

I'll have another pass to see if I can figure out an elegant solution otherwise will create an issue to track this.

at-nathan (approved):

examples/webpack/src/ui/custom-file-extension.customjsx
```diff
@@ -0,0 +1,27 @@
...
+  ':hover': {
+    color: 'red',
+  },
+
+  ':focus': {
+    color: 'blue',
+  },
+});
+
+export const CustomFileExtensionCss = {
```
MonicaOlejniczak, line 16: This should be lower case as `customFileExtensionCss`, same in the parcel example

I'd also use the explicit `css`

JakeLane: Unfortunately haven't got around to investigating yet but it appears the parcel loader breaks when importing any `css(...)` calls, not just in this PR. Will look soon as we'll need to raise a bug.

JakeLane: For later reference, the call stack is:
```🚨 Build failed.
@compiled/parcel-transformer: /Users/jlane2/Code/compiled/examples/parcel/src/app.js: Cannot read property 'buildError' of undefined
TypeError: /Users/jlane2/Code/compiled/examples/parcel/src/app.js: Cannot read property 'buildError' of undefined
    at NodePath.buildCodeFrameError (/Users/jlane2/Code/compiled/node_modules/@babel/traverse/lib/path/index.js:139:21)
    at Object.buildCodeFrameError (/Users/jlane2/Code/compiled/packages/babel-plugin/dist/utils/ast.js:59:52)
    at Object.buildCss (/Users/jlane2/Code/compiled/packages/babel-plugin/dist/utils/css-builders.js:621:17)
    at Object.buildCss (/Users/jlane2/Code/compiled/packages/babel-plugin/dist/utils/css-builders.js:567:30)
    at Object.visitCssPropPath (/Users/jlane2/Code/compiled/packages/babel-plugin/dist/css-prop/index.js:75:36)
    at PluginPass.JSXOpeningElement (/Users/jlane2/Code/compiled/packages/babel-plugin/dist/babel-plugin.js:197:28)
    at newFn (/Users/jlane2/Code/compiled/node_modules/@babel/traverse/lib/visitors.js:177:21)
    at NodePath._call (/Users/jlane2/Code/compiled/node_modules/@babel/traverse/lib/path/context.js:53:20)
    at NodePath.call (/Users/jlane2/Code/compiled/node_modules/@babel/traverse/lib/path/context.js:40:17)
    at NodePath.visit (/Users/jlane2/Code/compiled/node_modules/@babel/traverse/lib/path/context.js:100:31)
/Users/jlane2/Code/compiled/examples/parcel/src/app.js
```

examples/parcel/.parcelrc
JakeLane: Yep, we get a runtime error from the missing file without it

MonicaOlejniczak (approved):

examples/parcel/.compiledcssrc
JakeLane: Created an issue for it as I couldn't find an elegant way to address this without refactoring how we build the resolver https://github.com/atlassian-labs/compiled/issues/1000

Issue #1010, opened 2022-01-05
Babel Plugin cannot differentiate between reserved keyword and a variable while resolving bindings.


wilsonhou:
**Describe the bug**
Babel Plugin cannot differentiate between reserved keyword and a variable while resolving bindings - and will try to traverse both. Current workaround is to shortcircuit on modules beginning with `@compiled/`.

**To Reproduce**

1. Remove the following workaround in `resolve-binding.ts`:
```js
if (moduleImportSource.startsWith('@compiled/')) {
  // Ignore @compiled modules.
  return;
}
```

2. Run test in `/path/to/packages/react/keyframes/__tests__/index.test.tsx`.
3.  See Error
![Screen Shot 2022-01-05 at 2 41 11 pm](https://user-images.githubusercontent.com/52393809/148157214-40dfc252-da77-4e09-947d-5728f4cd590a.png)

**Expected behavior**
Babel Plugin should recognise `keyframes` as a reserved keyword and not traverse.

**Desktop (please complete the following information):**
- OS: macOS

**Additional context**
We currently use this workaround to stop traversal on compiled modules.
```js
if (moduleImportSource.startsWith('@compiled/')) {
  // Ignore @compiled modules.
  return;
}
```

PR #1015, opened 2022-01-06
Fix: Prevent usages of extraneous packages.

wilsonhou:
This pull request adds a new ESLint rule: `import/no-extraneous-dependencies`.

It makes a few changes to allow Compiled to work with this rule:
- Added imported extraneous packages as dependencies in relevant package.json files.
- Changed 'module-a' package to '@compiled-private/module-a' to prevent name clashes.
- Updated references to '@compiled-private/module-a'.

We then documented a workaround for issue #1010: https://github.com/atlassian-labs/compiled/issues/1010

### DIFF
--- a/packages/react/src/__tests__/ssr.test.tsx
+++ b/packages/react/src/__tests__/ssr.test.tsx
@@ -1,7 +1,9 @@
 /**
  * @jest-environment node
  */
+// eslint-disable-next-line import/no-extraneous-dependencies
 import { styled } from '@compiled/react';
+// eslint-disable-next-line import/no-extraneous-dependencies
 import { CC as CompiledRoot } from '@compiled/react/runtime';
 import React from 'react';
 import { renderToStaticMarkup } from 'react-dom/server';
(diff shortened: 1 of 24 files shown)


JakeLane (approved):

packages/react/src/__tests__/ssr.test.tsx
```diff
@@ -1,7 +1,9 @@
 /**
  * @jest-environment node
  */
+// eslint-disable-next-line import/no-extraneous-dependencies
```
pancaspe87, line 4: nit
You could also do
`/* eslint-disable import/no-extraneous-dependencies */`
import { CC as CompiledRoot } from '@compiled/react/runtime';
`/* eslint-enable import/no-extraneous-dependencies */`

pancaspe87 (dismissed):

packages/react/src/__tests__/ssr.test.tsx
wilsonhou: thanks! that's really neat.

JakeLane (approved):

PR #1032, opened 2022-01-12
Destructured prop fixes

at-nathan:
- Fixes de-structuring not happening when arrow function returns property: value pairs
- Add support for nested and alias de-structuring
- Removed behavior that replaces aliases with original prop reference as it may collide with external vars


packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
```diff
@@ -417,13 +431,13 @@ describe('styled component behaviour', () => {
       import { CUSTOM_WIDTH } from './constants';
 
       const ListItem = styled.div\`
-        width: \${(props) => props.useCustomWidth ? \`\${CUSTOM_WIDTH}px)\` : '100%'};
```
at-nathan, line 420: Removing the stray `)`

packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
```diff
@@ -433,13 +447,141 @@ describe('styled component behaviour', () => {
       import { CUSTOM_WIDTH } from './constants';
 
       const ListItem = styled.div\`
-        width: \${({useCustomWidth}) => useCustomWidth ? \`\${CUSTOM_WIDTH}px)\` : '100%'};
```
at-nathan, line 436: Ditto here too

packages/babel-plugin/src/utils/build-styled-component.ts
```diff
@@ -156,56 +209,6 @@ const handleMemberExpressionInStyledInterpolation = (path: NodePath<t.MemberExpr
   return propsToDestructure;
 };
 
-/**
```
at-nathan, line 159: As discussed with @MonicaOlejniczak, we're removing this logic as  it replaces any use of destructure aliasing with the original prop var name. It's intention was to reduce output size by preventing something like:
```
{
  name: nickName1,
  name: nickName2,
} = props;
```

However, this sort of occurrence would be quite low in a styled component But also removing aliases could cause collisions with outer scope vars, for example:

```
const width = 10;
const Comp = styled.div`
  width: $({ width }) => ${width + width}px;
`;
```
Without aliasing here things won't work as intended.

MonicaOlejniczak: Just to clarify, the second example you gave does work. Usually lint rules that ban shadowed variables are in place to improve code readability.

Regardless, this optimisation is not needed in compiled and it is generally safer for us to preserve as much user code as possible. Bundlers can optimise the final result as well.

packages/babel-plugin/src/utils/build-styled-component.ts
```diff
@@ -102,6 +101,60 @@ const traverseStyledBinaryExpression = (node: t.BinaryExpression, nestedVisitor:
...
+        // { width: alias } becomes width:alias
+        const propsUsed = path.toString().replace(/\s/g, '').slice(1, -1).split(',');
+
+        destructuredProps.push(...propsUsed);
+        path.stop();
+      }
+    },
+  };
+
+  const arrowFunctionVisitor = {
```
MonicaOlejniczak, line 140: I would inline this in the below `traverse`

packages/babel-plugin/src/utils/build-styled-component.ts
```diff
@@ -102,6 +101,60 @@ const traverseStyledBinaryExpression = (node: t.BinaryExpression, nestedVisitor:
...
+    noScope: true,
+    ArrowFunctionExpression(path: NodePath<t.ArrowFunctionExpression>) {
+      const {
+        params: [propsParam],
+      } = path.node;
+      // Is the param for props being destructured
+      if (t.isObjectPattern(propsParam)) {
+        path.traverse(appendDestructuredPropsVisitor);
```
MonicaOlejniczak, line 148: Instead of traversing again, which is quite inefficient, could we just inspect the first argument and check if it is an object pattern?

packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
```diff
@@ -72,7 +72,21 @@ describe('styled component behaviour', () => {
     expect(actual).toInclude('{font-weight:500}');
   });
 
-  it('should not pass down invalid html attributes to the node', () => {
+  it('should not destructure valid html attributes from props', () => {
+    const actual = transform(`
```
MonicaOlejniczak, line 76: Could you use the common `transform` from the `test-utils`?

.changeset/lemon-nails-sniff.md
```diff
@@ -0,0 +1,5 @@
+---
+'@compiled/babel-plugin': minor
+---
+
+Fixes destructuring not happening when arrow functions returns property: value pairs. Added support for nested and alias destructuring.
```
MonicaOlejniczak, line 5: What about this?

Handle destructuring `property: value` pairs returned from arrow functions, and add support for nested and alias destructuring

packages/babel-plugin/src/utils/build-styled-component.ts
at-nathan: I was trying say we can't use both `width`s within the arrow function as without aliasing we can only use the one in the function scope :)

stories/conditional-rules-styled.tsx
```diff
@@ -309,3 +317,21 @@ export const ConditionWithNoValuePath = (): JSX.Element => {
     </div>
   );
 };
+
+export const ConditionWithDeconstructedValues = (): JSX.Element => {
+  return (
```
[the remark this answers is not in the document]
at-nathan: Sorry, I gotta get out of the habit of copying and pasting 😓

packages/babel-plugin/src/utils/build-styled-component.ts
at-nathan: 👍  Ah I see it's only beneficial to split out nested visitors.

packages/babel-plugin/src/utils/build-styled-component.ts
at-nathan: Hmm in order to get the string source of a node I need to use it's path. Is there a cheaper way to get this?

packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
at-nathan: For the whole file or just this test? I'll give it a try.

packages/babel-plugin/src/utils/build-styled-component.ts
MonicaOlejniczak: Try `path.get('params')[0].toString()`

at-nathan: Ah thanks, I didn't know `get` will return a path too.

packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
at-nathan: Tried using `import { transform } from '../../test-utils';` and everything died :(

stories/conditional-rules-styled.tsx
```diff
@@ -93,6 +93,14 @@ const NoValuePath = styled.div<TextProps>`
   ${(props) => (props.isPrimary ? undefined : 'color: red')}
 `;
 
+const DeconstructedPropsValues = styled.div<TextProps>`
```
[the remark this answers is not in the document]
at-nathan: Variable shadowing isn't working as intended, it takes uses the value from the outer scope and statically evaluates the value. Looks like `evaluate-expression` is working too well 😓 Thoughts on creating a separate issue for this? Seems like an existing bug.

packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
MonicaOlejniczak: Right, I forgot it would affect non-snapshots too. Yeah, don't worry about it for now then.

MonicaOlejniczak (approved):

stories/conditional-rules-styled.tsx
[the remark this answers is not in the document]
MonicaOlejniczak: Separate issue is fine

JakeLane (approved):

PR #1043, opened 2022-01-17
fix:statically evaluates deconstructed values from deeply nested objects

pancaspe87:
Currently we fail to statically evaluate value when doing:

const { red } = theme.colors;

Works when member is not involved for example:

const { red } = colors;


packages/babel-plugin/src/utils/resolve-binding.ts
```diff
@@ -10,6 +10,7 @@ import resolve from 'resolve';
 import { DEFAULT_CODE_EXTENSIONS } from '../constants';
 import type { Metadata } from '../types';
 
+import { evaluateExpression } from './evaluate-expression';
```
at-nathan, line 13: I think this is a cyclic import as `evaluate-expression` imports `resolveBinding` too.

at-nathan:
Interested to see if this works with star imports, i.e.
```
import * as colors from 'colors';

const { red } = colors;
```

Can we try this out in `/packages/babel-plugin/src/__tests__/module-traversal.test.tsx`. Otherwise I'm happy for it to be addressed in another issue.

packages/babel-plugin/src/utils/resolve-binding.ts
MonicaOlejniczak: I too am surprised this works...

pancaspe87: good catch! I'll look for a different solution

at-nathan: One way I can think of is to pass in `evaluate-expression` as a param to `resolveBinding`. But this may cause cyclic imports in other places where `resolveBinding` is used.

pancaspe87: This should be fixed now, I should have thought to use `traverseMemberExpression` in the first place!

pancaspe87:
Fix doesn't work with star values. 
Created a [new issue](https://github.com/atlassian-labs/compiled/issues/1047) for it.

packages/babel-plugin/src/utils/resolve-binding.ts
```diff
@@ -10,6 +10,7 @@ import resolve from 'resolve';
 import { DEFAULT_CODE_EXTENSIONS } from '../constants';
 import type { Metadata } from '../types';
 
+import { traverseMemberExpression } from './evaluate-expression';
```
at-nathan, line 13: hmm this is still a cyclic import though, because `evaluate-expression` imports `resolve-binding` and vice versa.

packages/babel-plugin/src/utils/evaluate-expression/index.tsx
```diff
@@ -1,268 +1,17 @@
...
 
 import type { Metadata } from '../../types';
-import { getPathOfNode, wrapNodeInIIFE } from '../ast';
+import { getPathOfNode } from '../ast';
+import { createResultPair } from '../create-result-pair';
 import { isCompiledKeyframesCallExpression } from '../is-compiled';
-import { resolveBinding } from '../resolve-binding';
 
-import { createResultPair } from './common';
-import { traverseMemberAccessPath } from './traverse-access-path';
```
pancaspe87, line 11: @MonicaOlejniczak , @at-nathan I took all the methods traversing expressions, not only `traverseMemberExpression`, out of evaluate-expression. This also implied to take `createResultPair` out of it.

I know that you been reorganising the folder structure, let me know if this structure doesn't fit your vision

packages/babel-plugin/src/utils/traverse-expression/traverse-member-expression.tsx
```diff
@@ -0,0 +1,98 @@
+import traverse from '@babel/traverse';
```
at-nathan, line 1: hmm I think the refactor is good :) `traverseMemberAccessPath` is exclusively used by `traverseMemberExpression`. So I think everything in `traverseMemberAccessPath` should sit under it. Thoughts on making `traverse-member-expression`and folder?

packages/babel-plugin/src/utils/traverse-expression/traverse-call-expression.tsx
```diff
@@ -0,0 +1,116 @@
+import * as t from '@babel/types';
+
+import type { Metadata } from '../../types';
+import { getPathOfNode, wrapNodeInIIFE } from '../ast';
+import { createResultPair } from '../create-result-pair';
+import { evaluateExpression } from '../evaluate-expression';
+import { resolveBinding } from '../resolve-binding';
+
+import { traverseMemberExpression } from './traverse-member-expression';
```
at-nathan, line 9: This is a cyclic import :( `traverseMemberExpression` imports `traverseCallExpression` too.

packages/babel-plugin/src/utils/traverse-expression/traverse-member-expression.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const { accessPath, bindingIdentifier } = getMemberExpressionMeta(expression);
+
+  if (bindingIdentifier) {
+    return traverseMemberAccessPath(
+      t.identifier(bindingIdentifier.name),
+      meta,
+      bindingIdentifier.name,
+      accessPath,
+      expression,
+      { callExpression: traverseCallExpression, memberExpression: traverseMemberExpression }
```
at-nathan, line 93: The reason I did this was to prevent cyclic imports. Maybe we can do this in `traverseCallExpression` too.

packages/babel-plugin/src/utils/traverse-expression/traverse-member-expression.tsx
pancaspe87: sure, I'll do that!

packages/babel-plugin/src/utils/traverse-expression/traverse-call-expression.tsx
pancaspe87: - [ ] resolve cyclic import

packages/babel-plugin/src/utils/traverse-expression/traverse-member-expression.tsx
pancaspe87: Added traverseCallExpression directly in [traverseMemberAccessPath](https://github.com/atlassian-labs/compiled/pull/1043/files#diff-d65791a40983b8a189d518fc1d924a71ade13f378b12d80026f10a760aae2ee4R20)

JakeLane (approved):

at-nathan:
LGTM :) For future improvements we could maybe move `traverseMemberAccessPath` into `traverseMemeberExpression` in another PR so we don't bloat this one.

at-nathan (approved):

PR #1046, opened 2022-01-18
move attrs transformation after applyVisitor & match compiled styled …

blackheart-atlassian:
…local name

### DIFF
--- a/packages/codemods/src/transforms/styled-components-to-compiled/styled-components-to-compiled.tsx
+++ b/packages/codemods/src/transforms/styled-components-to-compiled/styled-components-to-compiled.tsx
@@ -1,5 +1,6 @@
 import type { API, FileInfo, Options, Program, TaggedTemplateExpression } from 'jscodeshift';
 
+import { COMPILED_IMPORT_PATH } from '../../constants';
 import defaultCodemodPlugin from '../../plugins/default';
 import type { CodemodPluginInstance } from '../../plugins/types';
 import {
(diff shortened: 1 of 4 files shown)


packages/codemods/src/utils/styled-components-attributes.ts
```diff
@@ -37,7 +37,7 @@ const getRestAttributeExpressions = (expression: ObjectExpression) =>
     ? (expression.properties.filter((property) => !isStyleProperty(property)) as ObjectProperty[])
     : null;
 
-const getAlternativeComponentName = (name: string) => `Compiled${name}`;
+const getAlternativeComponentName = (name: string) => `${name}Styles`;
```
at-nathan, line 40: We had a hard time coming up with a description name for this component. It's the component that's had any styles that came from `.attr` applied and awaiting the rest of props from `.attr` to be spread back in. Any thoughts on the naming of this?

at-nathan (dismissed):

pancaspe87 (dismissed):

packages/codemods/src/transforms/styled-components-to-compiled/styled-components-to-compiled.tsx
```diff
@@ -53,40 +54,55 @@ const transformer = (fileInfo: FileInfo, api: API, options: Options): string =>
...
-    });
-  }
-
   applyVisitor({
     plugins,
     originalProgram,
     currentProgram: collection.find(j.Program).get(),
   });
 
+  const styledImport = collection.find(
```
MonicaOlejniczak, line 63: Can't we just put this logic in the default plugin `buildAttributes`, omit the local `styled` name logic there, and then in Jira hook into `buildAttributes` so that it renames it to the local name? If that doesn't work, we should find a way so that it does.

blackheart-atlassian: we talked about it with @at-nathan and decided that in this particular case that makes sense, as we don't really want to transform any styled-component attrs usages but rather fix compiled incompatibilities. And in order to do that, we need to apply all transformations, including the program visitor, and only then transform attrs usages.

packages/codemods/src/transforms/styled-components-to-compiled/styled-components-to-compiled.tsx
```diff
@@ -53,40 +54,55 @@ const transformer = (fileInfo: FileInfo, api: API, options: Options): string =>
...
+    j.ImportSpecifier,
+    (specifier) => specifier.imported.name === 'styled'
+  );
+
+  styledImport.forEach((styledImport) => {
+    const isCompiledImport =
+      styledImport.parentPath.parentPath.value.source.value === COMPILED_IMPORT_PATH;
```
JakeLane, line 70: could this cause a null pointer issue?

packages/codemods/src/transforms/styled-components-to-compiled/styled-components-to-compiled.tsx
```diff
@@ -53,40 +54,55 @@ const transformer = (fileInfo: FileInfo, api: API, options: Options): string =>
...
+      const taggedTemplateExpressionsWithAttrs = collection.find(
+        j.TaggedTemplateExpression,
+        ({ tag }: TaggedTemplateExpression) => {
+          if (tag.type === 'CallExpression') {
+            const { callee } = tag;
+
+            return (
+              callee.type === 'MemberExpression' &&
+              callee.object.type === 'MemberExpression' &&
+              callee.object.object.type === 'Identifier' &&
```
JakeLane, line 84: same as above

packages/codemods/src/transforms/styled-components-to-compiled/styled-components-to-compiled.tsx
blackheart-atlassian: I don't think so, the object structure is pretty standard. We can use something safer here or just wrap it in try/catch if you think that's necessary.

packages/codemods/src/transforms/styled-components-to-compiled/styled-components-to-compiled.tsx
at-nathan: I think this one should be fine because above we check that `callee.object` is a member, which means `callee.object.object` will exist.

packages/codemods/src/transforms/styled-components-to-compiled/styled-components-to-compiled.tsx
at-nathan: Yeah maybe safer to do `styledImport?.parentPath?.parentPath`

packages/codemods/src/transforms/styled-components-to-compiled/styled-components-to-compiled.tsx
blackheart-atlassian: we do all the previous checks to unsure the object has all necessary fields. so, I'd say it' pretty safe. any ideas how we can make it better?

JakeLane: yeah looking at this one closer, the types look safe 👍

packages/codemods/src/transforms/styled-components-to-compiled/styled-components-to-compiled.tsx
JakeLane: I think the main problem is `any` type - we've been bitten a couple times in the past because of this same thing coming out of `jscodeshift` :/

JakeLane (approved):

PR #1052, opened 2022-01-24
Implement stylesheet extraction with Parcel

JakeLane:
Picking up where #695 left off.

Simple support for stylesheet extraction in parcel. The changes are introduced by using three different types of Parcel plugins.

First is the `@compiled/parcel-transformer` which is extended to add new imports including the CSS payload. e.g. `import "compiled-css:...";`. The second is the `@compiled/parcel-resolver` which resolves the imports into CSS code. The final is the @compiled/parcel-optimizer` which takes the CSS and sorts the rules.

These three plugins can be combined together with minimal configuration by utilising `@compiled/parcel-config` to include each into the codebase.


packages/parcel-resolver/src/resolver.tsx
```diff
@@ -0,0 +1,23 @@
+import { dirname, join } from 'path';
+
+import { Resolver } from '@parcel/plugin';
+
+const resolver = new Resolver({
+  async resolve({ dependency }) {
+    const { specifier, resolveFrom } = dependency;
+
+    const cssPrefix = 'compiled-css!';
```
hannahkmolloy, line 9: How come this prefix has an exclamation mark?

hannahkmolloy: actually nvm I read further down and can see here https://github.com/atlassian-labs/compiled/pull/1052/files#diff-c0b739275fb1afe7a00a93f95e32cf560fb7579d0d3a8e7360b8cf003a6e0f24R120 that the dependency is defined with an exclamation mark

hannahkmolloy (dismissed):

packages/parcel-config/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Installation
+
+```bash
+npm i @compiled/parcel-optimizer
+```
+
+## Usage
+
+Detailed docs and example usage can be [found on the documentation website](https://compiledcssinjs.com/docs/pkg-parcel).
```
pancaspe87, line 13: The page is currently blank. Are you also updating the docs?

JakeLane: Yep I plan to! In retrospect, I should probably update this to https://compiledcssinjs.com/docs/pkg-parcel-config so it corresponds to a real package

hannahkmolloy (approved):

pancaspe87 (dismissed):

packages/parcel-transformer/tsconfig.json
```diff
@@ -6,5 +6,10 @@
     "module": "commonjs",
     "target": "es2017"
   },
-  "references": [{ "path": "../babel-plugin" }, { "path": "../utils" }]
+  "references": [
+    { "path": "../babel-plugin-strip-runtime" },
+    { "path": "../babel-plugin" },
```
MonicaOlejniczak, line 11: /nit `babel-plugin` before `babel-plugin-strip-runtime`

packages/tsconfig.json
```diff
@@ -10,6 +10,9 @@
     { "path": "eslint-plugin/tsconfig.json" },
     { "path": "jest/tsconfig.json" },
     { "path": "parcel-transformer/tsconfig.json" },
```
MonicaOlejniczak, line 12: Remove this one since it's a duplicate

packages/parcel-transformer/src/transformer.tsx
```diff
@@ -37,8 +40,8 @@ export default new Transformer<UserlandOpts>({
     return false;
   },
 
-  async parse({ asset }) {
-    if (!asset.isSource) {
+  async parse({ asset, config }) {
+    if (!asset.isSource && !config.extract) {
```
MonicaOlejniczak, line 44: Shouldn't the asset be parsed when extraction is disabled?

packages/parcel-transformer/src/transformer.tsx
```diff
@@ -64,54 +67,63 @@ export default new Transformer<UserlandOpts>({
 
   async transform({ asset, config }) {
     const ast = await asset.getAST();
-    if (!asset.isSource || !ast) {
-      // We will only recieve ASTs for assets we're interested in.
+    if (!ast) {
```
MonicaOlejniczak, line 70: I believe this will bail out of the transform entirely when the asset is not source or we are extracting, since you check for the presence of the ast from `parse`. These scenarios should all be tested in code with the programmatic api...

packages/parcel-transformer/src/transformer.tsx
```diff
@@ -64,54 +67,63 @@ export default new Transformer<UserlandOpts>({
...
 
-    if (result?.ast) {
-      asset.setAST({
-        // TODO: Currently if we set this as `'babel'` the babel transformer blows up.
-        // Let's figure out what we can do to reuse it.
-        type: 'compiled',
-        version: '0.0.0',
-        program: result.ast,
+    if (config.extract && foundCSSRules.length) {
+      foundCSSRules.forEach((rule) => {
```
MonicaOlejniczak, line 118: Can we just use `for of`

packages/parcel-transformer/src/transformer.tsx
```diff
@@ -64,54 +67,63 @@ export default new Transformer<UserlandOpts>({
...
+        // Build imports with css to be resolved into CSS by `@compiled/parcel-resolver`
+        const params = encodeURIComponent(rule);
+        output = `\nimport 'compiled-css!${params}';\n${output}`;
```
MonicaOlejniczak, line 121: Can you check whether this blows up flow in Jira?

MonicaOlejniczak (changes requested):
I would like to see tests for these changes, like we have for the webpack loader

packages/parcel-transformer/src/transformer.tsx
JakeLane: When the asset is not source and extraction is disabled, the transformer has nothing to do so this should reduce build time

packages/parcel-optimizer/src/__tests__/optimizer.parceltest.tsx
```diff
@@ -0,0 +1,63 @@
...
+const rootPath = join(__dirname, '..', '..', '..', '..');
+const fixtureRoot = join(rootPath, 'fixtures/parcel-optimizer-test-app');
+
+const workerFarm = createWorkerFarm();
+const outputFS = new MemoryFS(workerFarm);
+
+const parcel = new Parcel({
+  config: join(fixtureRoot, '.parcelrc'),
+  entries: [join(fixtureRoot, 'src', 'index.html')],
+  workerFarm,
```
MonicaOlejniczak, line 19: `workerFarm` after `targets`

packages/parcel-transformer/src/__tests__/transformer.parceltest.tsx
```diff
@@ -0,0 +1,84 @@
...
+afterAll(() => {
+  workerFarm.end();
+});
+
+describe('transformer without extraction', () => {
+  const outputFS = new MemoryFS(workerFarm);
+  const parcel = new Parcel({
+    config: join(fixtureRoot, '.parcelrc'),
+    entries: [join(fixtureRoot, 'src', 'index.html')],
+    workerFarm,
```
MonicaOlejniczak, line 21: `workerFarm` after `targets`

packages/parcel-transformer/src/__tests__/transformer.parceltest.tsx
```diff
@@ -0,0 +1,84 @@
...
+    },
+  });
+
+  it('transforms assets with babel plugin', async () => {
+    const { changedAssets } = await parcel.run();
+
+    const asset = Array.from(changedAssets.values()).find(
+      (asset) => asset.filePath === join(fixtureRoot, 'src/index.js')
+    );
+    expect(asset).toBeDefined();
```
MonicaOlejniczak, line 36: Same as above, this can be removed

packages/parcel-transformer/src/__tests__/transformer.parceltest.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    const code = await asset?.getCode();
+    expect(code).toInclude('/* index.js generated by @compiled/babel-plugin v0.0.0 */');
```
MonicaOlejniczak, line 39: Are you able to take an inline snapshot? If there's too much code, maybe trim it to where you need

packages/parcel-transformer/src/__tests__/transformer.parceltest.tsx
```diff
@@ -0,0 +1,84 @@
...
+    },
+  });
+
+  it('transforms assets with compiled and extraction babel plugins', async () => {
+    const { changedAssets } = await parcel.run();
+
+    const asset = Array.from(changedAssets.values()).find(
+      (asset) => asset.filePath === join(extractionFixtureRoot, 'src/index.js')
+    );
+    expect(asset).toBeDefined();
```
MonicaOlejniczak, line 65: Same as above, this can be removed

packages/parcel-transformer/src/__tests__/transformer.parceltest.tsx
```diff
@@ -0,0 +1,84 @@
...
+    expect(code).toInclude(`require("@compiled/react/runtime")`);
+  });
+
+  it('css is extracted into inline imports', async () => {
+    const { changedAssets } = await parcel.run();
+
+    const asset = Array.from(changedAssets.values()).find(
+      (asset) => asset.filePath === join(extractionFixtureRoot, 'src/index.js')
+    );
+    expect(asset).toBeDefined();
```
MonicaOlejniczak, line 79: Same as above, this can be removed

packages/parcel-transformer/src/__tests__/transformer.parceltest.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    const code = await asset?.getCode();
+    expect(code).toInclude('/* index.js generated by @compiled/babel-plugin v0.0.0 */');
+    expect(code).toInclude('._syaz5scu{color:red}');
+    expect(code).toInclude(`require("@compiled/react/runtime")`);
+  });
+});
+
+describe('transformer with extraction', () => {
```
MonicaOlejniczak, line 45: Can you test `node_modules` extraction?

fixtures/parcel-transformer-test-extract-app/.parcelrc
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "@parcel/config-default",
+  "transformers": {
+    "*.js": ["@compiled/parcel-transformer", "..."]
+  },
+  "resolvers": ["@compiled/parcel-resolver-noop", "..."]
```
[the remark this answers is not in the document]
JakeLane: I wanted to keep the parcel plugin tests independent so that issues in plugins don't leak between tests. It does add a lot of fixtures but I did prefer that only the relevant tests were failing when I was working on this branch

MonicaOlejniczak (dismissed):

MonicaOlejniczak (approved):

PR #1058, opened 2022-01-27
Add property to loader options when the extract plugin is enabled 

hannahkmolloy:
When setting up the CompiledExtractPlugin (when `extract` is enabled), we set an option on the loader context to let the @compiled/webpack-loader know that everything was set up correctly. However, when the @compiled/webpack-loader is used in conjunction with the thread-loader, the CompiledExtractPlugin is set up in a different scope to where the @compiled/webpack-loader is run. This means the check to see if the plugin was set up correctly will always fail when using the thread-loader.

The approach is to modify the loader options (that are passed directly into the @compiled/webpack-loader) instead of the loader context.

### DIFF
--- a/packages/webpack-loader/src/utils.tsx
+++ b/packages/webpack-loader/src/utils.tsx
@@ -44,31 +44,6 @@ export const setPluginConfiguredOption = (
 };
 
 /**
- * Gets the normal module hook for webpack 4 & webpack 5.
- *
- * @returns
- */
-export const getNormalModuleHook = (
-  compiler: Compiler,
-  compilation: CompilationType
-): CompilationType['hooks']['normalModuleLoader'] => {
-  const { NormalModule } =
-    // Webpack 5 flow
-    compiler.webpack ||
-    // Webpack 4 flow
-    require('webpack');
-
-  const normalModuleHook =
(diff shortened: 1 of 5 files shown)


packages/webpack-loader/src/utils.tsx
```diff
@@ -1,28 +1,46 @@
...
-
-  const normalModuleHook =
-    NormalModule && typeof NormalModule.getCompilationHooks !== 'undefined'
-      ? // Webpack 5 flow
-        NormalModule.getCompilationHooks(compilation).loader
-      : // Webpack 4 flow
-        compilation.hooks.normalModuleLoader;
-
-  return normalModuleHook;
+export const setPluginConfiguredOption = (
```
hannahkmolloy, line 11: This function is so complex because the `use` property can be so many types. It can be a string or object or an array of strings + objects

MonicaOlejniczak (dismissed):

MonicaOlejniczak:
@hannahkmolloy Just noticed you need to add a changeset, just run `yarn changeset` and then select the webpack loader as a patch version.

JakeLane (dismissed):

MonicaOlejniczak (approved):

at-nathan (approved):

JakeLane (approved):

PR #1061, opened 2022-01-27
fix: Ensure that any leading comment is preserved at the top of the proc…

pancaspe87:
When running compiled, it is possible that the @compiled/babel-plugin generates the following output:

```
import { forwardRef } from 'react';
import { ax, ix } from "@compiled/react/runtime";
// @flow strict-local
import React, { type Node } from 'react';
```

This results in cryptic babel flow parser errors, since // @flow strict-local is not longer placed at the top of the file. 

This happens when [additional imports](https://github.com/atlassian-labs/compiled/blob/master/packages/babel-plugin/src/babel-plugin.tsx#L92-L100) are inserted during the build. As a result, the node including the flow comment is not longer the first one in the list


packages/babel-plugin/src/utils/preserve-flow-comment.tsx
```diff
@@ -0,0 +1,20 @@
...
+ * dynamically inserted during the build.
+ *
+ * @param path
+ */
+export const preserveFlowComment = (path: NodePath<t.Program>): void => {
+  const leadingComment =
+    path.node.body.length > 0 &&
+    path.node.body[0].leadingComments &&
+    path.node.body[0].leadingComments.length > 0 &&
+    path.node.body[0].leadingComments[0];
```
at-nathan, line 15: I think we can do this in 1 statement with optional chaining. Something like:
`path.node.body?.[0].leadingComments?.[0]`

packages/babel-plugin/src/babel-plugin.tsx
```diff
@@ -85,6 +86,8 @@ export default declare<State>((api) => {
             opts: { importReact: shouldImportReact = true },
           } = state;
 
+          preserveFlowComment(path);
```
at-nathan, line 89: Should we make this flow agnostic? Maybe `preserveLeadingComment`?

packages/babel-plugin/src/utils/preserve-flow-comment.tsx
MonicaOlejniczak: Flow is not guaranteed to be the top comment either, it's in any leading comment

packages/babel-plugin/src/babel-plugin.tsx
MonicaOlejniczak: Yep, the intention for the original ticket was for the solution to be agnostic. Flow was just a direct use-case. You should either change the logic to insert everything after the original `@compiled/react` import before it's removed or insert it first, then move all the following leading comments back up

packages/babel-plugin/src/utils/preserve-flow-comment.tsx
pancaspe87: @at-nathan Yes, we can. When I tried it, I was getting some errors, maybe I had a typo

@MonicaOlejniczak fair point, I had the assumption that we always place flow at the top

packages/babel-plugin/src/babel-plugin.tsx
pancaspe87: @MonicaOlejniczak Just to clarify with everything do you mean the existing imports or comments?

packages/babel-plugin/src/css-prop/__tests__/jsx-pragma.test.tsx
```diff
@@ -44,11 +44,11 @@ describe('local jsx namespace', () => {
     expect(actual).toMatchInlineSnapshot(`
       "/* File generated by @compiled/babel-plugin v0.0.0 */
 
-      import { ax, ix, CC, CS } from \\"@compiled/react/runtime\\";
-
       /** @jsxRuntime classic */
 
       /** @jsx jsx */
+
+      import { ax, ix, CC, CS } from \\"@compiled/react/runtime\\";
```
pancaspe87, line 51: I understand why you are[ recommending](https://github.com/atlassian-labs/compiled/pull/1061#discussion_r794093223) to insert everything after the original `@compiled/react` import as the safest option. This case is an example; ideally we want to keep comments in their related/closest node. 

However, I don't think that we can rely on` @compiled/react` import being always available. I am not very familiar with jsx-pragma, but it seems that `@compiled/react` import is not required when using `/** @jsxImportSource @compiled/react */` ( eg https://compiledcssinjs.com/docs/api-css-prop).

We could insert them after the first import, but this won't work if there are not import declaration within the file.
Hence, I think moving the leading comments back up regardless seems to be the best option. Open to other suggestions though

packages/babel-plugin/src/__tests__/index.test.tsx
```diff
@@ -129,6 +129,37 @@ describe('babel plugin', () => {
     `);
   });
 
+  it('should preserve comments at the top of the processed file before inserting runtime imports', () => {
+    const output = transformSync(
+      `
```
MonicaOlejniczak, line 134: /nit ` on the previous line

packages/babel-plugin/src/utils/preserve-leading-comments.tsx
```diff
@@ -0,0 +1,18 @@
...
+import type * as t from '@babel/types';
+
+/**
+ * Ensure that any leading comment is preserved at the top of the processed file before additional imports are
+ * dynamically inserted during the build.
+ *
+ * @param path
+ */
+export const preserveLeadingComments = (path: NodePath<t.Program>): void => {
+  const leadingComments = path.node.body?.[0]?.leadingComments as any;
```
MonicaOlejniczak, line 11: Can you remove `as any` and place it below like:

```
path.addComments('leading', leadingComments as any[]);
```

This isolates the type conversion to just `path.addComments`

packages/babel-plugin/src/__tests__/index.test.tsx
pancaspe87: - [ ]

pancaspe87: It's prettier placing the ` on a new line

MonicaOlejniczak (approved):

JakeLane (approved):

PR #1070, opened 2022-02-07
fix issues with conditional rules

pancaspe87:
This PR attempts to address two issues with conditional rules, such as:
1. CSS variables are created instead of classes
2. Style inside a pseudo class ( eg : hover) or pseudo element ( eg :before) gets applied to root component instead

In more details:

**1. CSS variables are created instead of classes**
When using conditionals as seen below, Compiled creates a CSS variable.

```
color: ${({ isTrue }) => isTrue ? 'red' :  'green'};

//will output:
style={{ ...style, "--_wopr5k": ix(isTrue ? "red" : "green") }}
```
This is not as efficient as creating 2 classes, as per other conditionals like:
```
${({ isTrue }) => isTrue ? 'color: red' :  'color: green'};

//will output:
className={ax([
  "",
  isTrue ? "_1e0c1kw7" : "_1e0cglyw",
  props.className,
])}
```

Currently we [expect the property to be defined within the expression
](https://github.com/atlassian-labs/compiled/blob/master/packages/babel-plugin/src/utils/css-builders.tsx#L320).

**2. Style inside a pseudo class ( eg : hover) or pseudo element ( eg :before) gets applied to root component instead**
Given:
```
const Component = styled.div`
  color: black;
  
  /* This won't get applied to :hover selector */
  :hover {
    ${({ isTrue }) => isTrue ? 'color: pink' : 'color: red'};
  }
  
  /* This is okay, we get ._30l31dwl:hover{color:var(--_114zjqh)} */
  :hover {
    color: ${({ isTrue }) => isTrue ? 'pink' : 'red'};
  }
`;
```
color: pink and color: red will get applied to Component instead of Component: hover

The second example works because we fall back to creating variables and keep track of any[ interpolation](https://github.com/atlassian-labs/compiled/blob/master/packages/babel-plugin/src/utils/css-builders.tsx#L618)

**THE SOLUTION**
All the issues can be addressed by manipulating the Template Literal so that property, pseudo element, or pseudo class are included within the expression rather than in the correspondent template element. 

Inputs: ( not working as expected )
`color: ({ isTrue }) => isTrue ? 'red' :  'green';` 
`hover: { ${({ isTrue }) => isTrue ? 'color: pink' : 'color: red'};}`

Ouputs:( working as expected )
`$({ isTrue }) => isTrue ? 'color:red' :  'color:green'; `
`${({ isTrue }) => isTrue ? ':hover {color: pink}' : ':hover{color: red}'};`


packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
```diff
@@ -396,17 +396,18 @@ describe('styled component behaviour', () => {
     `);
 
     expect(actual).toIncludeMultiple([
+      '._syaz5scu{color:red}',
```
pancaspe87, line 399: All the outputs should now have classes instead of CSS variables

packages/babel-plugin/src/styled/__tests__/tagged-template-expression.test.tsx
```diff
@@ -656,7 +656,7 @@ describe('styled tagged template expression', () => {
     ]);
   });
 
-  it('should only destructure a prop if hasnt been already', () => {
+  it.skip('should only destructure a prop if hasnt been already', () => {
```
pancaspe87, line 659: TODO: The example provide has now a different output and is not longer relevant. May need to replace it

packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
```diff
@@ -597,17 +603,19 @@ describe('styled component behaviour', () => {
...
     ]);
 
     expect(actual).toInclude(
-      `<C{...props}style={{...style,\"--_b3bzwz\":ix(isPrimary?'blue':'red'),\"--_ubmxli\":ix(isPrimary?'1px solid blue':'1px solid red')}}ref={ref}className={ax([\"_syazmuvv _19it1hus _1wyb1ul9\",props.className])}/>`
+      `ax([\"_1wyb1ul9\",props.isPrimary?\"_syaz13q2\":\"_syaz5scu\",props.isPrimary?\"_19it1nsd\":\"_19it107e\",props.className]`
     );
   });
 
-  it('should apply conditional CSS with nested ternary operators', () => {
+  it.only('should apply conditional CSS with nested ternary operators', () => {
```
at-nathan, line 618: Don't forget to remove this :)

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -299,6 +303,66 @@ const assertNoImportedCssVariables = (
...
+ */
+const moveCssPropertyInExpression = (
+  expression: t.ConditionalExpression,
+  quasi: t.TemplateElement,
+  nextQuasis: t.TemplateElement,
+  deleteValue = true
+): void => {
+  const conditionalPaths: ['consequent', 'alternate'] = ['consequent', 'alternate'];
+  const value = quasi.value.raw;
+  const re = /([^;/]+$)/; // everything after the last ';'
```
at-nathan, line 322: What was `re` stand for?

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -561,6 +625,20 @@ const extractTemplateLiteral = (node: t.TemplateLiteral, meta: Metadata): CSSOut
       return acc + quasi.value.raw + suffix;
     }
 
+    /**
+     * Manipulate the node if the CSS property, pseudo classes or pseudo elements are defined
+     * within a template element rather than in a expression.
+     * Eg :hover { color: ${({ isPrimary }) => (isPrimary ? 'green' : 'red')}; }
+     */
+    if (
```
at-nathan, line 633: We have a function called `extractConditionalExpression` would it make sense to move this logic there? Or does this behaviour only apply to template literals?

packages/css/src/utils/css-property.tsx
```diff
@@ -126,3 +128,17 @@ export function addUnitIfNeeded(
...
+
+/**
+ * Returns `true` if the CSS property is defined within a template element
+ * (the string pieces of the template literal)
+ * rather than in the expression.
+ * Eg color: ${({ isPrimary }) => (isPrimary ? 'green' : 'red')};
+ * @param node
+ */
+export const isCssPropertyInTemplateElement = (node: t.TemplateElement): boolean => {
+  const re = /(\s.[a-z.\-_]*?\:)/; // `property:`
```
at-nathan, line 140: Does this handle the case where the expression is on the next line? For example
```
color:
  ${() => ...};
```

packages/babel-plugin/src/utils/css-builders.tsx
MonicaOlejniczak, line 322: Can you use `split` for this too? It should be more performant and readable than a regex

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -299,6 +303,66 @@ const assertNoImportedCssVariables = (
...
 
+/**
+ * Manipulates template literal so that it resembles
+ * CSS `property: value` declaration where
+ * - before: all the CSS _before_ the interpolation
+ * - after: all the CSS _after_ the interpolation
+ * @param expression
+ * @param quasi
+ */
+const moveCssPropertyInExpression = (
```
MonicaOlejniczak, line 314: Since this is fairly standalone, could this util be moved into its own file?

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -561,6 +625,20 @@ const extractTemplateLiteral = (node: t.TemplateLiteral, meta: Metadata): CSSOut
...
+     * within a template element rather than in a expression.
+     * Eg :hover { color: ${({ isPrimary }) => (isPrimary ? 'green' : 'red')}; }
+     */
+    if (
+      t.isArrowFunctionExpression(nodeExpression) &&
+      t.isConditionalExpression(nodeExpression.body) &&
+      isCssPropertyInTemplateElement(quasi)
+    ) {
+      const nextQuasis = node.quasis[index + 1];
+      moveCssPropertyInExpression(nodeExpression.body, quasi, nextQuasis);
```
MonicaOlejniczak, line 639: Is it necessary to move the property in the expression? Can't you just directly add to `css`, `variables`, or `acc` when you find the property and value?

pancaspe87: @MonicaOlejniczak Currently we [expect the property to be defined within the expression](https://github.com/atlassian-labs/compiled/blob/master/packages/babel-plugin/src/utils/css-builders.tsx#L320).

Otherwise, given 
`color: ${({ isTrue }) => isTrue ? 'red' :  'green'};`
it will output 
```
cssOutput: {
css: [
{
css: 'red',
type: 'unconditional'
}
]}
```
and cause an error.

Are you suggesting to manipulate the css in [here](https://github.com/atlassian-labs/compiled/blob/master/packages/babel-plugin/src/utils/css-builders.tsx#L583), meaning that
- we let `extractConditionalExpression` evaluate all expressions, even without` :`
- add interpolation, if any, directly in the result so that we avoid creating an unconditional css for the blank CSS property?

For example, instead of having this output
image
we have
image

packages/babel-plugin/src/utils/css-builders.tsx
pancaspe87: regex, but I was meant to rename it at the end

packages/babel-plugin/src/utils/css-builders.tsx
pancaspe87: This extra logic is actually needed to allow  `extractConditionalExpression` evaluate all cases since we currently [expect the property to be defined within the expression](https://github.com/atlassian-labs/compiled/blob/master/packages/babel-plugin/src/utils/css-builders.tsx#L320). I think it should be within `extractTemplateLiteral` since we are manipulating both `expressions` and `quasis`

packages/css/src/utils/css-property.tsx
pancaspe87: good point!
- [ ] check if this handle the case where the expression is on the next line

packages/babel-plugin/src/styled/__tests__/behaviour.test.tsx
pancaspe87: - [ ] remove `only`

packages/css/src/utils/css-property.tsx
```diff
@@ -126,3 +128,17 @@ export function addUnitIfNeeded(
...
 }
+
+/**
+ * Returns `true` if the CSS property is defined within a template element
+ * (the string pieces of the template literal)
+ * rather than in the expression.
+ * Eg color: ${({ isPrimary }) => (isPrimary ? 'green' : 'red')};
+ * @param node
+ */
+export const isCssPropertyInTemplateElement = (node: t.TemplateElement): boolean => {
```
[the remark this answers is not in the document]
pancaspe87: yeahhh, I was considering to make it more generic

- [ ] place `isCssPropertyInTemplateElement` inside the babel plugin, unless it is not specific to the babel AST

packages/css/src/utils/css-property.tsx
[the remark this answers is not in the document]
pancaspe87: - [ ] try not to use regex

packages/babel-plugin/src/utils/css-builders.tsx
pancaspe87: I forgot to mention that this particular regex was a quick way to make things work. I am not 100% sure whether we should rely on the semicolon; what if we have a semicolon within the CSS value ( eg background: url ("this contains a semicolon")

- [ ]  try not to use regex

packages/babel-plugin/src/utils/css-builders.tsx
pancaspe87: - [ ] moved into its own file

packages/babel-plugin/src/utils/css-builders.tsx
MonicaOlejniczak: Yeah so the alternative would be to build a new context in meta that is aware of the property from the template literal, handle the logic in `extractCondtionalExpression` and then adjust the output within the `extractTemplateLiteral` logic. I think it complicates nested selectors more, so don't worry about it for the time being.

packages/css/src/utils/css-property.tsx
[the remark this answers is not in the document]
pancaspe87: This won't work if we have
```
:focus { ${({ isTrue }) => isTrue ? 'color: purple' : 'color: orange'}; }
```

since we want to manipulate the node only if the CSS property is within the quasi, for example when we have
```
:focus: color { ${({ isTrue }) => isTrue ? 'purple' : 'orange'}; }
```

pancaspe87: Yep, it does handle that case. 
A similar output is generated in the `quasis` regardless of where the expression is on the next line. 
 
image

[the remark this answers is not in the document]
pancaspe87: Now that we are also including pseudo classes/pseudo elements we can just check if `:` is included.
Sorry for the confusion, the regex was initially designed to address the first issue ( CSS variables are created instead of classes)

packages/babel-plugin/src/styled/__tests__/tagged-template-expression.test.tsx
```diff
@@ -661,14 +661,20 @@ describe('styled tagged template expression', () => {
       import { styled } from '@compiled/react';
 
       const ListItem = styled.div\`
```
pancaspe87, line 663: The expected output is
```
expect(actual).not.toInclude('propz.loading?colors.N100:colors.N200');

    'isLoading ? colors.N20 : colors.N40',
    'l ? colors.N50 : colors.N10',
    'props.loading ? colors.N100 : colors.N200',
  ]);
```

The previous example is not longer relevant; the fix will generate classes instead of inline style to evaluate at run time.

The fix is not applied for certain cases of nested conditional rules; hence, I slightly changed the example to preserve the expected output.

packages/babel-plugin/src/styled/__tests__/tagged-template-expression.test.tsx
```diff
@@ -792,9 +798,14 @@ describe('styled tagged template expression', () => {
       import colors from 'colors';
 
       export const BadgeSkeleton = styled.span\`
-        background-color: \${({ isLoading }) => (isLoading ? colors.N20 : colors.N40)};
```
pancaspe87, line 795: same as [above](https://github.com/atlassian-labs/compiled/pull/1070/files#r808500698)

packages/babel-plugin/src/utils/manipulate-template-literal.ts
```diff
@@ -0,0 +1,143 @@
...
+ *
+ * Returns `true` if
+ * - there is another conditional expression at the root of the current template literal
+ *   with a conditional expression to evaluate OR
+ * - there is a nested logical expression within the current template literal
+ *
+ * @param node
+ * @param meta {Metadata} The current metadata to use to find the parent node
+ */
+export const hasNestedTemplateLiteralsWithConditionalRules = (
```
pancaspe87, line 36: Let me know if there is a better way to traverse the nodes to check nested conditional rules.
I couldn't find anything to traverse bottom-up :(

at-nathan (approved):

packages/babel-plugin/src/utils/manipulate-template-literal.ts
```diff
@@ -0,0 +1,144 @@
...
+ * @param nextQuasis
+ */
+export const moveCssPropertyInExpression = (
+  expression: t.ConditionalExpression,
+  quasi: t.TemplateElement,
+  nextQuasis: t.TemplateElement,
+  deleteValue = true
+): void => {
+  const value = quasi.value.raw;
+  const parts = value.split(';');
```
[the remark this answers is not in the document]
pancaspe87: good point!

packages/babel-plugin/src/utils/manipulate-template-literal.ts
```diff
@@ -0,0 +1,144 @@
...
+      expression[path] = t.stringLiteral(before + pathNode.value + after);
+    } else if (t.isIdentifier(pathNode)) {
+      const identifierExpression = [pathNode];
+      const identifierQuasis = [
+        t.templateElement({ raw: before, cooked: before }),
+        t.templateElement({ raw: after, cooked: after }),
+      ];
+      expression[path] = t.templateLiteral(identifierQuasis, identifierExpression);
+    } else if (t.isMemberExpression(pathNode)) {
+      // We've found a member expression like `colors.N50`
```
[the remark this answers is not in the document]
pancaspe87: TBD at a different moment.

packages/babel-plugin/src/utils/manipulate-template-literal.ts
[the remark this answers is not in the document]
pancaspe87: That shouldn't be affected; with the following line (`const before = parts[parts.length - 1];`) we are interested in the last `;`  

For example, if we have a TemplateElement such as `'\n  :before {\n    content: ';';\n    display: `, we are going to evaluate only the last part ` \n    display: `

I can't think of any case where a CSS value including `;` could be an issue

JakeLane (dismissed):

packages/babel-plugin/src/utils/manipulate-template-literal.ts
```diff
@@ -0,0 +1,148 @@
...
+export const moveCssPropertyInExpression = (
+  expression: t.ConditionalExpression,
+  quasi: t.TemplateElement,
+  nextQuasis: t.TemplateElement,
+  deleteValue = true
+): void => {
+  const value = quasi.value.raw;
+  const parts = value.split(';');
+  const before = parts[parts.length - 1]; // everything after the last ';'. Hence, after the last CSS declaration
+  const nextQuasisExtractedCss = cssAfterInterpolation(nextQuasis.value.raw);
```
MonicaOlejniczak, line 90: Could you leave `cssAfterInterpolation` as private, and use `cssAffixInterpolation` where before is an empty string?

MonicaOlejniczak (approved):

PR #1072, opened 2022-02-10
Negative and positive numbers get put into same css var

csowden-atlassian:
- Added logic to manipulate AST when dealing with negative values 
- Added unit test


packages/babel-plugin/src/__tests__/css-builder.test.tsx
```diff
@@ -49,4 +49,51 @@ describe('css builder', () => {
       "
     `);
   });
+
+  it('Calculates a negative varable seperatly from a positive variable of the same value', () => {
```
pancaspe87, line 53: typo 'varable'

packages/babel-plugin/src/__tests__/css-builder.test.tsx
```diff
@@ -49,4 +49,51 @@ describe('css builder', () => {
...
+
+      const LayoutRight = styled.aside\`
+        margin-right: -\${gridSize * 5}px;
+        margin-left: \${gridSize * 5}px;
+      \`;
+
+      <LayoutRight>Layout Right</LayoutRight>;
+    `);
+
+    expect(actual).toIncludeMultiple([
```
pancaspe87, line 67: We could combine the two. you should be able to do

```
'._18u01xn1{margin-left:var(--_1l3fmvo)}',
'._2hwxsxb8{margin-right:var(--_1cakqv5)},
'--_1cakqv5\\": ix(-gridSize * 5, \\"px\\")',
'--_1l3fmvo\\": ix(gridSize * 5, \\"px\\"),
'ax([\\"_2hwxsxb8 _18u01xn1\\", props.className])
]);
```

pancaspe87:
Hey @csowden-atlassian I don't think we should include the issue number, unless it is on GitHub

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -550,87 +602,101 @@ const extractTemplateLiteral = (node: t.TemplateLiteral, meta: Metadata): CSSOut
...
-      const suffix = meta.context === 'keyframes' ? '' : ';';
-      return acc + quasi.value.raw + suffix;
-    }
+  const literalResult = node.quasis.reduce<string>(
+    (acc: string, quasi: t.TemplateLitteral, index: number): string => {
+      const nodeExpression = node.expressions[index] as t.Expression | undefined;
+
+      // Deal with any negative values such as:
+      // margin: -${gridSize}, top: -${gridSize} etc.
+      if (quasi.value.raw.endsWith('-') && !t.isArrowFunctionExpression(nodeExpression)) {
```
pancaspe87, line 611: If this cannot be placed after https://github.com/atlassian-labs/compiled/blob/master/packages/babel-plugin/src/utils/css-builders.tsx#L563, you may want to check whether nodeExpression exists

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -550,87 +602,101 @@ const extractTemplateLiteral = (node: t.TemplateLiteral, meta: Metadata): CSSOut
...
+        quasi.value.raw = quasi.value.raw.substring(0, quasi.value.raw.length - 1);
+        quasi.value.cooked = quasi.value.cooked.substring(0, quasi.value.cooked.length - 1);
+        convertNegativeCssValuesToUnaryExpression(nodeExpression);
```
pancaspe87, line 614: Any reason why 
```
quasi.value.raw = quasi.value.raw.substring(0, quasi.value.raw.length - 1);
quasi.value.cooked = quasi.value.cooked.substring(0, quasi.value.cooked.length - 1);
```
are not placed inside  `convertNegativeCssValuesToUnaryExpression`?

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -400,6 +400,58 @@ const extractLogicalExpression = (node: t.ArrowFunctionExpression, meta: Metadat
...
 };
 
+/**
+ * Manipulates the AST to ensure that CSS variables are generated correctly for negative values
+ *
+ * consider we have the following:
+ *
+ * const gridSize = 8;
+ *
+ * const LayoutRight = styled.aside`
```
pancaspe87, line 410: Does `margin-left: -${gridSize}px;` work?

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -400,6 +400,58 @@ const extractLogicalExpression = (node: t.ArrowFunctionExpression, meta: Metadat
...
+ * With this manipulation you end up with
+ * -40px for A
+ * 40px for B
+ * -64px for C
+ * 64px for D
+ *
+ * @param nodeExpression Node we're interested in manipulating
+ * @param quasi quasi information about the current node
+ */
+const convertNegativeCssValuesToUnaryExpression = (nodeExpression: t.Expression): void => {
```
pancaspe87, line 436: Can we use a more specific type, eg `BinaryExpression`, or the node could be anything?

packages/babel-plugin/src/utils/css-builders.tsx
at-nathan: I wonder if we handle `margin: -1 * ${gridsSize}px;` right now, if not do we want to potentially handle it in another issue?

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -550,87 +602,101 @@ const extractTemplateLiteral = (node: t.TemplateLiteral, meta: Metadata): CSSOut
...
-      return acc + quasi.value.raw + suffix;
-    }
+  const literalResult = node.quasis.reduce<string>(
+    (acc: string, quasi: t.TemplateLitteral, index: number): string => {
+      const nodeExpression = node.expressions[index] as t.Expression | undefined;
+
+      // Deal with any negative values such as:
+      // margin: -${gridSize}, top: -${gridSize} etc.
+      if (quasi.value.raw.endsWith('-') && !t.isArrowFunctionExpression(nodeExpression)) {
+        quasi.value.raw = quasi.value.raw.substring(0, quasi.value.raw.length - 1);
```
at-nathan, line 612: super nit:  I think `quasi.value.raw.slice(-1)` will also do the job.

packages/babel-plugin/src/utils/css-builders.tsx
pancaspe87: @najjiatlassian we actually discussed that workaround but thought that it was less performant of changing the evaluation at build time

stories/css-prop-dynamic-object.tsx
```diff
@@ -5,6 +5,10 @@ export default {
   title: 'css prop/dynamic object',
 };
 
+// No idea why I need to add this - but storybook crashes for this stroy if it's not here
+// after adding stories/css-negative-margins.tsx
```
at-nathan, line 9: Is there an error throw by Storybook? Might be a longshot but maybe because the exported component is named `Styled` in css-negative-margins and it has a title of `styled/negative margins`

packages/babel-plugin/src/utils/css-builders.tsx
at-nathan: The expression can be anything but we can do a type check on the expression on line 437 instead of `nodeExpression?.left`

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -550,87 +602,101 @@ const extractTemplateLiteral = (node: t.TemplateLiteral, meta: Metadata): CSSOut
...
-
-    if (
-      !nodeExpression ||
-      (t.isArrowFunctionExpression(nodeExpression) && t.isLogicalExpression(nodeExpression.body))
-    ) {
-      const suffix = meta.context === 'keyframes' ? '' : ';';
-      return acc + quasi.value.raw + suffix;
-    }
+  const literalResult = node.quasis.reduce<string>(
+    (acc: string, quasi: t.TemplateLitteral, index: number): string => {
```
at-nathan, line 606: Git Actions is just saying you have a typo `t.TemplateLitteral`

packages/babel-plugin/src/__tests__/css-builder.test.tsx
```diff
@@ -49,4 +49,51 @@ describe('css builder', () => {
...
+      'margin-right:var(--_1cakqv5)',
+      'margin-left:var(--_1l3fmvo)',
+      '"--_1l3fmvo": ix(gridSize * 5, "px")',
+      '"--_1cakqv5": ix(-gridSize * 5, "px"',
+    ]);
+
+    expect(actual).toMatchInlineSnapshot(`
```
at-nathan, line 74: Snapshot is probably necessary, as you've covered related output assertion above.

packages/babel-plugin/src/utils/css-builders.tsx
csowden-atlassian: @at-nathan

The original workaround was:

`margin-left: calc(${gridSize} * - 1px);` 

which did _work_ - but kinda felt a bit hacky (& as @pancaspe87  mentioned - potentially less performant) - the route of modifying the AST directly to become `margin-left: ${-gridSize}px` felt a bit cleaner all around

>If this cannot be placed after https://github.com/atlassian-labs/compiled/blob/master/packages/babel-plugin/src/utils/css-builders.tsx#L563, you may want to check whether nodeExpression exists

Yep - I've moved this will push this with other fixes listed on this PR in the next short while

packages/babel-plugin/src/utils/css-builders.tsx
csowden-atlassian: good catch

packages/babel-plugin/src/utils/css-builders.tsx
csowden-atlassian: I originally was passing the `quasi` in - now i'm returning the new `nodeExpression` from the `convertNegativeCssValuesToUnaryExpression` function


```
if (quasi.value.raw.endsWith('-') && !t.isArrowFunctionExpression(nodeExpression)) {
        quasi.value.raw = trimUnaryOperator(quasi.value.raw);
        quasi.value.cooked = trimUnaryOperator(quasi.value.cooked);
        nodeExpression = convertNegativeCssValuesToUnaryExpression(nodeExpression);
      }
```      

> Based on the comment @param quasi quasi information about the current node

I've gone ahead and removed this param

packages/babel-plugin/src/utils/css-builders.tsx
csowden-atlassian: so close :)

`quasi.value.raw.slice(0, -1).trim()` did the trick

packages/babel-plugin/src/utils/css-builders.tsx
csowden-atlassian: Turns out `TemplateLitteral` was the wrong type 😖  updated to `Templateelement`

packages/babel-plugin/src/__tests__/css-builder.test.tsx
csowden-atlassian: Updated

packages/babel-plugin/src/utils/css-builders.tsx
csowden-atlassian: added logic to handle this

PR #1074, opened 2022-02-11
fix:resolve another Razzle peerDependencies that depends on old versi…

pancaspe87:
`"webpack-dev-server": "^3.11.0"` is another Razzle peer dependencies. Unfortunately it relies on  `node-forge v0.10.0` which is  a vulnerability.
This PR will reesolve the peer dependency to a higher version

image


package.json
```diff
@@ -61,7 +61,8 @@
     "mini-css-extract-plugin": "^1.6.2",
     "nth-check": "^2.0.1",
     "react": "^17.0.2",
-    "typescript": "^4.5.5"
+    "typescript": "^4.5.5",
+    "webpack-dev-server": "^4.7.4"
```
JakeLane, line 65: should this be added in the ssr package?

pancaspe87: I thought you need to add it in both. It is in the SSR package

JakeLane: In theory, packages should only be in the root package.json if the dependency is not for a specific package and is required to run the repo.

pancaspe87: It seems that we'll still download the old version if I don't place it in the root package.json

JakeLane (approved):

Issue #1071, opened 2022-02-09
Codemod/cli is unable to find codemods


danieldelcore:
**Describe the bug**
Compiled default flow (no flags) is unable to find a codemod

**To Reproduce**

1. `npx @compiled/cli`
2. select `codemods`
3. Error

**Expected behavior**
The flow works as expected

![compiled-cli](https://user-images.githubusercontent.com/3030010/153305012-e4600eb9-b952-4c71-9a48-fc7e885dc49f.gif)

pancaspe87:
Hi @danieldelcore,
thanks for taking the time to report the issue.
We'll add it in our backlog and try to address it as soon as we have capabilities.

pancaspe87:
https://github.com/atlassian-labs/compiled/pull/1195 could be a potential workaround.

danieldelcore:
Can confirm the workaround works 👍👍👍👍 Closing

PR #1076, opened 2022-02-11
Refactor file structure in codemods package

MonicaOlejniczak:
## Changes
* Move main logic in `transformers/*` to `index.ts` instead
* Move `utils` only used in transformers to a sibling file
* Move all utils into individual files, and export them in `index.ts`
* Remove any unnecessary `export`
* Rename all TS only files to us `*.ts` instead of `*.tsx`
* Rename `mergeImportSpecifiersAlongWithTheirComments` to `mergeImportSpecifiers`
* Rename some test names for consistency


JakeLane (approved):

JakeLane:
Does this resolve #1071?

at-nathan (approved):

MonicaOlejniczak:
@JakeLane I don't think so, I believe that issue is a problem with the `isLocalRun` check. I can look into fixing that later.

PR #1107, opened 2022-02-22
Move adding require statements to inside babel-plugin-strip-runtime

hannahkmolloy:
Move inserting require statements for atomic css rules to inside babel-plugin-strip-runtime so we can preserve leading comments

When extract is enabled, we add some require statements when a file uses a css rule. Each rule gets its own require statement, like so:
`require("@compiled/webpack-loader/css-loader!@compiled/webpack-loader/css-loader/compiled-css.css?style=._f8pj13q2%3Afocus%7Bcolor%3Ablue%7D");`

We want to add these require statements after any leading comments in the file but we don't have access to babel's `leadingComments` property in the compiled-loader context. So I have moved adding these require statements into the babel-plugin-strip-runtime plugin. This only runs when `extract` is true.


packages/babel-plugin-strip-runtime/src/__tests__/strip-runtime-source-code.test.ts
```diff
@@ -149,16 +155,17 @@ describe('babel-plugin-strip-runtime using source code', () => {
...
-
-        transform(baked, { onFoundStyleRules, run: 'both', runtime });
-
-        expect(onFoundStyleRules).toHaveBeenCalledTimes(1);
-        expect(onFoundStyleRules).toHaveBeenCalledWith([
-          '._1wyb1fwx{font-size:12px}',
-          '._syaz13q2{color:blue}',
+        const actual = transform(baked, {
+          styleSheetName: testStyleSheetName,
+          run: 'extract',
```
hannahkmolloy, line 162: I updated this the be `extract` instead of `both` because this test falls under the `when run in subsequent steps` describe block, so I was pretty sure it was meant to bake and extract separately - we run bake on line 159 so I assumed this was meant to be extract, but let me know if I am incorrect.

packages/babel-plugin-strip-runtime/src/__tests__/strip-runtime-source-code.test.ts
```diff
@@ -187,16 +194,17 @@ describe('babel-plugin-strip-runtime using source code', () => {
...
```
hannahkmolloy, line 201: Same here

packages/babel-plugin-strip-runtime/src/utils/preserve-leading-comments.ts
```diff
@@ -0,0 +1,18 @@
+import type { NodePath } from '@babel/traverse';
```
pancaspe87, line 1: This looks familiar! :)

If it is the same as [packages/babel-plugin/src/utils/preserve-leading-comments.tsx](https://github.com/atlassian-labs/compiled/blob/master/packages/babel-plugin/src/utils/preserve-leading-comments.tsx) , can we place it under `packages/utils/src`?

hannahkmolloy: Ah nice one! Didn't see that folder before - I didn't want to introduce a dependency between the two babel plugins which is why I duplicated it. Will move it to the utils folder 👍

hannahkmolloy: I have done this in a separate PR so I can bump the @compiled/utils package here https://github.com/atlassian-labs/compiled/pull/1109

packages/babel-plugin-strip-runtime/src/index.tsx
```diff
@@ -18,9 +21,24 @@ export default declare<PluginPass>((api) => {
...
+            this.styleRules.forEach((rule) => {
+              // Each found atomic rule will create a new import that uses `@compiled/webpack-loader/css-loader`.
+              // The benefit is two fold:
+              // (1) thread safe collection of styles
+              // (2) caching -- resulting in faster builds (one import per rule!)
+              const params = toURIComponent(rule);
+              path.unshiftContainer(
+                'body',
+                template.ast(
+                  `require("@compiled/webpack-loader/css-loader!@compiled/webpack-loader/css-loader/${this.opts.styleSheetName}.css?style=${params}");`
```
JakeLane, line 36: can we make this bundler agnostic? The parcel plugins will need to handle this behaviour if the require logic is moved to babel

JakeLane (approved):

PR #1108, opened 2022-02-22
Bugfix/negative and positive numbers get put into same css var

csowden-atlassian:


packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -400,6 +400,80 @@ const extractLogicalExpression = (node: t.ArrowFunctionExpression, meta: Metadat
...
+ * With this manipulation you end up with
+ * -40px for A
+ * 40px for B
+ * -64px for C
+ * 64px for D
+ *
+ * @param nodeExpression Node we're interested in manipulating
+ */
+const convertNegativeCssValuesToUnaryExpression = (nodeExpression: t.Expression): t.Expression => {
+  let newNodeExpression: any = { ...nodeExpression } as t.Expression;
```
MonicaOlejniczak, line 436: Can you remove the `any`? Maybe use `nodeExpression.clone()`

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -400,6 +400,80 @@ const extractLogicalExpression = (node: t.ArrowFunctionExpression, meta: Metadat
...
+  if (newNodeExpression?.left) {
```
MonicaOlejniczak, line 437: Why don't you check for the type here?

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -400,6 +400,80 @@ const extractLogicalExpression = (node: t.ArrowFunctionExpression, meta: Metadat
...
+    const { name, start, end } = newNodeExpression.left;
+    newNodeExpression.left = {
```
MonicaOlejniczak, line 439: You could create a helper here, like so:

```
const createNegativeUnaryExpression = (from: Node) => ({
  type: 'UnaryExpression',
  start: node.start,
  end: node.end,
  operator: '-',
  prefix: true,
  argument: {
    type: 'Identifier',
    start: start + 1,
    end: node.end,
    name: node.name,
  },
});

// ...
newNodeExpression.left = createNegativeUnaryExpression(newNodeExpression.left);

newNodeExpression = createNegativeUnaryExpression(newNodeExpression);
```

packages/babel-plugin/src/utils/css-builders.tsx
```diff
@@ -550,87 +624,100 @@ const extractTemplateLiteral = (node: t.TemplateLiteral, meta: Metadata): CSSOut
...
+      ) {
+        const suffix = meta.context === 'keyframes' ? '' : ';';
+        return acc + quasi.value.raw + suffix;
+      }
 
-    const { value: interpolation, meta: updatedMeta } = evaluateExpression(nodeExpression, meta);
+      // Deal with any negative values such as:
+      // margin: -${gridSize}, top: -${gridSize} etc.
+      if (quasi.value.raw.endsWith('-') && !t.isArrowFunctionExpression(nodeExpression)) {
+        quasi.value.raw = trimUnaryOperator(quasi.value.raw);
```
MonicaOlejniczak, line 642: I would just inline this and add a comment above it, i.e.

```
// Remove the '-' from the quasi, as it will soon be moved to a unary expression
quasi.value.raw = quasi.value.raw.slice(0, -1).trim();

MonicaOlejniczak (changes requested):

csowden-atlassian:
moved [here](https://github.com/atlassian-labs/compiled/pull/1120)

Issue #1112, opened 2022-02-25
CSS Parsing Exception: 


onionhammer:
**Describe the bug**
When compiling CSS after upgrading from @compiled/webpack-loader 0.8.0 to 0.8.3, I get the following error on build:

```
Input CSS: {

    border: none;
    outline: none;
    background: #fd5b47;
    color: #fff;
    user-select: none;
    text-align: center;
    text-decoration: none;
    text-transform: uppercase;1.4rem;
    font-size: ;;
    font-weight: 700;
    line-height: 1.4;
    transition: color .2s, background-color .2s;
    border-radius: 2rem;
    box-sizing: border-box;
    appearance: none;
    display: inline-block;;

    :hover, :focus {
        background: #f0a202;
    }

    :disabled, :hover:disabled {
        background: #cdd7d6;
    }
;
  }

  Exception: <css input>:9:31: Unknown word
    at runMicrotasks (<anonymous>)
```

with this source

![image](https://user-images.githubusercontent.com/969938/155735903-e738b8cb-66e5-4dc9-820d-b549a2e2be6d.png)

onionhammer:
This occurs in 0.8.1 and 0.8.2, so something changed between 0.8.0 and 0.8.1 to break this

pancaspe87:
Hi @onionhammer,
thanks for reporting the bug. 

In [@compiled/babel-plugin@0.13.0](https://github.com/atlassian-labs/compiled/releases/tag/%40compiled%2Fbabel-plugin%400.13.0) we released a fix for conditional rules. However, I didn't consider the scenario of having duplicate properties in the same block.

Any particular reason why font-size is declared twice? As a temporary fix I would suggest you to remove `font-size: 1.4rem;`

@MonicaOlejniczak , @JakeLane  do we support duplicate properties in the same block? If so, I think we need to change the logic in [here](https://github.com/atlassian-labs/compiled/blob/master/packages/babel-plugin/src/utils/manipulate-template-literal.ts#L137)

pancaspe87:
Raise a PR

onionhammer:
@pancaspe87 no, no reason that duplicate should be there

Issue #848, opened 2021-10-05
CSS properties with a conditional rule that doesn't specify a default


JakeLane:
**Is your feature request related to a problem? Please describe.**
CSS properties with a conditional rule are currently required to define a default value. We currently throw an error when this assumption is broken as we consider this incorrect syntax.

https://github.com/atlassian-labs/compiled/blob/371ee8975be8608d0ffece49b1b7f0928c18b8c6/packages/babel-plugin/src/styled/index.tsx#L169

However, this does provide a usability concern. See the following example:

```
&:hover {
  css-property: ${props => condition && 'value'};
}
```

To achieve the same result in Compiled we need to do the following

```
${props => condition && {
  ':hover': {
    cssProperty: 'value'
  }
}}
```

**Describe the solution you'd like**
We can simply support the above syntax easily enough but we should consider the impact of supporting "invalid" syntax. We may want to provide a "preferred" linting pattern to ensure the dev experience isn't impacted.

pancaspe87:
Something else to consider https://github.com/atlassian-labs/compiled/issues/1240

PR #1113, opened 2022-02-25
fix: allow duplicate properties in block

pancaspe87:
Fix for https://github.com/atlassian-labs/compiled/issues/1112


at-nathan (approved):

packages/babel-plugin/src/utils/manipulate-template-literal.ts
```diff
@@ -133,9 +133,10 @@ export const moveCssPropertyInExpression = (
   // Remove CSS property so it doesn't get processed again
   if (deleteValue) {
     let nextQuasisNewCSS = nextQuasisExtractedCss.css;
+    const newValue = value.substring(0, value.lastIndexOf(before));
```
at-nathan, line 136: Does this also work if the expression is before the duplicate?
i.e.
```
font-weight: ${(props) => (props.isBolded ? 'bold' : props.isPrimary ? 'normal' : 'light')};
font-weight: initial;
```

at-nathan:
Even with this fix the new conditional evaluation is failing on expressions such as:
```
height: calc(
  ${({ customHeight }) => customHeight != null ?
    `100vh - ${PAGE_LAYOUT_OFFSET_TOP} - ${gridSize * 4 + gridSize * 8}px`
});
```
```
border-color: ${({ color }) => (color && COLOR_MAP[color] ? COLOR_MAP[color] : colors.B400)};
```

packages/babel-plugin/src/utils/manipulate-template-literal.ts
pancaspe87: Hey @JakeLane thanks for adding a story for it! :)

@najjiatlassian yes it does work with the above case.` font-weight: initial;` should be the nextQuasis ( the next template element); hence, it is not affected by this fix

pancaspe87:
Hey @at-nathan re https://github.com/atlassian-labs/compiled/pull/1113#issuecomment-1058785012
I wonder if those issues are related to other known issues such as
- https://github.com/atlassian-labs/compiled/issues/848
- groups of composable styles

packages/babel-plugin/src/utils/manipulate-template-literal.ts
at-nathan: I see, just trying to make sure CSS order doesn't get changed.

at-nathan:
@pancaspe87 Sorry I entered the expression wrongly. It has no default value, I've updated the comment. It's possibly caused by the `calc(...)`.

JakeLane:
> @pancaspe87 Sorry I entered the expression wrongly. It has no default value, I've updated the comment. It's possibly caused by the `calc(...)`.

I think the calc issue already exists - we should probably raise an issue for it. From my understanding, unitless (e.g. number types) sizes do not get the implicit `px` added and cause invalid CSS.

No luck reproducing the failure on the border-color issue

at-nathan:
I've put up another PR that will address the current problems introduced with conditional expression optimization. It will also fix the problem this PR is addressing
https://github.com/atlassian-labs/compiled/pull/1152

at-nathan:
Closing this as we just merged
https://github.com/atlassian-labs/compiled/commit/2ad385cfadd0a8fb3021b3777d402acc6b9c2eea

PR #1140, opened 2022-03-09
Add no-tagged-template-expression eslint rules

MonicaOlejniczak:
These changes introduce three new recommended and fixable eslint rules that disallow the use of tagged template expressions:
* `no-css-tagged-template-expression`
* `no-keyframes-tagged-template-expression`
* `no-styled-tagged-template-expression`

This is part of the longer-term vision to scale large codebases by ensuring styles are type-safe, can be autocompleted, and are easier to statically parse (which reduces the total amount of bugs).

## Changes
* Add new eslint rules and parameterised tests
* Update `@compiled/eslint-plugin` documentation

## Limitations
Comments are not fixable at this stage as it gets pretty complex and requires a large refactor, though it is quite trivial to manually change the incorrectly generated output.


packages/eslint-plugin/src/test-utils.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export const createDeclarationInvalidTestCases = (
+  test: RuleTester.InvalidTestCase,
+  name: string,
+  replaceCode: (code: string, prefix: string) => string,
+  replaceOutput: (output: string, prefix: string) => string
+): RuleTester.InvalidTestCase[] => {
+  const filename = basename(test.filename!);
+  const exportDefaultPrefix = 'export default ';
+  const exportNamedPrefix = `export const ${name} = `;
```
at-nathan, line 47: Do we have to assume there is trailing whitespace after the `=`?  If consumer doesn't use a code formatter or right side is on a new line will replacement still work?

packages/eslint-plugin/src/rules/no-css-tagged-template-expression/__tests__/rule.test.ts
```diff
@@ -0,0 +1,324 @@
...
+import {
+  createAliasedInvalidTestCase,
+  createDeclarationInvalidTestCases,
+  tester,
+} from '../../../test-utils';
+import { noCssTaggedTemplateExpressionRule } from '../index';
+
+type InvalidTestCase = Omit<RuleTester.InvalidTestCase, 'errors'>;
+
+const lastCssCallExpressionRe = /css\((?![\s\S]*css\()/g;
```
at-nathan, line 12: nit: Does `Re` stand for `Regex`?

MonicaOlejniczak: Yeah, it's a pretty common shorthand

packages/eslint-plugin/src/test-utils.ts
MonicaOlejniczak: Yes, it will still work because the rules detect the tagged template expression directly. The declaration variants here simply validate that the indentation is correct, and the object properties start just after `export`, `const`, or the tagged template.

packages/eslint-plugin/tsconfig.json
```diff
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "rootDir": "src",
     "outDir": "dist",
-    "module": "commonjs"
+    "module": "commonjs",
+    "target": "es2017"
```
JakeLane, line 7: Do we want to target an older ES version?

MonicaOlejniczak: I used `es2017` to get around TS errors with `.entries()` and because it's also used in the webpack loader package. Let me know if that's a problem.

JakeLane: According to https://node.green/, our minimum version supports all the features we need, so all good

packages/eslint-plugin/src/rules/no-styled-tagged-template-expression/README.md
```diff
@@ -0,0 +1,65 @@
...
+  {
+    color: (props) => props.color,
+  },
+  (props) => (props.disabled ? 'opacity: 0.8' : 'opacity: 1')
+);
+```
+
+## Limitations
+
+- Comments are not fixable
```
JakeLane, line 65: What happens if there's a comment? The best bet would be to skip fixing that usage, as we don't want to generate invalid code.

JakeLane (approved):
Looks good! Just left a comment about what we do in a known non-working scenario. Do you know if there's any other edge cases that will break when fixed?

packages/eslint-plugin/src/rules/no-styled-tagged-template-expression/README.md
MonicaOlejniczak: Yeah it can generate invalid code, I can skip over them if you like

JakeLane (approved):

PR #1145, opened 2022-03-10
Use common transform test util

MonicaOlejniczak:
## Changes
* Use common `transform` test util in `@compiled/babel-plugin`
* Fix indentation of all tests
* Fix pretty logic in `transform`
* Remove unnecessary `import React from 'react';` in tests
* Remove some redundant tests
* Update `transform` options, and make `snippet` off by default


.changeset/grumpy-laws-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+'@compiled/babel-plugin': patch
```
JakeLane, line 2: I don't think we need to bump babel plugin for this, since the changes are just tests

JakeLane (approved):

.changeset/grumpy-laws-rule.md
MonicaOlejniczak: Looks like changeset changed their recommendations, before it mentioned to include a changeset for everything even if it's not released

JakeLane (approved):

PR #1152, opened 2022-03-15
Optimize conditional expressions

### LINKED ISSUE #1112: CSS Parsing Exception:

at-nathan:
Resolves #1112 

Another attempt at evaluating conditional CSS into separate CSS classes.
This PR limits the optimization to scenarios like:
```
color: ${({ isSelected }) => isSelected ? 'green' : 'black' };
```
Scenarios where conditionals have sibling expressions or apart of strings will fallback to CSS variables. For example:
```
/* Technique cannot be applied to conditions like this */
height: calc(100% - ${identifier} - ${conditionalExpression});

/* Strings are unpredictable and can break parsing */
content: 'Code example: ${conditionalExpression};';
```

Additional fixes:
- Conditional CSS not working in nested selectors
- Conditional CSS breaking in selectors with static styles
- Conditional CSS breaking with duplicate CSS property
- Conditional CSS breaking when condition test contains an object


packages/babel-plugin/src/utils/css-builders.ts
```diff
@@ -656,13 +659,12 @@ const extractTemplateLiteral = (node: t.TemplateLiteral, meta: Metadata): CSSOut
...
+    const doesExpressionContainCssBlock =
       t.isObjectExpression(interpolation) ||
       isCompiledCSSTaggedTemplateExpression(interpolation, meta.state) ||
-      isCompiledCSSCallExpression(interpolation, meta.state) ||
-      (t.isArrowFunctionExpression(nodeExpression) &&
-        t.isConditionalExpression(nodeExpression.body))
-    ) {
+      isCompiledCSSCallExpression(interpolation, meta.state);
+
+    if ((!isMidStatement && doesExpressionContainCssBlock) || doesExpressionHaveConditionalCss) {
```
at-nathan, line 667: Without the `isMidStatement` this was breaking on styles like:
`${({ color }) => (color && COLOR_MAP[color] ? COLOR_MAP[color] : colors.B400)};`
Because it `color` is dynamic, Compiled can only work out what `COLOR_MAP` is. It saw it was an `ObjectExpression` and spread the whole Object into the stylesheet.

We might want to consider returning the original node/expression if we can't evaluate it the whole way through for certain expression types.

JakeLane (approved):

JakeLane (approved):

Issue #1137, opened 2022-03-07
"You forgot to add the 'CompiledExtractPlugin'"


onionhammer:
**Describe the bug**
After upgrading from 0.8.0 of  `@compiled/webpack-loader` to 0.8.3 or 0.8.4, I now receive this error

```
Failed to compile.

██████╗ ██████╗ ███╗   ███╗██████╗ ██╗██╗     ███████╗██████╗
██╔════╝██╔═══██╗████╗ ████║██╔══██╗██║██║     ██╔════╝██╔══██╗
██║     ██║   ██║██╔████╔██║██████╔╝██║██║     █████╗  ██║  ██║
██║     ██║   ██║██║╚██╔╝██║██╔═══╝ ██║██║     ██╔══╝  ██║  ██║
╚██████╗╚██████╔╝██║ ╚═╝ ██║██║     ██║███████╗███████╗██████╔╝
 ╚═════╝ ╚═════╝ ╚═╝     ╚═╝╚═╝     ╚═╝╚══════╝╚══════╝╚═════╝

  @compiled/webpack-loader

  You forgot to add the 'CompiledExtractPlugin' plugin (i.e `{ plugins: [new CompiledExtractPlugin()] }`), please read https://compiledcssinjs.com/docs/css-extraction-webpack

```

My webpack.config.js **does** have the `CompiledExtractPlugin`. It's essentially a standard react ejected webpack.config.js, plus compiled library, following your docs.

**Expected behavior**
Should continue to work largely as before upgrading from 0.8.0

![image](https://user-images.githubusercontent.com/969938/157060100-e444090f-4997-4656-89af-cf90cded3d29.png)

![image](https://user-images.githubusercontent.com/969938/157060170-5f3880d8-58c9-4605-9422-fc976855a6f8.png)

JakeLane:
@hannahkmolloy could this be related to https://github.com/atlassian-labs/compiled/commit/530b52d6b7e776a8a4605797fc84c2a91fee0951 ?

hannahkmolloy:
@JakeLane It seems likely. However, I am unable to reproduce. @onionhammer Would you mind posting your whole webpack config please?

onionhammer:
Sure @hannahkmolloy , here's a link to a gist

https://gist.github.com/onionhammer/e3495a0da97177c139bb98d1b59d2fb5#file-webpack-config-js

hannahkmolloy:
[...]

PR #1165, opened 2022-03-23
Account for nested rules in webpack config 

hannahkmolloy:
This PR refactors setting the plugin configured option on the compiled webpack loader and accounts for nested rules in the webpack config https://webpack.js.org/configuration/module/#nested-rules

Fixes this bug https://github.com/atlassian-labs/compiled/issues/1137

### DIFF
--- a/packages/webpack-loader/src/utils.ts
+++ b/packages/webpack-loader/src/utils.ts
@@ -1,6 +1,27 @@
 import type { Compilation as CompilationType, Compiler, sources, RuleSetRule } from 'webpack';
 
 /**
+ * Helper function to set plugin configured option on the @compiled/webpack-loader
+ *
+ * @param use
+ * @param pluginName
+ */
+const setOptionOnCompiledWebpackLoader = (use: RuleSetRule['use'], pluginName: string) => {
+  if (!use || !Array.isArray(use) || !use.length) {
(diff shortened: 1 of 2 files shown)


packages/webpack-loader/src/utils.ts
```diff
@@ -1,5 +1,26 @@
 import type { Compilation as CompilationType, Compiler, sources, RuleSetRule } from 'webpack';
 
+/**
+ * Helper function to set plugin configured option on the @compiled/webpack-loader
+ *
+ * @param use
+ * @param pluginName
+ */
+const setOptionOnCompiledWebpackLoader = (use: RuleSetRule['use'], pluginName: string) => {
+  if (!use || !Array.isArray(use) || !use.length) {
```
at-nathan, line 10: nit: Could this be simplified with `!use?.length || !Array.isArray(use)`?

at-nathan (approved):

JakeLane (approved):

packages/webpack-loader/src/utils.ts
hannahkmolloy: It could be but I think how it is now might be better for readability (i.e. checking it is an array before checking the length property)

Issue #1187, opened 2022-04-05
Restrict mixing of CSS types


JakeLane:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #1188, opened 2022-04-05
Adjust flow types to handle function calls

JakeLane:
Had a whack at improving our types to get us to a better state.

I raised #1187 as a future improvement, where we should move towards a much stricter type setup.


packages/react/src/class-names/index.ts
```diff
@@ -3,11 +3,18 @@ import type { ReactNode, CSSProperties } from 'react';
 import type { BasicTemplateInterpolations, CssFunction } from '../types';
 import { createSetupError } from '../utils/error';
 
-export type Interpolations = (BasicTemplateInterpolations | CssFunction | CssFunction[])[];
+export type Interpolations = (
+  | BasicTemplateInterpolations
+  | CssFunction<unknown>
```
at-nathan, line 8: nit: Is it possible for `Interpolations` to have generics so we don't have to use `unknown` here?

at-nathan (approved):

packages/react/src/class-names/index.ts
JakeLane: Yep, will do

at-nathan (dismissed):

packages/react/src/styled/__tests__/types.test.js.flow
```diff
@@ -1,30 +1,52 @@
...
-  const StyledTaggedTemplateExpression = styled.div({ fontSize: 12 }, `font-size: 12px`);
-  return <StyledTaggedTemplateExpression>red text</StyledTaggedTemplateExpression>;
+  const StyledComponent = styled.div({ fontSize: 12 }, `font-size: 12px`);
+  return <StyledComponent>red text</StyledComponent>;
+};
+
+// Dynamic function calls
+() => {
+  const StyledComponent = styled.div({
+    backgroundColor: token('elevation.surface', 'white'),
```
MonicaOlejniczak, line 49:
```suggestion
    backgroundColor: token('elevation.surface', 'black'),
```

packages/react/src/styled/__tests__/types.test.js.flow
```diff
@@ -1,30 +1,52 @@
 // @flow strict-local
```
MonicaOlejniczak, line 1: This file should contain some conditional styles

packages/react/src/class-names/index.ts
```diff
@@ -3,11 +3,18 @@ import type { ReactNode, CSSProperties } from 'react';
 import type { BasicTemplateInterpolations, CssFunction } from '../types';
 import { createSetupError } from '../utils/error';
 
-export type Interpolations = (BasicTemplateInterpolations | CssFunction | CssFunction[])[];
+export type Interpolations<TProps> = (
+  | BasicTemplateInterpolations
```
MonicaOlejniczak, line 7: `CssFunction` already contains `BasicTemplateInterpolations`

packages/react/src/class-names/index.ts
```diff
@@ -3,11 +3,18 @@ import type { ReactNode, CSSProperties } from 'react';
...
+  | CssFunction<TProps>
+  | CssFunction<TProps>[]
+)[];
 
-export interface ClassNamesProps {
+export interface ClassNamesProps<TProps> {
   children: (opts: {
-    css: (css: CssFunction | CssFunction[], ...interpolations: Interpolations) => string;
+    css: (
```
MonicaOlejniczak, line 14: Can this be an interface like in `styled` so that you separate the tagged template usage from the object call expression?

Or could you utilise the types from `css` instead? Then just change the return type

packages/react/src/css/index.js.flow
```diff
@@ -1,35 +1,36 @@
...
+ * @example ```
+ * const redText = css`
+ *   color: red;
+ * `;
+ *
+ * <div css={redText} />
+ * ```
  */
-declare export default function css(_css: CSSProps): CSSProps;
+declare export default function css<TProps>(
```
MonicaOlejniczak, line 32: Doesn't flow complain when you export two overloaded functions?

JakeLane: seems to work fine from my testing - I'll double-check while fixing the other issues

packages/react/src/class-names/index.ts
JakeLane: yeah the CSS types approach seems pretty elegant, will try that

packages/react/src/css/__tests__/types.test.js.flow
```diff
@@ -12,6 +12,31 @@ import { css } from '@compiled/react';
 
 // Object call expression
 (): React$Element<'div'> => {
-  const styles = css({ color: 'red' });
+  const styles = css({ color: 'yellow' });
```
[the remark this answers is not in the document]
JakeLane: Just a thing I forgot to remove from testing (flow server wasn't happy for some reason :/)

packages/react/src/styled/index.ts
```diff
@@ -1,60 +1,41 @@
...
-  <TProps extends unknown>(
-    // Allows either string or object (`` or ({}))
-    css: CssObject<TProps> | TemplateStringsArray,
-    ...interpolations: Interpolations<TProps>
-  ): React.ComponentType<TProps & StyledProps & TInheritedProps>;
+export interface StyledComponent<ComponentProps extends unknown> {
+  // Allows either string or object (`` or ({}))
+  // We disable the ban types rule here as we need to join the empty object default with other props
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  <TProps = {}>(...css: Interpolation<TProps>[]): React.ComponentType<
```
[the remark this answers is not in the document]
JakeLane: We need to support strings here which is why it was so generic before. I'll see if I can make the types more precise.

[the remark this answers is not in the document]
JakeLane: See https://github.com/atlassian-labs/compiled/blob/0abe64ee46b89c69bf0c9679bb9142c953916758/packages/css-in-js/src/styled/__tests__/index.test.tsx

packages/react/src/styled/__tests__/types.test.js.flow
```diff
@@ -1,30 +1,70 @@
...
+  });
+
+  return <StyledComponent>red text</StyledComponent>;
+};
+
+// Object call expression with conditional
+() => {
+  const StyledComponent = styled.div({
+    fontSize: ({ big }) => `${big ? 50 : 12}px`,
+    width: ({ big }) => (big ? 500 : 120),
```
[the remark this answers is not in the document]
JakeLane: I implemented something similar but wasn't sure exactly how this would work. Does it mean:
```
...(({ big }) => big && { height: '100px' })
```?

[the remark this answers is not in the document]
Comment (MonicaOlejniczak) [reply packages/react/src/styled/__tests__/types.test.js.flow:45]: Nah, it would be like this:

```
styled.div<{ big: number, children: Node }>(
  {
    fontSize: ({ big }) => `${big ? 50 : 12}px`,
    width: ({ big }) => (big ? 500 : 120),
  },
  (({ big }) => big && { height: '100px' }),
);
```

MonicaOlejniczak (approved):

PR #1205, opened 2022-04-29
Refactor expand flex shorthands to handle the CSS spec correctly

### LINKED ISSUE #1204: flex property expansion in CSS is handled incorrectly
**Describe the bug**
flex shorthands are not handled correctly and cause incorrect CSS.

**To Reproduce**
With the following CSS

```
css({
    flex: '1 1 auto'
});
```

The flex property is dropped entirely.

**Expected behavior**
The flex property should be expanded into atomic rules correctly.

**Additional context**
Workaround is to not use the `flex` property, instead using the direct CSS rules

```
css({
    flexGrow: 1,
    flexShrink: 1,
    flexBasis: 'auto',
});
```

JakeLane:
Resolves #1204

The current implementation of flex shorthand expansion is lossy on values. We end up removing valid fields.

This change follows the [csswg spec](https://drafts.csswg.org/css-flexbox-1/#flex-property) and fixes this functionality. Note that we probably don't want to be super strict at this stage as there's no benefit to adding difficult-to-maintain business logic that could be brought in from an external source for validation.


packages/css/src/plugins/expand-shorthands/utils.ts
```diff
@@ -15,38 +15,48 @@ export const isColor = (node: Node): boolean => {
   return (node.type === 'word' || node.type === 'func') && node.isColor;
 };
 
+const widthUnits = new Set([
```
pancaspe87, line 18: Hey @JakeLane do we know how these units were picked? There may be more units to add https://developer.mozilla.org/en-US/docs/Web/CSS/length

JakeLane: no idea, probably good to update - will do 👍

pancaspe87: Thanks for rearranging them in alphabetical order! :)

pancaspe87 (approved):

PR #1193, opened 2022-04-12
Bug - shadow variable clash with destructured props

harshmalde:
Issue Description:
When shadow variable name and deconstructed `props` property are same, babel is unable to differentiate between them (since both the property will be of type `Identifier` and you can have a look at the storybook that has been created with this PR).

Implementation:
This pull request helps to reconstruct the destructed props in the Arrow function.

Bug:
[APP-286](https://product-fabric.atlassian.net/browse/APP-286)

PR #1214, opened 2022-05-10
Update codeowners

JakeLane:
Removing members no longer actively working on compiled 😢

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,3 @@
-* @MonicaOlejniczak @JakeLane @at-nathan @blackheart-atlassian @pancaspe87 @TimeRaider @heliam1 @hannahkmolloy @harshmalde
+* @MonicaOlejniczak @JakeLane @at-nathan @pancaspe87
 **/package.json
 yarn.lock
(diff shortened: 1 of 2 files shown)


.github/CODEOWNERS
```diff
@@ -1,3 +1,3 @@
-* @MonicaOlejniczak @JakeLane @at-nathan @blackheart-atlassian @pancaspe87 @TimeRaider @heliam1 @hannahkmolloy @harshmalde
+* @MonicaOlejniczak @JakeLane @at-nathan @pancaspe87
```
pancaspe87, line 1: should we add Liam Ma?

.github/CODEOWNERS
```diff
@@ -1,3 +1,3 @@
...
 **/package.json
 yarn.lock
```
pancaspe87, line 3: Should we put the [console.log back to avoid the warnings below](https://github.com/atlassian-labs/compiled/pull/1193#discussion_r860519974)?

.github/CODEOWNERS
JakeLane: Just done 👍

.github/CODEOWNERS
JakeLane: I've suppressed the eslint warning instead which should stop the CI action prompt

pancaspe87 (approved):

PR #1218, opened 2022-05-16
Issue/1216 prevent side effects

pancaspe87:
**Background**
Potential side-effects in SSR have been observed when enabling stylesheet extraction. It seems that these are caused by the inclusion of extra modules in the bundle and the creation of an additional [chunk group](https://github.com/atlassian-labs/compiled/blob/master/packages/webpack-loader/src/extract-plugin.ts#L64).
While the [additional requires](https://github.com/atlassian-labs/compiled/blob/master/packages/babel-plugin-strip-runtime/src/index.ts#L35) created by @compiled/babel-plugin-strip-runtime are needed at the client level, they shouldn't be included in the SSR bundle.

**Changes**
With this PR we introduce the ability to configure whether
- we want to include Compiled[ require](https://github.com/atlassian-labs/compiled/pull/1218/files#diff-adc32ff3f35c91656f29b6a64ffc17ac245de26b38ee7ff427b4021738a7666cL25) in the bundle
- we want to have the extra [cacheGroup](https://github.com/atlassian-labs/compiled/pull/1218/files#diff-ba83ed35c316ae395037590865d54fd06a89073cd4ff97313b1c8e5b2b9a75c1L64) chunk

This PR should be approved along with https://github.com/compiled/website/pull/561 that documents the additional options in https://compiledcssinjs.com/docs/


packages/webpack-loader/src/extract-plugin.ts
```diff
@@ -44,6 +44,7 @@ const forceCSSIntoOneStyleSheet = (compiler: Compiler) => {
       // We merge only CSS from Compiled.
       test: /css-loader\/compiled-css\.css$/,
       enforce: true,
+      priority: Infinity,
```
pancaspe87, line 47: This is not related to the SSR side effects, but it is an additional patch we had to do to make stylesheet extraction work

JakeLane (approved):

PR #1225, opened 2022-05-27
Transform babel plugin config

### LINKED ISSUE #1171: [Feature Request] Provide babel configuration through webpack loader config

The feature request is similar to #1143.

Next.js 12 internally uses swc based compiler, too. However, in order for `compiled`'s `babel.parseAsync` to recognize tsx files, I need `@babel/plugin-syntax-typescript` and `@babel/plugin-syntax-jsx`.

Currently, `compiled` will let babel load the config through files. But if I supplement a `.babelrc`, Next.js will automatically fall back to babel for compiling.

In order to leverage Next.js's swc compiler, the absence of any babel configuration file is required.

Add `babel` to `compiled-loader`'s configuration, so I can provide babel configuration directly, without any `.babelrc` files.

JakeLane:
Resolves #1171 
Closes #1143

Simple change to add configuration for transform babel plugins to be ran before compiled.

This enables consumers to run pre-requisite babel plugins before the Compiled babel plugin is ran. This is important in cases where the source code is not pure JavaScript or TypeScript, such decorators (see the examples).

Long term, we will want to refactor to decouple the babel plugin from the loaders. This will mean that babel projects can configure the babel plugin normally without the painful workarounds we add into the loaders. We can probably also provide a way to run the babel plugin in projects that do not use babel loaders.


examples/webpack/webpack.config.js
```diff
@@ -36,7 +37,8 @@ module.exports = {
               extract: extractCSS,
               importReact: false,
               extensions: ['.js', '.jsx', '.ts', '.tsx', '.customjsx'],
-              babelPlugins: ['typescript'],
+              parserBabelPlugins: ['typescript'],
```
pancaspe87, line 40: Can we update the [docs](https://compiledcssinjs.com/docs/pkg-webpack-loader) with these extra configurable options?

examples/parcel/.compiledcssrc
```diff
@@ -7,8 +7,16 @@
     ".tsx",
     ".customjsx"
   ],
-  "babelPlugins": [
+  "parserBabelPlugins": [
```
pancaspe87, line 10: Can we update the [docs](https://compiledcssinjs.com/docs/pkg-parcel-transformer) with these extra configurable options?

liamqma (approved):

pancaspe87 (approved):

Issue #1237, opened 2022-06-02
`@compiled/react` compatibility issues with Vite


bienzaaron:
**Describe the bug**
Not sure if this is necessarily a bug with `@compiled/react` or if it's an issue with the way Vite pre-bundled dependencies, but I'm seeing my React app failing to render and logging the following console error when attempting to use `@compiled/react` and `@compiled/babel-plugin` with Vite:
```
Uncaught SyntaxError: The requested module '/node_modules/.vite/deps/@compiled_react_jsx-dev-runtime.js?v=e03c1b82' does not provide an export named 'jsxDEV' (at main.jsx:2:1)
```

I think it has something to do with the way things are re-exported from `react/jsx-dev-runtime` (wildcard export), because if I swap out `@compiled/react` with `@emotion/react`, everything works as expected - I believe because they [explicitly export `jsxDEV`](https://github.com/emotion-js/emotion/blob/main/packages/react/src/jsx-dev-runtime.js#L8).

I've tried excluding `@compiled/react` from vite's pre-bundling as well without success.

**To Reproduce**
See console error:
https://stackblitz.com/edit/vitejs-vite-1z9eeu?file=package.json,vite.config.js,index.html,src%2Fmain.jsx

**Expected behavior**
React app should render and apply styles appropriately.

**Desktop:**

- OS: [e.g. iOS] MacOS 11.6.6
- Browser [e.g. chrome, safari] Chrome
- Version [e.g. 22] 101

vidarc:
At least for me, I was able to resolve this by removing the jsxPragma I had specified at the top of my components using compiled and added a `"jsxImportSource": "@compiled/react",` to my tsconfig

tsconfig:
```json
{
  "extends": "@tsconfig/vite-react/tsconfig.json",
  "include": ["src"],
  "compilerOptions": {
    "baseUrl": "./src/client",
    "jsx": "react-jsx",
    "jsxImportSource": "@compiled/react",
    "skipLibCheck": true
  }
}
```

vite.config.js
```javascript
import react from '@vitejs/plugin-react';
import { defineConfig } from 'vite';

export default defineConfig({
  root: 'src',
  build: {
    outDir: '../dist/client',
    emptyOutDir: true,
  },
  plugins: [
    react({
      babel: {
        plugins: ['@compiled/babel-plugin'],
      },
    }),
  ],
});
```

bienzaaron:
[...]

PR #1246, opened 2022-06-18
fix for jsx-runtime when used with vitejs

vidarc:
fixes https://github.com/atlassian-labs/compiled/issues/1237

vitejs appears to optimize the needed functions away. have tried by manually adjusting the file at dist/esm/jsx/jsx-dev-runtime.js and this works

### DIFF
--- a/.changeset/forty-plants-return.md
+++ b/.changeset/forty-plants-return.md
@@ -1,0 +1,5 @@
+---
+'@compiled/react': patch
+---
+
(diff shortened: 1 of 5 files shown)


JakeLane:
Hi @vidarc, thanks for the contribution!

As you've noted in the last commit, the types seem to break down when importing from the react module. I think we need to resolve this somehow as we shouldn't intentionally export broken types - that will cause lots of consumers builds to break.

vidarc:
resolved on my end without this.

PR #1248, opened 2022-07-07
Add support for as expressions in TypeScript

JakeLane:
Currently we cannot handle files that use `as` on expressions we want to use in compiled.

For example:
```
const style = {
  backgroundColor: 'red'
} as const;
  "input": style,
});
```

### DIFF
--- a/packages/babel-plugin/src/__tests__/expression-evaluation.test.ts
+++ b/packages/babel-plugin/src/__tests__/expression-evaluation.test.ts
@@ -454,5 +454,46 @@ describe('import specifiers', () => {
         'ax(["_19pk19vg"])',
       ]);
     });
+
+    it('statically evaluates a TS const expression', () => {
+      const actual = transform(
+        `
+        import '@compiled/react';
+
+        const styles = { color: 'red' } as const;
+
+        <div css={{ ...styles }} />;
+      `,
+        {
+          parserBabelPlugins: ['typescript', 'jsx'],
+        }
+      );
+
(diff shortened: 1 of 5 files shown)


packages/babel-plugin/src/__tests__/expression-evaluation.test.ts
```diff
@@ -454,5 +454,46 @@ describe('import specifiers', () => {
...
+      `,
+        {
+          parserBabelPlugins: ['typescript', 'jsx'],
+        }
+      );
+
+      expect(actual).toIncludeMultiple(['._syaz5scu{color:red}', 'ax(["_syaz5scu"])']);
+    });
+
+    it('statically evaluates a TS const expression in a resolved binding', () => {
```
pancaspe87, line 475: Hey @JakeLane what's the difference with the test above? Are you just testing with different Compiled APIs?

pancaspe87 (approved):

packages/babel-plugin/src/utils/evaluate-expression.ts
```diff
@@ -128,41 +128,51 @@ export const evaluateExpression = (
   let value: t.Node | undefined | null = undefined;
   let updatedMeta: Metadata = meta;
 
+  // TypeScript AST nodes can be skipped as we don't care about types
+  let targetExpression = expression;
```
at-nathan, line 132:
```suggestion
  const targetExpression = t.isTSAsExpression(expression) ? expression.expression : expression;
```

at-nathan (approved):

Issue #1263, opened 2022-07-14
Add configurable options to optimise CSS


pancaspe87:
**Is your feature request related to a problem? Please describe.**
At current Compiled [applies additional plugins](https://github.com/atlassian-labs/compiled/blob/master/packages/css/src/plugins/normalize-css.ts#L55)) to optimise CSS based on the value of `process.env.NODE_ENV` .
However, if the variable is `undefined` in one of the environment this could cause a mismatch between server and browser if they have different configuration when stylesheet extraction is enabled.

For example, given margin-top: 32px; this will become margin-top: 2pc; in production mode.
Because of these different CSS values, Compiled will generate different CSS classes and cause browser/server not to be on sync

PR #1266, opened 2022-07-15
Added configurable options to optimize CSS

pancaspe87:
In the past it was decided to not let consumers "opt in" this behaviour ( see[ comment](https://github.com/atlassian-labs/compiled/pull/427#discussion_r547613478)). However, we observe a mismatch between browser and SSR if the variable is not defined in one of those environments and extraction is enabled.

I set default value of `optimizeCss` to` true` so that existing consumers won't be affected by the change in production. However, they may need to either update their tests, if using toHaveCompiledCss to test CSS values that can be optimised, or opt out the behaviour in their tests and storybook

See https://github.com/atlassian-labs/compiled/issues/1263

Docs have been updated, see [PR]( https://github.com/compiled/website/pull/605/files)


pancaspe87:
- [ ] Update all failing tests due to expecting a different classes when applying all the CSS plugins.

fixtures/babel-component/.babelrc
```diff
@@ -3,5 +3,5 @@
     ["@babel/preset-env", { "targets": { "browsers": "last 1 version" } }],
     ["@babel/preset-react", { "runtime": "automatic" }]
   ],
-  "plugins": [["@compiled/babel-plugin", { "importReact": false }]]
+  "plugins": [["@compiled/babel-plugin", { "importReact": false, "optimizeCss": false }]]
```
pancaspe87, line 6: Needed to test the Webpack example -` yarn start:webpack`
When set to false, units won't be optimised. Eg `32px` won't be changed to `2pc`

Screen Shot 2022-08-05 at 4 27 44 pm

packages/babel-plugin/src/css-prop/__tests__/object-literal.test.ts
```diff
@@ -125,7 +125,7 @@ describe('css prop object literal', () => {
     `);
 
     expect(actual).toInclude(
-      'background-image:linear-gradient(45deg,gray 25%,transparent 25%),linear-gradient(-45deg,gray 25%,transparent 25%),linear-gradient(45deg,transparent 75%,gray 75%),linear-gradient(-45deg,transparent 75%,gray 75%)'
+      'background-image:linear-gradient(45deg,gray 25%,transparent 0),linear-gradient(-45deg,gray 25%,transparent 0),linear-gradient(45deg,transparent 75%,gray 0),linear-gradient(-45deg,transparent 75%,gray 0)'
```
pancaspe87, line 128: These values are indeed the same 
Screen Shot 2022-08-05 at 4 33 02 pm

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.ts
```diff
@@ -293,7 +293,7 @@ describe('css prop behaviour', () => {
     `);
 
     expect(actual).toInclude(
-      `{background-image:linear-gradient(45deg,gray 25%,transparent 25%),linear-gradient(-45deg,gray 25%,transparent 25%),linear-gradient(45deg,transparent 75%,gray 75%),linear-gradient(-45deg,transparent 75%,gray 75%)}`
+      `{background-image:linear-gradient(45deg,gray 25%,transparent 0),linear-gradient(-45deg,gray 25%,transparent 0),linear-gradient(45deg,transparent 75%,gray 0),linear-gradient(-45deg,transparent 75%,gray 0)}`
```
liamqma, line 296: how comes 25% becomes 0 here?

liamqma: just saw your comment. so 

linear-gradient(45deg,gray 25%,transparent 25%) = linear-gradient(45deg,gray 25%,transparent 0)?

pancaspe87: Yeahhhh, I had the same reaction too and[ visually checked ](https://github.com/atlassian-labs/compiled/pull/1266#discussion_r938487726). It is handled by https://github.com/atlassian-labs/compiled/blob/356b120030ffdd36f1872ad62efe163fc309e0f2/packages/css/src/plugins/normalize-css.ts#L28 

haven't checked the plugin, but [here](https://github.com/cssnano/cssnano/tree/master/packages/postcss-minify-gradients) the docs if you want to know more

liamqma (dismissed):

.changeset/clever-icons-call.md
```diff
@@ -0,0 +1,7 @@
+---
+'@compiled/babel-plugin': patch
+'@compiled/css': patch
+'@compiled/webpack-loader': patch
+---
+
+Added configurable options to optimize CSS
```
at-nathan, line 7: nit: Add configurable options to optimize CSS
```suggestion
Add configurable options to optimize CSS
```

at-nathan:
Thoughts on setting optimization to `false` for tests and storybook?  Since optimization is not our code, there's no need to test it on our side IMO.

pancaspe87:
> Thoughts on setting optimization to `false` for tests and storybook? Since optimization is not our code, there's no need to test it on our side IMO. It would make this PR smaller for sure :) Also when we write tests it would be helpful not to worry about getting the optimized style correct. It might also confuse new devs in the repo.

Hey @at-nathan I think consumers can decide whether to opt out for their dev loop. In our case I would keep it as it is for consistency with the default behaviour; tests were updated only to reflect the new output now that we optimize everything.

@liamqma , @JakeLane thoughts?

at-nathan:
> Hey @at-nathan I think consumers can decide whether to opt out for their dev loop. In our case I would keep it as it is for consistency with the default behaviour; tests were updated only to reflect the new output now that we optimize everything.
> 
> @liamqma , @JakeLane thoughts?

Are we always interested in the optimized result in our tests though? As @liamqma pointed out, it was confusing to see the optimized gradient output. It gave a false positive that the test was broken.

fixtures/babel-component/src/index.jsx
```diff
@@ -4,7 +4,7 @@ const Button = styled.button`
   color: blue;
   font-size: 30px;
   border: 2px solid blue;
-  padding: 8px;
+  padding: 32px;
```
pancaspe87, line 7: `32px` will be changed to `2pc` if optimization is enabled. I changed the value for testing

packages/parcel-transformer/src/__tests__/transformer.parceltest.ts
```diff
@@ -121,10 +121,10 @@ it('transforms assets with compiled and extraction babel plugins', async () => {
   expect(extractedCss).toMatchInlineSnapshot(`
     "/* index.jsx generated by @compiled/babel-plugin v0.0.0 */ \\"use strict\\";
     require(\\"compiled-css!?style=._19pk1ul9%7Bmargin-top%3A30px%7D\\");
-    require(\\"compiled-css!?style=._19bvftgi%7Bpadding-left%3A8px%7D\\");
```
pancaspe87, line 124: Updated snapshots to reflect values changed in https://github.com/atlassian-labs/compiled/pull/1266/files#r944090914

JakeLane (approved):

liamqma (approved):

at-nathan (approved):

PR #1273, opened 2022-07-27
1260 fix prop filtering dom

### LINKED ISSUE #1086: Shadowed variables can clash with destructured props
Given this scenario:

```
const isPrimary = true;

const Component = styled.div<TextProps>`
  color: ${({ isPrimary }) => (isPrimary ? 'green' : 'red')};
  font-weight: ${() => (isPrimary ? 'bold' : undefined)};
`;
```
The outer scope `isPrimary` is taking precedence over the one used in the `color` arrow function, which is incorrect behaviour. 

It seems`color` is being statically evaluated, so it could be `evaluate-expression` is used the scope of the file to evaluate `isPrimary`.

### LINKED ISSUE #1260: Prop filtering on DOM elements does not work
**Describe the bug**
When passing a prop to a styled component, the props can be forwarded to the DOM element.

**Expected behavior**
Props should only be forwarded to DOM elements if they are a DOM prop

**Screenshots**
Screen Shot 2022-07-12 at 3 10 19 PM

pancaspe87:
Resolves #1260 
Resolves #1086

This PR: 
- Normalizes prop usage in styled API to make name clashing easier to handle
```
// For example:
(propz) => propz.width
({ width }) => width
({ width: w }) => w
(height, ...rest) => rest.width

// All become
(props) => props.width
```
- Fixes bug where all props are spread onto the output DOM element
- Fixes bug where destructing in styled API can remove values from props

Next steps: Rename props identifier to something else likely to clash


packages/babel-plugin/src/utils/build-styled-component.ts
```diff
@@ -150,22 +183,16 @@ const getDestructuredProps = (node: t.Node): string[] => {
  *
  * @param path MemberExpression path
  */
-const handleMemberExpressionInStyledInterpolation = (path: NodePath<t.MemberExpression>) => {
+const handleMemberExpressionInStyledInterpolation = (
```
pancaspe87, line 186: I like @at-nathan idea to assume all props will have same identifier to simplify `props` so that we can simplify most of the logic, not only here. Eg I guess we can do the transformation before getting here

liamqma (approved):

packages/babel-plugin/src/utils/build-styled-component.ts
```diff
@@ -210,13 +237,58 @@ const styledTemplate = (opts: StyledTemplateOpts, meta: Metadata): t.Node => {
...
+                  path,
+                  isMemberExpressionNameTheSameAsFunctionFirstParam
+                );
+
+              propsToDestructure.push(...propsToDestructureFromMemberExpression);
+            }
+          },
+        };
+
+        const nestedMemberExpressionVisitor = {
```
at-nathan, line 263: hmm instead of having a nested visitor would it be better to just look for any member expression in an arrow function?

packages/babel-plugin/src/utils/build-styled-component.ts
at-nathan: To me it looks like we're adding on more logic to an already cluttered and non straight forward way of handling props and other dependencies. Should we rethink it and refactor to a cleaner solution?

packages/babel-plugin/src/utils/build-styled-component.ts
```diff
@@ -101,6 +104,36 @@ const traverseStyledBinaryExpression = (node: t.BinaryExpression, nestedVisitor:
   return node;
 };
 
+/**
+ * Traverses a tagged template expression looking for any arrow functions,
```
at-nathan, line 108: hmm this only handles tagged template expression? Does that mean we have do it again for Object styles? :(

.changeset/few-bulldogs-sin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@compiled/babel-plugin': patch
+---
+
+fix props not filtering out in template
```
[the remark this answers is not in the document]
pancaspe87: What about "Ensure props are not displayed as HTML attributes, unless they are valid attributes?"

packages/babel-plugin/src/utils/build-styled-component.ts
pancaspe87: Is it something that we want to look at now or at a later stage?

packages/babel-plugin/src/utils/build-styled-component.ts
```diff
@@ -39,6 +39,9 @@ export interface StyledTemplateOpts {
   sheets: string[];
 }
 
+let conditionalMemberExpressionNode: NodePath<t.MemberExpression>;
+let isConditionalMemberExpressionNameTheSameAsFunctionFirstParam = false;
```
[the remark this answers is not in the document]
pancaspe87: Yes, the error only occurs for conditionals; `width: (props) => width` was already handled. 

It seems that conditionals were handled though, based on this comment https://github.com/atlassian-labs/compiled/blob/master/packages/babel-plugin/src/utils/build-styled-component.ts#L148
Maybe something happened or it never worked

packages/babel-plugin/src/utils/build-styled-component.ts
pancaspe87: I think this is possible if we simplify the logic and assume that all props will have` props` as identifier.
We need a nested visitor to check whether all the nodes (`ArrowFunctionExpression` and `MemberExpression`) [have the same identifiers](https://github.com/atlassian-labs/compiled/blob/master/packages/babel-plugin/src/utils/build-styled-component.ts#L169)
Eg we don't want to apply the logic for cases such as` color: ({ loading: l }) => l ? colors.N50 : colors.N10`

pancaspe87: If there is an alternative way, let me know! For now I tried to extend the existing logic to also handle conditionals

packages/babel-plugin/src/utils/build-styled-component.ts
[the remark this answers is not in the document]
pancaspe87: Agree with you. I was having a hard time with nested visitors. AFAIK we can pass the state down, but not up; I guess this is why we use meta a lot. 
Let me look at this again

packages/babel-plugin/src/utils/build-styled-component.ts
pancaspe87: Good point! Let me check

packages/babel-plugin/src/utils/build-styled-component.ts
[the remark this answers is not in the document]
pancaspe87: Checked a very old branch that I still had locally (issue-390-conditional-css-rules). It was working as expected, something may have broken it

packages/babel-plugin/src/utils/build-styled-component.ts
```diff
@@ -150,22 +183,16 @@ const getDestructuredProps = (node: t.Node): string[] => {
  *
  * @param path MemberExpression path
  */
-const handleMemberExpressionInStyledInterpolation = (path: NodePath<t.MemberExpression>) => {
+const handleMemberExpressionInStyledInterpolation = (
+  path: NodePath<t.MemberExpression>,
+  isMemberExpressionNameTheSameAsFunctionFirstParam: boolean
```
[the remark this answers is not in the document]
pancaspe87: - [ ] rename to `isMemberNameMatchPropParam`

stories/conditional-rules-styled.tsx
```diff
@@ -331,3 +345,20 @@ export const ConditionWithDestructuredPropsKeyValueString = (): JSX.Element => (
...
+      minHeight="100px"
+      width="320px"
+      border="1px solid green"
+      isPrimary>
+      {' '}
+      Has conditional rules
+    </NotBooleanProps>
+    <NotBooleanProps width="200px"> Does not have conditional rules</NotBooleanProps>
+  </div>
+);
```
pancaspe87, line 364: Output:
Screen Shot 2022-08-09 at 9 58 38 am

packages/babel-plugin/src/utils/build-styled-component.ts
```diff
@@ -210,13 +200,73 @@ const styledTemplate = (opts: StyledTemplateOpts, meta: Metadata): t.Node => {
         const nestedArrowFunctionExpressionVisitor = {
           noScope: true,
           MemberExpression(path: NodePath<t.MemberExpression>) {
-            const propsToDestructureFromMemberExpression =
-              handleMemberExpressionInStyledInterpolation(path);
+            const traversedUpFunctionPath: NodePath<t.Node> | null = path.find((parentPath) =>
+              parentPath.isFunction()
+            );
+
+            const memberExpressionKeyName = (path.node.object as t.Identifier).name;
```
JakeLane, line 207: Why the type assertion here?

pancaspe87: I thought I could simplify https://github.com/atlassian-labs/compiled/blob/master/packages/babel-plugin/src/utils/build-styled-component.ts#L154-L157 assuming that it is going to be an identifier ( I have never seen other types)

Looking at this again, it seems that [object should be an expression](https://babeljs.io/docs/en/babel-types#memberexpression); I'll keep the initial logic

pancaspe87: keep initial logic

packages/babel-plugin/src/styled/__tests__/behaviour.test.ts
```diff
@@ -269,20 +275,6 @@ describe('styled component behaviour', () => {
     expect(actual).toInclude('"--_1poneq5":ix((()=>{return props.color;})(),"\\"","\\"")');
   });
 
-  it('should move suffix and prefix of a dynamic arrow function with a body into an IIFE by preventing passing down invalid html attributes to the node', () => {
```
at-nathan, line 272: Removed some tests in this PR as they were already being tested else where or have now become redundant due to the changes.

JakeLane (approved):

liamqma (approved):

PR #1279, opened 2022-08-09
Remove @compiled/cli package

JakeLane:
As we've now enrolled in the [codeshift community](https://www.codeshiftcommunity.com/docs/) codemod command, we can remove the cli package if we update the docs to reference the new command.


packages/codemods/README.md
```diff
@@ -2,10 +2,23 @@
 
 > Codemods for easy migration from [styled components](https://styled-components.com/) and [emotion](https://emotion.sh/docs/introduction).
 
+## Usage
```
pancaspe87, line 5: Should we also update/remove https://compiledcssinjs.com/docs/pkg-cli ?

scripts/generate-renovate-changeset.ts
```diff
@@ -1,23 +0,0 @@
-// This script is designed so that Renovate can automatically add changesets to the PRs it
-// creates. It is designed to be run in the root of the repo.
```
pancaspe87, line 2: Why are we removing this?

liamqma (approved):

packages/codemods/README.md
JakeLane: yep will do after merging this PR

scripts/generate-renovate-changeset.ts
JakeLane: Unfortunately it doesn't work anymore due to internal security compliance. We'll need to come up with a different solution

pancaspe87: so not strictly related to this PR, but something to do anyway.

pancaspe87 (approved):

PR #1281, opened 2022-08-10
Omit rules with empty values from stylesheet

### LINKED ISSUE #1240: Ignore rule if value is not conditionally defined
 When using a ternary operator to conditionally apply CSS Compiled will generate either
- a blank CSS variable  (if `undefined`) OR
- a not valid CSS declaration ( if empty string)

This seems to happen only for object syntax

```
 background: 'white',
 position: iStickyHeaderEnabled() ? 'sticky' : undefined,
 top: iStickyHeaderEnabled() ? 0 : '',
});
```

We should ignore rules if the value is undefined or empty string.

 **Workaround**

1. Use spread operator
```
 background: 'white',
 ...(isStickyHeaderEnabled() && { 
  position: 'sticky',
  top: 0,
 })
});
```

2. use the default value.
```
position: isStickyHeaderEnabled() ? 'sticky' : 'static',
```
Need to ensure position is not set somewhere else.

3. use template
```
styled.div`
   ${isStickyHeaderEnabled() ? 'position: sticky' : ''}
`
```
However, [template literals are not recommended](https://github.com/atlassian-labs/compiled/tree/master/packages/eslint-plugin/src/rules/no-styled-tagged-template-expression)

at-nathan:
Resolves #1240


pancaspe87:
May worth checking if [this](https://github.com/atlassian-labs/compiled/issues/848) still happens and whether we want to address it in this PR or later on

at-nathan:
> worth checking if [this](https://github.com/atlassian-labs/compiled/issues/848) still happens and whether we want to address it in this PR or later on

Looks like we would have to add extra logic to cater for this. But it should be possible to do though.

.changeset/ten-horses-live.md
```diff
@@ -0,0 +1,6 @@
+---
+'@compiled/babel-plugin': minor
```
JakeLane, line 2: I think these could be `patch` level as we haven't done a major release yet.

packages/babel-plugin/src/utils/css-builders.ts
```diff
@@ -488,6 +489,16 @@ const extractObjectExpression = (node: t.ObjectExpression, meta: Metadata): CSSO
         return;
       }
 
+      if (isEmptyValue(propValue)) {
+        css.push({
+          type: 'unconditional',
+          // Let postCSS strip out later
+          css: `${kebabCase(key)}: undefined;`,
```
JakeLane, line 496: I'm a bit concerned by this as it is technically invalid CSS. That adds a lot of potential for bugs if we don't handle all of these cases safely.

Could you avoid pushing the CSS entirely in this case?

at-nathan: Ah dang that was really silly of me!  I can just return here. Nice catch!

packages/babel-plugin/src/styled/__tests__/behaviour.test.ts
```diff
@@ -1125,6 +1185,97 @@ describe('styled component behaviour', () => {
...
+      'className={ax(["",!props.isPrimary&&"_syazbf54",props.className])}',
+    ]);
+  });
+
+  it('should apply logical test to class when a conditional branch contains empty string value', () => {
+    const actual = transform(`
+      import { styled } from '@compiled/react';
+
+      const Component = styled.div({
+        color: props => props.isPrimary ? '' : 'green',
```
liamqma, line 1223: empty string is useful for ::before or ::after pseudo element.
e.g.
```
div:after {
    content:"";
}
```

packages/css/src/plugins/discard-empty-rules.ts
```diff
@@ -0,0 +1,25 @@
+import type { Plugin } from 'postcss';
```
liamqma, line 1: Just out of interest, since empty values are handled by babel-plugin, why do we need this plugin?

packages/babel-plugin/src/styled/__tests__/behaviour.test.ts
at-nathan: Good point, we already handle this case, then value will actually be `'""'`, I'll add it to a test case to demonstrate this.

packages/css/src/plugins/discard-empty-rules.ts
at-nathan: Empty values still generate CSS and CSS classes (i.e. color: undefined) , so we need this plugin to remove them from the style sheet.

JakeLane (approved):

PR #1293, opened 2022-09-12
1292 - Update loki & fix for Chrome CSS var behaviour

### LINKED ISSUE #1292: New Chrome ignores CSS vars with no name
**Describe the bug**
Current Chrome as of posting (105.0.5195.102) will ignore CSS var defined with no name (i.e. `--`)
```
/* Will be stripped by chrome */
--: black;
/* Will be ignored by chrome */
--named-var: var(--);
```
Compiled currently uses this var name for empty var values. This is preventing the upgrade of `loki` as VR tests are failing due to it using affected newer version of Chromium.

at-nathan:
Resolves #1292 
- Update loki
- Give a name to CSS var used for empty values


yarn.lock
```diff
@@ -1840,85 +1840,85 @@
     "@jridgewell/resolve-uri" "^3.0.3"
     "@jridgewell/sourcemap-codec" "^1.4.10"
 
-"@loki/browser@^0.29.0":
-  version "0.29.0"
-  resolved "https://registry.yarnpkg.com/@loki/browser/-/browser-0.29.0.tgz#39377f8da4eae642cf41bb999029549a1177f82b"
-  integrity sha512-w/Hpc95jFwrjfQhlk6Gb+vdanZCA4cg/qiarxhX3kVeVKVwEXgaHfHHpfKpl3f3bk3T4fhWfRV84T4lWD2vEow==
+"@loki/browser@^0.30.1":
+  version "0.30.1"
+  resolved "https://packages.atlassian.com/api/npm/npm-remote/@loki/browser/-/browser-0.30.1.tgz#f90ab52d63985b5e717fd6e79526044c02cb6649"
```
JakeLane, line 1845: Please update to the yarn mirror

JakeLane (changes requested):

.changeset/nine-colts-raise.md
```diff
@@ -0,0 +1,6 @@
+---
+'@compiled/react': patch
+---
+
+- Give a name to CSS var used for empty values
```
pancaspe87, line 5: curious to know why we need a name

pancaspe87 (approved):

.changeset/nine-colts-raise.md
at-nathan: Current Chrome will treat rules using empty CSS var names as invalid. This is also why new Loki version was breaking on VR, because it contained new Chromium :)

JakeLane (approved):

PR #1296, opened 2022-09-14
1295 - Improve support for computed object property keys

### LINKED ISSUE #1295: Can't use more than one import in template literal as object property key
**Describe the bug**
```
import { hiddenClass } from './hidden';
import { itemClass } from './item';

// Output selector becomes .hidden, itemClass
const Comp = styled.div({
  [`${hiddenClass}, ${itemClass}`] { ... },
});
```

at-nathan:
Resolves #1295
- Fixes bug where more than one import cannot be used in template literal object property key
- Support string binary operation in object property keys

### DIFF
--- a/packages/babel-plugin/src/utils/object-property-to-string.ts
+++ b/packages/babel-plugin/src/utils/object-property-to-string.ts
@@ -1,0 +1,88 @@
+import * as t from '@babel/types';
+
+import type { Metadata } from '../types';
+
+import { evaluateExpression } from './evaluate-expression';
+
+type ExpressionToString = (expression: t.Expression, meta: Metadata) => string;
+
+const templateLiteralToString = (
+  template: t.TemplateLiteral,
+  meta: Metadata,
+  expressionToString: ExpressionToString
+) => {
+  let result = '';
+
+  for (let i = 0; i < template.quasis.length; i += 1) {
+    result += template.quasis[i].value.raw;
+
+    if (i < template.expressions.length) {
+      const expression = template.expressions[i];
+      if (t.isTSType(expression)) {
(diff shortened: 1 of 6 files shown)


JakeLane (approved):

packages/babel-plugin/src/utils/object-property-to-string.ts
```diff
@@ -0,0 +1,88 @@
...
+    const rightValue = expressionToString(right, meta);
+
+    return `${leftValue}${rightValue}`;
+  }
+
+  throw new Error(`Cannot use ${operator} for string operation. Use + for string concatenation`);
+};
+
+const expressionToString: ExpressionToString = (expression, meta) => {
+  // handles {'key-name': 'value'} or {1: 'value'}
```
pancaspe87, line 52: Not sure if we should also handle `isNumericLiteral`. 
I can't think of any css property defined with a number, or any scenario where a selector could be a number. Keeping it shouldn't cause any issue though.

@JakeLane, @liamqma  any thoughts on this?

at-nathan: Yeah true, when I was writing it I was thinking about all Objects in general 😅 Maybe number should be unsupported so it can throw an error?

at-nathan: On second thought, I think we do need to keep the numeric evaluation, otherwise we can do things like:
`['field' + 1]: { ... }`

pancaspe87: I thought it was checking for numbers only. Yeahh, we should keep it if it deals with alphanumerics :)

PR #1298, opened 2022-09-23
fix(webpack-loader): add missing `options.parserBabelPlugins`

SukkaW:
The PR follows #1225.

In #1225 @JakeLane introduced two new options, `parserBabelPlugins` and `transformerBabelPlugins`. Although both two options were added to the parcel transformer, only `transformerBabelPlugins` was added to the webpack loader while `parserBabelPlugins` is missing. The PR fixes that.

### DIFF
--- a/.changeset/ten-trees-shave.md
+++ b/.changeset/ten-trees-shave.md
@@ -1,0 +1,5 @@
+---
+'@compiled/webpack-loader': patch
+---
+
+Add missing parserBabelPlugins options to webpack loader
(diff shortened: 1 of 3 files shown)


JakeLane (approved):

at-nathan (approved):

SukkaW:
I have also updated the webpack example, added the missing `jsx`. It should fix the webpack example build issue.

JakeLane:
Thanks for the contribution @SukkaW!

JakeLane (approved):

PR #1301, opened 2022-10-06
Props not forwarded when composing using styled API

at-nathan:
Fixes bug where props are not forwarded when composing styles from another styled component


packages/babel-plugin/src/utils/build-styled-component.ts
```diff
@@ -115,9 +115,10 @@ const styledTemplate = (opts: StyledTemplateOpts, meta: Metadata): t.Node => {
   const styleProp = opts.variables.length
     ? styledStyleProp(opts.variables)
     : t.identifier(STYLE_IDENTIFIER_NAME);
+  const isInBuiltComponent = opts.tag.type === 'InBuiltComponent';
   // This completely depends on meta.parentPath to be the styled component.
   // If this changes please pass the component in another way
-  const invalidDomProps = getInvalidDomProps(meta.parentPath);
+  const invalidDomProps = isInBuiltComponent ? getInvalidDomProps(meta.parentPath) : [];
```
liamqma, line 121:
```
        margin: props => { return props.foo; },
      });
      const ListItem = styled(Component)({
        padding: props => { return props.bar; },
      });
```

Will it still pick up invalid dom props if composing styles from another styled component?

at-nathan: Just tested this and they don't get applied 😌

pancaspe87: Can we add a test for it?

at-nathan: There is
packages/babel-plugin/src/styled/__tests__/behaviour.test.ts

liamqma: I just tested it. It seems invalid prop appear in the HTML in my test case. 🤔

```

      const Component = styled.div({});

      const ListItem = styled(Component)({
        fontSize: props => props.textSize,
      });
```

get transpiled to 
```
    import { forwardRef } from "react";
    const _ = "._1wyb1sqb{font-size:var(--_8t6091)}";
    const Component = forwardRef(
      ({ as: C = "div", style: __cmpls, ...__cmplp }, __cmplr) => {
        return (
          <CC>
            <CS>{[]}</CS>
            <C
              {...__cmplp}
              style={__cmpls}
              ref={__cmplr}
              className={ax(["", __cmplp.className])}
            />
          </CC>
        );
      }
    );

    if (process.env.NODE_ENV !== "production") {
      Component.displayName = "Component";
    }

    const ListItem = forwardRef(
      ({ as: C = Component, style: __cmpls, ...__cmplp }, __cmplr) => {
        return (
          <CC>
            <CS>{[_]}</CS>
            <C
              {...__cmplp} // `__cmplp` includes `textSize`
              style={{ ...__cmpls, "--_8t6091": ix(__cmplp.textSize) }}
              ref={__cmplr}
              className={ax(["_1wyb1sqb", __cmplp.className])}
            />
          </CC>
        );
      }
    );

    if (process.env.NODE_ENV !== "production") {
      ListItem.displayName = "ListItem";
    }
```

at-nathan: Hmm let me try again. Did you run `yarn clean && yarn build`?

liamqma: no, i ran the test so I guess it runs against the original code directly. but will try `yarn clean && yarn build`.

at-nathan: Sorry, I tested with boolean values :( That's why it didn't show up...

JakeLane (dismissed):

liamqma (approved):

PR #1302, opened 2022-10-07
pass a custom resolver from parcel-transformer to babel plugin

liamqma:
Similar to what Webpack's compiled-loader does, this PR passes a custom resolver from parcel-transformer to babel plugin.

Ideally, `@parcel/node-resolver-core` should be used so Tilde is supported. But as Monica said, all the methods in the `@parcel/node-resolver-core` perform asynchronous resolutions, so won't be able to use it in babel plugin.

I guess we will eventually develop our own resolver.


packages/parcel-transformer/src/types.ts
```diff
@@ -26,4 +27,9 @@ export interface ParcelTransformerOpts extends BabelPluginOpts {
    * Default to `true`.
    */
   optimizeCss?: boolean;
+
+  /**
+   * Override the default `resolve` passed into webpack, which is used to statically evaluate import declarations
```
JakeLane, line 32: should reference parcel?

packages/parcel-transformer/src/index.ts
```diff
@@ -103,6 +115,12 @@ export default new Transformer<ParcelTransformerOpts>({
           {
             ...config,
             onIncludedFiles: (files: string[]) => includedFiles.push(...files),
+            resolver: {
```
JakeLane, line 118: could you please add a test?

packages/parcel-transformer/src/types.ts
pancaspe87: should we also update the docs?

liamqma: Have created a PR to update the Website: https://github.com/compiled/website/pull/661

packages/parcel-transformer/src/index.ts
liamqma: Added

.changeset/twenty-colts-invite.md
```diff
@@ -0,0 +1,6 @@
+---
+'@compiled/parcel-transformer': minor
```
liamqma, line 2: Do you think if this change should be `minor`?

fixtures/parcel-transformer-test-custom-resolver-app/src/index.jsx
```diff
@@ -0,0 +1,8 @@
+import '@compiled/react';
+import { color } from '~/src/foo.js';
+
+const App = () => (
+  <>
+    <div css={{ fontSize: 50, color }}>CSS prop</div>
```
pancaspe87, line 6: I guess this was to visually check that font-size gets units

.changeset/twenty-colts-invite.md
JakeLane: yeah probably a good idea since we're potentially changing resolving behaviour by default

packages/parcel-transformer/src/__tests__/transformer.parceltest.ts
```diff
@@ -68,6 +72,19 @@ it('transforms assets with babel plugin', async () => {
...
+  const parcel = getParcelInstance(customResolverFixtureRoot);
+  const { changedAssets } = await parcel.run();
+
+  const asset = Array.from(changedAssets.values()).find(
+    (asset) => asset.filePath === join(customResolverFixtureRoot, 'src/index.jsx')
+  );
+
+  const code = await asset?.getCode();
+
+  expect(code).toInclude('color:red');
```
pancaspe87, line 85: nit should we also check for font-size?

packages/parcel-transformer/src/__tests__/transformer.parceltest.ts
```diff
@@ -68,6 +72,19 @@ it('transforms assets with babel plugin', async () => {
   `);
 }, 50000);
 
+it('transforms assets with custom resolver and statically evulates imports', async () => {
```
JakeLane, line 75:
```suggestion
it('transforms assets with custom resolver and statically evaluates imports', async () => {
```

JakeLane (dismissed):

packages/parcel-transformer/src/__tests__/transformer.parceltest.ts
liamqma: um..good catch. I think I will remove 'font-size' from this test case. The test case of font-size has already been tested at https://github.com/atlassian-labs/compiled/blob/master/packages/parcel-transformer/src/__tests__/transformer.parceltest.ts#L46. The purpose of this test is to ensure we are able to statically evaluate color from another file.

packages/parcel-transformer/src/__tests__/transformer.parceltest.ts
liamqma: Good catch!

pancaspe87 (approved):

PR #1303, opened 2022-10-18
update babel-plugin-strip-runtime to pass styleRules to meta, update …

liamqma:
This PR is still WIP but I'd like to validate the idea with you. 😃

### This PR solves two issues:

- Parcel build is slow when stylesheet extraction is enabled on a large project 
- In Parcel, currently the extraction produces one CSS bundle per JS bundle - that is, when there is an async split there is also a CSS split. With this PR, it produces a single CSS asset per entry point, optimizes the rules, and injects it to HTML.

### How it works:

- babel-plugin-strip-runtime adds stylesRules to meta, and pass it to parcel-transformer
- parcel-transformer adds stylesRules to asset.meta
- parcel-optimizer-html-inline-css (a new package) grabs styleRules, optimizes the rules and inject it to HTML
- parcel-resolver and parcel-optimiser are unused in this setup

### How it address the above two problems:
- The long build is caused by a large number of css imports (e.g. `require(._1e0cglyw%7Bdisplay%3Anone%7D.css)`). By using asset.meta to store these styles rules, the bundler avoids handling that many dependencies.
- We have the access to the bundleGraph at the optimizer so we can split / merge the css content however we want. This allows us to have one CSS asset per entry. Also, paves the way to CriticalCSS.

TODO:
- [x] Tests
- Will update documentation (https://github.com/compiled/website/pull/682) after merging the PR
- [x] Allow users to opt in inline-css


examples/parcel/src/app.jsx
```diff
@@ -24,5 +26,8 @@ export const App = () => (
     <CustomFileExtensionStyled>Custom File Extension Styled</CustomFileExtensionStyled>
     <div css={customFileExtensionCss}>Custom File Extension CSS</div>
     <Annotated />
+    <React.Suspense fallback={<div>Loading...</div>}>
```
liamqma, line 29: Add AsyncComponent so we have bundle splitting in the Parcel example app.

packages/babel-plugin-strip-runtime/src/index.ts
```diff
@@ -21,23 +21,33 @@ export default declare<PluginPass>((api) => {
     },
     visitor: {
       Program: {
-        exit(path) {
-          if (!this.opts.compiledRequireExclude && this.opts.styleSheetPath) {
-            preserveLeadingComments(path);
+        exit(path, { file } ) {
+
+          if (this.opts.compiledRequireExclude) {
```
liamqma, line 26: I've reused this variable `compiledRequireExclude`.
So if it's enabled, it
- does not add  `require("${this.opts.styleSheetPath}?style=${params}");`
- add styleRules to metadata for later use

packages/parcel-optimizer-html-inline-css/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+# @compiled/parcel-optimizer-html-inline-css
```
liamqma, line 1: I am not too sure about the name of this package. Its job is to retrieves asset.meta.styleRules, optimizes the rules, and inject the rules to HTML. Please let me know if you can think of a better name.

packages/parcel-transformer/src/index.ts
```diff
@@ -148,6 +152,11 @@ export default new Transformer<ParcelTransformerOpts>({
 
     asset.setCode(output);
 
+    if (!config.ssr) {
+      const metadata = result?.metadata as CustomBabelFileMetadata;
+      asset.meta.styleRules = metadata.styleRules;
```
liamqma, line 157: TODO: since Asset stuff needs to be serialised/de-serialised, not sure if there is less overhead to have a single property rather than an array (e.g. just make it a single string rather than an array)

marcins:
> babel-plugin-strip-runtime adds stylesRules to meta, and pass it to parcel-transformer
Can we fold this functionality into the Transformer? Maybe doesn't have to be done as part of this PR, but we should considering minimising our dependency on Babel.

packages/parcel-optimizer-html-inline-css/src/index.ts
```diff
@@ -0,0 +1,43 @@
...
+import { sort } from '@compiled/css';
+import { Optimizer } from '@parcel/plugin';
+import posthtml from "posthtml";
+import { insertAt } from "posthtml-insert-at";
+
+export default new Optimizer({
+  async optimize({ contents, map, bundleGraph }) {
+    const styleRules: string[] = [];
+
+    for (const bundle of bundleGraph.getBundles()) {
```
marcins, line 12: AFAIK this will get all the bundles, not just the bundles that this HTML is the entry for - is that expected? (i.e. every HTML file will end up with the same styles in it with this approach - which I know we did discuss, but can't remember where we ended up on whether that was what we wanted)

packages/parcel-transformer/src/index.ts
```diff
@@ -127,8 +132,7 @@ export default new Transformer<ParcelTransformerOpts>({
         config.extract && [
           '@compiled/babel-plugin-strip-runtime',
```
marcins, line 133: Ah, I see, this plugin is used by the Babel pass _inside_ the existing Compiled transformer - that makes sense, you're not adding an extra Babel plugin to the Babel config for the project using Compiled.

So this will be addressed by whatever future endeavour we go on in order to make Compiled when used with Parcel not have an internal dependency on Babel :)

marcins (dismissed):
Approach looks good! As you mentioned, curious if there's benefit to having a single asset meta property to (de)serialise rather than an array, but we can measure that later.

I'll let some of the other bundler team folks chime in with any other feedback I might've missed.

packages/parcel-transformer/src/index.ts
```diff
@@ -127,8 +132,7 @@ export default new Transformer<ParcelTransformerOpts>({
...
           {
-            styleSheetPath: 'compiled-css!',
-            compiledRequireExclude: config.ssr,
+            compiledRequireExclude: true,
```
liamqma, line 135: I am thinking of making a breaking change to the parcel packages as splitting css every async js split defeats the purpose of atomic css. It means 
- deprecate parcel-resolver 
- major bump parcel-transformer
- use parcel-optimizer to insert css to HTML, and provide an option to inline/externalise css.

liamqma: had a discussion with Jake,
We'll
Deprecate parcel-resolver
Update parcel-config
Rename 'parcel-optimizer-html-inline-css' to parcel-optimiser

packages/parcel-transformer/src/index.ts
```diff
@@ -77,14 +82,16 @@ export default new Transformer<ParcelTransformerOpts>({
...
+  async transform({ asset, config, options }) {
     const ast = await asset.getAST();
     if (!ast) {
       // We will only receive ASTs for assets we're interested in.
       // Since this is undefined (or in node modules) we aren't interested in it.
       return [asset];
     }
 
+    // Disable stylesheet extraction locally due to https://github.com/atlassian-labs/compiled/issues/1306
+    const extract = config.extract && options.mode !== 'development';
```
liamqma, line 94: Disable stylesheet extraction on local Parcel. I will update the documentation once this PR is merged.

JakeLane (approved):

JakeLane (approved):

packages/parcel-optimizer/src/index.ts
```diff
@@ -1,8 +1,108 @@
...
+        undefined // for TypeScript
+      );
+
+      newContents = (
+        await posthtml()
+          .use(
+            insertAt({
+              selector: 'head',
+              append:
+                '<link href="' + bundle.target.publicUrl + cssFileName + '" rel="stylesheet" />',
```
pancaspe87, line 98: Why was this previously manually added and now is automated?

pancaspe87 (approved):

packages/parcel-optimizer/src/index.ts
liamqma: Sorry, I don't get what you mean. Do you mean it was manually added in the fixture app https://github.com/atlassian-labs/compiled/pull/1303/files/83f34318edcce33f1d9ea0a95aa8f38a0cb70721#diff-8d93de8220a5ed7650b00d8791683374942b49b5fa6753bb833ba978af721596?

pancaspe87: Yep!

liamqma: Previously the fixture app didn't use `@compiled/react`, so for simplicity, it manually adds a CSS file in order to test that optimizer can sort the CSS. 
Now I cannot test optimizer independently because optimizer needs the metadata from transformer. I've added `@compiled/react` and `@compiled/parcel-transformer` to the fixture app. The test has become an integration test.
Worth mentioning [this file ](https://github.com/atlassian-labs/compiled/blob/master/fixtures/parcel-transformer-test-app/src/index.html) is the entrypoint (similar to webpack.config.js). Because previously the entrypoint references a CSS file, Parcel automatically adds CSS file to the output HTML.
Now since we've used CSS-in-JS (@compiled/react), we need to manually add the CSS file to output HTML.

I hope it makes sense 😄

pancaspe87: Yeahhh, I think I get it now. Thanks for the explanation, I was just curious :)

Issue #1312, opened 2022-11-17
Eslint plugin breaks when having multiple selectors across lines


pancaspe87:
`no-styled-tagged-template-expression` rule is not fixable in the following scenarios

1. multiple selectors across lines

Given

```
const Component = styled.div`
    display: none;
    ul, ol {
        margin: 0;
    }
`;
```
this will wok.

Given

```
const Component = styled.div`
    display: none;
    ul, 
    ol {
        margin: 0;
    }
`;
```
this will break.

2. invalid CSS

PR #1313, opened 2022-11-18
fix breaking rule

pancaspe87:
This resolves https://github.com/atlassian-labs/compiled/issues/1312

### DIFF
--- a/packages/eslint-plugin/src/rules/no-styled-tagged-template-expression/__tests__/rule.test.ts
+++ b/packages/eslint-plugin/src/rules/no-styled-tagged-template-expression/__tests__/rule.test.ts
@@ -899,5 +899,84 @@ tester.run('no-styled-tagged-template-expression', noStyledTaggedTemplateExpress
         );
       `,
     },
+    {
+      filename: 'multiple-selectors-across-lines.ts',
+      code: `
+        import { styled } from '@compiled/react';
+
+        styled.div\`
+          color: blue;
+          &:hover,
+          &:focus,
+          &:active {
+            text-decoration: inherit;
+          }
+          ul,
+          li {
+            color: red;
+          }
+        \`;
+      `,
+      output: `
+        import { styled } from '@compiled/react';
+
+        styled.div({
+          color: "blue",
+          "&:hover, &:focus, &:active": {
+            textDecoration: "inherit"
+          },
+          "ul, li": {
+            color: "red"
+          }
+        });
+      `,
+    },
+    {
+      filename: 'multiple-selectors-on-same-line.ts',
+      code: `
+        import { styled } from '@compiled/react';
+
+        styled.div\`
+          color: blue;
+          &:hover, &:focus,
+          &:active {
+            text-decoration: inherit;
+          }
+          ul, li {
+            color: red;
+          }
+        \`;
+      `,
+      output: `
+        import { styled } from '@compiled/react';
+
+        styled.div({
+          color: "blue",
+          "&:hover, &:focus, &:active": {
+            textDecoration: "inherit"
+          },
+          "ul, li": {
+            color: "red"
+          }
+        });
+      `,
+    },
+    {
+      filename: 'do-not-handle-invalid-css.ts',
+      code: `
+        import { styled } from '@compiled/react';
+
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/src/utils/create-no-tagged-template-expression-rule/index.ts
```diff
@@ -39,11 +39,18 @@ export const createNoTaggedTemplateExpressionRule =
             return;
           }
 
+          const args = toArguments(source, quasi);
+
+          // Skip invalid CSS
+          if (args.length < 1) {
```
pancaspe87, line 45: RFC: This is to prevent eslint rule to break when dealing with invalid CSS such as `margin-top 0;` ( missing colon symbol).
However, maybe we should throw an error instead.

packages/eslint-plugin/src/utils/create-no-tagged-template-expression-rule/to-arguments.ts
```diff
@@ -130,7 +135,10 @@ export const toArguments = (source: SourceCode, template: ESTree.TemplateLiteral
   };
 
   for (const [i, quasi] of template.quasis.entries()) {
-    for (const char of quasi.value.raw) {
+    // Deal with multi selectors
+    const styleTemplateElement = quasi.value.raw.replace(/\n/g, '').replace(/\s+/g, ' ');
```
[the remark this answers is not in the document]
JakeLane:
```suggestion
    const styleTemplateElement = quasi.value.raw.replace(/(\r\n|\n|\r)/gm, ' ').replace(/\s+/g, ' ');
```
In this case we should add a space, consider the following edge case:
```
div
h1 { ... }
```

Which will turn into:
```
divh1 { ... }
```

packages/eslint-plugin/src/utils/create-no-tagged-template-expression-rule/index.ts
JakeLane: Throwing an error will cause the lint rule to fail. This shouldn't be an error for the lint rule as it's expected to see invalid CSS in the AST. 

We should skip auto-fixing if we would end up with invalid code if transformed.

JakeLane (changes requested):

packages/eslint-plugin/src/rules/no-styled-tagged-template-expression/__tests__/rule.test.ts
```diff
@@ -899,5 +899,84 @@ tester.run('no-styled-tagged-template-expression', noStyledTaggedTemplateExpress
...
+      `,
+      output: `
+        import { styled } from '@compiled/react';
+
+        styled.div\`
+          color blue;
+        \`;
+      `,
+    },
   ]),
```
JakeLane, line 981: Can you please also add a test for https://github.com/atlassian-labs/compiled/pull/1313#discussion_r1027385229

packages/eslint-plugin/src/utils/create-no-tagged-template-expression-rule/to-arguments.ts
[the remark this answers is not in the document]
pancaspe87: both good point!

I'll add test cases for the edge case.

packages/eslint-plugin/src/utils/create-no-tagged-template-expression-rule/index.ts
pancaspe87: absolutely right re auto-fixing!

liamqma (approved):

pancaspe87:
Hey @liamqma re https://github.com/atlassian-labs/compiled/pull/1313#discussion_r1026077290

IIRC `return`; will prevent the autofixer to run. Hence, a `tagged-template-expression` won't be auto-fixed on save. `throw` will cause the lint rule to show the error only when running the command and may attempt to fix.

This is what @JakeLane meant in https://github.com/atlassian-labs/compiled/pull/1313#discussion_r1027385867

JakeLane (approved):

liamqma (approved):

PR #1315, opened 2022-11-21
Add Type support to no-styled-tagged-template-expression ESLint rule

liamqma:
This fixed the issue: `no-styled-tagged-template-expression` ESLint rule  strips out type.

Example:
```
const Foo = styled2.div<{ color: string }>`
    box-shadow: 0 0 0 2px ${({ color }) => color} inset;
`;
```
Becomes:
```
const Foo = styled2.div({
    boxShadow: `0 0 0 2px ${({ color }) => color} inset`,
});
```

### DIFF
--- a/packages/eslint-plugin/src/rules/no-styled-tagged-template-expression/__tests__/rule.test.ts
+++ b/packages/eslint-plugin/src/rules/no-styled-tagged-template-expression/__tests__/rule.test.ts
@@ -5,7 +5,8 @@ import type { RuleTester } from 'eslint';
 import {
   createAliasedInvalidTestCase,
   createDeclarationInvalidTestCases,
-  tester,
+  createTypedInvalidTestCase,
+  typeScriptTester as tester
 } from '../../../test-utils';
 import { noStyledTaggedTemplateExpressionRule } from '../index';
 
@@ -31,6 +32,9 @@ const replaceAlias = (str: string) =>
     .replace('styled.div', 'styled2.div')
     .replace('styled(Base)', 'styled2(Base)');
 
+const replaceType = (str: string) =>
+  str.replace('styled.div', 'styled.div<{color: string}>');
+
 const createInvalidTestCases = (tests: InvalidTestCase[]) =>
   tests
     .map((t) => ({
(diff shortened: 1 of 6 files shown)


yarn.lock
```diff
@@ -4317,6 +4317,16 @@
     "@typescript-eslint/typescript-estree" "4.33.0"
     debug "^4.3.1"
 
+"@typescript-eslint/parser@^5.43.0":
+  version "5.43.0"
+  resolved "https://packages.atlassian.com/api/npm/npm-remote/@typescript-eslint/parser/-/parser-5.43.0.tgz#9c86581234b88f2ba406f0b99a274a91c11630fd"
```
pancaspe87, line 4322: need to change npm registry, it is using atlassian

liamqma: ah, good catch!

packages/eslint-plugin/src/rules/no-styled-tagged-template-expression/__tests__/rule.test.ts
```diff
@@ -31,6 +32,8 @@ const replaceAlias = (str: string) =>
     .replace('styled.div', 'styled2.div')
     .replace('styled(Base)', 'styled2(Base)');
 
+const replaceType = (str: string) => str.replace('styled.div', 'styled.div<{color: string}>');
```
pancaspe87, line 35: Hey @liamqma can you help me to understand how we are testing that 
`styled.div<{color: string}>` doesn't become `styled.div`

Should we add a test case (input code/expected output) in https://github.com/atlassian-labs/compiled/blob/master/packages/eslint-plugin/src/rules/no-styled-tagged-template-expression/__tests__/rule.test.ts#L47

liamqma: We are testing `styled.div<{color: string}>` doesn't become `styled.div`, because
`flatMap((t) => [t, createTypedInvalidTestCase(t, replaceType, replaceType)])` creates a new test case for every test in the list.
e.g.
```
code: `

  styled.div\`color: blue\`;
`,
output: `

  styled.div({
    color: "blue"
  });
`,
```
becomes
```
code: `

  styled.div<{color: string}>\`color: blue\`;
`,
output: `

  styled.div<{color: string}>({
    color: "blue"
  });
`,
```

pancaspe87 (approved):

JakeLane (approved):

PR #1317, opened 2022-11-22
Use 0% as default value for flex-basis

liamqma:
When expanding flex, 0% is used as default value for flex-basis by browsers. Although this is not what the Spec says, we'd have to make it compatible with the browser behavior.


packages/css/src/plugins/expand-shorthands/flex.ts
```diff
@@ -7,6 +7,11 @@ const isFlexNumber = (node: ChildNode): node is Numeric => node.type === 'numeri
 const isFlexBasis = (node: ChildNode): node is Numeric | Word | Func =>
   (node.type === 'word' && node.value === 'content') || isWidth(node);
 
+// According to the spec, the default value of flex-basis is 0.
```
JakeLane, line 10: Nice pragmatic way to document this 🏆

JakeLane (approved):

JakeLane (approved):

pancaspe87 (approved):

Issue #1316, opened 2022-11-21
MemberExpression isn't a supported CSS type when using CSS prop


pancaspe87:
**Describe the bug**
CSS prop doesn't work with imported styles defined as object

**To Reproduce**
Given 

`component.tsx`
```
import { styles } from 'style';
....
 return (
    <main css={styles.default}>
        <article css={styles.listItem}> ... </article>
    </main>
)
```

`style.tsx`

```
import { css } from '@compiled/react';

export const styles = {
  default: css({
    display: 'flex',
    flexDirection: 'column',
  }),
  listItem: css({
    display: 'flex',
    flexDirection: 'row',
  }),
....
}

```

This will work with [styled api](https://compiledcssinjs.com/docs/api-styled).
This will break with [css API](https://compiledcssinjs.com/docs/api-css)

**Expected behaviour**
We should support MemberExpression also to ensure emotion migration

PR #1321, opened 2022-11-24
Fix/1316 CSS prop 

pancaspe87:
This PR addresses a variety of compilation errors - as described in https://github.com/atlassian-labs/compiled/issues/1316 

Types will be addressed in https://github.com/atlassian-labs/compiled/issues/1323


packages/react/types/css.d.ts
```diff
@@ -0,0 +1,7 @@
+import type { CssFunction } from '@compiled/react';
+
+declare module 'react' {
```
pancaspe87, line 3: To address error such as 
`Property 'css' does not exist on type 'DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>`

Not sure if `CssFunction` is the most appropriate type, let me know if there is better one!

packages/babel-plugin/src/utils/traverse-expression/traverse-call-expression.ts
```diff
@@ -50,6 +50,10 @@ export const traverseCallExpression = (
       // Right now we are only supported these 2 flavors. If we have complex case like `func('arg').fn().variable`,
       // it will not get evaluated.
       if (t.isIdentifier(callee)) {
+        if (callee.name === 'css' && t.isObjectExpression(expression.arguments[0])) {
+          return evaluateExpression(expression.arguments[0], updatedMeta);
+        }
+
         const resolvedBinding = resolveBinding(callee.name, updatedMeta, evaluateExpression);
```
pancaspe87, line 57: We have a util to check if [isCompiledCSSCallExpression]( https://github.com/atlassian-labs/compiled/blob/87efa985685720886da35766535efc68436f2cb6/packages/babel-plugin/src/utils/is-compiled.ts#L12), but it won't work with imported mixins as it checks for `css` within the given file

packages/babel-plugin/src/utils/traverse-expression/traverse-call-expression.ts
```diff
@@ -50,6 +50,10 @@ export const traverseCallExpression = (
       // Right now we are only supported these 2 flavors. If we have complex case like `func('arg').fn().variable`,
       // it will not get evaluated.
       if (t.isIdentifier(callee)) {
+        if (callee.name === 'css' && t.isObjectExpression(expression.arguments[0])) {
```
liamqma, line 53: css might be aliased?
```
import { css as css2 } from "@compiled/react"

css2(...)
```
Also, css might have multiple arguments?
```
css({
   color: 'red'
}, {
   background: 'red'
})
```

pancaspe87: Both very good points! 

- [ ] css might be aliased
- [ ] css might have multiple arguments

packages/react/types/css.d.ts
JakeLane: We need to think about how we're going do this, as this will break jsx runtime overrides

pancaspe87: Raised https://github.com/atlassian-labs/compiled/issues/1323

JakeLane (approved):

liamqma (approved):

packages/babel-plugin/src/__tests__/module-traversal.test.ts
```diff
@@ -589,5 +589,50 @@ describe('module traversal', () => {
 
       expect(actual).toInclude('{color:red}');
     });
+
+    it('should resolve member expression in CSS prop', () => {
```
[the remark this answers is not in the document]
pancaspe87: discussed offline. We didn't find any actual usages in products and we think that declarations should resemble CSS and be something like

css({
    color: 'red',
    background: 'red'
  })
  
  Array expressions are supported.

Issue #1324, opened 2022-11-29
[ESLINT] @compiled/no-exported-css should not throw if exported component makes reference to css({})


pancaspe87:
Looking at the source code of this ESLint rule [compiled/validate-definition.ts at master · atlassian-labs/compiled](https://github.com/atlassian-labs/compiled/blob/master/packages/eslint-plugin/src/utils/create-no-exported-rule/validate-definition.ts#L143) , it iterates through all of the references to css({}).

Example:

const foo = css` color: 'red'; `; 
// @compiled/no-exported-css throws because bar references foo and bar is exported 
export const bar = ` ${foo} ` 

Same thing in our case here

const foo = css` color: 'red'; `; 
// @compiled/no-exported-css throws as well export 
const bar = styled.div` ${foo} ` 

Adjust the linting rule to allows the 2nd scenario because exporting a component is legit.

leeguooooo:
This looks like it has been resolved by the refactor of `check-if-compiled-export.ts`. The current code in [`packages/eslint-plugin/src/utils/create-no-exported-rule/check-if-compiled-export.ts`](https://github.com/atlassian-labs/compiled/blob/master/packages/eslint-plugin/src/utils/create-no-exported-rule/check-if-compiled-export.ts) — line 126-131 — explicitly bails out when the surrounding usage is a styled component:

```ts
// Exporting a component with a css reference should be allowed
if (isStyledComponent(nodes, context)) {
  return {
    isExport: false,
  };
}
```

The `no-exported-css` rule's test suite at [`rule.test.ts:688-738`](https://github.com/atlassian-labs/compiled/blob/master/packages/eslint-plugin/src/rules/no-exported-css/__tests__/rule.test.ts#L688-L738) covers the exact second scenario from this issue and asserts it as **valid** (no errors expected):

```ts
import { css, styled } from '@compiled/react';

const styles = /* css(...) or css`...` */;

export const Component = styled.div`${styles}`;  // valid, no error
```

…along with the call-expression form (`styled.button(buttonStyle)` where `buttonStyle = css\`...\``), aliased `styled as styled2`, conditional-interpolation `${({ isXEnabled }) => isXEnabled ? css\`width: 100px\` : undefined}`, and the `styled(Component)(css({...}))` higher-order form.

[...]

PR #1339, opened 2022-12-12
Fix/1324 no exported css reference

pancaspe87:
Fix for https://github.com/atlassian-labs/compiled/issues/1324

At current` @compiled/no-exported-css` ESLint rule will throw an error when exporting components with references to `css({})`. This seems to happen only for the styled API, not for CSS prop (see [test cases](https://github.com/atlassian-labs/compiled/blob/master/packages/eslint-plugin/src/rules/no-exported-css/__tests__/rule.test.ts#L38-L89), they don't expect errors ).

Exporting components should be allowed

```
const foo = css`color: 'red'`;
export const bar = styled.div${foo}
```


packages/eslint-plugin/src/rules/no-emotion-css/index.ts
```diff
@@ -2,8 +2,8 @@ import type { Rule } from 'eslint';
 import type { ImportSpecifier, ImportDeclaration } from 'estree';
 
 import { buildImportDeclaration, buildNamedImport } from '../../utils/ast-to-string';
+import { COMPILED_IMPORT } from '../../utils/constants';
```
pancaspe87, line 5: Unrelated to the issue/change. However, I noticed that the constant was used several time. Hence, created a file for it

packages/eslint-plugin/src/rules/no-exported-css/__tests__/rule.test.ts
```diff
@@ -685,6 +685,39 @@ const createTestCases = (importSource: string) =>
...
+
+          const buttonStyle = css\`
+            display: flex;
+            align-items: center;
+            justify-content: center;\`;
+
+          export const Component = styled.button(buttonStyle);
+        `,
+        errors: [],
+      },
```
liamqma, line 720: I've seen something like below in Confluence.

```
export const Foo = styled.div`
    color: black;
    ${({ isXEnabled }) => isXEnabled ? css`width: 100px` : undefined
`
```

Can you please add a test case for it?

packages/eslint-plugin/src/utils/create-no-exported-rule/is-styled-component.ts
```diff
@@ -0,0 +1,59 @@
...
+  if (node && node.type === 'CallExpression' && node.callee.type === 'MemberExpression') {
+    return node.callee;
+  }
+
+  // Eg. const Component = styled.div`${styles}`;
+  if (node && node.type === 'TaggedTemplateExpression' && node.tag.type === 'MemberExpression') {
+    return node.tag;
+  }
+
+  return;
```
liamqma, line 30: What about `styled(button)(styled)`?

packages/eslint-plugin/src/rules/no-exported-css/__tests__/rule.test.ts
pancaspe87: - [ ] add test

packages/eslint-plugin/src/utils/create-no-exported-rule/is-styled-component.ts
pancaspe87: Can we do this in Compiled?

liamqma: oh,sry, what i mean is

```
const button = styled.div({ color: 'red' });
const button2 = styled(button)({ width: '100px' })
```

pancaspe87: I think I get it now. I'll check both cases

1. exporting a component that extends a component using css. Eg

```
const button = styled.div(css({ color: 'red' }));
export const button2 = styled(button)({ width: '100px' });
```

OR 

2. exporting a component that extends another component and has a css reference
```
const button = styled.div({ color: 'red' });
export const button2 = styled(button)(css({ width: '100px' }));
```

liamqma: That's a nice summary 👍

pancaspe87: they should be the same in terms of AST

liamqma (approved):

JakeLane (approved):

PR #1343, opened 2023-01-05
Preserve unresolved imports codemod

JakeLane:
Currently, we remove unresolved imports and add a comment explaining this. There are some imports that have no replacement so it makes more sense to preserve them, which will make code mods require less manual intervention to land changes.

We could consider still adding the comment for documentation purposes but it seemed a bit noisy to me.

### DIFF
--- a/.changeset/twenty-wasps-care.md
+++ b/.changeset/twenty-wasps-care.md
@@ -1,0 +1,5 @@
+---
+'@compiled/codemods': minor
+---
+
+Preserve unresolved imports in codemods instead of removing and adding a comment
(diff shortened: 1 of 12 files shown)


liamqma (dismissed):

pancaspe87:
Hey @JakeLane,

are unresolved imports documented anywhere else? If not, should we add a note in the codemods README or in https://compiledcssinjs.com/docs/migrating?

JakeLane:
@pancaspe87 good call! Yeah I'll add it to the docs after this PR and update the readme.

liamqma (approved):

PR #1351, opened 2023-01-10
Fix flex shorthand expansion when flex basis is 0

### LINKED ISSUE #1350: flex property expansion in CSS is handled incorrectly for flex-basis 0
**Describe the bug**
flex shorthands are not handled correctly and cause incorrect CSS.

**To Reproduce**

```js
css({
    flex: '1 1 0'
});
```

The flex property is dropped entirely.

**Expected behavior**
The flex property should be expanded into atomic rules correctly.

**Additional context**
Workaround is to add some unit, for example **px**

```js
css({
    flex: '1 1 0px'
});
```

JakeLane:
Fixes #1350

We should expand 0 flex-basis values to 0px despite it being different in the W3 spec. This is because browsers all transform to 0px so we should match that instead.

### DIFF
--- a/packages/css/src/plugins/expand-shorthands/__tests__/flex.test.ts
+++ b/packages/css/src/plugins/expand-shorthands/__tests__/flex.test.ts
@@ -137,7 +137,7 @@ describe('flex property expander', () => {
               "
                     flex-grow: 1;
                     flex-shrink: 1;
-                    flex-basis: 0px;
+                    flex-basis: 0%;
                   "
           `);
     });
--- a/packages/css/src/plugins/expand-shorthands/__tests__/flex.test.ts
+++ b/packages/css/src/plugins/expand-shorthands/__tests__/flex.test.ts
@@ -127,6 +127,20 @@ describe('flex property expander', () => {
                   "
           `);
     });
+
+    it('should expand flex 0 basis', () => {
+      const result = transform`
+      flex: 1 1 0;
+    `;
+
+      expect(result).toMatchInlineSnapshot(`
+              "
+                    flex-grow: 1;
+                    flex-shrink: 1;
+                    flex-basis: 0px;
+                  "
+          `);
+    });
   });
 
   describe('is invalid', () => {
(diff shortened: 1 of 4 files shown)


packages/css/src/plugins/expand-shorthands/__tests__/flex.test.ts
```diff
@@ -127,6 +127,20 @@ describe('flex property expander', () => {
...
+    it('should expand flex 0 basis', () => {
+      const result = transform`
+      flex: 1 1 0;
+    `;
+
+      expect(result).toMatchInlineSnapshot(`
+              "
+                    flex-grow: 1;
+                    flex-shrink: 1;
+                    flex-basis: 0px;
```
pancaspe87, line 140: I should probably try to replicate the[ issue](https://github.com/atlassian-labs/compiled/issues/1350) to understand the change, but should it be `0%` based on [previous fix](https://github.com/atlassian-labs/compiled/pull/1317/files#diff-5b489bd99f3b48b7aa7144ee849e92582c4a6e452ad02874b3216aa16ca01d73R10) or the latter is addressing a different issue?

JakeLane: This is the same issue but a different manifestation, as we didn't expect the expansion to be required for three-valued flex values.

JakeLane: Also I think the unit is irrelevant here but consistency would be better, will fix

packages/css/src/plugins/expand-shorthands/utils.ts
```diff
@@ -74,6 +83,9 @@ export const isWidth = (node: Node): boolean => {
  */
 export const getWidth = (node: Numeric | Word | Func): string => {
   if (node.type === 'numeric') {
+    if (node.value === '0' && node.unit === '') {
```
pancaspe87, line 86: I guess this will affect all unit based properties ( eg `height`, `width` not only `flex`).
However, unit should be irrelevant for those cases too

JakeLane: hrm yeah, these utils are just used for flex but they could be useful for any future shorthands. I'll refactor this a bit to be a bit more readable

pancaspe87 (approved):

PR #1354, opened 2023-01-16
Remove `ax` from runtime 

liamqma:
This PR is the first attempt to remove `ax` from runtime. It will remove `ax` if the below conditions are met:
- Use CSS props 
- Does not receive className
- Atomic CSS rules are statically defined

Example
```
```

### DIFF
--- a/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.ts
+++ b/packages/babel-plugin/src/css-prop/__tests__/behaviour.test.ts
@@ -152,7 +152,7 @@ describe('css prop behaviour', () => {
       <div css={{ fontSize: 20 }} {...props} />
     `);
 
-    expect(actual).toInclude('<div{...props}className={ax(["_1wybgktf"])}/>');
+    expect(actual).toInclude('<div{...props}className="_1wybgktf"/>');
   });
 
   it('should pass through static props', () => {
@@ -162,7 +162,7 @@ describe('css prop behaviour', () => {
       <div css={{ fontSize: 20 }} role="menu" />
     `);
 
-    expect(actual).toInclude('<div role="menu"className={ax(["_1wybgktf"])}/>');
+    expect(actual).toInclude('<div role="menu"className="_1wybgktf"/>');
   });
 
   it('should concat explicit use of class name prop from an identifier on an element', () => {
@@ -182,15 +182,15 @@ describe('css prop behaviour', () => {
       import '@compiled/react';
 
       const base = { color: 'black' };
-      const top = \` color: red; \`;
+      const top = \` marginTop: 10px; \`;
 
       <div css={[base, top]}>hello world</div>
     `);
 
     expect(actual).toIncludeMultiple([
-      '._syaz5scu{color:red}',
       '._syaz11x8{color:black}',
-      '<div className={ax(["_syaz11x8","_syaz5scu"])}>hello world</div>',
+      '._16f719bv{marginTop:10px}',
+      '<div className="_syaz11x8 _16f719bv">hello world</div>',
     ]);
   });
 
@@ -209,7 +209,7 @@ describe('css prop behaviour', () => {
       '._1wyb1fwx{font-size:12px}',
       '._syaz11x8{color:black}',
       '._1e0c1o8l{display:inline-block}',
-      '<div className={ax(["_1e0c1o8l _syaz11x8","_1wyb1fwx _1bsb12am"])}>hello world</div>',
+      '<div className="_1e0c1o8l _syaz11x8 _1wyb1fwx _1bsb12am">hello world</div>',
     ]);
   });
 
@@ -222,7 +222,7 @@ describe('css prop behaviour', () => {
 
     expect(actual).toInclude(`{color:blue}`);
     expect(actual).toInclude(
-      `<div style={{display:'block'}}className={ax([\"_syaz13q2\"])}>hello world</div>`
+      `<div style={{display:'block'}}className=\"_syaz13q2\">hello world</div>`
     );
   });
 
@@ -452,7 +452,7 @@ describe('css prop behaviour', () => {
       'const _2="._y44vk4ag{animation:fadeOut 2s ease-in-out}"',
       'const _="@keyframes fadeOut{0%{opacity:1}50%{opacity:0.5}to{opacity:0}}"',
       '<CS>{[_,_2]}</CS>',
-      'className={ax(["_y44vk4ag"])}',
+      'className="_y44vk4ag"',
     ]);
   });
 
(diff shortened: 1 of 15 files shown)


packages/babel-plugin/src/utils/build-compiled-component.ts
```diff
@@ -57,6 +58,49 @@ const getExpression = (
   return value;
 };
 
+/**
+ * Check if 'node' is a 'inBuiltComponent'.
+ * We assume it's a 'inBuiltComponent' if it starts with a lowercase letter (i.e. div).
+ *
+ * @param node Originating node
+ */
+const isInBuiltComponent = (node: t.JSXElement): boolean => {
```
liamqma, line 67: - [ ] Add a test case for "UserDefinedComponent"

packages/babel-plugin/src/utils/build-compiled-component.ts
```diff
@@ -57,6 +58,49 @@ const getExpression = (
...
+ * We assume it's required if
+ * - there is a class name prop statically defined
+ * - there is conditional css
+ * - 'node' is a 'UserDefinedComponent'
+ *
+ * @param node Originating node
+ * @param classNames The classNames for a component at runtime
+ *
+ */
+const isAxRequired = (node: t.JSXElement, classNames: t.Expression[]) => {
```
liamqma, line 97: - [ ] Add a test case for dynamic atomic declaration

packages/babel-plugin/src/utils/build-compiled-component.ts
```diff
@@ -1,6 +1,7 @@
 import generate from '@babel/generator';
 import template from '@babel/template';
 import * as t from '@babel/types';
+import { ax } from '@compiled/react/runtime';
```
liamqma, line 4: Not sure if a babel plugin should import stuff from runtime. I've been thinking about moving `ax` to `@compiled/utils`. But I noticed `@compiled/utils` requires some modules (e.g. source-map) which shouldn't exist in runtime. Although bundler might remove it from runtime code, I think it's still dangerous for runtime to depend on `@compiled/utils`.

It's safe other way around. (build time library depends on runtime library)

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.ts
```diff
@@ -182,15 +182,15 @@ describe('css prop behaviour', () => {
       import '@compiled/react';
 
       const base = { color: 'black' };
-      const top = \` color: red; \`;
+      const top = \` marginTop: 10px; \`;
```
pancaspe87, line 185: Why was this changed?

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.ts
```diff
@@ -182,15 +182,15 @@ describe('css prop behaviour', () => {
...
 
       <div css={[base, top]}>hello world</div>
     `);
 
     expect(actual).toIncludeMultiple([
-      '._syaz5scu{color:red}',
       '._syaz11x8{color:black}',
-      '<div className={ax(["_syaz11x8","_syaz5scu"])}>hello world</div>',
```
pancaspe87, line 193: I never looked into how [composition](https://compiledcssinjs.com/docs/composition#composition) works, but I would have expected only 1 class (`_syaz5scu `) in here in the original example

packages/babel-plugin/src/utils/build-compiled-component.ts
```diff
@@ -57,6 +58,49 @@ const getExpression = (
...
+ * - there is conditional css
+ * - 'node' is a 'UserDefinedComponent'
+ *
+ * @param node Originating node
+ * @param classNames The classNames for a component at runtime
+ *
+ */
+const isAxRequired = (node: t.JSXElement, classNames: t.Expression[]) => {
+  return (
+    isInBuiltComponent(node) &&
```
pancaspe87, line 99: maybe I misunderstood but thought that `ax` is not required in cases such as `<div css={css({color: 'red'})} />`

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.ts
liamqma: Yea, only 1 class would exist at runtime after `ax` is executed. Previously the babel-plugin wouldn't do any duplication. It just simply constructs something like `ax(["_syaz11x8","_syaz5scu"]` and leave it to runtime `ax` to ensure one atomic group (color) exists.

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.ts
liamqma: because I created another test case to [test deduping atomic group](https://github.com/atlassian-labs/compiled/pull/1354/files#diff-69558aafb510bb07e99ed2dc72efeb29d5ff2370b0a3910c2a996c71f4081fa9R737). So I updated this one to not have duplicated `color`.

packages/babel-plugin/src/utils/build-compiled-component.ts
liamqma: Sorry, I don't quite follow your question. Can you elaborate it please?

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.ts
pancaspe87: yeahhh, makes sense.

I think I was getting confused by the Styled API behaviour; IIRC it would generate only one class for each property

packages/babel-plugin/src/utils/build-compiled-component.ts
pancaspe87: nvm, I just read the description... I get it now :)

packages/babel-plugin/src/utils/build-compiled-component.ts
JakeLane: I think this is ok as we don't import from the babel plugin in the react package. 

As for side-effects in the runtime accidentally being imported, we should be safe if we use the `size-limit` tool to prevent regressions.

packages/babel-plugin/src/utils/build-compiled-component.ts
```diff
@@ -57,6 +58,49 @@ const getExpression = (
...
+  ) {
+    tagName = node.openingElement.name.name;
+  } else {
+    tagName = node.openingElement.name.property.name;
+  }
+
+  if (typeof tagName !== 'string') return false;
+
+  // Check if the first letter is lower case
+  return tagName.charCodeAt(0) >= 97 && tagName.charCodeAt(0) <= 122;
```
JakeLane, line 83: I don't think this is safe as we could have a user-defined component with a lower case name. We should find an alternative way to decide if this is a dom component.

packages/babel-plugin/src/utils/build-compiled-component.ts
```diff
@@ -57,6 +58,49 @@ const getExpression = (
...
+};
+
+/**
+ * Check if the `ax` runtime utility is required.
+ * We assume it's required if
+ * - there is a class name prop statically defined
+ * - there is conditional css
+ * - 'node' is a 'UserDefinedComponent'
```
JakeLane, line 91: props spreading is also another way for the component to have a defined classname

JakeLane (changes requested):

packages/babel-plugin/src/utils/build-compiled-component.ts
liamqma: Yup, you are right. 👍

packages/babel-plugin/src/utils/build-compiled-component.ts
liamqma: Good point! 👍

- [ ] consider props spreading 

I notice props spreading can cause issue in existing/master codebase too. Raised an issue: https://github.com/atlassian-labs/compiled/issues/1361

liamqma:
Close this PR for now because removing `ax` is no longer a prerequisite of compressing class names. [Instead, we update 'ax' to accept short class name](https://github.com/atlassian-labs/compiled/pull/1396). The reason is we can only remove `ax` in limited cases. Even in those cases, it requires AST tree modification, which is prone to bugs. 
We'll revisit removing `ax` as runtime improvement holistically in the future.

PR #1383, opened 2023-01-20
Add new ESLint rule for missing css function on styles

JakeLane:
Introducing a new ESLint rule that enforces the usage of the `css` import when styling with the CSS prop API.

We want to encourage this wherever possible as makes the build process more deterministic and improves code quality.


packages/eslint-plugin/src/rules/no-css-prop-without-css-function/__tests__/rule.test.ts
```diff
@@ -0,0 +1,185 @@
...
+
+        const styles = { backgroundColor: 'red' };
+
+        <div css={styles} />;
+      `,
+      output: outdent`
+        import React from 'react';
+        import { css } from '@compiled/react';
+
+        const styles = css({ backgroundColor: 'red' });
```
liamqma, line 106: Why is this invalid?

packages/eslint-plugin/src/rules/no-css-prop-without-css-function/__tests__/rule.test.ts
```diff
@@ -0,0 +1,185 @@
...
+
+        const styles = { backgroundColor: 'red' };
+
+        <div css={[someBoolean ? styles : undefined]} />;
+      `,
+      output: outdent`
+        import React from 'react';
+        import { css } from '@compiled/react';
+
+        const styles = css({ backgroundColor: 'red' });
```
liamqma, line 179: Would you mind adding a test of valid way of using Ternary?

packages/eslint-plugin/src/rules/no-css-prop-without-css-function/index.ts
```diff
@@ -0,0 +1,93 @@
...
+import type { Node } from 'estree';
+import type { JSXExpressionContainer } from 'estree-jsx';
+
+import { findCompiledImportDeclarations } from '../../utils/ast';
+import { addImportToDeclaration, buildImportDeclaration } from '../../utils/ast-to-string';
+
+type Q<T> = T extends Scope.Definition ? (T['type'] extends 'Variable' ? T : never) : never;
+type VariableDefinition = Q<Scope.Definition>;
+
+const findStyleNodes = (node: Node, references: Scope.Reference[], context: Rule.RuleContext) => {
```
liamqma, line 11: Nice recursive function!

liamqma (approved):

PR #1396, opened 2023-01-25
Update 'ax' to accept short class name

liamqma:
This PR updates `ax` function to accept short class name. It paves the way for compressing Atomic class names.

Example
```
// The below returns 'c'. 
// 'aaaa' is the Atomic group and 'c' is the compressed class name.
ax(['_aaaabbbb', '_aaaa_c']) 
```

Completely removing `ax` from runtime is difficult. The idea is to give `ax` both Atomic group and short class name. The Atomic group name is needed to ensure only one Atomic group exists. This idea is borrowed from https://medium.com/@a.sachin533/ab8bc09ee4da. 

Note there will be another PR where `@compiled/babel-plugin` updates

```
ax(['_aaaabbbb', '_aaaacccc']) 
```

to

```
ax(['_aaaabbbb', '_aaaa_c']) 
```


packages/react/src/runtime/ax.ts
```diff
@@ -27,10 +27,14 @@ const ATOMIC_GROUP_LENGTH = 5;
  *
  * @param classes
  */
-export default function ax(classNames: (string | undefined | false)[]): string | undefined {
+export default function ax(
```
JakeLane, line 30: Could you please do a benchmark for the function? I don't think there's anything controversial here but important to be careful

.eslintrc.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   root: true,
   ignorePatterns: [
     'dist',
+    'build',
```
liamqma, line 5: To ignore `examples/ssr/build`.

packages/react/src/runtime/ax.ts
liamqma: - `ax` on Master branch: 486,792 ops/sec
- `ax` on this branch: 212,283 ops/sec
- `ax` with compressed class names on this branch: 261,141 ops/sec

Based on the benchmark, this PR makes `ax` 2 times slower due to additional `if Array.isArray()`. Hopefully the performance gain from compressing the classnames outweighs the loss here.
Note `ax([['aaaa', 'b'], ['aaaa', 'c']])` is slightly more performant than `ax(['_aaaabbbb', '_aaaacccc'])`. I think it's because the former skips `.slice`.

JakeLane (approved):

packages/react/src/runtime/ax.ts
JakeLane: Awesome work!

liamqma:
Decline this PR, and re-create a [PR](https://github.com/atlassian-labs/compiled/pull/1408) which includes the changes of both `ax` and `babel-plugin`.

PR #1408, opened 2023-02-06
Compress atomic class names

liamqma:
This PR does

- updates ax function to accept short class name
- updates babel-plugin to compress class names based on pre-bundling hash map
- generates compression map


liamqma:
TODO

- [ ] Changeset
- [ ] Parcel integration

pancaspe87 (commented):
Are we testing overrides here?

packages/react/src/runtime/__tests__/ax.test.ts
```diff
@@ -1,70 +1,81 @@
...
+  const isEnabled: boolean = (() => false)();
+
+  it.each([
+    ['should handle empty array', [], undefined],
+    ['should handle array with undefined', [undefined], undefined],
+    ['should join single classes together', ['foo', 'bar'], 'foo bar'],
+    ['should join multi classes together', ['foo baz', 'bar'], 'foo baz bar'],
+    ['should remove undefined', ['foo', 'bar', undefined], 'foo bar'],
+    [
+      'should ensure the last atomic declaration of a single group wins',
```
pancaspe87, line 13: Are we testing overrides here?

packages/react/src/runtime/__tests__/ax.test.ts
```diff
@@ -1,70 +1,81 @@
...
+      ['_aaaabbbb', '_aaaacccc'],
+      '_aaaacccc',
+    ],
+    [
+      'should ensure the last atomic declaration of a single group with short class name wins',
```
pancaspe87, line 18: I haven't followed up the previous PRs and I am still wrapping my head around these changes. Hence, apologies if this was mentioned somewhere else

I am assuming that:
- `_aaaabbbb` is the CSS class with no manipulation
- `a` is the optimised class, which is stored in the map
- `_aaaa_a` is made of atomic group + optimised class -> in this way we are able to handle overrides

examples/webpack/class-name-compression-map.json
```diff
@@ -0,0 +1,27 @@
+{
+  "1wyb12am": "a",
```
pancaspe87, line 2: Where/how does this map get generated?

packages/babel-plugin/src/__tests__/index.test.ts
```diff
@@ -201,4 +201,166 @@ describe('babel plugin', () => {
 
     expect(actual).toInclude('c_MyDiv');
   });
+
```
pancaspe87, line 204: Do we still have cases where compression is not safe, hence we should bail out?

packages/react/src/runtime/__tests__/ax.test.ts
liamqma: Yes, spot on!

packages/react/src/runtime/__tests__/ax.test.ts
liamqma: I think so. It should be covered by the next test case.
```
[
      'should ensure the last atomic declaration of a single group with short class name wins',
      ['_aaaabbbb', '_aaaacccc', '_aaaa_a'],
      'a',
],
```
Is it what you refer to?

examples/webpack/class-name-compression-map.json
liamqma: Good question! For now it's assumed that library consumers manually curate the map and pass it to the babel-plugin. We will later create an automatic method to generate the map by statically analysing the codebase, and the map should be periodically updated by some a build.

packages/babel-plugin/src/class-names/index.ts
```diff
@@ -132,15 +132,12 @@ export const visitClassNamesPath = (path: NodePath<t.JSXElement>, meta: Metadata
       }
 
       const builtCss = buildCss(styles, meta);
-      const { sheets, classNames } = transformCss(
-        builtCss.css.map((x) => getItemCss(x)).join(''),
-        meta.state.opts
-      );
+      const { sheets, classNames } = transformCssItems(builtCss.css, meta);
```
liamqma, line 135: The change in this file is to make <ClassNames> handle conditional CSS rules.
i.e. 
```
<ClassNames>
          {({ css }) => (<div className={css({ ...props.isPrimary && { color: 'blue' }})}>hello, world!</div>)}
</ClassNames>
```

packages/css/src/plugins/atomicify-rules.ts
```diff
@@ -76,20 +77,28 @@ const replaceNestingSelector = (selector: string, parentClassName: string) => {
...
       ...opts,
       selectors: [normalizedSelector],
     });
-    const replacedSelector = replaceNestingSelector(normalizedSelector, className);
 
-    selectors.push(replacedSelector);
+    const compressedClassName =
+      classNameCompressionMap && classNameCompressionMap[fullClassName.slice(1)];
+
+    if (compressedClassName) {
```
liamqma, line 93: If compressed class name is available, we are replacing full class name with the compressed in the sheet.
So
```
.aaaabbbb { font-size: 10px }
```
becomes
```
.a { font-size: 10px }
```
It will reduce the size of stylesheet but will break if we don't replace every occurrence of `_aaaabbbb`, which I should have done in this PR. 🤞

packages/babel-plugin/src/__tests__/index.test.ts
liamqma: Compression should be safe everywhere now.

packages/react/src/runtime/__tests__/ax.test.ts
pancaspe87: yep!

packages/css/src/utils/class-name-generator.ts
```diff
@@ -0,0 +1,39 @@
+const acceptPrefix = 'abcdefghijklmnopqrstuvwxyz_'.split('');
```
JakeLane, line 1: Did you miss uppercase chars?

packages/babel-plugin/src/__tests__/index.test.ts
```diff
@@ -201,4 +201,160 @@ describe('babel plugin', () => {
...
+      \`;
+    `,
+      {
+        classNameCompressionMap: {
+          '1wyb1fwx': 'a',
+        },
+      }
+    );
+
+    expect(actual).toIncludeMultiple(['.a{font-size:12px}', 'ax(["_1wyb_a", __cmplp.className])']);
```
pancaspe87, line 221: Does the extracted CSS also contain
`._1wyb_a, .1wyb1fwx {font-size:12px}`?

packages/css/src/__tests__/generate-compression-map.test.ts
```diff
@@ -0,0 +1,139 @@
...
+    '121jagmp': 'rb',
+    '1bumglyw': 'sb',
+    sedtglyw: 'tb',
+    '14kw1hna': 'ub',
+    mc2h1hna: 'vb',
+    '14fy1hna': 'wb',
+    '1vdp1hna': 'xb',
+    '17gjpfqs': 'yb',
+  };
+  it('should generate class names without the old compression map', () => {
```
pancaspe87, line 123: Is the compression map supposed to be stable across builds? 
Here we are just testing that it gets generated; however, in the following test it seems that a different compressed class is generated for `17gjpfqs`

packages/css/src/utils/class-name-generator.ts
liamqma: Good point. I had a brief look at the spec. CSS class is case sensitive. We'll do more research to confirm it.

packages/babel-plugin/src/__tests__/index.test.ts
liamqma: No, the extracted CSS looks like
```
.a {font-size:12px}
```

packages/react/src/runtime/__perf__/ax.test.ts
```diff
@@ -39,4 +39,24 @@ describe('ax benchmark', () => {
       fastest: ['ax() string'],
     });
   }, 30000);
+
+  it('completes with ax() non-compresssed class names as the fastest', async () => {
```
dddlr, line 43: Nit: typo in "compresssed"

packages/css/src/__tests__/generate-compression-map.test.ts
liamqma: I hope I understand the question correctly 🙂. Let me explain it.

> Is the compression map supposed to be stable across builds?

Yes, the compression map is supposed to be stable across builds. The way it achieves this is to take the compression map from the previous build as parameter:

```
generateCompressionMap = (
  css: string,
  oldClassNameCompressionMap?: { [index: string]: string }
):
```

For example, let's say we have the CSS:
```
._aaaabbbb { color: red }
._aaaacccc { color: black }
```
oldClassNameCompressionMap
```
{
   "aaaabbbb": "ddd"
}
```
`generateCompressionMap` returns new compression map
```
{
   "aaaacccc": "a"
   "aaaabbbb": "ddd" // Force "aaaabbbb" to be "ddd", which is same as oldClassNameCompressionMap
}
```

In the following test case, without providing an oldCompressionMap, '17gjpfqs' is compressed to "yb", but because we provide oldCompressionMap as  { "17gjpfqs": "a", 1vdp1hna': 'b',  '14fy1hna': 'c'}, '17gjpfqs' is compressed to 'a'. We want to ensure the new compression map keeps the same class name mapping from the oldCompressionMap.

examples/parcel/.compiledcssrc
```diff
@@ -20,5 +20,6 @@
     ]
   ],
   "extract": true,
-  "optimizeCss": false
+  "optimizeCss": false,
+  "classNameCompressionMap": {"1wyb12am":"a","syaz5scu":"b","syazruxl":"c","k48pbfng":"d","30l35scu":"e","f8pj13q2":"f","1e0c1o8l":"g","ca0qftgi":"h","u5f3ftgi":"i","n3tdftgi":"j","19bvftgi":"k","19itak0l":"l","2rko1l7b":"m","syaz1aj3":"n","1p1dangw":"o","bfhkbf54":"p"}
```
JakeLane, line 24: this should probably be an external file right?

packages/babel-plugin/src/__tests__/index.test.ts
```diff
@@ -201,4 +201,160 @@ describe('babel plugin', () => {
...
+
+  it('should compress pseudo classes', () => {
+    const actual = transform(
+      `
+      import '@compiled/react';
+      <div css={{ "&:hover": { fontSize: 12 }, "&:active": { color: 'red' } }} />
+    `,
+      {
+        classNameCompressionMap: {
+          '9h8h5scu': 'a',
```
JakeLane, line 289: nit: inconsistent quotes?

packages/css/src/__tests__/generate-compression-map.test.ts
pancaspe87: thanks for the explanation! Now I get it! I was looking at the  key/pair values that I didn't realise what the comparison was about  :)

packages/css/src/generate-compression-map.ts
```diff
@@ -0,0 +1,68 @@
...
+ * @param oldClassNameCompressionMap the previous compression map, which ensures the compression is deterministic.
+ * @returns newClassNameCompressionMap
+ */
+export const generateCompressionMap = (
+  css: string,
+  oldClassNameCompressionMap?: { [index: string]: string }
+): undefined | { [index: string]: string } => {
+  let classNamesToCompress: string[] = [];
+  const classNameCompressionMap: { [index: string]: string } = {};
+  const reservedClassNames = ['ad', 'ads']; // Prevent AdBlocker from removing the class names
```
pancaspe87, line 22: Nice! 

Out of curiosity, was this based on Tom's comment or did you find any resource with a list of AdBlocker classes.
Maybe `adv` could be an issue too ( [eg](https://stackoverflow.com/a/20242141))? 

Also, given that we prefix classes with `_`, I wonder if compressed classes are safe, regardless

packages/react/src/runtime/__perf__/ax.test.ts
pancaspe87: good catch!

packages/css/src/utils/__tests__/class-name-generator.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    });
+  });
+
+  it('should skip reservedClassNames', () => {
+    const generator = new ClassNameGenerator({ reservedClassNames: ['a', 'b', 'c'] });
+    const className = generator.generateClassName('foo');
+    expect(className).toBe('d');
+  });
+
+  it('should generate class names starting with a number', () => {
```
pancaspe87, line 18: I guess this is safe because the class will be prefixed with `_`

packages/react/src/runtime/__perf__/ax.test.ts
```diff
@@ -3,24 +3,24 @@ import { runBenchmark } from '@compiled/benchmark';
...
+    '_n3tdftgi',
+    '_19bv19bv',
+    '_bfhk1mzw',
+    '_syazu67f',
+    '_k48p1nn1',
+    '_ect41kw7',
+    '_1wybdlk8',
+    '_irr3mlcl',
+    '_1di6vctu',
+    undefined,
```
pancaspe87, line 20: Why is there `undefined` in here?

packages/css/src/__tests__/generate-compression-map.test.ts
```diff
@@ -0,0 +1,139 @@
...
+    @media screen and (min-width:1300px){._1jhpoyl8{max-width:10vw}}
+`;
+
+  const baseResult = {
+    '154i14e6': 'a',
+    '14tk72c6': 'b',
+    '14n4stnw': 'c',
+    '13h81y44': 'd',
+    '1di6k6hx': 'e',
+    irr3k6hx: 'f',
```
dddlr, line 48: Would it make a difference if we change all of these keys (irr3k6hx, jomrk6hx, etc) to be in single quotes? I'm not quite sure why some are in single quotes and some are not

dddlr: Ohh I think I see why, the number(s) at the start of the key means that the single quotes are required right?

dddlr: Would personally prefer consistency, i.e. all of the keys being in single quotes

packages/css/src/generate-compression-map.ts
liamqma: Good point! I had a look at the easylist. I think I will take the approach of [styled-components](https://github.com/styled-components/styled-components/blob/main/packages/styled-components/src/utils/generateAlphabeticName.ts#L1), which avoids any word includes `ad`. Let me update the code.

dddlr: Should we add 'ddb' to this list as well? [EasyList](https://easylist.to/easylist/easylist.txt) (which is available for Adblock Plus, uBlock Origin etc) will block any class named "ddb"

> Also, given that we prefix classes with _, I wonder if compressed classes are safe, regardless

I'm curious about this too :)

packages/css/src/utils/__tests__/class-name-generator.test.ts
liamqma: Ah, good catch. The test case should be `should NOT generate class names starting with a number`. 🤦‍♂️
I still haven't made the mind of whether prefix with `_`. Prefixing with `_` guarantee the uniqueness but reduce the perf improvement. 
I think I will provide a option and let the library consumer to decide.

packages/react/src/runtime/__perf__/ax.test.ts
liamqma: `undefined` is an acceptable parameter so we want to include it in the test case.

```
ax([undefined, '.aaaabbbb'])
```

btw, I didn't add `undefined`.

packages/css/src/__tests__/generate-compression-map.test.ts
liamqma: The formatting is automatically done by [prettier](https://prettier.io/blog/2019/04/12/1.17.0.html#add-an-option-to-modify-when-prettier-quotes-object-properties-5934-by-azz). We don't customize this rule so it's the default "as-needed". I am not particular about the formatting but happy to change it if you prefer.

packages/babel-plugin/src/__tests__/index.test.ts
liamqma: https://github.com/atlassian-labs/compiled/pull/1408#discussion_r1106683169

examples/parcel/.compiledcssrc
liamqma: - [ ] externalize the map

packages/css/src/__tests__/generate-compression-map.test.ts
dddlr: Ah no worries, thank you for clarifying! Since it's being done by our linter, I'd be happy to leave it as-is

packages/css/src/utils/__tests__/class-name-generator.test.ts
pancaspe87: +1 on making it optional, sounds a perfect solution for different consumers :)

packages/react/src/runtime/__perf__/ax.test.ts
dddlr: Optional: I know the `undefined` wasn't added by your PR, but I think incorporating what you've said here (`undefined` is a valid parameter to `ax`) into a brief comment in the code itself might be helpful for future developers. I know that I'd definitely be confused to see it in a test case as well 😅

liamqma: - [ ] Add comment

dddlr:
Thank you Liam for updating the PR :) This looks good to me, but I'll wait for an approval from @JakeLane or @pancaspe87 too since I don't have a good understanding of the whole codebase

JakeLane (approved):

dddlr (approved):

PR #1422, opened 2023-02-16
Add support for all CSS at-rules

### LINKED ISSUE #1297: @-moz-document url-prefix() at rule is not supported
**Describe the bug**
```
import '@compiled/react';

<>
  <div css={{ '@-moz-document url-prefix()': { color: 'red' } }} />
</>
```

gets transformed to

```
const _ = "@-moz-document url-prefix() { color: red; }";
<>
  <CC>
    <CS>{[_]}</CS>
    {<div className={ax([])} />} // <= no class name is given to <div>
  </CC>
</>;
```

`@-moz-document url-prefix()` atrule isn’t supported by compiled:
https://github.com/atlassian-labs/compiled/blob/master/packages/css/src/plugins/atomicify-rules.ts#L219

**Expected behavior**
It should either support `@-moz-document url-prefix()` or throw error to warn the developers. It shouldn't silently does nothing.

dddlr:
Fixes #1297


packages/css/src/plugins/atomicify-rules.ts
```diff
@@ -216,7 +216,7 @@ export const atomicifyRules = (opts = {}): Plugin => {
       root.each((node) => {
         switch (node.type) {
           case 'atrule':
-            const supported = ['media', 'supports', 'document'];
+            const supported = ['media', 'supports', 'document', '-moz-document'];
```
liamqma, line 219: I was hoping that we can support all atrules instead of keeping adding them to the list in the future. At least, if we don't support a certain atrule, we should warn the developers (or even throw an error) rather than just not working silently (I think it's what it does currently).

Hey @itsdouges , it's probably a big throwback. From your [commit](https://github.com/atlassian-labs/compiled/pull/299/files#diff-55e7dd703c9c2042f0a2ba361fc8d6e322fe3de404cc2962ea005b424b4b3ab1R207), do you remember why we have a limited support for atrules? Can we remove `if (supported.includes(node.name))`?

liamqma: oh, actually I think i understand why now. I guess it's because some atrules don't make sense to be atomicified.

e.g. `@import url;`

packages/css/src/plugins/__tests__/atomicify-rules.test.ts
```diff
@@ -432,4 +432,16 @@ describe('atomicify rules', () => {
 
     expect(actual).toMatchInlineSnapshot(`"._syaz1qpq{color:red!important}._syaz5scu{color:red}"`);
   });
+
+  // Used for detecting that the current browser is Firefox
```
pancaspe87, line 436: nit. I think the comment can be removed

packages/css/src/plugins/atomicify-rules.ts
pancaspe87: > oh, actually I think i understand why now. I guess it's because some atrules don't make sense to be atomicified.
Yeahhh, I was just looking at [At-rules](https://developer.mozilla.org/en-US/docs/Web/CSS/At-rule); I think you are spot on.

Wondering if there is any other case that we are missing

dddlr: > At least, if we don't support a certain atrule, we should warn the developers (or even throw an error) rather than just not working silently (I think it's what it does currently).

Ah that's a good point @liamqma - I'll revert this PR to a draft and see if I can add some error messaging for unsupported at rules

dddlr: Out of curiosity, would it make sense for [`@container`](https://developer.mozilla.org/en-US/docs/Web/CSS/@container) and [`@layer`](https://developer.mozilla.org/en-US/docs/Web/CSS/@layer), to be atomificied?

Looks like they support syntax like this:
```
@layer state {
    .warning {
        background-color: brown;
    }
    p {
        text-align: right;
    }
}
```

liamqma: Yes, I think so. Both makes sense to be atomificied.
In fact, devs from Atlassian have asked for it. I'll dm you the message.

dddlr:
Some things about this PR I thought I'd get your thoughts on 🙂:

1. We should definitely support `@-moz-document` due to its use by Atlassian developers, but should we also support [`@document`](https://developer.mozilla.org/en-US/docs/Web/CSS/@document)? This isn't supported by any browsers (except for Firefox which requires the `-moz-` prefix) and is non-standard.
2. Should we add `@when` and `@else`? This is part of a [W3C working draft](https://www.w3.org/TR/css-conditional-5/#when-rule) but there is [no browser support](https://caniuse.com/css-when-else) as of writing

In the meantime, I'll look into how some other libraries (as suggested by @liamqma, thank you for the tip 😃) handle these at-rules

JakeLane:
> 1. We should definitely support `@-moz-document` due to its use by Atlassian developers, but should we also support [`@document`](https://developer.mozilla.org/en-US/docs/Web/CSS/@document)? This isn't supported by any browsers (except for Firefox which requires the `-moz-` prefix) and is non-standard.
> 2. Should we add `@when` and `@else`? This is part of a [W3C working draft](https://www.w3.org/TR/css-conditional-5/#when-rule) but there is [no browser support](https://caniuse.com/css-when-else) as of writing

I think it may be best to follow vanilla-extract in this case, as new rules could be added to the spec that won't work automatically (e.g. the syntax could be different). In cases where someone specifies an at-rule we don't support, we should clearly throw an error.

liamqma:
> * Linaria:
>   [Appears to handle all at-rules the same way](https://github.com/callstack/linaria/blob/3285ccc1d00449b78b3fc74087528cd38cbdd116/packages/atomic/src/processors/helpers/atomize.ts) without verifying that the at-rule name is valid. The exception is `@keyframes`, which is handled separately.
> * Emotion: Not sure
> * styled-components: Not sure
> * vanilla-extract:
>   [Appears to support](https://github.com/vanilla-extract-css/vanilla-extract/blob/d9db9bf7a638f8ec9ba1d1518b2a5be47352f397/packages/css/src/transformCss.ts) `@media`, `@supports`, `@container`, as well as `@font-face` and `@keyframes`.

Hi Grant, thanks for the investigation and summary. Here is my two cents: 
We should support everything that makes sense to be atomisized. So it includes `@when`, `@else`, and `@document`. On the other hand, throws an error if unsupported atrules are used (e.g. `@import`). It might be easier to have a black list rather than a whitelist.

It will require you to find out the entire list of atrules, understand what each rule does, and decide if it makes sense to be atomisized. I did a quick research but couldn't find a list of all the atrules from specs. If the specs doesn't provide such thing, maybe just use MDN. As for how to decide if a certain rule is okay to atomisize, I found the general rule is to check if the rule has a selector.

Compiled CSS rules must be applied to a selector (except for keyframes). Example:

```
```

So `@charset "utf-8";` doesn't make sense to be atomisized as it doesn't apply to a selector.
It's just my thought. I am happy to budge. 🙂

dddlr:
Thank you for your advice! @JakeLane @liamqma 

> We should support everything that makes sense to be atomisized. So it includes @when, @else, and @document. On the other hand, throws an error if unsupported atrules are used (e.g. @import). It might be easier to have a black list rather than a whitelist.

Are there potentially use cases where using `@import` (and similar at-rules such as `@charset` and `@namespace`) within the input to `@compiled/css` might make sense? I noticed that throwing an error for these at-rules, where it doesn't make sense to atomicise the contents, would break [this test case](https://github.com/atlassian-labs/compiled/blob/master/packages/css/src/plugins/__tests__/atomicify-rules.test.ts#L401)

liamqma:
> Are there potentially use cases where using @import (and similar at-rules such as @charset and @namespace) within the input to @compiled/css might make sense

I cannot think of any use case now. It's a good question. I noticed [PostCSS allows plugins to attach warnings to the result](https://github.com/postcss/postcss/blob/main/lib/result.d.ts#L195). I think it makes more sense for `@compiled/css` passes the warnings (e.g.  "You cannot use `@import`") to its consumer (`@compiled/babel-plugin` in this case), and let `@compiled/babel-plugin` decide how to deal with the warnings. But that seems complicating things. I am not too sure. Happy for your decide. 😜

pancaspe87:
> We should support everything that makes sense to be atomisized. So it includes https://github.com/when, @else, and @document. On the other hand, throws an error if unsupported atrules are used (e.g. https://github.com/import). It might be easier to have a black list rather than a whitelist.

`@document` seems [deprecated](https://developer.mozilla.org/en-US/docs/Web/CSS/@document). I wouldn't include it, most recent browsers won't supported https://caniuse.com/?search=%40document, except for limited support in FIrefox that I believe you are handling it already with `-moz-document`

packages/css/src/plugins/atomicify-rules.ts
```diff
@@ -151,6 +151,47 @@ const atomicifyRule = (node: Rule, opts: AtomicifyOpts): Rule[] => {
...
+    'supports',
+    'when',
+  ];
+  const forbidden = ['charset', 'import', 'namespace'];
+  const ignored = [
+    'color-profile',
+    'counter-style',
+    'font-face',
+    'font-palette-values',
+    'keyframes',
```
liamqma, line 179: Can we handle `keyframes` as a special case?

Instead of ignoring it, I suggest we throw an error. The error is something like
```
You cannot use @keyframes at-rule here. Please see the [documentation](https://compiledcssinjs.com/docs/api-keyframes).
```

dddlr: Sure thing!

dddlr: Are there any other at-rules for which we have specific recommendations in the form of documentation (or some other external-facing document)? I'm happy to add some more specific errors for different at-rules if so

liamqma: Not that I am aware of.

packages/css/src/plugins/atomicify-rules.ts
```diff
@@ -151,6 +151,47 @@ const atomicifyRule = (node: Rule, opts: AtomicifyOpts): Rule[] => {
...
+    'page',
+    'property',
+  ];
+
+  if (canBeAtomificied.includes(node.name)) {
```
liamqma, line 184: Nice way of handling different at-rules 👍

JakeLane (approved):
nice work!

packages/css/src/plugins/atomicify-rules.ts
dddlr: Did some digging into this, and unfortunately I wasn't able to find a way to throw an error for `@keyframes` usage here without breaking [our own keyframes API](https://compiledcssinjs.com/docs/api-keyframes) at the same time. This is because keyframes({...}) and keyframes\`...\` [are converted](https://github.com/atlassian-labs/compiled/blob/14f2091b30287c21b618ab4b3c26aec1f12f9ab8/packages/babel-plugin/src/utils/css-builders.ts#L573) to `@keyframes ... { ... }` before the contents of `atomicify-rules.ts` is run.

Will mark this suggestion as out-of-scope (maybe something for another PR) if you're ok with this @liamqma

liamqma: Sounds good. Thanks for looking into it.

liamqma (dismissed):

liamqma (approved):

PR #1426, opened 2023-02-20
Add more CSS properties to list of unitless properties

dddlr:
Fixes a bug where `-webkit-line-clamp` (aka `WebkitLineClamp`) has a `px` appended to its value automatically, for example here:

```
const CompactTooltipInnerContainerExperiment = styled2.div<any>({
    WebkitLineClamp: 2
});
```

Previously, `@compiled/css` would erroneously output `-webkit-line-clamp: 2px` instead of the correct `-webkit-line-clamp: 2`.

This PR also adds `base-palette` and `font-size-adjust` to the list of CSS properties that support unitless values. These were the only standardised properties I could find that (1) allowed a unitless number as its value, (2) had any browser support as of writing, and (3) was not already accounted by the existing list in `@compiled/css`.


liamqma (approved):

liamqma (approved):

packages/css/src/utils/css-property.ts
```diff
@@ -23,13 +26,13 @@ const unitless = {
   gridColumnEnd: true,
   gridColumnSpan: true,
   gridColumnStart: true,
-  fontWeight: true,
```
JakeLane, line 26: why has this changed? Is it incorrect?

Just thinking about the implications of bumping a change like this for something that's heavily used

dddlr: It was moved to line 19 to keep the document pseudo-alphabetical 😉

JakeLane: oops cheers for the explanation

JakeLane (approved):

PR #1432, opened 2023-03-07
Only invalidate cache on startup when using js config files

mattcompiles:
As the compiled transformer runs on all JS/TS files, calling `invalidateOnStartup` essentially makes an entire Parcel project unable to be cached. While I'd suggest that calling `invalidateOnStartup` is overkill for a basic config file as it's unlikely to return non-deterministic results, we should definitely not call it when the config file type being used is not a JS type.


JakeLane (approved):

pancaspe87 (approved):

JakeLane:
Thanks for the contribution! I believe @liamqma will be introducing a JS config file into Jira soon for a recent change, we might want to discuss the implication of this for caching.

packages/parcel-optimizer/src/index.ts
```diff
@@ -27,7 +27,10 @@ export default new Optimizer<ParcelOptimizerOpts, unknown>({
     };
 
     if (conf) {
-      config.invalidateOnStartup();
+      if (conf.filePath.endsWith('.js')) {
```
liamqma, line 30: The supported config files are `['.compiledcssrc', '.compiledcssrc.json', 'compiledcss.js','compiledcss.config.js']`

If the config file is `compiledcss.config.js` or `compiledcss.js'`, will it cause any issue?

mattcompiles: We'll basically end up back with the cache being invalidated constantly. We can remove the `invalidateOnStartup` completely, but I'm not sure about handling any internal requires done by the config. e.g. compression map lookups. 

Could we consider making the compression map an internal lookup? That way we could safely add both files to the cache tree.
e.g. 
```js
    classNameCompressionMap: './compiled-class-name-compression-map.json'
```

dddlr (approved):

packages/parcel-optimizer/src/index.ts
liamqma: Yup, you are right. I will add it to our TODO list. Let's assume this is already done, should we remove `invalidateOnStartup` completely in this PR? 

Just to clarify, if we remove `invalidateOnStartup` completely, do people need to `rm -rf .parcel_cache` when config is changed on local development? I think it's reasonable even if it's the case.

Also, checking other plugin, e.g. vue-transformer, it does the same as compiled transform. I guess it suffers the same problem.
https://github.com/parcel-bundler/parcel/blob/cf3a75265789d2eec4f8f6ed0805efd95869e38e/packages/transformers/vue/src/VueTransformer.js#L38

mattcompiles: Actually we can use the `addDevDependency` API for handling the requires. However, I'd still suggest against using a JS file as we can't detect when it will return something different. e.g. `process.env` or `fs.readFileSync`.

mattcompiles: Yeah I guess Vue suffers the same issue. Parcel goes to great lengths to try make the cache as correct and reusable as possible so I'd suggest keeping the `invalidateOnStartup` in but it's upto you. 

The issue with `rm -rf .parcel_cache` is people don't know why and when they need to do it and stop trusting the cache at all. From a Parcel perspective I'd suggest against using JS configs at all. Potentially even warning compiled users against having them when using the Parcel plugins.

PR #1437, opened 2023-03-28
Introduce ac

liamqma:
This PR creates an a new runtime function called `ac`. It substitutes `ax` (at build time) if class name compression is enabled. 

The purpose of `ac` is to solve the below issue:

```
ax([ax('_aaaa_a'), '_aaaa_b']) // it should return 'b'. But it currently returns 'a b'. 👎
```

so `ac` solves this problem

```
ac([ac('_aaaa_a'), '_aaaa_b']) // returns 'b' 👍
```

Although `ac` does what `ax` does plus handling compressed class names, `ax` is more performant than `ac`.
Therefore, we use `ax` by default unless classNameCompressionMap is provided.


packages/react/src/runtime/ac.ts
```diff
@@ -0,0 +1,83 @@
...
+      }
+    } else {
+      // if cls is an instance of AtomicGroups, transfer its values to `atomicGroups`
+      for (const key in cls.values) {
+        atomicGroups[key] = cls.values[key];
+      }
+    }
+  }
+
+  return new AtomicGroups(atomicGroups);
```
liamqma, line 82: `ac` returns an object instead of a primitive. The object holds the knowledge of atomic group and shortened class names so it can be chained.

liamqma:
- [ ] think about edge cases e.g. https://github.com/JedWatson/classnames/issues/153
- [ ] benchmark ax vs ac

packages/react/src/runtime/__perf__/ax.test.ts
```diff
@@ -41,24 +41,4 @@ describe('ax benchmark', () => {
       fastest: ['ax() string'],
     });
   }, 30000);
-
-  it('completes with ax() non-compressed class names as the fastest', async () => {
```
liamqma, line 45: This test has been moved to `packages/react/src/runtime/__perf__/ac.test.ts`.

packages/react/src/runtime/__tests__/ax.test.ts
```diff
@@ -14,51 +14,31 @@ describe('ax', () => {
       ['_aaaabbbb', '_aaaacccc'],
       '_aaaacccc',
     ],
-    [
```
liamqma, line 17: Remove the tests as `ax` can no longer handle compressed class names.

packages/react/src/runtime/ac.ts
```diff
@@ -0,0 +1,131 @@
...
+  const cacheKey = classNames.reduce((accumulator: string, currentValue) => {
+    // if current is undefined, false, or ""
+    if (!currentValue) return accumulator;
+    if (typeof currentValue === 'string') {
+      return accumulator + currentValue;
+    }
+    return accumulator + currentValue.cacheKey;
+  }, '');
+
+  if (cache.has(cacheKey)) return cache.get(cacheKey);
```
liamqma, line 81: Memoize `ac` so if it is called with the same args, it returns immediately. This has two benefits:
1. In case react re-renders with same attributes, memorisation improves performance.
2. Re-use the same object reference to avoid useless React re-renders.

packages/react/src/runtime/ax.ts
```diff
@@ -28,8 +28,10 @@ const ATOMIC_GROUP_LENGTH = 5;
  * @param classes
  */
 export default function ax(classNames: (string | undefined | false)[]): string | undefined {
-  // short circuit if there's no class names.
-  if (classNames.length <= 1 && !classNames[0]) return undefined;
+  if (classNames.length <= 1 && (!classNames[0] || classNames[0].indexOf(' ') === -1)) {
```
liamqma, line 31: Revert `ax` back to before this [PR](https://github.com/atlassian-labs/compiled/pull/1408). So It won't have performance impact if class name compression is disabled.

packages/react/src/runtime/ac.ts
```diff
@@ -0,0 +1,131 @@
...
+export default function ac(
+  classNames: (AtomicGroups | string | undefined | false)[]
+): AtomicGroups | undefined {
+  // short circuit if there's no class names.
+  if (classNames.length <= 1 && !classNames[0]) return undefined;
+
+  // build the cacheKey based on the function argument
+  // e.g. if the argument is ["_aaaabbbb", "_aaaa_a", "some-class-name"],
+  // then the cacheKey is "_aaaabbbb_aaaa_asome-class-name"
+  const cacheKey = classNames.reduce((accumulator: string, currentValue) => {
```
JakeLane, line 72: replace reduce with a conventional for loop for better performance in the javascript engine:

```
for (let i = 0; i < n; i += 1) {}
```

packages/react/src/runtime/ac.ts
```diff
@@ -0,0 +1,131 @@
...
+ * Also, to prevent useless React rerenders
+ */
+const cache = new Map();
+
+/**
+ * `ac` returns an instance of AtomicGroups. The instance holds the knowledge of Atomic Group so we can chain `ac`.
+ * e.g. <div className={ax([ax(['_aaaa_b']), '_aaaa_c'])} />
+ */
+class AtomicGroups {
+  values: Record<string, string> | undefined;
```
JakeLane, line 20: Change to a `Map` type for better performance characteristics, ref https://www.zhenghao.io/posts/object-vs-map#performance-extravaganza

packages/react/src/runtime/ac.ts
```diff
@@ -0,0 +1,131 @@
...
+ * This length includes the underscore,
+ * e.g. `"_1s4A"` would be a valid atomic group hash.
+ */
+const ATOMIC_GROUP_LENGTH = 5;
+
+/**
+ * Memoize the result of ac so if it is called with the same args, it returns immediately.
+ * Also, to prevent useless React rerenders
+ */
+const cache = new Map();
```
JakeLane, line 13: Did you look into the memory footprint and whether it was worth using a LRU cache? https://en.wikipedia.org/wiki/Cache_replacement_policies#Least_recently_used_(LRU)

packages/react/src/runtime/ac.ts
```diff
@@ -0,0 +1,131 @@
...
+        const isCompressed = isAtomic && atomic.charCodeAt(5) === UNDERSCORE_UNICODE;
+
+        const atomicGroupName = isAtomic ? atomic.slice(0, ATOMIC_GROUP_LENGTH) : atomic;
+        atomicGroups[atomicGroupName] = isCompressed
+          ? atomic.slice(ATOMIC_GROUP_LENGTH + 1)
+          : atomic;
+      }
+    } else {
+      // if cls is an instance of AtomicGroups, transfer its values to `atomicGroups`
+      for (const key in cls.values) {
```
JakeLane, line 106: conventional for loop here as well I think

JakeLane (changes requested):
nice work! Just some comments about perf

packages/react/src/runtime/ac.ts
```diff
@@ -0,0 +1,131 @@
...
+ * `ac` returns an instance of AtomicGroups. The instance holds the knowledge of Atomic Group so we can chain `ac`.
+ * e.g. <div className={ax([ax(['_aaaa_b']), '_aaaa_c'])} />
+ */
+class AtomicGroups {
+  values: Record<string, string> | undefined;
+  cacheKey: string;
+  constructor(values: Record<string, string>, cacheKey: string) {
+    // An object stores the relation between Atomic group and actual class name
+    // e.g. { "aaaa": "a" } `aaaa` is the Atomic group and `a` is the actual class name
+    this.values = values;
```
dddlr, line 25: Just to confirm, do we assume that the keys and values of the Record `AtomicGroups.values` never themselves contain underscores? (e.g. "aaaa_bbbb" would never be a valid key in `AtomicGroups.values`)

I'm curious whether it is possible to get two different AtomicGroups with the same cache key

packages/react/src/runtime/__tests__/ac.test.ts
```diff
@@ -0,0 +1,116 @@
...
+      'should not remove any atomic declarations if there are no duplicate groups',
+      ['_aaaabbbb', '_bbbbcccc'],
+      '_aaaabbbb _bbbbcccc',
+    ],
+    [
+      'should not remove any atomic declarations if there are short class name and no duplicate groups',
+      ['_eeee_e', '_aaaabbbb', '_bbbbcccc'],
+      'e _aaaabbbb _bbbbcccc',
+    ],
+    ['should not apply conditional class', [isEnabled && 'foo', 'bar'], 'bar'],
```
dddlr, line 66: Having a bit of trouble understanding this test case - what would be an example where we might use `&&` and `||` operators in the input to `ac()`, and why would `isEnabled && 'foo'` not be part of `ac()`'s output?

packages/react/src/runtime/ac.ts
dddlr: Awesome work!! 🙌

packages/react/src/runtime/__tests__/ac.test.ts
liamqma: example:

```

const MyDiv = styled.div({
  color: ({ isActive }) =>  isActive ? 'red' : undefined
});
```

which gets compiled to 
```
   const _ = "._syaz5scu{color:red}";
    const MyDiv = forwardRef(
      ({ as: C = "div", style: __cmpls, ...__cmplp }, __cmplr) => {
        const { isActive, ...__cmpldp } = __cmplp;
        return (
          <CC>
            <CS>{[_]}</CS>
            <C
              {...__cmpldp}
              style={__cmpls}
              ref={__cmplr}
              className={ac([
                __cmplp.isActive && "_syaz5scu",
                __cmplp.className,
              ])}
            />
          </CC>
        );
      }
    );
```

`ac([__cmplp.isActive && "_syaz5scu",__cmplp.className,])` would be an example where we might use `&&`.

As for `why would isEnabled && 'foo' not be part of ac()'s output`, I don't understand why you mean. Would you mind elaborating it?

packages/react/src/runtime/ac.ts
liamqma: "aaaa" is the atomic group. It's always 4 chars and doesn't not include underscores. You can see how it's generated at [here](https://github.com/atlassian-labs/compiled/blob/master/packages/css/src/plugins/atomicify-rules.ts#L30).

packages/react/src/runtime/ac.ts
liamqma: Initially I thought it is fine because `style-cache` uses [a same pattern](https://github.com/atlassian-labs/compiled/blob/master/packages/react/src/runtime/style-cache.tsx#L12). But you've raised a very good point. Let me look into it futher.

packages/react/src/runtime/ac.ts
liamqma: It's been updated.

packages/react/src/runtime/ac.ts
liamqma: Done.

packages/react/src/runtime/ac.ts
liamqma: > Did you look into the memory footprint and whether it was worth using a LRU cache? 

Thanks for pointing it out. I've tested it on my staging instance. The memory size of cache on issue view is ~50kb. It grows up to 100kb after clicking around and jumping between pages. I am leaning towards **not** implement LRU because
- The cache shouldn't grow indefinitely even if the page stays open for a long time as there are limited number of Compiled components
- LRU increases runtime cost

What do you think? @JakeLane @dddlr

liamqma: > also are there any implications for the bundler when using a side-effect singleton to store the cache? Do we need to initialise it somewhere to prevent multiple cache instances and breaking tree-shaking?

I've implemented https://github.com/atlassian-labs/compiled/pull/1437/files#diff-33bd4912ff7f9f36d0dd9ce38ba892dd1407cb49bb890319400819bd70c12047 to ensure only one runtime exist in the bundle.

packages/react/src/runtime/ac.ts
liamqma: Updated.

packages/tsconfig.options.json
```diff
@@ -16,7 +16,7 @@
     "resolveJsonModule": true,
     "sourceMap": true,
     "strict": true,
-    "target": "es5",
+    "target": "es6",
```
liamqma, line 19: The reason for this change is that TS adds polyfills for `Class` and `Map` if target is 'es5'. The polyfills increases runtime cost and bundle size. 

Looking into other Atlassian owned libraries, [target](https://github.com/atlassian-labs/react-resource-router/blob/master/tsconfig.json#L16) has been set above es5. Therefore, I think it's safe to change target in Compiled as well.

JakeLane: You might want to make another changeset with a minor bump that includes this change. Any packages that have a different distributable should have it otherwise we'll break semver

liamqma: oh, good point!

packages/react/src/runtime/__tests__/ac.test.ts
dddlr: Thank you for the detailed explanation! :)

> As for why would isEnabled && 'foo' not be part of ac()'s output, sorry, I don't follow it. Would you mind elaborating it?

Ah sorry I wasn't very clear - I was wondering why the output would be just `'bar'` and not something like `'foo bar'` (or some other output that would incorporate `'foo'` somehow)

liamqma: oh, because of the line [here](https://github.com/atlassian-labs/compiled/pull/1437/files/fc90687444a03ceb28dec0fc30cd222f7371a39d#diff-53e489e609eaf51087b24f1e67079bc574f2cc03404278dccc22227241b2354fR4) , isEnabled is bound to return false in this test.

packages/react/src/runtime/ac.ts
dddlr: > Thanks for pointing it out. I've tested it on my staging instance. The memory size of cache on issue view is ~50kb. It grows up to 100kb after clicking around and jumping between pages.

Thank you for looking into it and for your insight @liamqma! Just to confirm, would you anticipate that the cache won't grow much beyond ~100 kB even in a (worst-case?) scenario where the user interacts with many different pages over a long span of time?

Also interested to hear what Jake thinks

JakeLane (approved):

dddlr (approved):

PR #1440, opened 2023-04-06
Allow static evaluation of variables inside nested template literals

dddlr:
- [x] Write test cases, including those for edge cases
- [x] Write changeset
- [x] Do a rebase and update the branch
- [x] We don't know whether `result` is guaranteed to be an `UnconditionalCssItem` or not (as opposed to a ConditionalCssItem, LogicalCssItem, or SheetCssItem).
    - It looks like if we have a logical expression or arrow function in a template expression (a.k.a. the "expression" part of \`quasi1 ${expression} quasi2\`), `extractObjectExpression` runs instead of `extractTemplateLiteral`, so we never seem to get `t.isTemplateLiteral(nodeExpression)` and `result.css is not UnconditionalCssItem` true at the same time in `extractTemplateLiteral`... best to double check with someone else though?


packages/babel-plugin/src/utils/css-builders.ts
```diff
@@ -679,9 +679,28 @@ const extractTemplateLiteral = (node: t.TemplateLiteral, meta: Metadata): CSSOut
...
+        context: 'fragment',
+        state: updatedMeta.state,
+        parentPath: updatedMeta.parentPath,
+      };
+
+      const buildCssMeta = t.isTemplateLiteral(nodeExpression)
+        ? nestedTemplateLiteralMeta
+        : updatedMeta;
+
+      const result = buildCss(interpolation, buildCssMeta);
```
dddlr, line 703: Is it safe to assume that the type of `result.css` is `UnconditionalCssItem` here? (This matters because of line 707 below)

In Jira frontend, `t.isTemplateLiteral(nodeExpression)` and `result.css is not UnconditionalCssItem` are never simultaneously true in this part of the function, but I'm not sure whether it's *theoretically* possible for both to be true

dddlr: To answer my own question: yes, because of `extractObjectExpression`:

https://github.com/atlassian-labs/compiled/blob/0448f52bb0fae6cca0e8217d9662d936eeeb6b1a/packages/babel-plugin/src/utils/css-builders.ts#L530

Code such as

```js
const ListItem = styled.div<{ hide: boolean }>({
    display: (props) => (props.hide ? 'none' : 'inline-block')
});
```

will have the `display` line converted to something like

```js
`display: ${(props) => (props.hide ? 'none' : 'inline-block')}`
```

and the `extractTemplateLiteral` function (the function that this PR modifies) will run afterwards

packages/babel-plugin/src/styled/__tests__/call-expression.test.ts
```diff
@@ -538,4 +538,87 @@ describe('styled object call expression', () => {
...
+      const Component = styled.div<{ isActive: boolean }>({
+        boxShadow: (props) =>
+            props.isActive
+                ? \`0 \${size}px \${\`var(--my-variable, \${color})\`}\`
+                : \`var(--my-other-variable, \${color2})\`,
+      });
+    `);
+
+    // We currently don't statically evaluate color, color2, or size here
+    expect(actual).toMatchInlineSnapshot(`
```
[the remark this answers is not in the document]
dddlr: Ah sorry yep, this is one of the cases where this PR doesn't have any effect. I added it to check that nothing was being silently deleted. Should I still keep test cases like these? 🤔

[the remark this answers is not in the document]
liamqma: If I understand it correctly, you want to make sure nothing gets accidentally deleted by `@compiled/babel-plugin`. The thing is `@compiled/babel-plugin` won't even recognise the syntax because the code has TypeScript `<{ isActive: boolean }>`. However,[ transform function](https://github.com/atlassian-labs/compiled/blob/master/packages/babel-plugin/src/test-utils.ts#L15) isn't configured to  parse TypeScript file so `@compiled/babel-plugin` will just bail out.

packages/babel-plugin/src/styled/__tests__/call-expression.test.ts
```diff
@@ -538,4 +538,87 @@ describe('styled object call expression', () => {
 
     expect(actual).toIncludeMultiple(['color:black', 'font-weight:400', 'background-color:white']);
   });
+
+  it('should transform variable in a nested template literal', () => {
+    // this is the output of applying @atlaskit/tokens babel plugin
+    // (as of v1.0.0) to some code similar to this:
+    //     styled.div({
```
liamqma, line 545: Thanks for providing what looks like before applying `@atlaskit/tokens` 👍. That really helps me understand the code.

packages/babel-plugin/src/styled/__tests__/call-expression.test.ts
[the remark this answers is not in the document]
JakeLane: you can configure the transform function to support TS:

```
const actual = transform(
  `
  import '@compiled/react';

  const styles = { color: 'red' } as const;

  <div css={{ ...styles }} />;
`,
  {
    parserBabelPlugins: ['typescript', 'jsx'],
  }
);
```

JakeLane (approved):
Nice work!

packages/babel-plugin/src/styled/__tests__/tagged-template-expression.test.ts
```diff
@@ -819,6 +918,120 @@ describe('styled tagged template expression', () => {
...
+          return (
+            <CC>
+              <CS>{[_]}</CS>
+              <C
+                {...__cmpldp}
+                style={{
+                  ...__cmpls,
+                  "--_hcoy15": ix(
+                    \`\${
+                      __cmplp.isActive
```
liamqma, line 1018: Just wondering if `var(--my-variable, color)` and `var(--my-other-variable, color2)` are supposed to be part of style attribute?

dddlr: Ah good question! To summarise and expand on our Zoom discussion from yesterday, yep that is correct - my understanding is that since we have a ternary expression here, it makes sense to not try to evaluate it to a regular string at build time. This is because we don't really know what the value of `background-color` will be at build time, which in turn is because we might not know whether `isActive` is true or false before runtime.

Slight tangent: I suppose we *could* evaluate just `color` and `color2` at build time so that instead of having `var(--my-variable, ${color})` in the babel plugin output, we would have `var(--my-variable, red)`. (We still keep the ternary expression in the babel plugin output of course.) However from what I can tell, both versions generate the exact same HTML `style="--_hudwc7:var(--my-variable, red);"`.

Thank you for bringing this up - looking at this again, I see there's actually a typo in this test case that I need to fix 😆 `var(--my-variable, color)` should be `var(--my-variable, ${color})`

liamqma (approved):

dddlr:
Close and re-open PR to fix snyk test failures (hopefully)

PR #1444, opened 2023-04-21
Add classNameCompressionMapFilePath as an option to Parcel transformer.

liamqma:
per discussion https://github.com/atlassian-labs/compiled/pull/1432#discussion_r1127289296, add `classNameCompressionMapFilePath` as an option to Parcel transformer.


examples/parcel/.compiledcssrc.json
```diff
@@ -0,0 +1,16 @@
+{
```
dddlr, line 1: Is `.compiledcssrc.json` preferred over `compiledcss.js`?

packages/parcel-transformer/src/types.ts
```diff
@@ -47,4 +47,13 @@ export interface ParcelTransformerOpts extends BabelPluginOpts {
...
   classNameCompressionMap?: { [index: string]: string };
+
+  /**
+   * File path to classNameCompressionMap. It's relative to project root.
+   * Note `classNameCompressionMap` and `classNameCompressionMapFilePath` are mutually exclusive.
+   * If both are provided, classNameCompressionMap takes precedence.
+   *
+   * Default to `undefined`
+   */
+  classNameCompressionMapFilePath?: 'string';
```
dddlr, line 58: Not sure if `'string'` works here (I expected `string` without quotes)

packages/parcel-transformer/src/index.ts
```diff
@@ -42,6 +48,26 @@ export default new Transformer<ParcelTransformerOpts>({
         config.invalidateOnStartup();
       }
 
+      // Use `classNameCompressionMapFilePath` to get classNameCompressionMap
+      // Note `classNameCompressionMap` and `classNameCompressionMapFilePath` are mutually exclusive.
+      // If both are provided, classNameCompressionMap takes precedence.
```
dddlr, line 53: 👍 I like these clarification comments

packages/parcel-transformer/src/types.ts
dddlr: Ah yeah I think the quotes should be removed from `'string'`, otherwise `'string'` would be the only valid string value for `classNameCompressionMapFilePath`

examples/parcel/.compiledcssrc.json
liamqma: Yes!

packages/parcel-transformer/src/types.ts
liamqma: > Not sure if `'string'` (with quotes) works here (I expected `string` without quotes)

Good catch!

JakeLane (approved):

PR #1446, opened 2023-04-24
Bugfix generate compression map correctly

liamqma:
If a compressed class name is in both stylesheet and old compression map, keep it in the new compression map. Let me give an example.

If the stylesheet looks like
```
.a { color: red }
._aaaabbbb { color: green }
```
oldCompressionMap:
```
{ "aaaacccc": "a" }
```
We should keep `{ "aaaacccc": "a" }` in the returned compression map.

### DIFF
--- a/packages/css/src/generate-compression-map.ts
+++ b/packages/css/src/generate-compression-map.ts
@@ -19,15 +19,17 @@ export const generateCompressionMap = (
 ): undefined | { [index: string]: string } => {
   const { oldClassNameCompressionMap, prefix } = opts || {};
 
-  let classNamesToCompress: string[] = [];
+  let atomicClassNames: string[] = [];
+  let nonAtomicClassNames: string[] = [];
   const classNameCompressionMap: { [index: string]: string } = {};
   const reservedClassNames: string[] = [];
 
   const selectorProcessor = selectorParser((selectors) => {
     selectors.walkClasses((node: selectorParser.ClassName | selectorParser.Identifier) => {
-      // Only compress Atomic class names, which has the format of `_aaaabbbb`.
       if (node.value.charCodeAt(0) === UNDERSCORE_UNICODE && node.value.length === 9) {
-        classNamesToCompress.push(node.value.slice(1));
+        atomicClassNames.push(node.value.slice(1));
+      } else {
+        nonAtomicClassNames.push(node.value);
       }
     });
   });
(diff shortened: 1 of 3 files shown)


packages/css/src/generate-compression-map.ts
```diff
@@ -45,23 +47,56 @@ export const generateCompressionMap = (
...
         return false;
       }
       return true;
     });
+
+    if (Object.keys(classNameCompressionMap).length) {
+      // We've found class names which exist in old compression map but not in the stylesheet.
+      // Theoretically this should not happen. We want to warn the developers so they can investigate it. There is likely some race condition in the builds.
+      console.warn(
+        `${JSON.stringify(
```
JakeLane, line 70: maybe I'm misunderstanding the code here, but are you logging the right reference?

JakeLane: also couldn't this happen if someone deletes a style? Doesn't seem like a bug if that happens

liamqma: > also couldn't this happen if someone deletes a style? Doesn't seem like a bug if that happens

The input stylesheet is `compiled-css.css` from Webpacking so it's likely to include compressed class names already. This check is checking whether a style is supposed to be compressed (exists in old compression map) but isn't compressed in the stylesheet. Let me give an example to explain it.

### First pass
#### stylesheet
```
.aaaabbbb {
    color: red
}
.aaaacccc {
    color: green 
}
```
#### old compression map
undefined
#### generated compression map
```
{
  "aaaabbbb": "a",
  "aaaacccc": "c"
}
```
### second pass:
#### stylesheet
Note `.a` is compressed from `.aaaabbbb`
```
.a {
    color: red
}
.aaaacccc {
    color: green 
}
```
#### old compression map
```
{
  "aaaabbbb": "a",
  "aaaacccc": "c"
}
```
We found `.aaaacccc` exists in the old compression map, which means it should become ".c" instead. The code can handle this, but we want to warn developers that something is not right. At this point of time, `classNameCompressionMap` should be empty (means no unexpectedly uncompressed class names). The cause of this (that I can think of) is some sort of race condition in builds. e.g. multiple generateCompressionMap are running in parallel.

I feel my brain capacity reached to a limit when thinking about different cases 🤯. Does it make any sense?

JakeLane: right sounds good, maybe it would be good to comment on the different map variables to describe what they are for code readability. The comments for the errors are good but the context for the variable is missing

JakeLane (approved):

dddlr (approved):

PR #1447, opened 2023-04-27
remove runtime side effects

liamqma:
We added a runtime side-effect to ensure only one copy of `@compiled/react` exists in the bundle at [PR](https://github.com/atlassian-labs/compiled/pull/1437). 

However, the side-effect relies on the assumption that modules are cached once loaded. This assumption is incorrect in many cases. It will throw a runtime error if `@compiled/react/runtime` is loaded twice. 

I think the risk outweighs the benefits of this side-effect. I will remove this side-effect in this PR and explore other ways of ensuring only one `@compiled/react` in the bundle.

### DIFF
--- a/.changeset/thirty-days-tie.md
+++ b/.changeset/thirty-days-tie.md
@@ -1,0 +1,5 @@
+---
+'@compiled/react': patch
+---
+
+Remove `@compiled/react` runtime side-effect to ensure no error if the module is reloaded.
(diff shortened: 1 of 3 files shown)


liamqma:
I can think of the below cases, which will cause `@compiled/react/runtime` to re-load.

- jest automatic mocking
- user explicitly force reload a module 
- Webpack hot module reloading

JakeLane (approved):

PR #1450, opened 2023-05-01
Throw error when expressionToString runs infinite loop

dddlr:
When a variable cannot be statically evaluated (e.g. due to being from an external package that Compiled isn't aware about), we currently get an unhelpful `Error: Maximum call stack size exceeded` error.

This PR replaces this with a more helpful error that we can more easily diagnose


packages/babel-plugin/src/utils/object-property-to-string.ts
```diff
@@ -56,6 +56,12 @@ const expressionToString: ExpressionToString = (expression, meta) => {
   // handles {[key]: 'value'}
   if (t.isIdentifier(expression)) {
     const evaluatedExpression = evaluateExpression(expression, meta);
+    if (evaluatedExpression.value === expression) {
+      throw new Error(
+        `Cannot statically evaluate the value of "${expression.name}"; ` +
+          `are you importing this variable from an external package?`
```
liamqma, line 62: Do you mean `@compiled/babel-plugin` can’t static evaluate a variable imported from an external package?

dddlr: Ah you're right, thank you for mentioning this - I'll remove the second part of the error until we find what causes this recursion

liamqma (approved):

JakeLane (approved):

PR #1484, opened 2023-07-27
Support default parameters in arrow functions

### LINKED ISSUE #1468: Default value in arrow function does not work
## Describe the bug

Compiled supports using arrow functions as the value for a CSS property. This works fine in the majority of cases.

However, when a [default value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) (default function parameter) is provided for one of the arrow function's parameters, Compiled removes the default value during the transform stage (`@compiled/babel-plugin`?). As a result, default values do not work correctly.

## To reproduce

```tsx
const Button = styled.div({
  height: ({ height = 15 }) => `${height}px`,
});

export const Example = () => {
  return <Button>Test</Button>;
};
```

Another example is below. After #1331 is fixed, `margin` and "Hello world1" will work correctly. However, `padding` will still give a wrong value for "Hello world2" due to the use of default value.

```tsx

const Container = styled.div({
  margin: ({ customPadding }) => customPadding,
  padding: ({ customPadding = '10px 20px' }) => customPadding,
});

export const App = () => (
  <>
    <Container customPadding="10px 20px 30px 40px">Hello world1</Container>
    <Container>Hello world2</Container>
  </>
);
```

## Screenshots

Button - we expect `15px`, but instead we get `undefinedpx`:

Screenshot 2023-06-30 at 13 16 25

Hello world example (after #1331 is fixed) - we expect `10px 20px`, but instead we get `var(--c-, )`:

Screenshot 2023-06-30 at 13 15 43

## Workarounds

There exists a workaround: instead of using a default value, one can use a ternary expression. For example:

```tsx
// before
const Container = styled.div({
    padding: ({ customPadding = '10px 20px' }) => customPadding,
});

// after
const Container = styled.div({
    padding: ({ customPadding }) => customPadding ? customPadding : '10px 20px',
});
```

dddlr:
Fixes #1468

### DIFF
--- a/packages/babel-plugin/src/utils/normalize-props-usage.ts
+++ b/packages/babel-plugin/src/utils/normalize-props-usage.ts
@@ -1,4 +1,4 @@
-import type { NodePath } from '@babel/traverse';
+import type { Binding, NodePath } from '@babel/traverse';
 import * as t from '@babel/types';
 
 import { PROPS_IDENTIFIER_NAME } from '../constants';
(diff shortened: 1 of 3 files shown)


liamqma (approved):
Nice work ❤️

packages/babel-plugin/src/utils/normalize-props-usage.ts
```diff
@@ -15,18 +15,129 @@ const buildObjectChain = (path: NodePath<t.Node> | null, currentChain: string[]
...
+): Record<string, t.Expression> => {
+  const node = path.node;
+  const assignments: Record<string, t.Expression> = {};
+
+  const FindAllAssignmentsVisitor = {
+    AssignmentPattern(path: NodePath<t.AssignmentPattern>) {
+      const { node } = path;
+      if (t.isIdentifier(node.left)) {
+        assignments[node.left.name] = node.right;
+      } else {
```
dddlr, line 71: I couldn't think of any valid examples that would cause these two errors (lines 72 and 85) to be thrown 🤔 However, it's nonetheless theoretically possible according to the babel types

JakeLane (dismissed):

JakeLane (approved):

PR #1491, opened 2023-08-07
Forbid imported styles and function parameters in css attribute

dddlr:
Update `no-css-prop-without-css-function` eslint rule to forbid imported values, function parameters/props, and expressions containing undefined variables in the `css` prop.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-css-prop-without-css-function/index.ts
+++ b/packages/eslint-plugin/src/rules/no-css-prop-without-css-function/index.ts
@@ -85,7 +85,7 @@ const handleMemberExpression = (
   }
 };
 
-function fixWrapper(node: CSSValue, context: Context) {
+const fixWrapper = (node: CSSValue, context: Context) => {
   function* fix(fixer: TSESLint.RuleFixer) {
     const compiledImports = findTSCompiledImportDeclarations(context);
     const source = context.getSourceCode();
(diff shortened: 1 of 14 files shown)


JakeLane (dismissed):

packages/eslint-plugin/src/rules/no-css-prop-without-css-function/index.ts
```diff
@@ -33,6 +43,34 @@ const findStyleNodes = (node: Node, references: Scope.Reference[], context: Rule
     // Traverse to the variable value
     if (definition && definition.node.init) {
       findStyleNodes(definition.node.init, references, context);
+    } else {
+      const isImported = reference?.resolved?.defs.find((def) => def.type === 'ImportBinding');
+      const isFunctionParameter = reference?.resolved?.defs.find((def) => def.type === 'Parameter');
```
liamqma, line 48: I think that won't cover
```
import React from 'react';

const CoolComponent = (props) => {
  return <MyComponent css={props.styles} />;
}
```

dddlr: Ah good catch! Let me fix that

dddlr:
- [ ] Check that the addition of eslint errors for this situation doesn't break JFE

```
const CoolComponent = (props) => {
  return <MyComponent css={props.styles} />;
}
```

liamqma (approved):

PR #1492, opened 2023-08-10
Add CSS Map

liamqma:
Add the support for `cssMap`

```
import { css, cssMap } from '@compiled/react';

const styles = cssMap({
  danger: {
      color: 'red',
      backgroundColor: 'red'
  },
  success: {
    color: 'green',
    backgroundColor: 'green'
  }
});

<div css={css(styles[variant])} />;
```


packages/babel-plugin/src/css-prop/__tests__/css-map.test.ts
```diff
@@ -0,0 +1,142 @@
...
+
+        <div css={css(styles[variant])} />;
+      `);
+
+      expect(actual).toInclude(
+        '<div className={ax([variant==="danger"&&"_syaz5scu _bfhk5scu",variant==="success"&&"_syazbf54 _bfhkbf54"])}/>'
+      );
+    });
+
+    it('should evulate css map when variant is statically defined', () => {
```
JakeLane, line 47:
```suggestion
    it('should evaluate css map when variant is statically defined', () => {
```

packages/babel-plugin/src/css-prop/__tests__/css-map.test.ts
```diff
@@ -0,0 +1,142 @@
...
+        backgroundColor: 'green'
+      }
+    });
+  `;
+
+  const defaultErrorMessage = 'Using a CSS Map in this manner is incorrect.';
+  const nestedErrorMessage = 'You cannot access a nested CSS Map';
+
+  describe('valid syntax', () => {
+    it.only('should evulate css map when variant is a runtime variable', () => {
```
JakeLane, line 35:
```suggestion
    it('should evaluate css map when variant is a runtime variable', () => {
```

packages/babel-plugin/src/utils/css-map.ts
```diff
@@ -0,0 +1,192 @@
...
+  if (
+    t.isCallExpression(value) &&
+    t.isIdentifier(value.callee) &&
+    value.callee.name === meta.state.compiledImports?.cssMap &&
+    value.arguments.length > 0 &&
+    t.isObjectExpression(value.arguments[0])
+  ) {
+    // It's CSS Map! We now need to check if the use of the CSS Map is correct.
+    if (t.isCallExpression(expression.object)) {
+      throw buildCodeFrameError(createErrorMessage(), expression, updatedMeta.parentPath);
```
JakeLane, line 96: I think it would be better to be specific on these errors. Throwing the error on `expression.object` and saying why (call expression isn't supported) etc

packages/babel-plugin/src/utils/css-map.ts
```diff
@@ -0,0 +1,192 @@
...
+ */
+export const evaluateCSSMap = (
+  expression: t.Expression,
+  meta: Metadata,
+  evaluateExpression: EvaluateExpression
+): ReturnType<typeof createResultPair> => {
+  const result = getCSSmap(expression, meta, evaluateExpression);
+
+  // It should never happen because `isCSSMap` should have been checked already.
+  if (!result) throw buildCodeFrameError(createErrorMessage(), expression, meta.parentPath);
```
JakeLane, line 161: Good to include an assertion error here

packages/react/src/css-map/index.ts
```diff
@@ -0,0 +1,8 @@
+import type { CSSProps, CssObject } from '../types';
+import { createSetupError } from '../utils/error';
+
+export default function cssMap<T extends string, TProps = unknown>(
```
JakeLane, line 4: You can probably handle this when you make the types stricter but a TSDoc here would be great (like your existing one for `evaluateCSSMap`)

JakeLane (changes requested):

packages/babel-plugin/src/css-prop/__tests__/css-map.test.ts
liamqma: ah, good catch

packages/babel-plugin/src/utils/css-map.ts
```diff
@@ -0,0 +1,192 @@
...
+          !t.isIdentifier(property.key) ||
+          !t.isExpression(property.value)
+        )
+          throw buildCodeFrameError(createErrorMessage(), expression, updatedMeta.parentPath);
+
+        return t.logicalExpression(
+          '&&',
+          t.binaryExpression(
+            '===',
+            t.isStringLiteral(objectProperty)
```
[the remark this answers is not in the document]
dddlr: Or perhaps by extracting this ternary expression into a separate variable:

```
const someVariable = A ? B : (C ? D : E);
// ...
return ...
    t.binaryExpression(
        '===',
        someVariable,
        t.stringLiteral(property.key.name)
    )
    ...
```

packages/babel-plugin/src/utils/css-map.ts
```diff
@@ -0,0 +1,192 @@
...
+export const isCSSMap = (
+  expression: t.Expression,
+  meta: Metadata,
+  evaluateExpression: EvaluateExpression
+): boolean => {
+  return getCSSmap(expression, meta, evaluateExpression) !== undefined;
+};
+
+/**
+ * Transform expression that uses a CSS Map into an array of logical expressions.
```
dddlr, line 132: This documentation is great 👍 good work

dddlr (changes requested):

examples/webpack/src/app.jsx
```diff
@@ -23,5 +24,6 @@ export const App = () => (
     <CustomFileExtensionStyled>Custom File Extension Styled</CustomFileExtensionStyled>
     <div css={customFileExtensionCss}>Custom File Extension CSS</div>
     <Annotated />
+    <CSSMap variant="danger">CSS Map</CSSMap>
```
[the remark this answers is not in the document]
liamqma: I think it's okay not to add this CSS map example to `examples/parcel/`, because this change is not bundler specific.

packages/babel-plugin/src/utils/css-map.ts
```diff
@@ -0,0 +1,192 @@
...
+};
+
+/**
+ * Retrieves the CSS Map related information from a given expression.
+ *
+ * @param expression The expression to be evaluated.
+ * @param meta {Metadata} Useful metadata that can be used during the transformation
+ * @param evaluateExpression {EvaluateExpression} Function that evaluates an expression
+ */
+const getCSSmap = (
```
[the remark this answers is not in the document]
liamqma: Good idea. I'll take `getCssMap`.

packages/babel-plugin/src/utils/css-map.ts
```diff
@@ -0,0 +1,192 @@
...
+    meta,
+    evaluateExpression
+  );
+
+  // Ensure cssMap is used in a correct format.
+  if (
+    t.isCallExpression(value) &&
+    t.isIdentifier(value.callee) &&
+    value.callee.name === meta.state.compiledImports?.cssMap &&
+    value.arguments.length > 0 &&
```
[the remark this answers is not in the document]
liamqma: Good catch. Let me error it out if no arguments.

packages/babel-plugin/src/utils/css-map.ts
liamqma: Good idea.

packages/babel-plugin/src/utils/css-map.ts
liamqma: Thanks

packages/babel-plugin/src/utils/css-map.ts
liamqma: 👍

packages/react/src/css-map/index.ts
liamqma: Good idea.

liamqma: I encounter an issue with Flowgen when trying to make the types stricter. [Flowgen crushes](https://github.com/joarwilk/flowgen/issues/188) when it converts template literal types.
The only thing (super hacky) I can think of is to skip this file during flowgen. And then I manually create cssmap.js.flow. 
I am going to leave as-is, and revisit it later if it's okay with you?

JakeLane (dismissed):

packages/babel-plugin/src/utils/css-map.ts
```diff
@@ -0,0 +1,263 @@
...
+ *   solid: { borderStyle: 'solid' },
+ * });
+ * const Component = ({ borderStyle }) => <div css={css(borderStyleMap[borderStyle])} />
+ * ```
+ * gets transformed into:
+ * ```js
+ * const Component = ({ borderStyle }) => <div css={css([
+ *   borderStyle === 'none' && { borderStyle: 'none' },
+ *   borderStyle === 'solid' && { borderStyle: 'solid'}
+ * ])} />
```
tomgasson, line 216: This looks like it's outputting an intermediate form. It's preferable for us to consolidate things into a 1-pass compilation rather than having multiple passes, because this is lossy and prevents us from understanding the original context (which is useful for optimisations like dropping the `"success"==="danger"` case. What's inhibiting us from doing this in the same pass as the `css`-prop -> `className` transform?

examples/parcel/src/ui/css-map.jsx
```diff
@@ -0,0 +1,12 @@
...
+const styles = cssMap({
+  danger: {
+    color: 'red',
+  },
+  success: {
+    color: 'green',
+  },
+});
+
+export default ({ variant, children }) => <div css={css(styles[variant])}>{children}</div>;
```
tomgasson, line 12: Just walking my way through this...

1. Just using javascript to choose the style
```js
const styles = css.create({
    danger: {
       color: 'red'
    },
    success: {
      color: 'green'
    }
});

const map = {
  danger: styles.danger,
  success: styles.success
}

export default ({variant, children}) => <div css={css(map[variant])} >{children}</div>
```
Can compile to
```js
const styles = {
    danger: {
       color: '<hash_color_red>'
    },
    success: {
      color: '<hash_color_green>'
    }
});

const map = {
  danger: styles.danger, // == {color: <hash_color_red>}
  success: styles.success // == {color: <hash_color_green>}
}

export default ({variant, children}) => <div css={css(map[variant])} >{children}</div>
```

2. Consumer passes in a _style_ itself
```js
export default ({ extendedStyle, children }) => <div css={css(extendedStyle)}>{children}</div>;
```
can compile to
```js
export default ({ extendedStyle, children }) => <div css={css(extendedStyle)}>{children}</div>;
```

You're combining those two cases together... to form
3. 
```js
const styles = css.match({
    danger: {
       color: 'red'
    },
    success: {
      color: 'green'
    }
});
export default ({ variant, children }) => <div css={css(styles[variant])}>{children}</div>;
```
seems like a nice shortcut to me, but I'm unsure if we need it to be a _different and special_ construct vs what we could make possible with 1

tomgasson (commented):
A few thoughts

packages/babel-plugin/src/utils/css-map.ts
```diff
@@ -0,0 +1,263 @@
...
+import type { EvaluateExpression } from './types';
+
+// The messages are exported for testing.
+export enum ErrorMessages {
+  NUMBER_OF_ARGUMENT = 'You must provide cssMap with only one argument.',
+  ARGUMENT_TYPE = 'cssMap can receive object only.',
+  NESTED_VARIANT = 'You cannot access a nested CSS Map.',
+  VARIANT_CALL_EXPRESSION = 'You cannot use a function call to access a CSS Map.',
+  VARIANT_ACCESS = "You cannot access a CSS Map this way. Please use a string literal (e.g. styles['variantName']) or an identifier (e.g. styles[variantName]).",
+  STATIC_VARIANT_OBJECT = 'You must statically define variant object.',
```
dddlr, line 19: Nit: "The variant object must be statically defined" maybe

packages/babel-plugin/src/utils/css-map.ts
```diff
@@ -0,0 +1,263 @@
...
+ * @param expression The expression to be evaluated.
+ * @param meta {Metadata} Useful metadata that can be used during the transformation
+ * @param evaluateExpression {EvaluateExpression} Function that evaluates an expression
+ */
+const getCssMap = (
+  expression: t.Expression,
+  meta: Metadata,
+  evaluateExpression: EvaluateExpression
+):
+  | {
```
dddlr, line 106: Nit: not sure what this extra pipe character does here

packages/babel-plugin/src/utils/css-map.ts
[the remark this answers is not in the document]
dddlr: Thank you ❤️ love the new errors

dddlr (commented):
LGTM, just got some nitpicks

packages/babel-plugin/src/utils/css-map.ts
dddlr: I wonder why the eslint plugin doesn't catch this 🤔

examples/parcel/src/ui/css-map.jsx
liamqma: Compiled essentially transforms one CSS rule into three pieces.
1. classNames
2. sheets
3. CSS variables (if there are variables cannot be resolved statically). For simplicity, let's ignore it.

We can conditionally apply classNames to DOM element, but sheets cannot. Hence a special construct help identifying variants (will be useful for lints too), and piece sheets together. 

```
const styles = cssMap({
    danger: {
       color: 'red'
    },
    success: {
      color: 'green'
    }
});
export default ({ variant }) => <div css={css(styles[variant])} />;
``` 

gets compiled to

```
<CS>".a {color: red}", ".b {color: green}"</CS> // sheets need to include all possible CSS outcome
...
```

I think the approach you mentioned from 1 is more leaning on local tranforms (which is roadmapped 🙂). If we have local transforms:
```
const foo = css({ 
    color: 'red' 
})

const bar = cssMap({
    danger: {
       color: 'red'
    },
    success: {
      color: 'green'
    }
});
```
gets compiled to
```
const foo = {
  className: ".a",
  sheet: ".a{color:red}",
};

const bar = {
    danger: {
         className: ".a",
         sheet: ".a{color:red}",
    },
    success: {
         className: ".b",
         sheet: ".b{color:green}",
    }
}

```
We can easily use javascript to choose the style
```
<div css={css([foo, bar[variant]])} />
```
gets compiled to
```
<CS>{[foo, bar]}</CS>
<div className={ax(foo.className, bar[variant].className])} />
```
If I misunderstood the comment, let's have a quick zoom call to clarify it (maybe)?

examples/parcel/src/ui/css-map.jsx
```diff
@@ -0,0 +1,12 @@
+import { css, cssMap } from '@compiled/react';
+
+const styles = cssMap({
```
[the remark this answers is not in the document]
liamqma: Having `cssMap` simplifies types. What's the reason that `css.create` is preferred?

packages/babel-plugin/src/css-prop/__tests__/css-map.test.ts
```diff
@@ -0,0 +1,199 @@
...
+    it('should evaluate css map when variant is statically defined', () => {
+      const actual = transform(`
+       ${styles}
+        
+        <div css={css(styles.success)} />;
+        <div css={css(styles['danger'])} />;
+      `);
+
+      expect(actual).toInclude(
+        '<div className={ax(["success"==="danger"&&"_syaz5scu _bfhk5scu","success"==="success"&&"_syazbf54 _bfhkbf54"])}/>'
```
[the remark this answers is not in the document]
liamqma: Yea, you are right.
But I think the selected variant is runtime variable at most of cases. Statically access to a variant is "by-product". Users probably should use

```
const a = {
   color: red
}
<div css={css(a)} />
```  

instead of 

```
const styles = {
   a: {
       color: red
   }
}
<div css={css(styles.a)} />
```  
Also, bundler drops the dead code `('apple' === 'orange') && '.aaaabbbb'`

packages/babel-plugin/src/utils/css-map.ts
liamqma: um...good point. Let me look into it more.

liamqma: > consolidate things into a 1-pass compilation rather than having multiple passes

It's been architectural thing. 
- Compiled firstly evaluates user input, and compiles it to expression. e.g. `foo ? 'color: red': 'color: green'`
- At later stage, then compiles it to classNames and sheets

I think we can change this two-pass compilation into 1-pass by using local transform, as mentioned from another comment. I will make sure that I will spike this 1-pass compilation before PR is merged.

packages/babel-plugin/src/css-prop/__tests__/css-map.test.ts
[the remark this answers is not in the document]
liamqma: I think it makes sense to be in `packages/babel-plugin/src/css-prop`, because they aren't unit tests against the functions of `css-map.ts`. These tests rely on many things from `css-prop`.

packages/babel-plugin/src/utils/css-map.ts
liamqma: believe it or not. The extra pipe is enforced by prettier. 🤷‍♂️

dddlr: TIL

packages/babel-plugin/src/css-prop/__tests__/css-map.test.ts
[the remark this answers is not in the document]
dddlr: Ah no worries 👍

dddlr (approved):

liamqma:
An alternative implementation has been adopted.

PR #1494, opened 2023-08-23
eslint plugin: Allow function parameters/imported values in edge case

dddlr:
Follow up PR to #1491 

There is a case where we should allow function parameters and imported values in the `css` attribute: when the variable is the left hand side of an AND (`&&`) expression. Consider an example like this:

```jsx

const styles = css(...);
const otherStyles = css(...);
const Component = ({ myBoolean }) => <MyComponent css={[myBoolean && styles, otherStyles]} />;

<Component myBoolean />
```

Compiled handles this perfectly fine because `styles` and `otherStyles` are statically evaluable, but #1491 erroneously marks `myBoolean` as problematic. This PR fixes this.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-css-prop-without-css-function/index.ts
+++ b/packages/eslint-plugin/src/rules/no-css-prop-without-css-function/index.ts
@@ -40,6 +40,18 @@ const handleIdentifier = (node: TSESTree.Identifier, references: Reference[], co
     const isImported = reference?.resolved?.defs.find((def) => def.type === 'ImportBinding');
     const isFunctionParameter = reference?.resolved?.defs.find((def) => def.type === 'Parameter');
 
+    // We assume that anything that is not on the right side of an AND logical expression
+    // (e.g. the A in css={A && B}) is only used as a boolean (or de-facto booleans) for
+    // short-circuiting, and do not form part of the final styling for our component.
+    const isNotRightSideOfAndExpression =
+      node.parent?.type === 'LogicalExpression' &&
+      node.parent.operator === '&&' &&
+      node !== node.parent.right;
+
+    if (isNotRightSideOfAndExpression) {
+      return;
+    }
+
     const jsxElement = traverseUpToJSXOpeningElement(node);
 
     // css property on DOM elements are always fine, e.g.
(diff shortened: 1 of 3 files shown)


dddlr:
- [ ] Run this eslint rule in JFE and make sure it doesn't break things

packages/eslint-plugin/src/rules/no-css-prop-without-css-function/index.ts
```diff
@@ -40,6 +40,18 @@ const handleIdentifier = (node: TSESTree.Identifier, references: Reference[], co
     const isImported = reference?.resolved?.defs.find((def) => def.type === 'ImportBinding');
     const isFunctionParameter = reference?.resolved?.defs.find((def) => def.type === 'Parameter');
 
+    // We assume that anything that is on the left side of an AND logical expression
+    // (e.g. the A in css={A && B}) is only used as a boolean (or de-facto booleans) for
+    // short-circuiting, and do not form part of the final styling for our component.
+    const isLeftSideOfAndExpression =
```
liamqma, line 46: These checks don't cover the below cases.
```
foo || styles
foo ?? styles
foo() && styles
props.foo && styles
```

I suggest we simply remove line https://github.com/atlassian-labs/compiled/pull/1494/files#diff-635cc557082fc781a06aa84d129246f01c0e73e12470127ab1f6525217ec16acR144. Can you think of situations where we need it?

dddlr: Ah excluding `css={foo || styles}` and `css={foo ?? styles}` is deliberate; this is because in these cases, `foo` is returned when it is truthy. To me, this would seem to be equivalent to `css={foo}` and so the same `importedInvalidCssUsage / functionParameterInvalidCssUsage / otherInvalidCssUsage` checks should apply - what do you think?

Thank you for mentioning `foo() && styles` and `props.foo && styles`, I had missed that

> I suggest we simply remove line https://github.com/atlassian-labs/compiled/pull/1494/files#diff-635cc557082fc781a06aa84d129246f01c0e73e12470127ab1f6525217ec16acR144. Can you think of situations where we need it?

 Good idea, let me think about this

dddlr: Never mind, you're right - this does work in Compiled

```
const MyComponent = styled.div({});
const MyOtherComponent = ({ someParameter, children }) => <MyComponent css={someParameter || styles}>{children}</MyComponent>;
```

Very fascinating 😮 I'll update the PR to support everything you've mentioned

dddlr: Made a bug report to improve the error message if `someParameter` in `css={someParameter || styles}` is not wrapped in `css()`

https://github.com/atlassian-labs/compiled/issues/1495

liamqma: > (as long as someParameter is wrapped in css() - otherwise there is a cryptic runtime error 😅)
um. It shouldn't work. Yea, improving the error message will be good. 👍

liamqma (approved):

JakeLane (approved):

PR #1496, opened 2023-08-23
CSS Map alternative compilation approach

liamqma:
This PR presents a different compilation approach compared to https://github.com/atlassian-labs/compiled/pull/1492.

It transforms
```
const styles = cssMap({
  danger: {
      color: 'red',
      backgroundColor: 'red'
  },
  success: {
    color: 'green',
    backgroundColor: 'green'
  }
});
<div css={css([styles[variant]) />
```
to
```
const styles = {
  danger: "_syaz5scu _bfhk5scu",
  success: "_syazbf54 _bfhkbf54",
};
<div className={ax([styles[variant]) />
```

The benefits are:
- dramatically simplify evaluation as we no longer evaluate `styles[variant]`. 
- paves the way for transforming `css` the same way as `cssMap`. 
In the future, we can transform
```
const foo = css({ 
    color: 'red' 
})

const bar = cssMap({
    danger: {
       color: 'red'
    },
    success: {
      color: 'green'
    }
});

<div css={css([foo, bar[variant]])} />
```
to
```
const foo = ".a";

const bar = {
    danger:".a",
    success: ".b",
}
<div className={ax([foo, bar[variant]])} />
```


packages/babel-plugin/src/css-prop/__tests__/css-map.test.ts
```diff
@@ -0,0 +1,85 @@
...
+        };
+        <CC>
+          <CS>{[_, _2, _3, _4, _5]}</CS>
+          {
+            <div
+              className={ax([
+                foo && styles["danger"],
+                props.foo && styles["danger"],
+                styles.success,
+                styles["danger"],
```
JakeLane, line 71: Is it worth statically embedding this? We can reduce the amount of runtime code by directly injecting the classnames we know here

packages/babel-plugin/src/types.ts
```diff
@@ -141,6 +142,11 @@ export interface State extends PluginPass {
    * Files that have been included in this pass.
    */
   includedFiles: string[];
+
+  /**
+   * Holds a record of currently evulated CSS Map and its sheets in the module.
```
JakeLane, line 147:
```suggestion
   * Holds a record of currently evaluated CSS Map and its sheets in the module.
```

packages/babel-plugin/src/css-prop/__tests__/css-map.test.ts
liamqma: Good idea 👍

JakeLane:
Can you please add some unit tests for correct usage of the API and some VR tests?

packages/babel-plugin/src/css-map/__tests__/index.test.ts
JakeLane, line 1: What happens if the key is not in the css map?

liamqma: then class name will be undefined and stripped out by `ax`. The idea is to use Types to prevent this.

dddlr: makes sense, I assume that should help reduce the impact on build time

packages/babel-plugin/src/css-prop/__tests__/css-map.test.ts
liamqma: I want to minimise statical evaluation. I am going to leave this job to the bundler.

I just tested a JFE Webpack prod build. It inlines the variable well. 
```
const foo = { bar: '.aaaabbbb' }
<div className={foo.bar} />
```
to
```
<div className={'.aaaabbbb'} />
```

liamqma:
> Can you please add some unit tests for correct usage of the API and some VR tests?

Sure!

dddlr (approved):

JakeLane (approved):
Nice work!

PR #1500, opened 2023-08-31
ESLint rule for CSS map

dddlr:
Refactoring
- [x] extract as many of the functions I use as possible into the util folder
- [x] change functions in `index.ts` into an actual class, since there's a lot of shared state between the functions

Implementation
- [x] detect `import { X as Y } from Z`
- [x] move `packages/eslint-plugin/src/utils/create-no-exported-rule/validate-definition.ts` to a place less tied to `create-no-exported-rule`

Documentation
- [x] add comprehensive incorrect code examples to README


dddlr:
Rule implementation is ready for review, just needs some additional examples in the README

packages/eslint-plugin/src/rules/no-invalid-css-map/__tests__/rule.test.ts
```diff
@@ -0,0 +1,639 @@
...
+            backgroundColor: 'red'
+          },
+          success: {
+            color: 'green',
+            backgroundColor: 'green'
+          }
+        });
+      `,
+    },
+    {
```
liamqma, line 26: Can you add a test case when "ImportDefaultSpecifier" is allowFunctionCalls

```
import something from 'something';

const styles = cssMap({
  danger: {
      color: something('red', 'blue'),
  },
});
```

packages/eslint-plugin/src/utils/is-compiled-import.ts
```diff
@@ -0,0 +1,32 @@
+import type { Rule, Scope } from 'eslint';
```
liamqma, line 1: yea, there is duplicated code between isCss and isKeyFrame. It's good that you refactored them. 👍

packages/eslint-plugin/src/rules/no-invalid-css-map/__tests__/rule.test.ts
```diff
@@ -0,0 +1,639 @@
...
+      errors: [
+        {
+          messageId: 'noNonStaticallyEvaluable',
+        },
+      ],
+      code: outdent`
+        import { firstFunction } from 'example';
+        import React from 'react';
+        import { cssMap } from '@compiled/react';
+        import { bap } from '@atlaskit/token';
```
liamqma, line 456: It's possible that `bap` is evaluable even if it's imported from another package/file.
It's programatically supported by Compiled, although we advocate colocating styles with their consumers,
You can find examples like the below in design system components.
```
import { N200, N300 } from '@atlaskit/theme/colors';

const colorMap = {
  light: {
    textColor: token('color.text.subtlest', N200),
    separatorColor: token('color.text.subtlest', N200),
  },
  dark: {
    textColor: token('color.text.subtlest', N300),
    separatorColor: token('color.text.subtlest', N300),
  },
};
```
, where N200 and N300 are from another package. We will likely continue to support that to ease the migration from emotion to Compiled.

dddlr: Ah that's a good point. Would it be ok to allow all imported variables but continue to forbid imported functions that aren't listed in `allowedFunctionCalls`?

liamqma: yup, that sounds good. 👍

liamqma (dismissed):

dddlr:
did a rebase

JakeLane (approved):

liamqma (approved):

PR #1502, opened 2023-09-01
tighten up cssMap type

liamqma:
Further tighten up `cssMap` type to prevent some invalid usages.

Examples
No Arrays
```
const styles = cssMap({
   foo: [{ color: 'red' }]
})
```

No tagged template
```
const styles = cssMap({
   foo: `
     color: red
   `
})
```

No arrow function
```
const styles = cssMap({
   foo: {
      color: (props) => props.a ? 'red' : 'green'
   }
})
```

Only string and number are valid CSS values. No boolean/undefined.
```
const styles = cssMap({
   foo: {
      color: possibleUndefined && 'red'
   }
})
```


packages/react/src/css-map/index.ts
```diff
@@ -1,6 +1,44 @@
...
+ * const style: CssObject = {
+ *  "@media screen and (min-width: 480px)": {
+ *    ":hover": {
+ *      color: 'red'
+ *   }
+ *  }
+ * }
+ * ```
+ */
+type CssObject = Readonly<
```
dddlr, line 33: I wonder if there's a way to ensure that the user can't nest objects indefinitely, e.g. stuff like

```tsx
const styles = cssMap({
  danger: {
      veryDanger: {
        color: 'red',
      }
  },
})
```

without affecting valid use cases such as `@media` queries

packages/react/src/css-map/index.ts
```diff
@@ -18,10 +56,9 @@ import { createSetupError } from '../utils/error';
  * <Component borderStyle="solid" />
  * ```
  */
-type returnType<T extends string, P> = Record<T, CSSProps<P>>;
 
-export default function cssMap<T extends string, TProps = unknown>(
-  _styles: Record<T, CssObject<TProps> | CssObject<TProps>[]>
-): Readonly<returnType<T, TProps>> {
+export default function cssMap<T extends string>(
+  _styles: Record<T, CssObject>
```
dddlr, line 61: thank you ❤️ sorry for not catching this in the original PR

packages/react/src/css-map/index.ts
dddlr: I'm not at my laptop so I can't test this out, but it does seem like there's nothing stopping the user from doing stuff like

```tsx
const styles = cssMap({
  danger: {
      veryDanger: {
        lotsofDanger: {
          evenmoreDanger: {
            spectacularDanger: {
              color: 'red',
            }
          }
        }
      }
  },
})
```

liamqma: yea, we've to allow it unless we impose stricter types.
The problem is that we don't know if the _object key_ is a selector.
In the below example, we don't know if `header` is a variant name or a selector.
```
const styles = cssMap({
  success: {
      header: { // Is it <header /> or a variant name? 🤷‍♂️
         div: {
             color: 'red'
         }
      }
  },
})
```
I thought about disallowing targeting selectors completely, but I think disallowing it is too specific to Atlassian CSS standards.

dddlr: I see, thank you for clarifying 👍

dddlr (approved):

JakeLane (approved):

packages/react/src/css-map/index.ts
JakeLane: I think we can do better here, it may be worth a play with the structure

I've made a quick spike based on vanilla extract types https://github.com/atlassian-labs/compiled/pull/new/css-map-type-spike

In https://vanilla-extract.style/documentation/styling/#media-queries, they specify some bounded keys that allow the types to be much stricter

dddlr:
Todo:

- [ ] Rebase
- [ ] Double check type checking that nested selectors still work, as long as they're all targeting the same element
- [ ] Unit tests - thoroughly test both StringLiteral and Identifier
- [ ] VR tests (potentially)
- [ ] Investigate why having an empty object as a variant value gives an unhelpful error
- [ ] Replace `throw new Error` with `buildCodeFrameError`

packages/react/src/css-map/pseudos.ts
```diff
@@ -0,0 +1,143 @@
...
+  | '&::first-letter'
+  | '&::first-line'
+  | '&::grammar-error'
+  | '&::marker'
+  | '&::placeholder'
+  | '&::selection'
+  | '&::spelling-error'
+  | '&::target-text'
+  | '&::view-transition'
+  | '&:active'
```
itsdouges, line 88: Would be great to reduce the total amount of declarations. So instead of allowing both ::active and :active we settle on one.

dddlr: Ah good catch, didn't notice the duplicated pseudo-elements (:before, :after, :first-letter, :first-line)

liamqma:
Looks good. I wanted to give approval but notice I cannot approve my own PR 😬.

dddlr (approved):

dddlr:
all good @liamqma i've approved it now (despite most of the changes being from myself 😆 )

dddlr (approved):

PR #1503, opened 2023-09-01
Extract styles to file in babel plugin

JakeLane:
Simple implementation to allow us to emit css files in a babel environment.

This removes the need for the Compiled runtime to be used in an env where we're not running a bundler. This could either be consumers of a library that aren't running Compiled, or consumers that are running compiled, but don't need to re-build from source.


packages/babel-plugin-strip-runtime/src/index.ts
```diff
@@ -47,6 +51,44 @@ export default declare<PluginPass>((api) => {
...
+
+          if (this.opts.extractStylesToDirectory && this.styleRules.length > 0) {
+            // Build filename of the css file
+            const currentFilename = basename(filename);
+            const cssFilename = `${currentFilename.substring(
+              0,
+              currentFilename.lastIndexOf('.')
+            )}.compiled.css`;
+
+            if (!file.opts.generatorOpts?.sourceFileName) {
```
liamqma, line 63: `souceFileName` seems returning the baseName. Are you meant to use `filenameRelative`?

JakeLane: I use `sourceFileName` as that specifies the source file path based on where we are

For example, we're transforming `src/my-file.tsx` which will go to `dist/my-file.tsx`. `sourceFileName` should be `../src/my-file.tsx`.

Does that all make sense? I think I can leave some comments to make that clearer, as this was the bulk of the complexity (juggling the output file path)

JakeLane: Ah I see `filenameRelative` is in the implementation but not the types, I'll see if I can use that 💯

liamqma:
Looks good! I will make a note for myself. When we re-start classname compression, I will need to make some update to accomodate third-party compiled components.

packages/babel-plugin-strip-runtime/src/index.ts
JakeLane: Looking at https://babeljs.io/docs/options#filenamerelative, it seems `sourceFilename` is the variable we get from that configuration option

packages/babel-plugin-strip-runtime/src/index.ts
```diff
@@ -47,6 +51,44 @@ export default declare<PluginPass>((api) => {
...
+              throw new Error(`Source filename was not defined`);
+            }
+            const sourceFileName = `${file.opts.generatorOpts.sourceFileName}`;
```
dddlr, line 66: is there a reason why `file.opts.generatorOpts.sourceFileName` is wrapped in a template string?

JakeLane: good catch! That was something leftover from debugging I forgot to simplify

JakeLane: - [ ] Remove template string

packages/babel-plugin-strip-runtime/src/__tests__/strip-runtime-source-code.test.ts
```diff
@@ -21,21 +23,32 @@ const transformSync = (
     compiledRequireExclude?: boolean;
     run: 'both' | 'bake' | 'extract';
     runtime: 'automatic' | 'classic';
+    extractStylesToDirectory?: { source: string; dest: string };
```
dddlr, line 26: Does Babel enforce that `source` and `dest` are defined, or is it just (presumably) typescript?

dddlr:
Great work, overall looks good

JakeLane:
Released in #1513

PR #1512, opened 2023-09-13
Implement xcss

liamqma:
```
import { xcss } from '@compiled/react';

const styles = xcss({
      color: 'green',
});

<Component xcss={styles} />;
```
gets transformed to
```
const _ = "._syazbf54{color:green}";
const styles = "_syazbf54";
const Component = () => (
  <CC>
    <CS>{[_]}</CS>
    {<div xcss={styles} />}
  </CC>
);
```


liamqma:
TODO:

- [ ] Types
- [ ] Codegen

itsdouges:
Since this has conflicts introduced from #1533 and #1541 I'll close this. It'll keep existing if we want to pull any inspiration from it.

PR #1513, opened 2023-09-13
Feature/handle extracted compiled files

liamqma:


liamqma:
- [ ] Test parcel-transformer in jira-frontend
- [ ] Test webpack-loader in jira-frontend

packages/babel-plugin-strip-runtime/src/index.ts
```diff
@@ -47,6 +51,44 @@ export default declare<PluginPass>((api) => {
...
+              this.cwd,
+              this.opts.extractStylesToDirectory.dest,
+              dirname(relativePath),
+              cssFilename
+            );
+            mkdirSync(dirname(cssFilePath), { recursive: true });
+            writeFileSync(cssFilePath, sort(this.styleRules.sort().join('\n')));
+
+            // Add css import to file
+            path.unshiftContainer('body', template.ast(`import "./${cssFilename}";`));
```
dddlr, line 90: Can we guarantee that cssFilename doesn't contain any literal double quotes? (or any other character that is valid in a filename, but invalid in an import statement)

JakeLane: I think we'd be safe to just escape the double quotes `.replaceAll('"', '\\"')`

packages/babel-plugin-strip-runtime/src/types.ts
```diff
@@ -10,6 +10,11 @@ export interface PluginOptions {
    * When set will prevent additional require (one import per rule) in the bundle.
    */
   compiledRequireExclude?: boolean;
+
+  /**
+   * When set, extract styles to an external CSS file
+   */
+  extractStylesToDirectory?: { source: string; dest: string };
```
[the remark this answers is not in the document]
JakeLane: Just TypeScript, which is useful when calling the plugin programatically

JakeLane (changes requested):

packages/webpack-loader/src/__tests__/extract-plugin.test.ts
```diff
@@ -182,4 +182,22 @@ describe('CompiledExtractPlugin', () => {
       "
     `);
   }, 10000);
+
+  it('should hanlde extracted styles', async () => {
```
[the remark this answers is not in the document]
liamqma: ah, good catch

fixtures/babel-component-extracted/src/index.jsx
```diff
@@ -0,0 +1,17 @@
+import { styled, css } from '@compiled/react';
+
+const Button = styled.button`
```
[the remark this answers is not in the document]
liamqma: yup, you are right.

packages/babel-plugin-strip-runtime/src/index.ts
dddlr: It's just double quotes and backslashes that we need to be worried about, right?

packages/babel-plugin-strip-runtime/src/index.ts
```diff
@@ -47,6 +51,44 @@ export default declare<PluginPass>((api) => {
               // If we used ESM it would blow up with CJS source, unfortunately.
             });
           }
+
+          if (this.opts.extractStylesToDirectory && this.styleRules.length > 0) {
+            // Build filename of the css file
+            const currentFilename = basename(filename);
+            const cssFilename = `${currentFilename.substring(
```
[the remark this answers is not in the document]
liamqma: Good catch. Has updated it.

packages/babel-plugin-strip-runtime/src/index.ts
liamqma: Has escaped the invalid.

packages/babel-plugin-strip-runtime/src/index.ts
```diff
@@ -53,17 +53,16 @@ export default declare<PluginPass>((api) => {
           }
 
           if (this.opts.extractStylesToDirectory && this.styleRules.length > 0) {
-            // Build filename of the css file
-            const currentFilename = basename(filename);
-            const cssFilename = `${currentFilename.substring(
-              0,
-              currentFilename.lastIndexOf('.')
-            )}.compiled.css`;
```
dddlr: Ah is there protection against accidental overwriting of an existing `.compiled.css` file?

For example, if we have `app+v1.ts` and `app=v1.ts`, would the generated `.compiled.css` file for the second file overwrite the generated `.compiled.css` file for the first file?

packages/babel-plugin-strip-runtime/src/index.ts
```diff
@@ -53,17 +53,16 @@ export default declare<PluginPass>((api) => {
...
+            // Build and sanitize filename of the css file
+            const cssFilename = `${parse(filename).name}.compiled.css`.replace(
```
dddlr, line 57: Ah is there protection against accidental overwriting of an existing .compiled.css file?

For example, if we have app+v1.ts and app=v1.ts, would the generated .compiled.css file for the second file overwrite the generated .compiled.css file for the first file?

liamqma: I haven't thought about this. Good point. Let me write some code to protect this.

dddlr: No prob

Up to you how to handle it, but even something simple like erroring out should be ok i think

packages/babel-plugin-strip-runtime/src/index.ts
```diff
@@ -86,7 +87,7 @@ export default declare<PluginPass>((api) => {
             writeFileSync(cssFilePath, sort(this.styleRules.sort().join('\n')));
 
             // Add css import to file
-            path.unshiftContainer('body', template.ast(`import "./${cssFilename}";`));
+            path.unshiftContainer('body', template.ast(`import "./${addslashes(cssFilename)}";`));
```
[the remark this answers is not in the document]
liamqma: oh yea, that's a good idea.

JakeLane (approved):

dddlr (approved):

PR #1521, opened 2023-10-09
Add example type test

itsdouges:
This pull request installs the `expect-type` package and writes a simple test. It's a portable library with no boiler plate to have it configured.

It does not throw errors at runtime (meaning: all tests with only type assertions will always pass). It leans on the TypeScript compiler to raise errors at compile time. That means we need to make sure we're type checking tests else the tests won't actually do anything. 

We do throw the build if there are type errors in tests currently, so happy days!

```
➜  compiled git:(expect-type-assertions) yarn build
yarn run v1.22.19
$ yarn build:babel-fixture && yarn build:browser && yarn build:cjs && yarn build:esm && yarn flow-types build && yarn postbuild
$ yarn workspace @compiled/babel-component-fixture build && yarn workspace @compiled/babel-component-extracted-fixture build
$ TEST_PKG_VERSION='0.0.0' babel ./src --out-dir=./dist
Successfully compiled 1 file with Babel (1405ms).
$ TEST_PKG_VERSION='0.0.0' babel ./src --out-dir=./dist
Successfully compiled 1 file with Babel (1387ms).
$ ttsc --build packages/tsconfig.browser.json
packages/react/src/css/__tests__/types.test.ts:10:39 - error TS2344: Type 'string' does not satisfy the constraint '{ [x: number]: "Expected: string, Actual: never"; [iterator]: "Expected: function, Actual: never"; replace: "Expected: function, Actual: never"; fixed: "Expected: function, Actual: never"; ... 853 more ...; valueOf: "Expected: function, Actual: never"; }'.

10     expectTypeOf(style).toMatchTypeOf<string>();
                                         ~~~~~~


Found 1 error.

error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

### DIFF
--- a/packages/react/src/css/__tests__/types.test.ts
+++ b/packages/react/src/css/__tests__/types.test.ts
@@ -1,0 +1,12 @@
+/** @jsxImportSource @compiled/react */
+import { css } from '@compiled/react';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { expectTypeOf } from 'expect-type';
+
+describe('css func tests', () => {
+  it('should type violate when given invalid types', () => {
+    const style = css({ color: 'red' });
+
+    expectTypeOf(style).not.toMatchTypeOf<string>();
+  });
+});
--- a/packages/react/src/css/__tests__/types.test.ts
+++ b/packages/react/src/css/__tests__/types.test.ts
@@ -1,6 +1,5 @@
 /** @jsxImportSource @compiled/react */
 import { css } from '@compiled/react';
-// eslint-disable-next-line import/no-extraneous-dependencies
 import { expectTypeOf } from 'expect-type';
 
 describe('css func tests', () => {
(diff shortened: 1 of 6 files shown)


DarkPurple141 (approved):
LGTM will integrate into my branch

packages/react/src/css/__tests__/types.test.ts
```diff
@@ -0,0 +1,12 @@
+/** @jsxImportSource @compiled/react */
+import { css } from '@compiled/react';
+// eslint-disable-next-line import/no-extraneous-dependencies
```
itsdouges, line 3: We aren't handling this lint rule very well currently. Looking at the config options we have we can either:

- set an override just for tests to just disable the rule
- have every package declare test deps as dev dependencies
- continue supressing

Any thoughts?

liamqma: #2 sounds like the right thing to me. Is there any disadvantage of doing this apart from receptively specifying this dev dep in every needed package?

itsdouges: No disadvantage. I'll pull the pin and update this PR for this new dep.

liamqma (approved):

PR #1522, opened 2023-10-09
[WIP] Introduce scaffold of stricter types for css

DarkPurple141:
Applied here via `xcss` to make it clear what the proposed direction is. Key outputs are the introductin of:

`XCSSProp` type which is designed to restrict the unbound usage of styles generated via `xcss (or css)`.

You can see example usages of the type restriction in the test file.


packages/react/src/xcss/xcss-prop.ts
```diff
@@ -0,0 +1,50 @@
...
+ * // Valid usage
+ * <Component xcss={{ color: 'red', '&:hover': { color: 'blue' } }} />
+ *
+ * // Invalid usage
+ * <Component xcss={{ backgroundColor: 'red' }} />
+ * ```
+ *
+ */
+export type XCSSProp<AllowedProperties extends keyof CSSProperties, TPseudos extends CSSPseudos> =
+  | (XCSSItem<AllowedProperties> & XCSSPseudos<AllowedProperties, TPseudos>)
```
liamqma, line 47: Just to confirm that there won't be need for atrules overrides?

packages/react/src/cx/index.ts
```diff
@@ -0,0 +1,41 @@
...
+ * ```tsx
+ * const style = xcss({
+ *   color: 'red',
+ *   '&:active': {
+ *     color: 'red',
+ *   }
+ * })
+ * ```
+ */
+const cx = <Collection extends [...XCSSProp<any, any>[]]>(
```
liamqma, line 34: We use a function `cx` instead of a plain array to join overrides. I guess it's due to some type constraints?

itsdouges: Yes exactly. By default TypeScript will treat it as a union array, which isn't what we want. We want it to be treated as a const tuple (instead of an array).

Now we could leave it as a tuple, but that means `xcss` prop would allow arrays to be defined, which introduces risk into the system that we'd need to block by lint rules. Instead, we make `xcss` prop only ever take an object, and the `cx` function reduces style args down to a union of object styles.

See: [TypeScript playground](https://www.typescriptlang.org/play?#code/MYewdgzgLgBNCeAbApgFWQD1gXhgbxlERACcAuGAchOQBNKYBfAbgChRJYEUAhAcxi4CAIwCGwANZ8SIAK5haAYRDFyVYYlnIGLVu3DQ4UJMgiCYAbW5pMUADRGT-ALpsY7jzAD0XmAD0Afj0OQ2AMcwAeAEEYW2QFMwsAOhTRMHgLZ2cAPgAKFKTgCAgKKIBKUoswWQBbYWQSZ0Fs-FZ3KAALGQB3GDBkXoBREhkSXIAiAEswADdREkm0qHGytl19TkcUCAAxSYw6czDc63QsB2t+VbbPdx9-AKA)

packages/react/src/xcss/xcss-prop.ts
DarkPurple141: No I need to add some I was actually going to merge your approach in `cssMap` so they speak the same language.

packages/react/src/cx/index.ts
liamqma: Aha, thanks for explaining this.

packages/react/src/xcss/xcss-prop.ts
liamqma: Sounds good 👍

package.json
```diff
@@ -92,6 +92,7 @@
     "eslint-plugin-json-files": "^2.2.0",
     "eslint-plugin-react": "^7.32.2",
     "eslint-plugin-react-hooks": "^4.6.0",
+    "expect-type": "^0.17.3",
```
itsdouges, line 95: If you rebase you'll get this in (and the eslint improvements)

packages/react/src/css-map/index.ts
```diff
@@ -16,22 +15,22 @@ import type { Pseudos } from './pseudos';
...
  * ```
  */
-type CssProps = Readonly<Properties<string | number>>;
+type CSSProperties = Readonly<Properties<string | number>>;
 
-type AllPseudos = { [key in Pseudos]?: CssProps & AllPseudos };
+type AllPseudos = { [key in Pseudos]?: CSSProperties & AllPseudos };
 
 // The `screen and (max-width: 768px)` part of `@media screen and (max-width: 768px)`.
 // Ideally we would do type checking to forbid this from containing the `@media` part,
```
itsdouges, line 23: What's this comment touching on? Do template literal types solve this?

packages/react/src/cx/index.ts
itsdouges: There already exists the `cx` function internally - would make sense to have it be updated with these types.

So then usage would look like this in some form:

```
import { cssMap, cx } from '@compiled/react';

const styles = cssMap({
  bg: { color: 'red' }
  selected: { textTransform: 'uppercase' }
});

<Component xcss={cx(styles.bg, selected && styles.selected)} />
```

Might want to bikeshed the API though if everyone is using it directly for more use cases we could expose it as a property under `css` to streamline usage? 

```

const singleStyle = css({ color: 'red' });

const styles = css.create({
  bg: { color: 'red' }
  selected: { textTransform: 'uppercase' }
});

<Component xcss={css.cx(styles.bg, selected && styles.selected)} />
```

🤔 

@liamqma @JakeLane WDYT?

packages/react/src/xcss/__tests__/types.test.ts
itsdouges, line 1: How'd you find using expect type

packages/react/src/xcss/pseudos.ts
```diff
@@ -0,0 +1,27 @@
+import type { Pseudos as CSSPseudos } from 'csstype';
+// List of pseudo-classes and pseudo-elements are from csstype
+// but with & added in the front, so that we target the current element
+// (instead of a child element)
+type SafePseudos = Exclude<
```
itsdouges, line 5: nit: Can we invert this and declare them ourselves so when new pseudos are introduced we can control the roll out ensuring Compiled handles it?

@JakeLane wdyt?

packages/react/src/xcss/pseudos.ts
```diff
@@ -0,0 +1,27 @@
...
+  Exclude<
+    CSSPseudos,
+    | ':first-child'
+    | ':first-of-type'
+    | ':last-child'
+    | ':last-of-type'
+    | ':only-child'
+    | ':only-of-type'
+  >;
+export type Pseudos = { [Pseudo in SafePseudos]: `&${Pseudo}` }[SafePseudos];
```
itsdouges, line 27: Compiled handles dangling pseudos (without the need for `&`) - but I like the sentiment to lean on one.

packages/react/src/xcss/xcss-prop.ts
```diff
@@ -0,0 +1,50 @@
...
+ * type Props = {
+ *   xcss: XCSSProp<'color', '&:hover'>
+ * }
+ *
+ * const Component = ({ xcss }: Props) => {
+ *
+ * }
+ *
+ * // Valid usage
+ * <Component xcss={{ color: 'red', '&:hover': { color: 'blue' } }} />
```
itsdouges, line 39: This isn't actually correct they need to be flushed through `css` or equivalent.

packages/react/src/xcss/__tests__/types.test.ts
```diff
@@ -0,0 +1,74 @@
...
+          color: 'red',
+        },
+      })
+    ).toMatchTypeOf<XCSSProp<'color', '&:active'>>();
+  });
+
+  it('should allow composition via cx', () => {
+    expectTypeOf(
+      cx(
+        xcss({
```
itsdouges, line 27: We'd need to teach Compiled about xcss. Can we flush it through css instead? @liamqma only `cssMap` would be compatible with component declared styles right now because of the current compilation method yeah? (css doesn't get compiled down to class names, cssMap does?)

packages/react/src/xcss/xcss-prop.ts
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * // Invalid usage
+ * <Component xcss={{ backgroundColor: 'red' }} />
+ * ```
+ *
+ */
+export type XCSSProp<AllowedProperties extends keyof CSSProperties, TPseudos extends CSSPseudos> =
```
itsdouges, line 46: We'll also need to define media queries here. And we need to enable TPSeudos and TMediaQuery to be void - such that you can opt out from having them at all.

packages/react/src/xcss/__tests__/types.test.ts
liamqma: > css doesn't get compiled down to class names, cssMap does?

Yes, this is correct.

packages/react/src/xcss/pseudos.ts
JakeLane: Agree, we want to own these. New CSS rules seem to get used almost immediately which is a huge problem for browser support - types are a good layer to stop that

packages/react/src/xcss/pseudos.ts
JakeLane: Yeah no need to handle this in lint layer when types can fix that for us

packages/react/src/cx/index.ts
DarkPurple141: I had a look for the `cx` function and I'm not entirely sure what you're referring to @itsdouges?

package.json
DarkPurple141: yeah no worries will rebase a bit later.

packages/react/src/css-map/index.ts
DarkPurple141: This is an old comment from @liamqma - do u have context?

packages/react/src/cx/index.ts
itsdouges: Ah I think I'm talking about ax...

packages/react/src/xcss/__tests__/types.test.ts
DarkPurple141: Yeah overall I found it pretty good - the only thing is technically I needed to handle the error being thrown from the type test - so to workaround I had to remove the error but this isn't strictly the correct thing to do.

packages/react/src/css-map/pseudos.ts
DarkPurple141, line 1: So are you @JakeLane / @itsdouges saying you want this hand declared file?

packages/react/src/cx/index.ts
DarkPurple141: I don't think `ax` is quite the same either it takes an array unless you're saying to adjust the signature and and port any internal existing usage?

packages/react/src/xcss/__tests__/types.test.ts
DarkPurple141: @itsdouges can you explain what you're saying here in more detail? There are quite a few additional elements in `css` as it currently stands including support for fn interpolations which complicate the defintiion of the types. Tried to keep it strictly separate for now just to demonstrate logic and get it working. I can remove `xcss` here (although in practice it is just trying to emulate how it would work if it was `css`)

packages/react/src/xcss/pseudos.ts
DarkPurple141: Can have a look. Happy to invert the logic and hardcode the pseudo list as it was previously. Interestingly @liamqma had added the `&` prefix in the source of the current implementattion for `cssMap` which is partly why I pulled it out so had access to both types.

packages/react/src/xcss/pseudos.ts
DarkPurple141: So is this only to say... keep it like it is?

packages/react/src/xcss/xcss-prop.ts
DarkPurple141: Yep good shout.

packages/react/src/xcss/xcss-prop.ts
DarkPurple141: Third param (also as per Liam's comment media or all @ in general?)

packages/react/src/css-map/pseudos.ts
itsdouges: Yep

packages/react/src/cx/index.ts
itsdouges: Iirc it's exported so introducing another function that is very similar in capability isn't productive. We should have an answer to consolidate the two sometime down the line (doesn't have to be now).

packages/react/src/xcss/__tests__/types.test.ts
itsdouges: I'm saying can we get away from introducing another API to generate styles and instead lean on what already exists (css / cssMap)

packages/react/src/xcss/pseudos.ts
itsdouges: Yep, one or the other. Preference would be lean on the spec (e.g https://developer.chrome.com/articles/css-nesting/#precision-and-flexibility-with)

packages/react/src/xcss/xcss-prop.ts
itsdouges: Third param

packages/react/src/css-map/index.ts
liamqma: To write a media query in CSS map, we gonna do 
```
const foo = cssMap({
   "@media": {
       "screen and (max-width: 768px)": {
           color: 'red'
       }
   }
})
```
Note `@media screen and (max-width: 768px)` is segmented into two parts. This idea is borrowed from vanilla-extract https://vanilla-extract.style/documentation/styling/#media-queries.

The comment says ideally we should check the second segment does not include "@media" as it's redundant.

packages/react/src/cx/index.ts
liamqma: Sorry, I don't follow you @itsdouges Maybe a quick huddle?

packages/react/src/xcss/pseudos.ts
liamqma: Can change the type to "without &" for CSS Map too as long as it's consistent. @dddlr , do you remember why we choose "with &"?

packages/react/src/xcss/__tests__/types.test.ts
liamqma: Yes, xcss will pretty much use the compilation method of cssMap. I've created a prelimary PR for xcss implementation: https://github.com/atlassian-labs/compiled/pull/1512

packages/react/src/xcss/pseudos.ts
dddlr: > Interestingly @liamqma had added the & prefix in the source of the current implementattion for cssMap which is partly why I pulled it out so had access to both types.

Oh @DarkPurple141 I added the type checking related stuff in cssMap haha, my cssMap changes were in the same commit as Liam's so everything got a bit murky in the commit history

&-ful and &-less forms work the same in Compiled - both the below forms target the same element (not a child element):
```
css({ ':hover': { color: 'blue' } })
css({ '&:hover': { color: 'blue' } })
```

The `(instead of a child element)` in the comment is incorrect and was written before I realised this - the comment should probably be changed tbh

I went with only allowing the form with `&` for the sake of consistency (one way to write `:hover` is better than two) and because it more clearly showed that it was targetting the same element. But happy to change it to be without `&` if that's easier for migration from emotion

dddlr: Also agree with @JakeLane about hardcoding what pseudos we allow - gives us the ability to know exactly what selectors developers will be using (without having to monitor any changes being made to `csstype`)

packages/react/src/cx/index.ts
dddlr: `cx` is a new function that hasn't been implemented yet right? I couldn't find any existing usages of `cx` in Compiled or ADS, but I'm not really familiar with the latter

Merging `cx` and `ax` together later down the line sounds ok to me though

packages/react/src/xcss/xcss-prop.ts
dddlr: Also just wanted to confirm, the plan is that it will be impossible by design to specify more complex selectors (e.g. `:not(:active)`) and child selectors (e.g. `div span`) using `xcss` right?

cssMap currently does support these through the `selectors` object (https://compiledcssinjs.com/docs/api-cssmap#nested-selectors-and-advanced-selectors) but this can be forbidden through linting rules

itsdouges:
I'll be taking over this feature, will put up another PR soon.

PR #1533, opened 2023-10-20
Component declared styles

itsdouges:
This pull requests adds the ability for components to declare styles they accept via the "xcss" prop. Look at the tests (both babel + runtime) for examples of how it works. It currently works with the `cssMap()` API, as well as a single static inline object.

A component declares a `xcss` prop, during which they can specify what properties they accept, and if any, what pseudo selectors / elements they accept. When declaring the prop you _must_ declare both generics.

```tsx
import { XCSSProp, XCSSAllProperties, XCSSAllPseudos } from '@compiled/react';

// Allow everything
function Component(props: { xcss?: XCSSProp<XCSSAllProperties, XCSSAllPseudos> }): JSX.Element;

// Allow all properties and a single pseudo
function Component(props: { xcss?: XCSSProp<XCSSAllProperties,  '&::after'> }): JSX.Element;

// Allow only color, and no pseudos
function Component(props: { xcss?: XCSSProp<'color', never> }): JSX.Element;

// Allow nothing - :-)
function Component(props: { xcss?: XCSSProp<never, never> }): JSX.Element;
```

Using the prop looks like:

```tsx
// Inline object
<Component xcss={{ color: 'red' }} />

// cssMap() API
const styles = cssMap({ primary: { color: 'red' } });
<Component xcss={styles.primary} />

// Concatenating styles together
const styles = cssMap({ primary: { color: 'red' }, secondary: { background: 'blue' } });
<Component xcss={cx(styles.primary, styles.secondary)} />

// Defining xcss prop on host elements
const styles = cssMap({ primary: { color: 'red' } });
function Component({ xcss }: { xcss?: XCSSProp<'color', never> }) {
  return <div css={styles.primary} className={xcss} />;
}
```

- If you pass a value that is not accepted you will get an error at the `xcss` prop call site.
- The `xcss` prop can also be named, e.g: `containerXcss`

---

**FAQ**

- Q: Why don't we use array syntax for concatenating styles with xcss prop?
  A: Using array syntax leads to callsites that should be violations slip through the cracks. Here is an [example sandbox](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wChSYBPMJOAYQgBto4BeOAIjSegH0YkADxgc4AH07dmUXgCN0AawDmOAK4A7ACZymmjgG4K1WgGVgAL1rt8AZxApGjfOIKMUUJUnyGqNOCbB0KwIARgAGMOcJfAAmCO9yX1oAIWhNJCgTKkZg2yZgTSiCTQgYAUKXfE0UGwALJELDIz8GaQA1dzY4AAMAN3cACgBaIYASAG9W6ABfAEpun2N-CyQOqC6+wZGJs0s5hebTQLRVzvZNqGGx8YCg-aak+hMTBFUcgBEkTGB1YBhgCDqLrjUhwMFwIgoErqRiUOBSaAAfgAXPQeFA1i4bDAoD8lIZwRDUNDYXB5GhlGotFMoCi0e1OhJsbj1PjQeDISS4fYPD8ACoQMB024nNYEjnEwGknlKH6pMq4YXHU5QcVgzlS7nuWXqLLuGBKoJi9nqyUwrW89QAUS0htF7jVRKhmrggU0mjxAqFqJFKsdGvNrqhHtZ8pgip9yuNhIDpLdIaUetgdr9JqdXKD7rxNs0KejEudgdkaQyXqTMAA6gUYLVhSt-WbScWoOkoF6c1XNDW65YG4WmyWoGHcOXO92ffW07G4c3W8OQB3q7WJ72p42Z4OyzB9TS6TT86b+xuW6XBTnd6j9w610eyYP5+WL-ToAf0y7Zxl5+f0Xv0a-p3eJ5toK5ZZJQOR0qkQFgTkfYZh+wFgDmMFIJBg4oXB773qU4YgKB2SoaiUGthhN7wdhCoLloKFodBBGYUWm4gdusAIFCwCqDYPapjG66Aa27ZaGxHqcdx-58QhD4sTAwkcVxK48QW5FAV+QnsaJCnGtMDxLAACjYSCqCUNhdPgyK1BAvQZEUZnoP8VkJKQ6RoG4RDwoC2JwLwvAuTUNhEJgqIaMAACOqi0DYlAgMWjCLC0uBgMAOSaChdBuDYNjIJgGRIOoJzAnAADa3m+RlAUALqoji4VwNp5CHHAAAadDPAAkgIIAADwANJwEIAhaCZChIJQECYE8LxvEgnzfL8-yAgAfMCaaFQAinAPxwMNo3jS1k0fF8Px-AC6iVXA639blmgmb1iJouASUNKl6WZV8OV5bQqLqEgVmqqQdUNc1bUdTEPV9cIV1DSNY0Ta8B2zcdi1dEDJjtUgXXdUtEiYA4BkuOobyMC4GjpLNDSJEsKNg5dg1bdDu3PHD02HXNJ1LewKNoyAoOYy4AGcyDPULYV5VNKQmAaBgJ3whlnV8uDA3XbDU0zUd83qAtAzYuBSConysyoiChJrRtQLbTDfJnQwD3Jc9flZe9Jz-XARtwDWOAAO5wN9XtWlAOCXBwAByEAQho-wgEgHCzP95AS3l6twO8ECqJ4NhWoIKDgDkMQDOMghoBl0zIvnhc2KiVP4AiUD4Atcwu2uMCqFAQKdR6vRwAA9AtseA3tJi1CgfjsK7YKFdtqLMniVsJY9KUEWl9tvUQH299w6ieWgghdJ1ACCCuQ0VAB0J8owPQ9ICL5WayfR9l6iu8G3A++sEtrvuxAXs+3AfsBwM+A-H6LiFA6gYD4FmIYaY9V16b3RChLoZc85pmrqifAQCrhEE0LMMgcwmgwPgNXZIiYCJgnYIg0e7lpCoP6DYQoAAaNM5JKQp2pOiahWwhiYOwf9CB5BOo90JJ3Tu-gLJvE0N7UofV-bQDTJ1ZOqckDp0ztnJAMQ4AFwyqwcY1cULTC7j3NMQj6AgPwPADIAcySqHgHUFOjBxGyFoDyYa4iahwF6ACNw6sTLYFwHARgPx4BQCmjYWR8i04ZyzmAHO6iy5aMKjogidDKHQCIShcqeju7kEEcIgetjxHmJkYSORKdwnKKiaomJmjtHolSQRDJAjwTFIUUoyJ0SNE2C0VvAYCSdZJMIcQnWsx6mkE6pk0gQA) that demonstrates the issue.
- Q: What Compiled APIs does xcss prop work with?
  A: `cssMap()`, `inline object`
---

**TODO**

- [x] handle "NoPseudos" generic
- [x] fix types for inline object given to xcss prop
- [x] double check if we can or can't use arrays in xcss prop (and document the findings)
- [x] block at rules from xcss prop
- [x] add jsdoc to new APIs
- [x] confirm ax / ac func usage
- [x] bikeshed new APIs
- [x] fix types
- [x] add config for pragma usage with xcss

---

**FOLLOW UP**

- [ ] XCSSProp type helpers (required properties)
- [ ] type factory for compiled APIs (all APIs can be taught what property values / media queries should be API)
- [ ] bounded media query / at rule support for xcss prop
- [ ] lint rules for xcss prop (anything that could break out of the system):
  - no-supress-xcss (violate when ts-ignore / ts-expect error xcss prop)
  - local-cx-xcss (ensure cx is used inside xcss prop only)
  - no-any-xcss (violate when xcss prop has any in either declaration or usage)
  - xcss-pragma (ensure compiled pragma is inscope when setting className={xcss} on host elements so xcss prop value can be passed without error)
    - mode: always — will always require a pragma to be there for both declaration and usage (used for migrations)
    - mode: declaration — will only be required when declaring xcss prop on a class name
  - no-classname-object (violate anyone declaring an object in class name ??)
  - no-js-xcss (typescript must be used when using xcss prop)
- [ ] documentation
- [ ] update css prop to be able to take output from xcss prop (work around: pass into class name prop)


packages/react/src/types.ts
```diff
@@ -31,5 +31,65 @@ export type CssObject<TProps> = Readonly<{
...
   | undefined; // Something like `undefined && styles`
+
+// List of pseudo-classes and pseudo-elements are from csstype
+// but with & added in the front, so that we target the current element
+// (instead of a child element)
+
+// We also exclude anything that requires providing an argument
+// (e.g. &:not(...) ), and anything that uses information from elements
+// outside of the current element (e.g. &:first-of-type)
+export type CSSPseudos =
```
itsdouges, line 45: Moved from `pseudos.ts` to here

packages/babel-plugin/src/babel-plugin.ts
```diff
@@ -65,6 +66,8 @@ export default declare<State>((api) => {
           paths: [state.opts.root ?? this.cwd],
         }));
       }
+
+      this.transformCache = new WeakMap();
```
itsdouges, line 70: We use this to cache transformed nodes so we don't transform again. Let me know if there is a more appropriate direction.

packages/babel-plugin/src/babel-plugin.ts
```diff
@@ -87,6 +90,11 @@ export default declare<State>((api) => {
               }
             }
           }
+
+          if (/(x|X)css={/.exec(file.code)) {
+            // xcss prop was found turn on Compiled
+            state.compiledImports = {};
```
itsdouges, line 96: Turn on xcss prop if we have found prop usage.

packages/babel-plugin/src/xcss-prop/index.ts
```diff
@@ -0,0 +1,80 @@
...
+    const cssOutput = buildCss(container.expression, meta);
+    const { sheets, classNames } = transformCssItems(cssOutput.css, meta);
+
+    // Replace xcss prop with class names
+    // The object has a type constraint to always be a basic object with no values.
+    container.expression = classNames[0];
+    path.parentPath.replaceWith(compiledTemplate(jsxElementNode, sheets, meta));
+  } else {
+    const sheets = collectPassStyles(meta);
+    path.parentPath.replaceWith(compiledTemplate(jsxElementNode, sheets, meta));
```
itsdouges, line 78: This greatly simplifies compilation as if it isn't a inline object we assume no extra compilation is needed (so we get the best behaviour of local transforms that works in any scenarios).

packages/react/src/css-map/index.ts
```diff
@@ -112,15 +106,20 @@ type ReturnType<VariantName extends string> = Record<VariantName, CssProps>;
...
+ *  solid: { borderStyle: 'solid' },
  * });
  * const Component = ({ borderStyle }) => <div css={css(borderStyleMap[borderStyle])} />
  *
  * <Component borderStyle="solid" />
  * ```
  */
-
-export default function cssMap<T extends string>(_styles: Variants<T>): Readonly<ReturnType<T>> {
+export default function cssMap<
```
itsdouges, line 117: cssMap now works with xcss prop

packages/react/src/jsx/jsx-local-namespace.ts
```diff
@@ -68,7 +68,14 @@ export namespace CompiledJSX {
...
   export type IntrinsicElements = {
-    [K in keyof ReactJSXIntrinsicElements]: ReactJSXIntrinsicElements[K] & {
+    [K in keyof ReactJSXIntrinsicElements]: Omit<ReactJSXIntrinsicElements[K], 'className'> & {
+      // We override class name so we can pass xcss prop to it. We opt to do this instead of
+      // Making the output of cssMap() a string intersection so we can also have an inline object
+      // be declared.
+      /**
+       * The class name prop now can be given the output of xcss prop from `@compiled/react`.
+       */
+      className?: string | Record<string, any> | null | false;
```
itsdouges, line 78: This is needed to enable passing xcss prop to className:

```jsx
function Component({ xcss }: { xcss?: XCSSProp<'color', never> }) {
  return <div className={xcss} />;
}
```

Else it is a type error. There was an alternate solution which was enabling xcss prop / css map types be intersected with a string `& string` but that would then not work with inline objects:

```jsx
<Component xcss={{ color: 'red' }} />
```

Since we own both ends (jsx pragma + compiled APIs) I opted for this solution to get it all working.

packages/react/src/xcss-prop/index.ts
```diff
@@ -0,0 +1,141 @@
...
+ * We currently block all at rules from xcss prop.
+ * This needs us to decide on what the final API is across Compiled to be able to set.
+ */
+type XCSSAtRules = {
+  [Q in CSS.AtRules]?: never;
+};
+
+export type CSSRuleDefinition = CSSProperties & CSSPseudoRule;
+
+declare const __classref: unique symbol;
```
itsdouges, line 30: One thought: This is going to be referentially equivalent on both ends. In a monorepo world this is OK. In a platform vs. ecosystem world where Compiled might be outdated in Ecosystem 2p/3p but latest in platform it will be a problem as they could be on two different major versions (our more likely right now, two different minor versions pre 1.0!).

I'll see if I can move this to a string instead.

packages/react/src/types.ts
MonicaOlejniczak: /nit could use css type as this:

```js
import type { SimplePsuedos } from 'csstype';

export type CSSPseudos = `&${Exclude<
    SimplePseudos,
    ':first' | ':first-child' | ':first-of-type' | ':last' | ':last-child' | ':last-of-type' | ':only-child' | ':only-of-type'
>}`;
```

itsdouges: We definitely can — instead we're opting to explicitly define the psuedos so we can handle them as appropriate. E.g. there might be new pseudos we don't want to support, or something we want to delay.

packages/babel-plugin/src/babel-plugin.ts
MonicaOlejniczak: Why not lean in on how the CSS prop enables compiled with the above logic? Or check for the existence in `JSXOpeningElement`?

packages/react/src/types.ts
```diff
@@ -31,5 +31,65 @@ export type CssObject<TProps> = Readonly<{
 export type CssFunction<TProps = unknown> =
   | CssType<TProps>
   | BasicTemplateInterpolations // CSS values in tagged template expression
+  | null
```
dddlr, line 34: I think I'm missing a bit of context here - in what kind of situation would we get `null`? Is it something like `null && styles`?

packages/react/src/types.ts
```diff
@@ -31,5 +31,65 @@ export type CssObject<TProps> = Readonly<{
...
   | boolean // Something like `false && styles`
   | undefined; // Something like `undefined && styles`
+
+// List of pseudo-classes and pseudo-elements are from csstype
+// but with & added in the front, so that we target the current element
+// (instead of a child element)
```
dddlr, line 40: I originally wrote this comment and it's wrong haha, both `:hover` and `&:hover` target the current element

Didn't DST decide to go with the `&`-less syntax in the end? I don't think it's too late to change our syntax as nobody is using cssMap yet

packages/react/src/xcss-prop/index.ts
```diff
@@ -0,0 +1,143 @@
+import type * as CSS from 'csstype';
+
+import { ac } from '../runtime';
+import type { CSSPseudos } from '../types';
+
+type CSSProperties = Readonly<CSS.PropertiesFallback<number | string>>;
```
dddlr, line 6: TIL PropertiesFallback - is providing a CSS value as an array something we support? e.g.

```ts
{ display: ['-webkit-flex', 'flex'], }```

File: packages/react/src/css/__tests__/types.test.ts [hunk 11]
```diff
@@ -1,4 +1,5 @@
 /** @jsxImportSource @compiled/react */
+// eslint-disable-next-line import/no-extraneous-dependencies
```
dddlr, line 2: Thank you for fixing this :pray:

packages/babel-plugin/src/babel-plugin.ts
itsdouges: Because when someone is using the inline object exclusively:

```
<Component xcss={{ color: 'red' }} />
```

Having to turn on the jsx pragma or import Compiled APIs would be a point of friction.

I can go both ways for turning it on with the regex check or existence in JSXOpeningElement. What would the team prefer?

packages/react/src/types.ts
itsdouges: Yes doing conditional styles (ternary / binary / etc).

packages/react/src/types.ts
itsdouges: We're flexible. If we lean on `&` it is at least following the nested spec which means we can do less transformations.

I'd say let's do whatever the nested spec does.

packages/react/src/xcss-prop/index.ts
itsdouges: I can't remember. Not with these APIs at least. I'd say we don't need fallbacks thanks to autoprefixer. I'll remove the fallbacks for now!

packages/react/src/xcss-prop/__tests__/xcss-prop.test.tsx
```diff
@@ -0,0 +1,196 @@
...
+    const styles = cssMap({
+      redColor: { color: 'red', '&::after': { backgroundColor: 'green' } },
+    });
+
+    const { getByText } = render(<CSSPropComponent xcss={styles.redColor} />);
+
+    expect(getByText('foo')).toHaveCompiledCss('color', 'red');
+  });
+
+  it('should type error when given a pseduo and none are allowed', () => {
```
dddlr, line 24: Nit: pseudo

packages/react/src/types.ts
dddlr: Ah all good, `cssMap` is also pretty flexible (the restrictions on using `&:hover` instead of `:hover` are pretty arbitrary) so feel free to make changes to that if needed

liamqma (approved):

itsdouges:
@liamqma @MonicaOlejniczak @dddlr any thoughts on new API btw?

packages/react/src/types.ts
kylorhall-atlassian: In my past, I've seen `& :hover` meaning "self child:hover" so `:hover` could mean "child:hover" while `&:hover` explicitly means "self:hover", so my vote is for `&:pseudo`

packages/react/src/xcss-prop/index.ts
```diff
@@ -0,0 +1,143 @@
...
+ *
+ * <Component xcss={cx(isPrimary && styles.text, !isPrimary && styles.primary)} />
+ * ```
+ */
+export const cx = <TStyles extends [...XCSSProp<any, any>[]]>(
+  ...styles: TStyles
+): TStyles[number] & string => {
+  // Types won't match here as type-time are always objects.
+  // At runtime however they will be an array of strings.
+  return ac(styles as unknown as string[]) as unknown as TStyles[number] & string;
```
kylorhall-atlassian, line 142: Without diving in too much as this is all my first time seeing the _rest_ of the Compiled code, this does feel wrong of course, but I can't seem to fix it in 30s in a codespace, so I'll just believe the comment.

I will note that `ac` doesn't support `null`, eg. `classNames: (AtomicGroups | string | undefined | false)[]` — will it be handled fine, eg. is there a `typeof === 'object'` that might trigger on a null?

packages/react/src/xcss-prop/index.ts
```diff
@@ -0,0 +1,143 @@
...
+ * <Component xcss={styles.text} />
+ * ```
+ *
+ * To concatenate and conditonally apply styles use the {@link cssMap} {@link cx} functions.
+ */
+export type XCSSProp<
+  TAllowedProperties extends keyof CSSProperties,
+  TAllowedPseudos extends CSSPseudos
+> =
+  | (XCSSItem<TAllowedProperties> & XCSSPseudos<TAllowedProperties, TAllowedPseudos> & XCSSAtRules)
```
kylorhall-atlassian, line 116: We've got a `XCSSAtRules` used which is `{ […]?: never }`, should this be a defaulted generic if we want this by default, eg.:
```suggestion
  TAllowedPseudos extends CSSPseudos,
  TAllowedAtRules extends XCSSAtRules = XCSSAtRules
> =
  | (XCSSItem<TAllowedProperties> & XCSSPseudos<TAllowedProperties, TAllowedPseudos> & TAllowedAtRules)
```

Not sure what the plan to get AtRules working is all.  Maybe this is more of a TODO piece?

packages/react/src/types.ts
kylorhall-atlassian: I will note that `null` isn't present in some types, eg. in the `ac()` functions.  Does it need wider propagation?

kylorhall-atlassian (commented):
Had to login to this account for the first time for this; very minor comments as I don't know much about Compiled, glad you folks do!

packages/react/src/types.ts
itsdouges: Let me take a look. In this case it might not matter because it's all falsy values 🤔.

packages/react/src/types.ts
itsdouges: Yeah exactly. We'll keep it as is.

packages/react/src/xcss-prop/index.ts
itsdouges: 😅 

So `ac` doesn't return a string, it returns a class that has a `toString()` method. To make sure we can just pass it to areas that accept strings (...like `className`) we intersect it with the `string` literal.

@liamqma is `ac` safe taking `null`? To Kylors point around the typeof check.

- I'll look into adding `null` to `ac` type.

packages/react/src/xcss-prop/index.ts
itsdouges: Yeah it's a TODO in a follow up. For now we completely block it. It's because `css` `cssMap` have different APIs for the same thing — need to agree on a path forward.

dddlr:
> @liamqma @MonicaOlejniczak @dddlr any thoughts on new API btw?

API looks ok to me, will wait for others to share their thoughts (cc @JakeLane for when he gets back)

packages/react/src/xcss-prop/__tests__/xcss-prop.test.tsx
dddlr, line 24:
```suggestion
  it('should type error when given a pseudo and none are allowed', () => {
```

packages/react/src/xcss-prop/index.ts
liamqma: ac should handle `null` as it's doing truthy check. 

> I'll look into adding null to ac type.
Thank you. Can you add it to ax as well. It should be the same.

liamqma:
> @liamqma @MonicaOlejniczak @dddlr any thoughts on new API btw?

Looks good to me. 👍

packages/react/src/types.ts
dddlr, line 40:
```suggestion
// List of pseudo-classes and pseudo-elements are from csstype
// but with & added in the front. Compiled supports both &-ful
// and &-less forms and both will target the current element
// (`&:hover` <==> `:hover`), however we force the use of the
// &-ful form for consistency.
```

or something like this

itsdouges: Sweet will add in

packages/babel-plugin/src/babel-plugin.ts
MonicaOlejniczak: JSX element is more consistent with css prop, but @JakeLane can make the call

packages/react/src/xcss-prop/index.ts
```diff
@@ -0,0 +1,143 @@
...
+};
+
+/**
+ * Please think twice before using this type, you're better off declaring explicitly
+ * what your API should be, for example only defining `color`.
+ *
+ * Use in conjunction with {@link XCSSProp} to allow all properties to be given to
+ * your component.
+ */
+export type XCSSAllProperties = keyof CSSProperties;
```
MonicaOlejniczak, line 48: What's the reasoning behind using the word `All` in the type name rather than just `XSSProperties`? Is it to make it clearer you don't want it used?

packages/react/src/xcss-prop/index.ts
itsdouges: (If we didn't define it, it would mean anyone can pass anything! Which we want to avoid)

packages/react/src/xcss-prop/index.ts
itsdouges: Pushed up improvements.

packages/babel-plugin/src/babel-plugin.ts
itsdouges: Alright will catch up. Both are valid IMO – not needing it if you're exclusively only using xcss prop is great for just one less thing to remember to do.

packages/react/src/types.ts
```diff
@@ -31,5 +31,73 @@ export type CssObject<TProps> = Readonly<{
...
+  | '&:valid'
+  | '&:visited';
+
+/**
+ * The xcss prop must be given all known available properties even
+ * if it takes a subset of them. This is ensure the (lack-of an)
+ * excess property check doesn't enable makers to circumvent the
+ * system and pass in values they shouldn't.
+ */
+export type CSSProperties = Readonly<CSS.Properties<string | number>>;
```
itsdouges, line 103: Moved this here to share with `xcss` prop and `cssMap`

packages/react/src/xcss-prop/index.ts
itsdouges: Yeah to make it clear you're opting into everything. It's really only for makers declaring xcss prop, not for internal Compiled usage.

packages/babel-plugin/src/babel-plugin.ts
JakeLane: This is dicey because prop naming isn't required for `xcss` to work, we'd be leaning on convention here. I think I'd prefer to keep build perf here ideal so we can rely on plain text.

itsdouges: > This is dicey because prop naming isn't required for xcss to work

In what sense? Currently / by design inline object compilation will only work on a prop called `xcss` or `yzXcss`.

JakeLane: meaning we could turn on the API in numerous ways, rather than just looking at the prop name

JakeLane (approved):

itsdouges:
Will merge and fast follow with the type feature and the must have lint rules.

Once they are in let's cut a release and test this out in AFE.

PR #1536, opened 2023-10-23
Delete flow types

dddlr:
Two reasons:

* Flow is missing types that are blocking potential future Compiled changes, such as template literal types and `unique symbol`.
* Maintaining Flow types is tedious and requires several hacks: [scripts/flow-types.sh](https://github.com/atlassian-labs/compiled/blob/master/scripts/flow-types.sh)
* In internal packages that are/will be using Compiled, we've already switched from Flow to TypeScript, so maintaining support for Flow is no longer necessary


itsdouges (dismissed):
LGTM

MonicaOlejniczak (approved):

liamqma (approved):

PR #1541, opened 2023-10-26
XCSS prop options arg

itsdouges:
This pull request adds a third arg to the `XCSSProp` type to declare what properties and / or pseudos should be required. It looks like this:

```ts
XCSSProp<
  'color' | 'backgroundColor',
  '&hover',
  { requiredProperties: 'color'; requiredPseudos: '&:hover' }
>
```

In that example `'color'` property and `'&:hover'` pseudo is required, `'backgroundColor'` is optional. Look at tests under xcss prop for more examples.

- Both `requiredProperties` and `requiredPseudos` are constrained to the values passed in the first two generic args. If you pass something that isn't valid you'll get a type error.
- Passing `never` flags you want everything to be optional.
- Passing specific string literals flags each one should be required.

While I was here I also fixed two bugs:

- fixed passing an empty inline object to xcss prop threw an exception
- fixed `selectors` arg from cssMap sometimes being able to be passed to an xcss prop

---

**TODO**

- [x] jsdocs

### DIFF
--- a/packages/babel-plugin/src/xcss-prop/index.ts
+++ b/packages/babel-plugin/src/xcss-prop/index.ts
@@ -69,9 +69,26 @@ export const visitXcssPropPath = (path: NodePath<t.JSXOpeningElement>, meta: Met
     const cssOutput = buildCss(container.expression, meta);
     const { sheets, classNames } = transformCssItems(cssOutput.css, meta);
 
-    // Replace xcss prop with class names
-    // The object has a type constraint to always be a basic object with no values.
-    container.expression = classNames[0];
+    switch (classNames.length) {
+      case 1:
+        // Replace xcss prop with class names
+        // Remeber: The object has a type constraint to always be a basic object with no values.
+        container.expression = classNames[0];
+        break;
+
+      case 0:
+        // No styles were merged so we replace with an undefined identifier.
+        container.expression = t.identifier('undefined');
+        break;
+
+      default:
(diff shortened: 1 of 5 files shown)


packages/babel-plugin/src/xcss-prop/index.ts
```diff
@@ -69,9 +69,26 @@ export const visitXcssPropPath = (path: NodePath<t.JSXOpeningElement>, meta: Met
...
+        break;
+
+      case 0:
+        // No styles were merged so we replace with an undefined identifier.
+        container.expression = t.identifier('undefined');
+        break;
+
+      default:
+        throw buildCodeFrameError(
+          'Unexpected count of class names please raise an issue on Github',
```
liamqma, line 86: If classNames.length > 1, it probably means cssOutput.css includes some CSS, which cannot be statically evaluated. Do you think we should be more explicit about it?

liamqma (dismissed):

packages/babel-plugin/src/xcss-prop/index.ts
itsdouges: That should be covered by the static invariant above, no?

liamqma: yup, you are right.

liamqma (approved):

JakeLane (approved):

PR #1545, opened 2023-10-31
No suppress XCSS prop lint rule

itsdouges:
This pull request adds a new lint rule `no-suppress-xcss` that marks any ts-ignore / ts-expect-errors on the preceding line as violations when used in conjunction with xcss prop.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-suppress-xcss/__tests__/rule.test.ts
+++ b/packages/eslint-plugin/src/rules/no-suppress-xcss/__tests__/rule.test.ts
@@ -2,6 +2,101 @@ import { typeScriptTester as tester } from '../../../test-utils';
 import { noSuppressXCSS } from '../index';
 
 tester.run('no-styled-tagged-template-expression', noSuppressXCSS, {
-  valid: [],
-  invalid: [],
+  valid: [
+    `
+      <Component xcss={{ fill: 'red' }} />
+    `,
+    `
+      <Component innerXcss={{ fill: 'red' }} />
+    `,
+    `
+      // @ts-expect-error
+      function Foo() {
+        // @ts-ignore
+        return (
+          <>
+            <Component xcss={{ fill: 'red' }} />
+          </>
+        );
+      }
+    `,
+  ],
+  invalid: [
+    {
+      code: `
+      // @ts-expect-error
+      <Component xcss={{ fill: 'red' }} />
+    `,
+      errors: [{ messageId: 'no-suppress-xcss' }],
+    },
+    {
+      code: `
+      // @ts-expect-error
+      <Component innerXcss={{ fill: 'red' }} />
+    `,
+      errors: [{ messageId: 'no-suppress-xcss' }],
+    },
+    {
+      code: `
+      // @ts-ignore
+      <Component xcss={{ fill: 'red' }} />
+    `,
+      errors: [{ messageId: 'no-suppress-xcss' }],
+    },
+    {
+      code: `
+
+      <Component
+        // @ts-expect-error
+        innerXcss={{ fill: 'red' }}
+      />
+    `,
+      errors: [{ messageId: 'no-suppress-xcss' }],
+    },
+    {
+      code: `
+
+      <Component
+        // @ts-ignore
+        xcss={{ fill: 'red' }}
+      />
+    `,
+      errors: [{ messageId: 'no-suppress-xcss' }],
+    },
+    {
+      code: `
+      // @ts-ignore
+      <Component innerXcss={{ fill: 'red' }} />
+    `,
+      errors: [{ messageId: 'no-suppress-xcss' }],
+    },
+    {
+      only: true,
+      code: `
+      function Foo() {
+        return (
+          <>
+            {/* @ts-ignore */}
+            <Component xcss={{ fill: 'red' }} />
+          </>
+        );
+      }
+    `,
+      errors: [{ messageId: 'no-suppress-xcss' }],
+    },
+    {
+      only: true,
+      code: `
+      function Foo() {
+        return (
+          <>
+            {/* @ts-expect-error */}
+            <Component xcss={{ fill: 'red' }} />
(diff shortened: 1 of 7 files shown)


packages/eslint-plugin/src/rules/no-suppress-xcss/index.ts
```diff
@@ -0,0 +1,48 @@
...
+export const noSuppressXCSS: Rule.RuleModule = {
+  meta: {
+    docs: {
+      recommended: true,
+      description:
+        'The xcss prop is predicated on adhering to the type contract. Supressing it breaks this contract and thus is not allowed.',
+      url: 'https://github.com/atlassian-labs/compiled/tree/master/packages/eslint-plugin/src/rules/no-supress-xcss',
+    },
+    messages: {
+      'no-suppress-xcss': 'bad!',
```
MonicaOlejniczak, line 12:
```suggestion
      'no-suppress-xcss': '{{name}} should not contain a preceding {{tag}} as it breaks the type contract',
```

where name = the found `xcss` attribute and `tag` = `@ts-expect-error` or `@ts-ignore`

packages/eslint-plugin/src/rules/no-suppress-xcss/index.ts
```diff
@@ -0,0 +1,48 @@
...
+  create(context) {
+    return {
+      'JSXAttribute[name.name=/[xX]css$/]': (node: Rule.Node) => {
+        if (!node.loc) {
+          return;
+        }
+
+        const comments = context.sourceCode.getAllComments();
+
+        for (let i = 0; i < comments.length; i++) {
```
MonicaOlejniczak, line 25: Just use a for of loop?

packages/eslint-plugin/src/rules/no-suppress-xcss/__tests__/rule.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    },
+    {
+      code: `
+      // @ts-ignore
+      <Component innerXcss={{ fill: 'red' }} />
+    `,
+      errors: [{ messageId: 'no-suppress-xcss' }],
+    },
+    {
+      only: true,
```
MonicaOlejniczak, line 74: Remove the only here and below

packages/eslint-plugin/src/rules/no-suppress-xcss/__tests__/rule.test.ts
```diff
@@ -0,0 +1,102 @@
...
+        return (
+          <>
+            {/* @ts-expect-error */}
+            <Component xcss={{ fill: 'red' }} />
+          </>
+        );
+      }
+    `,
+      errors: [{ messageId: 'no-suppress-xcss' }],
+    },
```
MonicaOlejniczak, line 100: Missing test cases for individual properties?

```
function Foo() {
  return (
    <Component
      xcss={{
        // @ts-expect-error
        fill: 'red',
      }}
    />
  );
}
```

packages/eslint-plugin/src/rules/no-suppress-xcss/index.ts
dddlr: Great suggestion, I would also suggest briefly mentioning the consequences if they ignore the lint error (e.g. "Ignoring this error will break the styling for this component.")

packages/eslint-plugin/src/rules/no-suppress-xcss/index.ts
```diff
@@ -0,0 +1,48 @@
...
+            continue;
+          }
+
+          const commentLine = comment.loc.start.line;
+          const nodeLine = node.loc.start.line;
+          const isCommentOnPreviousLine = nodeLine - 1 === commentLine;
+
+          if (
+            isCommentOnPreviousLine &&
+            ['@ts-expect-error', '@ts-ignore'].some((tag) => comment.value.includes(tag))
```
dddlr, line 37: Sorry not at my laptop right now, are there any usages of ts-nocheck in jfe/cfe/afe? Wondering if it's worth adding that to this rule as well

itsdouges:
Thanks for the comments everyone! Will resolve today.

packages/eslint-plugin/src/rules/no-suppress-xcss/__tests__/rule.test.ts
itsdouges: Great catch

packages/eslint-plugin/src/rules/no-suppress-xcss/index.ts
itsdouges: Cheers

itsdouges:
Resolved all comments thanks folks, ready for another review!

MonicaOlejniczak (approved):

liamqma (approved):

packages/eslint-plugin/src/rules/no-suppress-xcss/index.ts
```diff
@@ -0,0 +1,79 @@
...
+      continue;
+    }
+
+    const commentLine = comment.loc.start.line;
+    const nodeLine = node.loc.start.line;
+    const isCommentOnPreviousLine = nodeLine - 1 === commentLine;
+
+    if (
+      isCommentOnPreviousLine &&
+      ['@ts-expect-error', '@ts-ignore', '@ts-nocheck'].some((tag) => comment.value.includes(tag))
```
dddlr, line 21: Ah for context, `@ts-nocheck` applies to the entire file, so `isCommentOnPreviousLine` could be false and the type checking would still be suppressed

dddlr: I believe it's used at the top of the file? Would be good to double check

itsdouges: Ah, fair shout. I'll check!

PR #1547, opened 2023-11-02
XCSS prop no js lint

itsdouges:
This pull request adds a new lint rule `no-js-xcss` that marks non tsx files as violations when using xcss prop.

### DIFF
--- a/packages/eslint-plugin/src/rules/no-js-xcss/index.ts
+++ b/packages/eslint-plugin/src/rules/no-js-xcss/index.ts
@@ -1,6 +1,6 @@
 import type { Rule } from 'eslint';
 
-export const noSuppressXCSS: Rule.RuleModule = {
+export const noJavaScriptXCSSRule: Rule.RuleModule = {
   meta: {
     docs: {
       recommended: true,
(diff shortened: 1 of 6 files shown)


MonicaOlejniczak (approved):

packages/eslint-plugin/src/rules/no-js-xcss/index.ts
```diff
@@ -0,0 +1,32 @@
...
+      url: 'https://github.com/atlassian-labs/compiled/tree/master/packages/eslint-plugin/src/rules/no-js-xcss',
+    },
+    messages: {
+      'no-js-xcss':
+        'Using xcss in JavaScript risks incidents and unintended behaviour when code changes — only use inside TypeScript files',
+    },
+    type: 'problem',
+  },
+  create(context) {
+    const violations = new WeakSet<Rule.Node>();
```
dddlr, line 18: where is `violations` used?

itsdouges: In the create() function scope to check if a parent had a violation or not, if it does bail out

itsdouges: Oops good shout this is from copy paste code. Will remove!

itsdouges: Updated

JakeLane (approved):

PR #1550, opened 2023-11-02
Update jsx-pragma rule to ensure it exists when using xcss prop

itsdouges:
This pull request updates the jsx pragma rule to mark violations when passing a `className` prop an `xcss` or named xcss prop value. To resolve your define the jsx pragma. This rule exists as a DX improvement otherwise it will be confusing how to resolve the type error when passing `className` the xcss typed object.

Violation:

```jsx
<div className={xcss} />
```

Success:

```jsx
/* @jsxImportSource @compiled/react */
<div className={xcss} />
```

**TODO**

- [x] Refactor to have an autofixer (re-use already existing logic!)

### DIFF
--- a/packages/eslint-plugin/src/rules/jsx-pragma/index.ts
+++ b/packages/eslint-plugin/src/rules/jsx-pragma/index.ts
@@ -27,6 +27,50 @@ const findReactDeclarationWithDefaultImport = (
   return undefined;
 };
 
+function createFixer(context: Rule.RuleContext, source: SourceCode, options: Options) {
+  return function* fix(fixer: Rule.RuleFixer) {
+    const pragma = options.runtime === 'classic' ? '@jsx jsx' : '@jsxImportSource @compiled/react';
+    const reactImport = findReactDeclarationWithDefaultImport(source);
+    if (reactImport) {
+      const [declaration, defaultImport] = reactImport;
+      const [defaultImportVariable] = context.getDeclaredVariables(defaultImport);
+
+      if (defaultImportVariable && defaultImportVariable.references.length === 0) {
+        if (declaration.specifiers.length === 1) {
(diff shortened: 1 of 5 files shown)


packages/eslint-plugin/src/rules/jsx-pragma/index.ts
```diff
@@ -27,13 +27,58 @@ const findReactDeclarationWithDefaultImport = (
   return undefined;
 };
 
+function createFixer(context: Rule.RuleContext, source: SourceCode, options: Options) {
+  return function* fix(fixer: Rule.RuleFixer) {
```
itsdouges, line 31: This was just lifted and shifted, no changes outside of moving `const pragma = ...` inside here.

dddlr: ah cool, thank you for refactoring it

packages/eslint-plugin/src/rules/jsx-pragma/index.ts
```diff
@@ -121,66 +166,38 @@ export const jsxPragmaRule: Rule.RuleModule = {
...
+        node: Rule.Node
+      ) => {
+        if (node.type !== 'JSXAttribute' || jsxPragma || jsxImportSourcePragma) {
+          return;
+        }
+
+        if (
+          node.value?.type === 'JSXExpressionContainer' &&
+          node.value.expression.type === 'Identifier' &&
+          /[Xx]css$/.exec(node.value.expression.name)
```
dddlr, line 179: using `test` instead of `exec` may be a bit faster

itsdouges: Great, will update.

dddlr (approved):
lgtm, ty douges!

PR #1557, opened 2023-11-10
Fix jsx pragma breaking @compiled/babel-plugin-strip-runtime stylesheet extraction

dddlr:
Currently, using `import { jsx } from '@compiled/react'` results in stylesheet extraction not working when combined with the `@babel/preset-react`, because our Babel plugin only looks for `React.createElement`.

This PR extends our plugin so it checks whether a classic JSX pragma (`/** @jsx jsx */`) exists. If it does, it will instead look for usages of the function listed in the JSX pragma (so long as the JSX pragma is the `jsx` function from Compiled).

Other changes:

* If the classic JSX pragma is detected and it refers to Compiled's `jsx` function, both the JSX pragma and the Compiled `jsx` import will be deleted from the code. This is so we avoid accidentally importing `@compiled/react`
* (Internal change) Improved types in `packages/babel-plugin-strip-runtime/src/types.ts` to be more accurate to what properties are actually available

- [x] Replace `jsx` with `React.createElement`
- [x] add tests for styled and css APIs with classic and automatic runtime, and with/without pragma comment

Will do in a follow up PR:
- Create linting rule to prevent usages of `jsx` in file


packages/babel-plugin-strip-runtime/src/types.ts
```diff
@@ -28,26 +38,19 @@ export interface PluginPass {
...
-    };
-  };
+  file: BabelFile;
 
   /**
-   * Current filename
+   * The name of the jsx function specified in the JSX pragma
+   * "@ jsx nameOfJsxFunction". (Classic syntax)
    */
-  filename: string;
```
dddlr, line 45: filename and cwd still exist, they're in `OriginalPluginPass` now

dddlr: https://github.com/babel/babel/blob/master/packages/babel-core/src/transformation/plugin-pass.js

JakeLane (approved):

PR #1568, opened 2023-11-30
Created ESLint rule to disallow/discourage usage of empty styled.div() calls

zerosicx:
- Raises an error for all styled.element() calls where either an empty object argument is passed, or no argument at all.
- Wrote tests for valid/invalid cases
- Moved common utility function `isStyledImportSpecifier` to the `eslint-plugin/src/utils` folder
- Added to recommended Compiled eslint rules as a warning
- Worked on code quality with Nathan

### DIFF
--- a/packages/eslint-plugin/src/configs/recommended.ts
+++ b/packages/eslint-plugin/src/configs/recommended.ts
@@ -11,5 +11,6 @@ export const recommended = {
     '@compiled/no-keyframes-tagged-template-expression': 'error',
     '@compiled/no-styled-tagged-template-expression': 'error',
     '@compiled/no-suppress-xcss': 'error',
+    '@compiled/no-empty-styled-expression': 'warning',
   },
 };
--- a/packages/eslint-plugin/src/configs/recommended.ts
+++ b/packages/eslint-plugin/src/configs/recommended.ts
@@ -11,6 +11,6 @@ export const recommended = {
     '@compiled/no-keyframes-tagged-template-expression': 'error',
     '@compiled/no-styled-tagged-template-expression': 'error',
     '@compiled/no-suppress-xcss': 'error',
-    '@compiled/no-empty-styled-expression': 'warning',
+    '@compiled/no-empty-styled-expression': 'error',
   },
 };
--- a/packages/eslint-plugin/src/rules/no-empty-styled-expression/README.md
+++ b/packages/eslint-plugin/src/rules/no-empty-styled-expression/README.md
@@ -11,27 +11,13 @@ Passing an empty object or no object at all causes Compiled to build extra `div/
 👎 Examples of **incorrect** code for this rule:
 
 ```
-const Wrapper = styled.div();
-
-function Button() {
-  return <Wrapper>
-    <MyComponent>Hello</MyComponent>
-    <MyOtherComponent>world</MyOtherComponent>
-  </Wrapper>;
-}
+const EmptyStyledExpression = styled.div();
 ```
 
 and
 
 ```
-const Wrapper = styled.div({});
-
-function Button() {
-  return <Wrapper>
-    <MyComponent>Hello</MyComponent>
-    <MyOtherComponent>world</MyOtherComponent>
-  </Wrapper>;
-}
+const EmptyStyledExpressionArgument = styled.div({});
 ```
 
 👍 Examples of **correct** code for this rule:
@@ -43,16 +29,9 @@ const Wrapper = styled.div({
     backgroundColor: 'green',
   },
 });
-
-function Button() {
-  return <Wrapper>
-    <MyComponent>Hello</MyComponent>
-    <MyOtherComponent>world</MyOtherComponent>
-  </Wrapper>;
-}
 ```
 
-🔀 Work Arounds and Recommendations
+🔀 Recommendations
 
 1. Write your code in a way that doesn't require a Wrapper
 2. Use the empty React fragment: `<> <YourComponentHere></YourComponentHere> </>`
--- a/packages/eslint-plugin/src/rules/no-empty-styled-expression/README.md
+++ b/packages/eslint-plugin/src/rules/no-empty-styled-expression/README.md
@@ -2,7 +2,7 @@
 
 Discourages any `styled` expression to be used when passing empty arguments in `@compiled/react`.
 
-Passing an empty object or no object at all causes Compiled to build extra `div` elements as opposed to simply using a `div`. This leads to reduced performance and is greatly discouraged. If a wrapper is necessary, opt to use a `div` or wrap it in the empty React fragment `<> <YourComponentHere></YourComponentHere> </>`.
+Passing an empty object or no object at all causes Compiled to build extra `div/span` elements, as opposed to simply using a `div`. This leads to reduced performance and is greatly discouraged. If a wrapper is necessary, opt to use a `div` or wrap it in the empty React fragment `<> <YourComponentHere></YourComponentHere> </>`.
 
 ---
 
--- a/packages/eslint-plugin/src/rules/no-empty-styled-expression/README.md
+++ b/packages/eslint-plugin/src/rules/no-empty-styled-expression/README.md
@@ -31,9 +31,9 @@ const Wrapper = styled.div({
 });
 ```
 
-# 🔀 Recommendations
+## 🔀 Recommendations
 
-## Use elements directly
+### Use elements directly
 
 ```diff
 - const Wrapper = styled.div({});
@@ -44,7 +44,7 @@ const Wrapper = styled.div({
   }
 ```
 
-## Use a React fragment
+### Use a React fragment
 
 ```diff
 - const Wrapper = styled.div({});
--- a/packages/eslint-plugin/src/rules/no-empty-styled-expression/README.md
+++ b/packages/eslint-plugin/src/rules/no-empty-styled-expression/README.md
@@ -12,12 +12,10 @@ Passing an empty object or no object at all causes Compiled to build extra `div/
 
 ```javascript
 const EmptyStyledExpression = styled.div();
-```
 
-and
-
-```javascript
 const EmptyStyledExpressionArgument = styled.div({});
+
+const EmptyStyledExpressionArgument = styled.div([]);
 ```
 
 👍 Examples of **correct** code for this rule:
--- a/packages/eslint-plugin/src/rules/no-empty-styled-expression/README.md
+++ b/packages/eslint-plugin/src/rules/no-empty-styled-expression/README.md
@@ -1,6 +1,6 @@
 # `no-empty-styled-expression`
 
-Discourages any `styled` expression to be used when passing empty arguments in `@compiled/react`.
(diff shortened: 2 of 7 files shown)


packages/eslint-plugin/src/configs/recommended.ts
```diff
@@ -11,5 +11,6 @@ export const recommended = {
     '@compiled/no-keyframes-tagged-template-expression': 'error',
     '@compiled/no-styled-tagged-template-expression': 'error',
     '@compiled/no-suppress-xcss': 'error',
+    '@compiled/no-empty-styled-expression': 'warning',
```
MonicaOlejniczak, line 14: This should be an error as it is considered best practice, and if it causes too many violations in Jira then you can just override it back to warning in product.

packages/eslint-plugin/src/rules/no-empty-styled-expression/README.md
```diff
@@ -0,0 +1,59 @@
...
+---
+
+## Rule details
+
+👎 Examples of **incorrect** code for this rule:
+
+```
+const Wrapper = styled.div();
+
+function Button() {
```
MonicaOlejniczak, line 16: The `Button` doesn't add any value to the example imo, should just be

```jsx
const EmptyStyledExpression = styled.div();
```

Same for all the others

packages/eslint-plugin/src/rules/no-empty-styled-expression/README.md
```diff
@@ -0,0 +1,59 @@
...
+  return <Wrapper>
+    <MyComponent>Hello</MyComponent>
+    <MyOtherComponent>world</MyOtherComponent>
+  </Wrapper>;
+}
+```
+
+🔀 Work Arounds and Recommendations
```
MonicaOlejniczak, line 55: Just say `Recommendations`

packages/eslint-plugin/src/rules/no-empty-styled-expression/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+1. Write your code in a way that doesn't require a Wrapper
```
MonicaOlejniczak, line 57: Maybe something more concrete to work with?

---

### Use elements directly

```diff
- const Wrapper = styled.div({});

   function App() {
-    return <Wrapper>hello world</Wrapper>;
+    return <div>hello world</div>;
  }
```

### Use a React fragment

```diff
- const Wrapper = styled.div({});

   function App() {
-    return <Wrapper>hello world</Wrapper>;
+    return <>hello world</>;
  }
```

packages/eslint-plugin/src/rules/no-empty-styled-expression/index.ts
```diff
@@ -0,0 +1,57 @@
...
+  ): RuleModule['create'] =>
+  (context) => {
+    return {
+      'CallExpression[callee.type="MemberExpression"]': (node: CallExpression) => {
+        const { references } = context.getScope();
+
+        const isStyledImported = references.some((reference) =>
+          reference.resolved?.defs.some(isStyledImportSpecifier)
+        );
+        if (!isStyledImported) {
```
MonicaOlejniczak, line 29:
```suggestion
        
        if (!isStyledImported || !isEmptyStyledExpression(node)) {
```

packages/eslint-plugin/src/rules/no-empty-styled-expression/README.md
```diff
@@ -0,0 +1,55 @@
...
+- const Wrapper = styled.div({});
+
+   function App() {
+-    return <Wrapper>hello world</Wrapper>;
++    return <div>hello world</div>;
+  }
+```
+
+```diff
+Use a React fragment
```
MonicaOlejniczak, line 48: Should be under a heading, not in the diff

packages/eslint-plugin/src/rules/no-empty-styled-expression/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+Passing an empty object or no object at all causes Compiled to build extra `div/span` elements, as opposed to simply using a `div`. This leads to reduced performance and is greatly discouraged. If a wrapper is necessary, opt to use a `div` or wrap it in the empty React fragment `<> <YourComponentHere></YourComponentHere> </>`.
+
+---
+
+## Rule details
+
+👎 Examples of **incorrect** code for this rule:
+
+```
```
MonicaOlejniczak, line 13: You should make this ```javascript to get code highlighting, same for below

packages/eslint-plugin/src/rules/no-empty-styled-expression/index.ts
```diff
@@ -0,0 +1,53 @@
...
+import { isStyledImportSpecifier } from '../../utils/styled-import';
+
+type RuleModule = Rule.RuleModule;
+
+const isEmptyStyledExpression = (node: CallExpression): boolean => {
+  const [firstArg] = node.arguments;
+  if (firstArg?.type === 'ObjectExpression') {
+    return firstArg.properties.length === 0;
+  }
+  return true;
```
JakeLane, line 13: There could be something else passed that is not an `ObjectExpression` and we'd throw an error. I think it's better that we only return true here if the args are empty (`node.arguments.length === 0`) or if there's a single arg with something we know is empty.

packages/eslint-plugin/src/rules/no-empty-styled-expression/__tests__/rule.test.ts
```diff
@@ -0,0 +1,65 @@
...
+      });
+    `,
+    `
+      import { styled } from 'styled';
+
+      styled.span({
+        color: blue
+      });
+    `,
+  ],
```
JakeLane, line 30: Can you please add a test case where you pass something that isn't an object expression?

JakeLane (changes requested):

packages/eslint-plugin/src/rules/no-empty-styled-expression/index.ts
zerosicx: That's fair enough. I'm happy to add these changes. Could I also ask for context on what kind of objects can we expect to be passed into styled.div(), and the impact of what happens when those objects are passed in?

packages/eslint-plugin/src/rules/no-empty-styled-expression/__tests__/rule.test.ts
```diff
@@ -0,0 +1,75 @@
...
+        import { styled } from 'styled';
+
+        styled.span({
+            color: blue
+        });
+    `,
+    `
+        import { styled } from '@compiled/react';
+
+        styled.span("hello world");
```
dddlr, line 33: Not sure if styled.span("Hello world") is valid Compiled syntax, will need to double check

Can we also add some examples of using a template string too?

```js
styled.span`color: blue`;

styled.h2`
    color: blue;
    background-color: pink;
`;
```

This was my mistake (sorry!) for not mentioning it in the ticket, but this is a loophole that you may want to fix with this rule as well:

```js
styled.div``; // equivalent to styled.div({})
styled.span``; // equivalent to styled.span({})
```

packages/eslint-plugin/src/rules/no-empty-styled-expression/__tests__/rule.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    `,
+    `
+        import { styled } from '@compiled/react';
+
+        styled.span({}, {});
```
dddlr, line 38: Thank you for catching this edge case! <3 Shouldn't we lint against these as well? I could be wrong here (not on laptop rn) but this seems equivalent to `styled.span({})`

packages/eslint-plugin/src/rules/no-empty-styled-expression/__tests__/rule.test.ts
dddlr: Yeah e.g. arrays

https://github.com/atlassian-labs/compiled/blob/master/packages/react/src/styled/index.ts#L64

Perhaps

```
styled.div([])
```

might be another loophole we need to address too?

And maybe booleans like styled.div(false && {}), but that's probably getting a bit obscure

https://github.com/atlassian-labs/compiled/blob/master/packages/react/src/types.ts#L35

packages/eslint-plugin/src/rules/no-empty-styled-expression/index.ts
```diff
@@ -0,0 +1,54 @@
...
+  return false;
+};
+
+export const createNoEmptyStyledExpressionRule =
+  (
+    isEmptyStyledExpression: (node: CallExpression) => boolean,
+    messageId: string
+  ): RuleModule['create'] =>
+  (context) => {
+    return {
```
dddlr, line 23: Well written 👍

packages/eslint-plugin/src/rules/no-empty-styled-expression/__tests__/rule.test.ts
zerosicx: I did think so as well. @JakeLane mentioned different types of objects other than ObjectExpressions can be passed into the styled.span({}) call, but I don't know exactly what => thus he asked to only check for the case where it is (1) empty or (2) contains exactly one object, but the object is empty. Realistically, I don't know what edge cases people will attempt - this kind of blurs the line between using the function _**wrong**_ and using it a _**not recommended/best practice**_ way.

packages/eslint-plugin/src/rules/no-empty-styled-expression/__tests__/rule.test.ts
zerosicx: Isn't the template usage disallowed in the first place? Covered [by this rule](https://github.com/atlassian-labs/compiled/tree/master/packages/eslint-plugin/src/rules/no-styled-tagged-template-expression). It doesn't check for empty template expressions, but it _**changes template expressions => CallExpressions**_ where this no-empty-syled-expression rule will then apply. It seems a little redundant to check this case, but please correct me if I'm wrong!

zerosicx: Discussed with Tom and Liam -- as the aforementioned syntax is covered by [another ESLint rule](https://github.com/atlassian-labs/compiled/tree/master/packages/eslint-plugin/src/rules/no-styled-tagged-template-expression), checks for the TagTemplateExpression will not be added to this PR.

dddlr: ah sounds good, ty for the extra context!

packages/eslint-plugin/src/rules/no-empty-styled-expression/__tests__/rule.test.ts
dddlr: yeah `styled` API is kinda scuffed in how it's a bit too flexible 🫠🫠🫠

happy to leave it as-is, since you mentioned on slack about how it would be too complex to handle all of the potential edge cases `styled` supports

dddlr (approved):

PR #1525, opened 2023-10-12
Support resolver module path

MonicaOlejniczak:
Supports module paths in the `@compiled/babel-plugin` resolver so that custom files and packages can be referenced in JSON configuration files (i.e. `babel.config.json` and `.compiledcssrc` for parcel)

## Changes
* Add relevant test fixtures
* Rename `fixtures/parcel-transformer-test-custom-resolve-app` to `fixtures/parcel-transformer-test-custom-resolver-app` making it represent its corresponding config option (`resolve`)
* Add `resolver` option to parcel transformer, and deprecate `resolve` since the former allows greater flexibility
* Add support for resolver string resolution in the babel plugin

PR #1569, opened 2023-12-01
Support resolver module path

bobrovnik:
Same as https://github.com/atlassian-labs/compiled/pull/1525 but now for webpack-loader, since Jira's SSR is still on webpack :(

### DIFF
--- a/packages/webpack-loader/src/compiled-loader.ts
+++ b/packages/webpack-loader/src/compiled-loader.ts
@@ -1,9 +1,7 @@
-import fs from 'fs';
-import { dirname, normalize } from 'path';
+import { normalize } from 'path';
 
 import { parseAsync, transformFromAstAsync } from '@babel/core';
 import { createError, toBoolean } from '@compiled/utils';
-import { CachedInputFileSystem, ResolverFactory } from 'enhanced-resolve';
 import { getOptions } from 'loader-utils';
 import type { LoaderContext } from 'webpack';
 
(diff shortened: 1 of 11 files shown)


packages/webpack-loader/src/compiled-loader.ts
```diff
@@ -177,12 +169,12 @@ export default async function compiledLoader(
...
             classNameCompressionMap: options.extract && options.classNameCompressionMap,
             onIncludedFiles: (files: string[]) => includedFiles.push(...files),
-            resolver: {
-              // The resolver needs to be synchronous, as babel plugins must be synchronous
-              resolveSync: (context: string, request: string) => {
-                return resolver.resolveSync({}, dirname(context), request);
-              },
-            },
+            resolver: options.resolver
+              ? options.resolver
```
liamqma, line 173: Do you need to `require(options.resolver)`?

JakeLane (approved):

liamqma (approved):

packages/webpack-loader/src/compiled-loader.ts
liamqma: I see. 
```
    if (typeof this.opts.resolver === 'object') {
        this.resolver = this.opts.resolver;
      } else if (typeof this.opts.resolver === 'string') {
        this.resolver = require(require.resolve(this.opts.resolver, {
          paths: [state.opts.root ?? this.cwd],
        }));
      }
```

PR #1574, opened 2023-12-05
Feature/add no empty styled expressions rule declaration

zerosicx:
- Disabled `no-extraneous-dependencies` check for the strict-api fie that's causing pre-commit errors
- Added `no-empty-styled-expression` rule to plugin declaration
- Added changeset

### DIFF
--- a/.changeset/clean-beds-laugh.md
+++ b/.changeset/clean-beds-laugh.md
@@ -1,0 +1,6 @@
+---
+'@compiled/eslint-plugin': minor
+'@compiled/react': minor
+---
+
+Added no-empty-styled-expression rule to eslint plugin rule declarations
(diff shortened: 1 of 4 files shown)


.changeset/clean-beds-laugh.md
```diff
@@ -0,0 +1,6 @@
+---
+'@compiled/eslint-plugin': minor
+'@compiled/react': minor
+---
```
dddlr, line 4:
```suggestion
'@compiled/eslint-plugin': patch
---
```

dddlr: no need for changeset for `@compiled/react` cuz the customer's code doesn't change

dddlr (approved):

packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api.ts
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
 import { createStrictAPI } from '@compiled/react';
```
liamqma, line 2:
```
import { createStrictAPI } from '../../index';
```
instead of disabling linting rule.

zerosicx: Oh no, I didn't see this! Would I be able to add this change to he Version Bump PR?

PR #1577, opened 2023-12-05
strict-api to import from local instead of @compiled/react

zerosicx:
- updated `strict-api.ts` to import from local rather than `@compiled/react`
- removed changeset due to unnecessary version bump


.changeset/config.json
liamqma, line 1: Undo it

dddlr (approved):
LGTM! ❤️

PR #1578, opened 2023-12-06
Fixed eslint rule declaration typo

zerosicx:
- Caused JFE/CFE "cannot find rule ___" error

### DIFF
--- a/packages/eslint-plugin/src/index.ts
+++ b/packages/eslint-plugin/src/index.ts
@@ -26,7 +26,7 @@ export const rules = {
   'no-keyframes-tagged-template-expression': noKeyframesTaggedTemplateExpressionRule,
   'no-styled-tagged-template-expression': noStyledTaggedTemplateExpressionRule,
   'no-suppress-xcss': noSuppressXCSS,
-  'no-empry-styled-expression': noEmptyStyledExpressionRule,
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/src/index.ts
```diff
@@ -26,7 +26,7 @@ export const rules = {
   'no-keyframes-tagged-template-expression': noKeyframesTaggedTemplateExpressionRule,
   'no-styled-tagged-template-expression': noStyledTaggedTemplateExpressionRule,
   'no-suppress-xcss': noSuppressXCSS,
-  'no-empry-styled-expression': noEmptyStyledExpressionRule,
+  'no-empty-styled-expression': noEmptyStyledExpressionRule,
```
MonicaOlejniczak, line 29: This will need a bugfix changeset since it has already been released with the typo

dddlr: ah yeah, you'll need to run `yarn changeset` (with `patch` instead of `minor`) cuz we want to release a new version

zerosicx: Amazing, thank you for the fast review :)

dddlr (approved):

PR #1585, opened 2023-12-07
Fix #1557 breaking Emotion usages

dddlr:
#1557 had a bug where `/** @jsx jsx */` would be deleted in _all_ cases, not just cases where this comment was being used for Compiled. This meant that when `@compiled/babel-plugin-strip-runtime` was turned on in a repository, many Emotion usages were broken:

```tsx
/** @jsx jsx */
import { jsx } from '@emotion/react';

<div css={{ color: 'blue' }} />
```

causing a `React is not defined` error, because [Babel's `@babel/preset-react` plugin](https://babeljs.io/docs/babel-plugin-transform-react-jsx#react-classic-runtime) will convert the `<div>` elements to `React.createElement` calls, instead of `jsx` calls.

In this PR:
* First commit is just the contents of #1557
* Second commit fixes the bug described above. `/** @jsx jsx */` will only be deleted if Compiled is being imported in the file.
* Third commit adds some Emotion-related tests

---

Todo:

- [x] Import React from react package
    - Not needed, this is handled by the `importReact` option in `@compiled/babel-plugin`
    - Will update documentation for `@compiled/babel-plugin-strip-runtime` to clarify that `importReact` needs to be true for this package to work (separate PR)
- [x] Test classic JSX with some packages that use Emotion
- [x] ~~Test automatic JSX on AFE~~ automatic JSX transform not supported on AFE, tested Emotion usages on Compiled repo instead


dddlr:
reverting to draft to fix automatic runtime

dddlr:
turns out automatic runtime is fine, I was just testing on a repository that was on React 16 and thus didn't support the automatic runtime in the first place 😆 ended up testing on the Compiled repo instead

packages/babel-plugin-strip-runtime/src/__tests__/strip-runtime-source-code.test.ts
```diff
@@ -247,6 +247,293 @@ describe('babel-plugin-strip-runtime using source code', () => {
         );
```
liamqma, line 247: Can we add a test where both `@emotion/react` and `@compiled/react` co-exist in the same file?

dddlr: Hmm explicitly importing `@emotion/react` and `@compiled/react` in the same file isn't something we support 😅
especially given that which library should parse `css={{ color: 'blue' }}` is currently a bit ambiguous (also see [eslint rule](https://github.com/atlassian-labs/compiled/tree/master/packages/eslint-plugin/src/rules/jsx-pragma))

If what you mean is Compiled `xcss` and Emotion `css` being used in the same file, happy to look into adding a test for that

liamqma (dismissed):

packages/babel-plugin-strip-runtime/src/__tests__/strip-runtime-source-code.test.ts
liamqma: > Hmm explicitly importing @emotion/react and @compiled/react in the same file isn't something we support 😅
especially given that which library should parse css={{ color: 'blue' }} is currently a bit ambiguous (also see [eslint rule](https://github.com/atlassian-labs/compiled/tree/master/packages/eslint-plugin/src/rules/jsx-pragma))

Got it. I forgot about it

itsdouges: Can we throw instead of relying on implicit behaviour?

dddlr: > Unless the behaviour is remove jsx pragma when using jsx pragma from compiled then all good.

yup thats correct

dddlr: > If what you mean is Compiled xcss and Emotion css being used in the same file, happy to look into adding a test for that

ah just realised such a test exists already haha

https://github.com/atlassian-labs/compiled/blob/master/packages/babel-plugin/src/xcss-prop/__tests__/transformation.test.ts#L340

liamqma (approved):

PR #1588, opened 2023-12-11
Fix Compiled runtime being added unexpectedly for xcss prop usage

itsdouges:
Currently Compiled will add its runtime even if there is no direct Compiled xcss prop usage (e.g. xcss prop from ADS primitive components!).

This pull request resolves that by moving some checks around (thanks @dddlr for putting in the initial plumbing!).

### DIFF
--- a/packages/babel-plugin/src/babel-plugin.ts
+++ b/packages/babel-plugin/src/babel-plugin.ts
@@ -108,14 +108,6 @@ export default declare<State>((api) => {
               }
             }
           }
-
-          // Default to true
-          const processXcss = state.opts.processXcss ?? true;
-
-          if (processXcss && /(x|X)css={/.test(file.code)) {
-            // xcss prop was found, turn on Compiled but just for xcss
-            state.usesXcss = true;
-          }
         },
         exit(path, state) {
           if (!state.compiledImports && !state.usesXcss) {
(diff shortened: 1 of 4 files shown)


JakeLane (approved):

packages/babel-plugin/src/babel-plugin.ts
```diff
@@ -274,7 +266,8 @@ export default declare<State>((api) => {
         visitClassNamesPath(path, { context: 'root', state, parentPath: path });
       },
       JSXOpeningElement(path, state) {
-        if (state.usesXcss) {
+        const compiledXCSSProp = state.opts.processXcss ?? true;
+        if (compiledXCSSProp) {
```
itsdouges, line 270: usesXcss / processXcss are now considered two flows.

- process turns on the compilation
- uses makes compiled bring in the runtime

JakeLane (dismissed):

JakeLane (approved):

packages/babel-plugin/src/babel-plugin.ts
liamqma: /nit should we rename `useXcss` to `injectXcssRuntime` to be more explicit?

liamqma (approved):

PR #1590, opened 2023-12-12
Fix no-empty-styled-expression rule that's linting false positives

zerosicx:
- Cause of the bug is checking the CallExpression itself for the styled import reference, rather than the MemberExpression.
   - This would result in checking the `Program` level on direct function calls (false positive) along with triggering the lint rule within nested empty function calls inside a `styled` TemplateExpression.
- Fix: checking the import reference from within the MemberExpression (much like the no-styled-tag-template rule -- this fix was implemented as this solution makes use of one of the utility functions in that rule).
- Added edge test cases from the false positives found on JFE


packages/eslint-plugin/src/rules/no-empty-styled-expression/index.ts
```diff
@@ -25,9 +25,20 @@ const createNoEmptyStyledExpressionRule =
     return {
       'CallExpression[callee.type="MemberExpression"]': (node: CallExpression) => {
         const { references } = context.getScope();
-        const isStyledImported = references.some((reference) =>
-          reference.resolved?.defs.some(isStyledImportSpecifier)
-        );
+
+        if (node.callee.type !== 'MemberExpression') {
```
MonicaOlejniczak, line 29: Is this needed? Shouldn't `CallExpression[callee.type="MemberExpression"]` already filter on that?

packages/eslint-plugin/src/rules/no-empty-styled-expression/index.ts
```diff
@@ -25,9 +25,20 @@ const createNoEmptyStyledExpressionRule =
     return {
       'CallExpression[callee.type="MemberExpression"]': (node: CallExpression) => {
         const { references } = context.getScope();
```
MonicaOlejniczak, line 27: If you leave the callee check in, this should be placed after the early return. It will ensure you do less work, by not having to evaluate `getScope()` until it's needed.

packages/eslint-plugin/src/rules/no-empty-styled-expression/index.ts
```diff
@@ -25,9 +25,20 @@ const createNoEmptyStyledExpressionRule =
...
-        const isStyledImported = references.some((reference) =>
-          reference.resolved?.defs.some(isStyledImportSpecifier)
-        );
+
+        if (node.callee.type !== 'MemberExpression') {
+          return;
+        }
+
+        const membEx = node.callee;
```
MonicaOlejniczak, line 33: I'd just call it callee via `const { callee } = node;` or make it longhand `const { callee: memberExpression } = node;`

packages/eslint-plugin/src/rules/no-empty-styled-expression/__tests__/rule.test.ts
```diff
@@ -45,6 +45,29 @@ tester.run('no-styled-tagged-template-expression', noEmptyStyledExpressionRule,
...
+        import { styled } from '@compiled/react';
+
+        const HeaderTitle = styled.h1<any>\`
+        \${typography.h500()};
+        \${typography.h400()};\` 
+    `,
+    `
+        import { styled } from '@compiled/react';
+
+        ufologger.enable();
```
MonicaOlejniczak, line 64: This is too specific to Atlassian, it should be renamed. Maybe one of these, or anything else you prefer.

```
object.property(); // ast
foo.bar();         // metasyntactic
```

packages/eslint-plugin/src/rules/no-empty-styled-expression/__tests__/rule.test.ts
```diff
@@ -45,6 +45,29 @@ tester.run('no-styled-tagged-template-expression', noEmptyStyledExpressionRule,
...
+    `,
+    `
+        import { styled } from '@compiled/react';
+
+        provideHealthcheck();
```
MonicaOlejniczak, line 69: Same as above, just use `test()` or `call()` or `foo()`

packages/eslint-plugin/src/rules/no-empty-styled-expression/__tests__/rule.test.ts
```diff
@@ -45,6 +45,29 @@ tester.run('no-styled-tagged-template-expression', noEmptyStyledExpressionRule,
...
+        import { styled } from '@compiled/react';
+
+        const HeaderTitle = styled.h1<any>\`
+        \${typography.h500()}\` 
+    `,
+    `   
+        import { styled } from '@compiled/react';
+
+        const HeaderTitle = styled.h1<any>\`
+        \${typography.h500()};
```
MonicaOlejniczak, line 58: Indent these by a couple

packages/eslint-plugin/src/rules/no-empty-styled-expression/index.ts
zerosicx: I also thought this, but the code is giving type errors when I don't check for the node.callee.type either implicitly or explicitly. I will try to find an alternative if I can!

MonicaOlejniczak: Oh I see, it's for type checking. Probably not much you can do, since the types don't accept generic arguments.

liamqma:
It seems the diffs include more than your commits, which makes it a tad hard to know what's actually in the PR.

liamqma (approved):

dddlr:
looks good!

PR #1582, opened 2023-12-06
Block invalid properties on `cssMap` when extended type was met.

kylorhall-atlassian:
Previously, `asdf` in this example would never error because part of the interface existed, so this was "extended", eg. `{ color: …, asdf: … }` extends `{ color: … }` was true.

So instead of using the inferred generic `TStyles extends …` to enforce, we enforce the type explicitly, and use the inferred generic `TStyles` to build our output type.  This maintains the proper output type while blocking the input type.

More easily visualized in the playground:

[...]

PR #1596, opened 2023-12-18
Fix XCSSProp types

itsdouges:
This pull request fixes a missed use case introduced in #1582 or later. The change in types resulted in output type of `cssMap` from `createStrictAPI` no longer being generic, meaning "every value" was considered to be used breaking usage in XCSSProp (as all values breaks most XCSSProp declarations).

The fix is to add more tests to cover the known use cases and to bring the generic back in as an intersection type. My brain hurts to try and understand it but all the tests pass so I'm not too worried. Potentially we might be missing test cases, I'll go over more in my head just in case.

### DIFF
--- a/packages/react/src/create-strict-api/__tests__/index.test.tsx
+++ b/packages/react/src/create-strict-api/__tests__/index.test.tsx
@@ -331,7 +331,7 @@ describe('createStrictAPI()', () => {
         />
       );
 
-      expect(getByTestId('button')).toHaveCompiledCss('background', 'var(--ds-surface)');
+      expect(getByTestId('button')).toHaveCompiledCss('color', 'var(--ds-text)');
     });
 
     it('should error with values not in the strict `CompiledAPI`', () => {
--- a/packages/react/src/create-strict-api/__tests__/index.test.tsx
+++ b/packages/react/src/create-strict-api/__tests__/index.test.tsx
@@ -271,6 +271,46 @@ describe('createStrictAPI()', () => {
   });
 
   describe('XCSSProp', () => {
+    it('should allow valid values from cssMap', () => {
+      function Button({ xcss }: { xcss: ReturnType<typeof XCSSProp<'background', never>> }) {
+        return <button data-testid="button" className={xcss} />;
+      }
+
+      const styles = cssMap({ bg: { background: 'var(--ds-surface)' } });
+      const { getByTestId } = render(<Button xcss={styles.bg} />);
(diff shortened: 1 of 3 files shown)


packages/react/src/create-strict-api/__tests__/index.test.tsx
```diff
@@ -271,6 +271,142 @@ describe('createStrictAPI()', () => {
...
+    });
+
+    it('should disallow invalid values from cssMap', () => {
+      function Button({ xcss }: { xcss: ReturnType<typeof XCSSProp<'background', never>> }) {
+        return <button data-testid="button" className={xcss} />;
+      }
+
+      const styles = cssMap({ bg: { accentColor: 'red' } });
+      // @ts-expect-error — Type 'CompiledStyles<{ accentColor: "red"; }>' is not assignable to type ...
+      const { getByTestId } = render(<Button xcss={styles.bg} />);
```
kylorhall-atlassian, line 292: Only minor nitpick is this could error in a few ways; I'd suggest this (plus or minus formatting), as you do below:
```suggestion
      const { getByTestId } = render(
        <Button
          // @ts-expect-error — Type 'CompiledStyles<{ accentColor: "red"; }>' is not assignable to type ...
          xcss={styles.bg}
        />
      );
```

kylorhall-atlassian (approved):
LGTM, but a nitpick on a test.

kylorhall-atlassian (dismissed):

packages/react/src/create-strict-api/index.ts
```diff
@@ -59,7 +59,8 @@ interface CompiledAPI<TSchema extends CompiledSchema> {
    */
   cssMap<TStylesMap extends CSSMapStyles<TSchema>>(
     // We intersection type the generic both with the concrete type and the generic to ensure the output has the generic applied.
-    // Without both it would allow either the input to now have excess property check kick in, or have all values set as the output.
+    // Without both it would either have the input arg not have excess property check kick in allowing unexpected values or
```
kylorhall-atlassian, line 62: `Without both, it would … kick in, allowing …` :trollface: nah, I got a 🅱️ in English, what do I know :shipit:

kylorhall-atlassian (approved):

packages/react/src/create-strict-api/index.ts
itsdouges: 🙉

PR #1600, opened 2023-12-21
Export `CSSPseudos, CSSProperties, StrictCSSProperties` types for a strict `XCSSProp` implementation.

kylorhall-atlassian:
To avoid the complexity of requiring `ReturnType<typeof XCSSProp<…>>` when creating a `createStrictAPI().XCSSProp` and `props.xcss` implementation, being able to use these types in generics are helpful.

The example usage we want is:
```tsx
import { createStrictAPI, type CSSPseudos, type StrictCSSProperties, type XCSSAllProperties, type XCSSAllPseudos } from '@compiled/react';

const { XCSSProp, css, cssMap, cx } = createStrictAPI<…>();

/** Re-implementing the jsdoc */
export type StrictXCSSProp<
  TAllowedProperties extends keyof StrictCSSProperties,
  TAllowedPseudos extends CSSPseudos,
  TRequiredProperties extends {
  } = never,
> = ReturnType<
  typeof XCSSProp<TAllowedProperties, TAllowedPseudos, TRequiredProperties>
>;
```

Then this becomes a bit easier of a type to understand:
```diff
-export const Component = ({ xcss }: { xcss?: ReturnType<typeof XCSSProp<…>> }) =>
+export const Component = ({ xcss }: { xcss?: StrictXCSSProp<…> }) =>
  <div className={xcss} />
```

Particularly when contrasted against the non-createStrictAPI `XCSSProp` type, which is different.

---

Draft, pending final validation of the idea.

### DIFF
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -1,16 +1,21 @@
 import { createElement } from 'react';
 
 import type { CompiledJSX } from './jsx/jsx-local-namespace';
-import type { CssFunction, CSSProps, CssType } from './types';
 
-export type { CSSProps, CssFunction, CssType };
-
-export { keyframes } from './keyframes';
-export { styled } from './styled';
 export { ClassNames } from './class-names';
+export { createStrictAPI } from './create-strict-api';
 export { default as css } from './css';
 export { default as cssMap } from './css-map';
-export { createStrictAPI } from './create-strict-api';
+export { keyframes } from './keyframes';
+export { styled } from './styled';
+export type {
+  CSSProperties,
+  CSSProps,
+  CSSPseudos,
+  CssFunction,
+  CssType,
+  StrictCSSProperties,
+} from './types';
 export { type XCSSAllProperties, type XCSSAllPseudos, type XCSSProp, cx } from './xcss-prop';
 
 // Pass through the (classic) jsx runtime.
--- a/.changeset/gorgeous-melons-sniff.md
+++ b/.changeset/gorgeous-melons-sniff.md
@@ -1,0 +1,5 @@
+---
+'@compiled/react': patch
+---
+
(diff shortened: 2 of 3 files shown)


packages/react/src/index.ts
```diff
@@ -1,16 +1,21 @@
...
+export { styled } from './styled';
+export type {
+  CSSProperties,
+  CSSProps,
+  CSSPseudos,
+  CssFunction,
+  CssType,
+  StrictCSSProperties,
+} from './types';
 export { type XCSSAllProperties, type XCSSAllPseudos, type XCSSProp, cx } from './xcss-prop';
```
itsdouges, line 19: While we are here I noticed `XCSSAllProperties` doesn't type check when passed to the `XCSSProp` from `createStrictAPI`. Can you update it to be sourced from `StrictCSSProperties` instead of `CSSProperties`.

kylorhall-atlassian: Not catching what you mean—I don't think we're using `CSSProperties` in the output of `createStrictAPI`, right?  Don't see anything in https://github.com/atlassian-labs/compiled/blob/34674ae1a005c4bf2d4b227504b22e5b64ffe669/packages/react/src/create-strict-api/index.ts#L147

itsdouges: Yeah exactly — this line https://github.com/atlassian-labs/compiled/blob/master/packages/react/src/create-strict-api/index.ts#L147 is a subst of `CSSProperties` so it type errors if you try and do:

```
XCSSProp<XCSSAllProperties>
```

kylorhall-atlassian: I got the context now; done in 2a93b6c2.

itsdouges (approved):
LGTM

PR #1603, opened 2023-12-22
Fix @compiled/babel-plugin handling of classic JSX pragma

dddlr:
`@compiled/babel-plugin-strip-runtime` removes the JSX pragma `/** @jsx jsx */` from the file before the [Babel React plugins](https://babeljs.io/docs/babel-plugin-transform-react-jsx) run, resulting in the output containing `React.createElement` calls instead of `jsx` calls (where `jsx` is imported from `@compiled/react`). This is because the `jsx` function is just a wrapper around `React.createElement` anyway, which is unnecessary.

An implicit assumption is that `@compiled/babel-plugin` will have imported React, which is actually incorrect 😱 This wasn't caught in previous PRs as `@compiled/babel-plugin` erroneously only detects JSX pragmas that match `/** @jsx jsx */` exactly, and not other forms like `/** @jsx myJsx */` (which was the form used in our tests).

Potential todos:

* [x] Changeset
* [x] Test with a package that uses Compiled
* [x] Clean up code

Non-blocking todos:

* [ ] Create follow-up PR to update website
* [x] ~Create follow-up PR to refactor tests for readability~

#### What if the classic pragma is set through the `pragma` option in the [`@babel/plugin-transform-react-jsx`](https://babeljs.io/docs/babel-plugin-transform-react-jsx) config?

* This gives invalid output 😦 
* Personally this doesn't seem worth supporting, as [Emotion has never recommended this approach](https://web.archive.org/web/20181130084308/https://emotion.sh/docs/css-prop) (looks like they now prefer either `jsxImportRuntime` or use of their own plugin)
* We throw an error if it detects that the classic pragma is set in `@babel/plugin-transform-react-jsx` config.

#### Should we remove `/** @jsxImportSource @compiled/react */` comments as well, so that the Babel React plugins don't see these either?

Done 👍

#### What if `jsxImportSource` is set through the `importSource` option in the [`@babel/plugin-transform-react-jsx`](https://babeljs.io/docs/babel-plugin-transform-react-jsx) Babel config?

* Currently this works fine, but we get some `@compiled/react/jsx-runtime` imports in the Babel output, e.g.
```js
import { ax, ix } from '@compiled/react/runtime';
import { jsxs as _jsxs } from '@compiled/react/jsx-runtime';
import { jsx as _jsx } from '@compiled/react/jsx-runtime';
const Component = () =>
  _jsx('div', {
    className: ax(['_1wyb1fwx _syaz13q2']),
    children: 'hello world',
  });
```
* We could potentially throw an error here because this is inefficient - `@compiled/react/jsx-runtime` is just a wrapper around `react/jsx-runtime`. Instead I've decided to leave it as-is.
  * This is because unlike the classic JSX pragma which would add `@compiled/react` imports, `@compiled/react/jsx-runtime` is a distinct namespace.
  * Additionally, leaving it as-is would make migrating from an Emotion setup to a Compiled setup a bit easier.
* Future work might involve a separate plugin that runs after `@babel/preset-react` and removes all traces of `@compiled/react/jsx-runtime`, and/or add the `/*#__PURE__*/` comment before the `_jsx` calls.


dddlr:
I've also refactored the tests in `packages/babel-plugin-strip-runtime/src/__tests__/` so there won't be files with 900 lines of tests anymore :)

But currently this is in a separate commit to be added to a follow-up PR, so as to keep this PR somewhat small. That said, I'm happy to add it to this PR if it makes things easier - just let me know

zerosicx (approved):
Logically it makes sense for me, but I don't understand most of this stuff at a low level :'D

packages/babel-plugin-strip-runtime/src/__tests__/strip-runtime-source-code.test.ts
```diff
@@ -250,6 +251,53 @@ describe('babel-plugin-strip-runtime using source code', () => {
 
     describe('with jsx pragma', () => {
       it('work with classic jsx pragma', () => {
```
itsdouges, line 253: nit: can we be specific on the behaviour we're observing? What does it work mean?

packages/babel-plugin-strip-runtime/src/__tests__/strip-runtime-source-code.test.ts
```diff
@@ -250,6 +251,53 @@ describe('babel-plugin-strip-runtime using source code', () => {
...
+          );
+        `;
+
+        const actual = transform(codeWithPragma, {
+          run: 'both',
+          runtime: 'classic',
+        });
+
+        expect(actual).toMatchInlineSnapshot(`
+          "/* app.tsx generated by @compiled/babel-plugin v0.0.0 */
```
itsdouges, line 277: And then for the snapshot. Is the assertion that the jsx pragma is removed and replaced by React? Can we assert that?

packages/babel-plugin/src/babel-plugin.ts
```diff
@@ -223,6 +288,22 @@ export default declare<State>((api) => {
         path: NodePath<t.TaggedTemplateExpression> | NodePath<t.CallExpression>,
         state: State
       ) {
+        if (isJsx(path, state)) {
+          throw buildCodeFrameError(
+            `Found a \`jsx\` function call in the Babel output where one should not have been generated. Was Compiled was not set up correctly?
```
itsdouges, line 293:
```suggestion
            `Found a \`jsx\` function call in the Babel output where one should not have been generated. Was Compiled not set up correctly?
```

packages/babel-plugin/src/babel-plugin.ts
```diff
@@ -223,6 +288,22 @@ export default declare<State>((api) => {
...
+
+Reasons this might happen:
+* Importing \`jsx\` from a library other than Compiled CSS-in-JS - please only import from \`@compiled/react\`.
+* Both \`runtime: classic\` and \`pragma: 'jsx'\` were specified in your Babel configuration - please use the /** @jsx jsx */ syntax instead, or switch to \`runtime: automatic\`.
+
+See https://compiledcssinjs.com/docs/installation for how to set up the Compiled Babel plugin.`,
```
itsdouges, line 299: nit: Should we mention anything around Webpack / Parcel?

packages/babel-plugin/src/types.ts
```diff
@@ -131,6 +131,9 @@ export interface State extends PluginPass {
   pragma: {
     jsx?: boolean;
     jsxImportSource?: boolean;
+
```
itsdouges, line 134:
```suggestion
```

packages/babel-plugin/src/utils/is-jsx-function.ts
```diff
@@ -0,0 +1,23 @@
+import type { NodePath } from '@babel/traverse';
+import * as t from '@babel/types';
+
+import type { State } from '../types';
+
+export const isJsx = (
```
itsdouges, line 6: Ah so this is looking for the transformed jsx function? Can we call the function that?

```suggestion
export const isTransformedJsxFunction = (
```

packages/babel-plugin/src/babel-plugin.ts
dddlr: oops ty for the catch, I probably should revise my english grammar 😆

packages/babel-plugin-strip-runtime/src/__tests__/strip-runtime-source-code.test.ts
dddlr: ah hmm, it's more of a general check that `@compiled/babel-plugin`, `@compiled/babel-plugin-strip-runtime`, and the JSX-related plugins in [@babel/preset-react](https://babeljs.io/docs/babel-preset-react) all work together to create output that isn't invalid i suppose 🤔 

that said, i notice some things that i do want to specifically check for (e.g. jsx pragma is converted correctly), i'll see if i can make the snapshots more specific

incidentally, i've moved these JSX pragma-related tests into their own file `packages/babel-plugin-strip-runtime/src/__tests__/jsx-pragma.test.ts`, might as well include that in this PR since i'll be editing the tests anyway 😅

packages/babel-plugin/src/babel-plugin.ts
dddlr: hmm probs not necessary?

`@babel/plugin-transform-react-jsx` is the plugin where one would specify `runtime: classic` and `pragma: 'jsx'` (and potentially run into this error), and

* I think with parcel you would need a `babel.config.json` to specify the above config options (not 100% sure on this, best to fact-check haha)
* with webpack you can include the config options within your `webpack.config.js` (e.g. [this person's setup](https://stackoverflow.com/q/52565118)) but it's probably clear enough what "Babel configuration" would refer to here

packages/babel-plugin-strip-runtime/src/__tests__/extract-styles.test.ts
```diff
@@ -0,0 +1,155 @@
...
+      `;
+
+      it('adds styles to directory', () => {
+        const actual = transform(code, {
+          run: 'both',
+          runtime,
+          extractStylesToDirectory: { source: 'src/', dest: 'dist/' },
+        });
+
+        expect(actual).toMatchInlineSnapshot(`
```
itsdouges, line 25: Is the main assertion the existence of the CSS import? Do we want to assert any specific CSS output? Could they regress?

packages/babel-plugin-strip-runtime/src/__tests__/jsx-pragma.test.ts
```diff
@@ -0,0 +1,370 @@
...
+          // All components should look like this:
+          //     const Component = () => /*#__PURE__*/ React.createElement(...)
+          expect(actual).toMatch(
+            /Component = \(\) =>\n(\s)*\/\*#__PURE__\*\/ React.createElement/m
+          );
+          expect(actual).toMatch(
+            /Component2 = \(\) =>\n(\s)*\/\*#__PURE__\*\/ React.createElement/m
+          );
+          // All traces of JSX pragma or JSX imports are removed
+          expect(actual).not.toContain('jsx');
```
itsdouges, line 68: Nice

packages/babel-plugin-strip-runtime/src/__tests__/jsx-pragma.test.ts
```diff
@@ -0,0 +1,370 @@
...
+            </div>
+          );
+        `;
+
+        const actual = transform(code, {
+          run: 'both',
+          runtime: 'classic',
+        });
+
+        expect(actual).toMatchInlineSnapshot(`
```
itsdouges, line 97: Instead of a snapshot can we do a exact match?

```suggestion
        expect(actual).toEqual(code);
```

packages/babel-plugin-strip-runtime/src/__tests__/jsx-pragma.test.ts
```diff
@@ -0,0 +1,370 @@
...
+                  color: 'pink',
+                }),
+              },
+              'hello world 2'
+            );
+          "
+        `);
+      });
+
+      it('does not process file (alt.)', () => {
```
itsdouges, line 126:
```suggestion
      it('does not process file with remapped jsx pragma', () => {
```

packages/babel-plugin-strip-runtime/src/__tests__/jsx-pragma.test.ts
```diff
@@ -0,0 +1,370 @@
...
+            </div>
+          );
+        `;
+
+        const actual = transform(code, {
+          run: 'both',
+          runtime: 'classic',
+        });
+
+        expect(actual).toMatchInlineSnapshot(`
```
itsdouges, line 150:
```suggestion
        expect(actual).toEqual(code);
```

packages/babel-plugin-strip-runtime/src/__tests__/jsx-pragma.test.ts
```diff
@@ -0,0 +1,370 @@
...
+        });
+
+        expect(actual).toContain("import { jsx as _jsx } from 'react/jsx-runtime';");
+        // jsx function calls from React have "PURE" comments
+        // added beforehand from @babel/preset-react. This helps
+        // with tree-shaking by Webpack or Parcel
+        expect(actual).toContain('/*#__PURE__*/ _jsx');
+
+        expect(actual).not.toContain(
+          "import { jsxs as _jsxs } from '@compiled/react/jsx-runtime';"
```
itsdouges, line 292: We should also check for jsx — perhaps we just check the jsx-runtime module isn't in scope?

```suggestion
          "@compiled/react/jsx-runtime"
```

packages/babel-plugin-strip-runtime/src/__tests__/jsx-pragma.test.ts
```diff
@@ -0,0 +1,370 @@
...
+
+        const actual = transform(code, {
+          run: 'both',
+          runtime: 'automatic',
+        });
+
+        expect(actual).toContain("import { jsx as _jsx } from '@emotion/react/jsx-runtime';");
+
+        expect(actual).not.toContain(
+          "import { jsxs as _jsxs } from '@compiled/react/jsx-runtime';"
```
itsdouges, line 325:
```suggestion
          "@compiled/react/jsx-runtime"
```

packages/babel-plugin-strip-runtime/src/__tests__/jsx-pragma.test.ts
```diff
@@ -0,0 +1,370 @@
...
+        );
+        expect(actual).not.toContain("import { jsx as _jsx } from '@compiled/react/jsx-runtime';");
+
+        // If any of the below statements are present, that means we accidentally
+        // got rid of the /** @jsxImportSource @emotion/react */ and the default React
+        // JSX runtime is being used instead of Emotion -- whoops!
+        expect(actual).not.toContain("import { jsxs as _jsxs } from 'react/jsx-runtime';");
```
itsdouges, line 332: We could just check for the module as formatting / exports could change over time.

packages/babel-plugin-strip-runtime/src/__tests__/jsx-pragma.test.ts
```diff
@@ -0,0 +1,370 @@
...
+        );
+      `;
+
+      const actual = transform(code, {
+        run: 'both',
+        runtime: 'automatic',
+        babelJSXImportSource: '@compiled/react',
+      });
+
+      expect(actual).toContain("import { jsxs as _jsxs } from '@compiled/react/jsx-runtime';");
```
itsdouges, line 363: As above. Could just check for module existence.

packages/babel-plugin/src/babel-plugin.ts
```diff
@@ -23,17 +24,42 @@ import {
...
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const packageJson = require('../package.json');
 const JSX_SOURCE_ANNOTATION_REGEX = /\*?\s*@jsxImportSource\s+([^\s]+)/;
-const JSX_ANNOTATION_REGEX = /\*?\s*@jsx\s+([^\s]+)/;
 const DEFAULT_IMPORT_SOURCE = '@compiled/react';
 
 let globalCache: Cache | undefined;
 
+const FindClassicJsxPragmaImport: Visitor<State> = {
```
itsdouges, line 38: nit:

```suggestion
const findClassicJsxPragmaImport: Visitor<State> = {
```

packages/babel-plugin/src/babel-plugin.ts
```diff
@@ -23,17 +24,42 @@ import {
...
+  ImportSpecifier(path, state) {
+    const specifier = path.node;
+
+    t.assertImportDeclaration(path.parent);
```
itsdouges, line 42: Does this throw if it isn't an import declaration?

packages/babel-plugin/src/babel-plugin.ts
itsdouges: I guess the point I'm making is most people wouldn't use the Babel plugin directly right, but through a bundler plugin?

itsdouges (commented):
Looking good mate! Here's my review.

packages/babel-plugin/src/babel-plugin.ts
dddlr: hmm babel plugin handbook sometimes capitalises visitors

https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md#toc-paths-in-visitors

I guess there's no consistency, even in our repository haha

packages/babel-plugin/src/babel-plugin.ts
dddlr: Screenshot 2023-12-29 at 14 44 38

yeah, provides both type checking and a runtime check

packages/babel-plugin-strip-runtime/src/__tests__/extract-styles.test.ts
dddlr: this test was added in https://github.com/atlassian-labs/compiled/pull/1513/files#diff-a3198cfeec7834168f1ae1a2e61d0f416af2b7167289b727aa4a260d93f59579 i think

it's testing both

* `app.compiled.css` import was added
* css styling no longer exists in any form in the babel output

it doesn't look like the contents of `app.compiled.css` is tested (@liamqma can you confirm?), maybe i can add it?

packages/babel-plugin-strip-runtime/src/__tests__/jsx-pragma.test.ts
dddlr: `actual` isn't equal to `code` though 🤔

packages/babel-plugin-strip-runtime/src/__tests__/jsx-pragma.test.ts
dddlr: ah good idea, makes things less verbose

packages/babel-plugin-strip-runtime/src/__tests__/jsx-pragma.test.ts
itsdouges: Does this mean it is being transformed? 😅

packages/babel-plugin-strip-runtime/src/__tests__/extract-styles.test.ts
dddlr: this test was added in https://github.com/atlassian-labs/compiled/pull/1513/files#diff-a3198cfeec7834168f1ae1a2e61d0f416af2b7167289b727aa4a260d93f59579 i think

it's testing both

app.compiled.css import was added
css styling no longer exists in any form in the babel output
it doesn't look like the contents of app.compiled.css is tested (@liamqma can you confirm?), maybe i can add it?

packages/babel-plugin-strip-runtime/src/__tests__/jsx-pragma.test.ts
dddlr: yes but not by Compiled, but rather by `@babel/preset-react`

i'll update the wording of the test name

packages/babel-plugin/src/babel-plugin.ts
dddlr: updated message for clarity 👍 let me know what you think

packages/babel-plugin-strip-runtime/src/__tests__/jsx-pragma.test.ts
dddlr: addressed by https://github.com/atlassian-labs/compiled/pull/1603#discussion_r1437904318

itsdouges (approved):
LGTM

dddlr:
just in time for new years 😄

PR #1604, opened 2023-12-22
Fix nested types output from createStrictAPI() being unintentionally marked as required

itsdouges:
This pull request updates types and adds some tests for this specific use case.

### DIFF
--- a/packages/babel-plugin/src/css-map/index.ts
+++ b/packages/babel-plugin/src/css-map/index.ts
@@ -98,7 +98,7 @@ export const visitCssMapPath = (
         const { sheets, classNames } = transformCssItems(css, meta);
         totalSheets.push(...sheets);
 
-        if (classNames.length !== 1) {
+        if (classNames.length > 1) {
           throw buildCodeFrameError(
             createErrorMessage(ErrorMessages.STATIC_VARIANT_OBJECT),
             property,
@@ -106,7 +106,7 @@ export const visitCssMapPath = (
           );
         }
 
-        return t.objectProperty(property.key, classNames[0]);
+        return t.objectProperty(property.key, classNames[0] || t.stringLiteral(''));
       })
     )
   );
--- a/packages/react/src/create-strict-api/__tests__/generics.test.tsx
+++ b/packages/react/src/create-strict-api/__tests__/generics.test.tsx
@@ -5,6 +5,58 @@ import type { XCSSProp } from './__fixtures__/strict-api-recursive';
 import { css, cssMap } from './__fixtures__/strict-api-recursive';
 
 describe('createStrictAPI()', () => {
+  it('should mark all styles as optional in css()', () => {
+    const styles = css({
+      '&:hover': {},
+      '&:active': {},
+      '&::before': {},
+      '&::after': {},
+    });
+
+    const { getByTestId } = render(<div css={styles} data-testid="div" />);
+
+    expect(getByTestId('div')).toBeDefined();
+  });
+
+  it('should mark all styles as optional in cssMap()', () => {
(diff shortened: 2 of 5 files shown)


packages/react/src/create-strict-api/__tests__/generics.test.tsx
```diff
@@ -5,6 +5,58 @@ import type { XCSSProp } from './__fixtures__/strict-api-recursive';
 import { css, cssMap } from './__fixtures__/strict-api-recursive';
 
 describe('createStrictAPI()', () => {
+  it('should mark all styles as optional in css()', () => {
+    const styles = css({
+      '&:hover': {},
```
itsdouges, line 10: This was a type violation previously.

packages/react/src/create-strict-api/index.ts
```diff
@@ -6,11 +6,11 @@ type PseudosDeclarations = {
...
 };
 
-type EnforceSchema<TObject> = {
-  [P in keyof TObject]?: P extends keyof CompiledSchema
-    ? TObject[P] extends Record<string, unknown>
-      ? EnforceSchema<TObject[P]>
-      : TObject[P]
+type EnforceSchema<TSchema> = {
+  [P in keyof TSchema]?: P extends keyof CompiledSchema
+    ? TSchema[P] extends Record<string, any>
```
itsdouges, line 11: Main fix: `unknown` -> `any`

kylorhall-atlassian: Interesting, do you know why this works? I've seen similar things, just never seen like a TS issue or explanation.

packages/babel-plugin/src/css-map/index.ts
```diff
@@ -98,15 +98,15 @@ export const visitCssMapPath = (
         const { sheets, classNames } = transformCssItems(css, meta);
         totalSheets.push(...sheets);
 
-        if (classNames.length !== 1) {
+        if (classNames.length > 1) {
```
itsdouges, line 101: This allows us to make empty styles now (at least for testing).

packages/babel-plugin/src/css-map/index.ts
```diff
@@ -98,15 +98,15 @@ export const visitCssMapPath = (
...
           throw buildCodeFrameError(
             createErrorMessage(ErrorMessages.STATIC_VARIANT_OBJECT),
             property,
             meta.parentPath
           );
         }
 
-        return t.objectProperty(property.key, classNames[0]);
+        return t.objectProperty(property.key, classNames[0] || t.stringLiteral(''));
```
kylorhall-atlassian, line 109: Got it, so having `'&:hover'` empty threw this?

kylorhall-atlassian (approved):
Seems straightforward from my understanding, and I know there's tests for `'&:hover': { … }` having valid css in them, so LGTM.

packages/babel-plugin/src/css-map/index.ts
itsdouges: Yeah exactly.

itsdouges: > Interesting, do you know why this works? I've seen similar things, just never seen like a TS issue or explanation.

I think it's an edge case with complex nested types (interface + omits + ?). At a glace `unknown` should be fine but in this case it's obviously not.

I'm not sure!

packages/react/src/create-strict-api/index.ts
itsdouges: I think it's an edge case with complex nested types (interface + omits + ?). At a glace unknown should be fine but in this case it's obviously not.

I'm not sure!

PR #1608, opened 2023-12-28
Option to increase specificity of generated styles to enable XCSSProp migration

itsdouges:
This pull request introduces a new option to Compiled for increasing the specificity of all styles generated by Compiled, without changing the class name. This enables Compiled to remain deterministic when you have styles built from two or more sources that may or may not have increased specificity turned on.

We're going to be using this for migration to XCSSProp where the platform components haven't yet been migrated to Compiled (away from Emotion). We need to ensure that styles passed through XCSSProp win over Emotion styles in a deterministic way. This was chosen over other solutions such as prepending Emotion styles to the head instead of appending as the unknown of legacy styles in Jira and beyond being unintentionally affected.

To enable set `increaseSpecificity` to true in your Compiled config:

```diff
+increaseSpecificity: true
```

---

**TODO**

- [x] Investigate VR test failure
- [x] Edge case tests

### DIFF
--- a/.storybook/.babelrc.json
+++ b/.storybook/.babelrc.json
@@ -5,6 +5,7 @@
       {
         "importReact": false,
         "optimizeCss": false,
+        "increaseSpecificity": true,
         "transformerBabelPlugins": [
           [
             "@babel/plugin-proposal-decorators",
--- a/packages/css/src/__tests__/transform.test.ts
+++ b/packages/css/src/__tests__/transform.test.ts
@@ -452,27 +452,48 @@ describe('#css-transform', () => {
     expect(actual.classNames).toEqual(expected.classNames);
   });
 
-  it('should add extra specificity to declarations', () => {
-    const styles = `
-      padding: 8px;
-      color: red;
-      :before {
-        content: var(--hello-world);
-        margin-right: 8px;
-        color: pink;
-      }
-    `;
-    const { sheets: actual } = transformCss(styles, { increaseSpecificity: true });
+  describe('increased specificity', () => {
+    it('should add extra specificity to declarations', () => {
+      const styles = `
+        padding: 8px;
+        color: red;
+        :before {
+          content: var(--hello-world);
+          margin-right: 8px;
+          color: pink;
+        }
+        ::after {
+          color: red;
+        }
+      `;
+      const { sheets: actual } = transformCss(styles, { increaseSpecificity: true });
 
-    expect(actual.join('\n')).toMatchInlineSnapshot(`
-      "._ca0qftgi:not(#\\9){padding-top:8px}
-      ._u5f3ftgi:not(#\\9){padding-right:8px}
-      ._n3tdftgi:not(#\\9){padding-bottom:8px}
-      ._19bvftgi:not(#\\9){padding-left:8px}
-      ._syaz5scu:not(#\\9){color:red}
-      ._1kt9o5oc:not(#\\9):before{content:var(--hello-world)}
-      ._eid3ftgi:not(#\\9):before{margin-right:8px}
-      ._is0632ev:not(#\\9):before{color:pink}"
-    `);
+      expect(actual.join('\n')).toMatchInlineSnapshot(`
+        "._ca0qftgi:not(#\\9){padding-top:8px}
+        ._u5f3ftgi:not(#\\9){padding-right:8px}
+        ._n3tdftgi:not(#\\9){padding-bottom:8px}
+        ._19bvftgi:not(#\\9){padding-left:8px}
+        ._syaz5scu:not(#\\9){color:red}
+        ._1kt9o5oc:not(#\\9):before{content:var(--hello-world)}
+        ._eid3ftgi:not(#\\9):before{margin-right:8px}
+        ._is0632ev:not(#\\9):before{color:pink}
+        ._14rn5scu:not(#\\9):after{color:red}"
+      `);
+    });
+
+    it('should increase & selector specificity', () => {
+      const styles = `
+        div & { color: red; }
+        div:hover & { color: red; }
+        div &:hover { color: red; }
+      `;
+      const { sheets: actual } = transformCss(styles, { increaseSpecificity: true });
+
+      expect(actual.join('\n')).toMatchInlineSnapshot(`
+        "div ._kqan5scu:not(#\\9){color:red}
+        div:hover ._12hc5scu:not(#\\9){color:red}
+        div ._wntz5scu:not(#\\9):hover{color:red}"
+      `);
+    });
   });
 });
--- a/packages/css/src/__tests__/transform.test.ts
+++ b/packages/css/src/__tests__/transform.test.ts
@@ -1,8 +1,4 @@
-import { transformCss as transform } from '../transform';
-
-interface TransformOpts {
-  optimizeCss?: boolean;
-}
+import { transformCss as transform, type TransformOpts } from '../transform';
 
 const transformCss = (code: string, opts: TransformOpts = { optimizeCss: false }) =>
   transform(code, opts);
@@ -438,5 +434,18 @@ describe('#css-transform', () => {
         ._otyridpf{margin-bottom:0}"
       `);
     });
+  });
+
+  it('should add extra specificity after atomicizing without affecting class names', () => {
+    const actual = transformCss(`margin-left: 0;`, { increaseSpecificity: true });
+    const expected = transformCss(`margin-left: 0;`, { increaseSpecificity: false });
+
+    expect(actual.classNames).toEqual(expected.classNames);
+  });
+
+  it('should add extra specificity to declarations', () => {
+    const { sheets: actual } = transformCss(`margin-left: 0;`, { increaseSpecificity: true });
+
+    expect(actual.join()).toMatchInlineSnapshot(`"._18u0idpf:not(#\\9){margin-left:0}"`);
   });
 });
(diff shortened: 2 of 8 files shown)


packages/css/src/plugins/atomicify-rules.ts
```diff
@@ -3,11 +3,8 @@ import type { Plugin, ChildNode, Declaration, Container, Rule, AtRule } from 'po
 import { rule } from 'postcss';
 
 interface PluginOpts {
-  classNameCompressionMap?: { [index: string]: string };
+  classNameCompressionMap?: Record<string, string>;
```
itsdouges, line 6: Fixed this type up while I was here

packages/css/src/plugins/atomicify-rules.ts
```diff
@@ -3,11 +3,8 @@ import type { Plugin, ChildNode, Declaration, Container, Rule, AtRule } from 'po
...
   callback?: (className: string) => void;
-}
-
-interface AtomicifyOpts extends PluginOpts {
```
itsdouges, line 10: Merged this into PluginOpts

packages/css/src/plugins/atomicify-rules.ts
```diff
@@ -264,7 +261,7 @@ const atomicifyAtRule = (node: AtRule, opts: AtomicifyOpts): AtRule => {
  *
  * 1. No nested rules allowed - normalize them with the `parent-orphaned-pseudos` and `nested` plugins first.
  */
-export const atomicifyRules = (opts = {}): Plugin => {
+export const atomicifyRules = (opts: PluginOpts = {}): Plugin => {
```
itsdouges, line 264: Declared this type so it's actually... typed!

packages/css/src/plugins/increase-specificity.ts
```diff
@@ -0,0 +1,31 @@
...
+ * This means generated class names with / without the increased specificity are the same,
+ * so when running Compiled together with (in product) + without it (in platform) they will
+ * keep the deterministic behaviour of the last declared style wins, but enjoying increased
+ * specificity when used in non-Compiled contexts for migration purposes, e.g. xcss prop
+ * being passed to Emotion CSS-in-JS!
+ */
+export const increaseSpecificity = (): Plugin => {
+  return {
+    postcssPlugin: 'increase-specificity',
+    OnceExit(root) {
```
itsdouges, line 19: Because we want this to run after the atomic plugin we need to use the same hook - OnceExit!

packages/css/src/__tests__/transform.test.ts
```diff
@@ -1,8 +1,4 @@
-import { transformCss as transform } from '../transform';
-
-interface TransformOpts {
```
itsdouges, line 3: Removed this type and brought in the exported one instead.

.storybook/.babelrc.json
```diff
@@ -5,6 +5,7 @@
       {
         "importReact": false,
         "optimizeCss": false,
+        "increaseSpecificity": true,
```
itsdouges, line 8: Turn on the new feature so we can test the storybook remains unaffected.

packages/css/src/plugins/__tests__/increase-specificity.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const transform = (css: TemplateStringsArray) => {
+  const result = postcss([increaseSpecificity()]).process(css[0], {
+    from: undefined,
+  });
+
+  return result.css;
+};
+
+describe('increase specicifity plugin', () => {
```
kylorhall-atlassian, line 13: Worth a test for some pseudos, eg. this is expected?
```css
@media (…) {
  .abc123:hover:not(#\9) {
    color: red;
  }
}
```

What about keyframes, supports, other at-rules?  No conerns?

itsdouges: I'll add some tests to cover them (specifically: they shouldn't directly be affected, only the class names!).

kylorhall-atlassian: 86bd8eb5d7c548a0a4a8d5630bcc246a8eb0e9ac was there an issue 🤔 I figured this could go first.

How about a case where `selector = '.abc123 button'` or something nested?  Anything like that?

itsdouges: Yeah there was! Before/after weren't applied if the `has:` was at the end. I'll do more edge cases exploration tests.

packages/css/src/__tests__/transform.test.ts
```diff
@@ -439,4 +435,44 @@ describe('#css-transform', () => {
       `);
     });
   });
+
+  it('should add extra specificity after atomicizing without affecting class names', () => {
+    const styles = `
+      padding: 8px;
+      color: red;
+      :before {
```
dddlr, line 443: would it be worthwhile to check selectors with `&` at the front as well? (e.g. `&::before`)

also nit: `/:before/::before/g`

packages/css/src/plugins/__tests__/increase-specificity.test.ts
dddlr: since kylor already mentioned nesting (`.abc123 { button { ... } }`), I'll also mention that some tests for funky cases involving `&` would be good too

examples:

```css
div & { /* ... */ }
div:hover & { /* ... */ }
div &:hover { /* ... */ }
```

packages/css/src/plugins/increase-specificity.ts
dddlr: TIL OnceExit, will be handy next time I do some babel plugin stuff :)

packages/css/src/plugins/increase-specificity.ts
```diff
@@ -0,0 +1,39 @@
+import type { Plugin } from 'postcss';
+
+const INCREASE_SPECIFICITY_SELECTOR = ':not(#\\9)';
```
dddlr, line 3: I understand this is inspired by [stylex's PR earlier today](https://github.com/facebook/stylex/pull/271), but is there any docs about this kind of CSS syntax? I still don't really understand how this is a valid selector 😆 

Also, why `\9` specifically (which I assume is U+0009?) and not a different escape character?

dddlr:
LGTM assuming that above comments are addressed 👍

packages/css/src/plugins/increase-specificity.ts
itsdouges: This is postcss ;-)

packages/css/src/plugins/__tests__/increase-specificity.test.ts
itsdouges: All of those will still have a class name prefixing it post atomic-ification. Will add them though! Push soon.

packages/css/src/plugins/increase-specificity.ts
itsdouges: Inspired by this old postcss plugin for the selector https://github.com/MadLittleMods/postcss-increase-specificity/tree/master and then stylex was just another signal that isn't not the worst idea ;-).

\9 is a tab character, pretty much just a grab bag of whatever works https://github.com/MadLittleMods/postcss-increase-specificity/tree/master

itsdouges:
Okay I've pushed up an improved implementation LMK what folks think.

packages/css/src/__tests__/transform.test.ts
```diff
@@ -439,4 +435,65 @@ describe('#css-transform', () => {
...
+      expect(actual.join('\n')).toMatchInlineSnapshot(`
+        "._ca0qftgi:not(#\\9){padding-top:8px}
+        ._u5f3ftgi:not(#\\9){padding-right:8px}
+        ._n3tdftgi:not(#\\9){padding-bottom:8px}
+        ._19bvftgi:not(#\\9){padding-left:8px}
+        ._syaz5scu:not(#\\9){color:red}
+        ._1kt9o5oc:not(#\\9):before{content:var(--hello-world)}
+        ._eid3ftgi:not(#\\9):before{margin-right:8px}
+        ._is0632ev:not(#\\9):before{color:pink}
+        ._14rn5scu:not(#\\9):after{color:red}"
```
kylorhall-atlassian, line 480: I mean technically `::after` isn't `&::after` now that's properly valid modern CSS—but that's a CSS vs. Compiled spec issue, not your change, right?

packages/css/src/plugins/__tests__/increase-specificity.test.ts
```diff
@@ -0,0 +1,79 @@
...
+            @media {
+              .foo:not(#\\9) {}
+            }
+          "
+    `);
+  });
+
+  it('should ignore root elements', () => {
+    const actual = transform`
+      html {}
```
kylorhall-atlassian, line 44: What would happen if I just had this—and what's expected to happen?  Will this not increase specificity?
```tsx
css({
  button: { color: 'red' },
});
```

This is the same as the `& button` (which you could fit into the `div` test above), right?

For root/html/body or "global css" this makes sense, I guess, but for nested selectors…?  I'm less certain…

itsdouges:
Reminder for myself: make sure this doesn't affect classes that aren't defined by Compiled, just in case.

packages/css/src/__tests__/transform.test.ts
itsdouges: Correct, in this case it's lingering behaviour remnant of other styling libraries.

packages/css/src/plugins/__tests__/increase-specificity.test.ts
itsdouges: It will. As it increases the specificity of the compiled class name.

itsdouges: Yes same as div!

kylorhall-atlassian (approved):

PR #1612, opened 2024-01-05
Fix Compiled throwing when multiple API imports exist in the same module

itsdouges:
With the introduction of the strict API and custom importSources Compiled can now be used from different imports in the same module. One use case is when migrating to the new strict API but being blocked on something, so having to fallback to the current top-level API.

To ensure a smooth ride we need to enable Compiled to work when multiple imports exist in a module. This pull request changes the internal data structure to capture the local names of all Compiled APIs.

**TODO**

- [x] Update css state to check for multiple modules

### DIFF
--- a/packages/babel-plugin/src/styled/index.ts
+++ b/packages/babel-plugin/src/styled/index.ts
@@ -35,7 +35,7 @@ const extractStyledDataFromTemplateLiteral = (
   if (
     t.isMemberExpression(node.tag) &&
     t.isIdentifier(node.tag.object) &&
-    node.tag.object.name === meta.state.compiledImports?.styled &&
+    meta.state.compiledImports?.styled?.includes(node.tag.object.name) &&
     t.isIdentifier(node.tag.property)
   ) {
     const tagName = node.tag.property.name;
@@ -47,7 +47,7 @@ const extractStyledDataFromTemplateLiteral = (
   if (
     t.isCallExpression(node.tag) &&
     t.isIdentifier(node.tag.callee) &&
-    node.tag.callee.name === meta.state.compiledImports?.styled &&
+    meta.state.compiledImports?.styled?.includes(node.tag.callee.name) &&
     t.isIdentifier(node.tag.arguments[0])
   ) {
     const tagName = node.tag.arguments[0].name;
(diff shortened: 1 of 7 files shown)


packages/babel-plugin/src/types.ts
```diff
@@ -116,11 +116,11 @@ export interface State extends PluginPass {
...
-    styled?: string;
-    cssMap?: string;
+    ClassNames?: string[];
+    css?: string[];
+    keyframes?: string[];
+    styled?: string[];
+    cssMap?: string[];
   };
 
   usesXcss?: boolean;
```
itsdouges, line 126: There is also `importedCompiledImports` — @pancaspe87 you're last to touch that line of code... what does it do? It doesn't have a jsdoc description. Should we give it the same treatment or it's something else?

packages/babel-plugin/src/styled/index.ts
```diff
@@ -35,7 +35,7 @@ const extractStyledDataFromTemplateLiteral = (
   if (
     t.isMemberExpression(node.tag) &&
     t.isIdentifier(node.tag.object) &&
-    node.tag.object.name === meta.state.compiledImports?.styled &&
+    meta.state.compiledImports?.styled?.includes(node.tag.object.name) &&
```
pancaspe87, line 38: out of curiosity, why did you change all of these? is the latter easier to read?

packages/babel-plugin/src/types.ts
pancaspe87: don't remember, give me some time to dig into it

packages/babel-plugin/src/styled/index.ts
itsdouges: No, because they're arrays now instead of strings 😃

pancaspe87: riiiiight, I saw the types now 😅

packages/babel-plugin/src/types.ts
pancaspe87: Answer is possibly [here](https://github.com/atlassian-labs/compiled/pull/1321/files#diff-5b63ee465fe228fb576db7b039cc976de17575ad2fe33c5172fb1e2239818aa1R7)

pancaspe87: In terms of giving the same treatment, I am not really sure. I think I introduced the change to address [this issue](https://github.com/atlassian-labs/compiled/issues/1316) - which seems specific to imported mixins

itsdouges: So if the imported mixins can also use APIs from another module, then... probably should be given the same treatment 😄.

I'll fix it up and add context around the internal state.

itsdouges: After digging through the code it doesn't actually check for a specific import specifier. I'll leave it for now.

dddlr (approved):
LGTM, thank you for your patience :pray:

PR #1613, opened 2024-01-05
Add missing typecheck command

itsdouges:
This pull request adds a typecheck command and adds it to the CI run. It also fixes a ton of violations. One thing I haven't reconciled yet is how to typecheck the project references with the same command, any ideas let me know! Perhaps we just exclude packages and assume they're typechecked when generating the types?...

### DIFF
--- a/packages/tsconfig.options.json
+++ b/packages/tsconfig.options.json
@@ -16,6 +16,7 @@
     "resolveJsonModule": true,
     "sourceMap": true,
     "strict": true,
+    "skipLibCheck": true,
     "target": "es6",
     "typeRoots": ["../node_modules/@types", "../types"]
   }
--- a/stories/css-prop-static-object.tsx
+++ b/stories/css-prop-static-object.tsx
@@ -1,4 +1,4 @@
-import '@compiled/react';
+/** @jsxImportSource @compiled/react */
 import { hoverObjectLiteral } from './mixins';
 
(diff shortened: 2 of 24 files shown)


packages/tsconfig.options.json
```diff
@@ -16,6 +16,7 @@
     "resolveJsonModule": true,
     "sourceMap": true,
     "strict": true,
+    "skipLibCheck": true,
```
pancaspe87, line 19: What does this do?

stories/css-prop-static-object.tsx
```diff
@@ -23,7 +23,6 @@ export const ObjectLiteralSpreadFromFunc = (): JSX.Element => {
       css={{
         display: 'flex',
         fontSize: '50px',
-        color: 'blue',
```
pancaspe87, line 26: Why are we removing this?

packages/tsconfig.options.json
itsdouges: Uh was just testing, I should remove it.

stories/css-prop-static-object.tsx
itsdouges: It's a type error :-) "color is defined multiple times and this won't be used"

dddlr:
Ooh thanks for the fix

LGTM assuming that `skipLibCheck` is removed and you get green builds

PR #1622, opened 2024-02-05
Add `importSources` option to jsx-pragma rule

dddlr:
Currently, the `jsx-pragma` rule only supports detecting and handling `@compiled/react` to ensure the JSX pragma is there.

However, with the `createStrictAPI` from contributions by @kylorhall-atlassian and @itsdouges, we might have Compiled APIs being imported by other libraries (e.g. `@atlaskit/css`). We add an `importSources` option to specify what these additional libraries are. This option affects two other options:

* What is counted as a "Compiled import" in the `detectConflictWithOtherLibraries` option, where we detect whether Compiled imports conflict with Emotion imports
* What is counted as a "Compiled import" in the `onlyRunIfImportingCompiled` option, where we only run this rule if a Compiled import is used in the file.

### DIFF
--- a/.changeset/purple-squids-switch.md
+++ b/.changeset/purple-squids-switch.md
@@ -2,4 +2,4 @@
 '@compiled/eslint-plugin': patch
 ---
 
-Add `onlyRunIfImportingCompiled` option to `jsx-pragma` rule, to specify additional libraries that should be considered Compiled imports
+Add `importSources` option to `jsx-pragma` rule, to specify additional libraries that should be considered Compiled imports
--- a/.changeset/purple-squids-switch.md
+++ b/.changeset/purple-squids-switch.md
@@ -1,0 +1,5 @@
+---
+'@compiled/eslint-plugin': patch
+---
+
+Add `onlyRunIfImportingCompiled` option to `jsx-pragma` rule, to specify additional libraries that should be considered Compiled imports
--- a/packages/eslint-plugin/src/rules/jsx-pragma/__tests__/rule.test.ts
+++ b/packages/eslint-plugin/src/rules/jsx-pragma/__tests__/rule.test.ts
@@ -232,7 +232,7 @@ import { css } from '@compiled/react';
         // check the raw message, not the messageId, to ensure that this
         // says "jsxImportSource pragma" and not "jsx pragma"
         {
-          message: 'To use the `css` prop you must set the jsx pragma.',
+          message: 'To use the `css` prop you must set the jsxImportSource pragma.',
         },
       ],
     },
--- a/packages/eslint-plugin/src/rules/jsx-pragma/__tests__/rule.test.ts
+++ b/packages/eslint-plugin/src/rules/jsx-pragma/__tests__/rule.test.ts
@@ -105,6 +105,16 @@ tester.run('jsx-pragma', jsxPragmaRule, {
       `,
       options: [{ onlyRunIfImportingCompiled: true }],
     },
+    {
+      name: 'when importSources is non-empty, onlyRunIfImportingCompiled should automatically be set to true',
+      // If we get an error here, then onlyRunIfImportingCompiled was not set to true.
+      code: `
+        /** @jsx jsx */
+        import { css, jsx } from '@emotion/react';
+        <div css={css({ display: 'block' })} />
+      `,
+      options: [{ importSources: ['@atlaskit/css'] }],
+    },
   ],
   invalid: [
     {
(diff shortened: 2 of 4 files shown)


packages/eslint-plugin/src/rules/jsx-pragma/__tests__/rule.test.ts
```diff
@@ -300,30 +332,7 @@ import { useState } from 'react';
       ],
     },
     {
-      code: `
```
dddlr, line 303: deleted this test case as it appears to be a duplicate of the previous one (albeit with different spacing that doesn't actually impact eslint's AST parsing)

packages/eslint-plugin/src/rules/jsx-pragma/__tests__/rule.test.ts
```diff
@@ -173,14 +197,15 @@ import { jsx } from '@compiled/react';
       options: [{ runtime: 'classic' }],
       errors: [
         {
-          message: 'To use the `css` prop you must set the jsx pragma.',
```
dddlr, line 176: not sure why `message` was used here at all when the rest of the test cases were using `messageId`

packages/eslint-plugin/src/rules/jsx-pragma/__tests__/rule.test.ts
```diff
@@ -3,32 +3,49 @@ import { jsxPragmaRule } from '../index';
...
-    `
-      /** @jsxImportSource @compiled/react */
-      <div className={xcss} />
-    `,
-    `
-      /** @jsxImportSource @compiled/react */
-      <div className={innerXcss} />
-    `,
     {
+      name: 'should be valid if automatic jsx pragma is used',
```
dddlr, line 7: `name` requires eslint 8.10+

packages/eslint-plugin/src/rules/jsx-pragma/__tests__/rule.test.ts
itsdouges: Was there a variable it was asserting in the message

packages/eslint-plugin/src/rules/jsx-pragma/index.ts
```diff
@@ -144,14 +154,20 @@ export const jsxPragmaRule: Rule.RuleModule = {
...
+      alsoAddCompiledPragmaFor: optionsRaw.alsoAddCompiledPragmaFor ?? [],
     };
 
     const source = context.getSourceCode();
     const comments = source.getAllComments();
 
     const compiledImports = findLibraryImportDeclarations(context);
+    const otherCompiledImports = findLibraryImportDeclarations(
+      context,
+      options.alsoAddCompiledPragmaFor
```
itsdouges, line 169: Can we call this `importSources`? Then it's aligned with other API usages in the Compiled compiler.

packages/eslint-plugin/src/rules/jsx-pragma/__tests__/rule.test.ts
dddlr: ahh you're right, i had missed that

dddlr: i'll add some comments clarifying this

packages/eslint-plugin/src/rules/jsx-pragma/index.ts
```diff
@@ -144,14 +154,20 @@ export const jsxPragmaRule: Rule.RuleModule = {
     const optionsRaw = context.options[0] || {};
     const options: Options = {
       detectConflictWithOtherLibraries: optionsRaw.detectConflictWithOtherLibraries ?? true,
-      onlyRunIfImportingCompiled: optionsRaw.onlyRunIfImportingCompiled ?? false,
+      onlyRunIfImportingCompiled:
+        optionsRaw.onlyRunIfImportingCompiled ?? optionsRaw.importSources ?? false,
```
kylorhall-atlassian, line 158: Should this at all be checking length, do you get into a weird place where `onlyRunIfImportingCompiled = […]` and it should be a boolean?  Probably doesn't matter, just I think this would fail type-checking?

eg.
```suggestion
        optionsRaw.onlyRunIfImportingCompiled ?? !!optionsRaw.importSources?.length || false,
```

.changeset/purple-squids-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@compiled/eslint-plugin': patch
+---
+
+Add `onlyRunIfImportingCompiled` option to `jsx-pragma` rule, to specify additional libraries that should be considered Compiled imports
```
kylorhall-atlassian, line 5: Should this also mention `importSources`?

kylorhall-atlassian (commented):
One brief thought, the change looks fine to me, I just don't know if anything else might be missing.

packages/eslint-plugin/src/rules/jsx-pragma/index.ts
dddlr: ah that's a good point, `optionsRaw.importSources` was typed as `any` (it's an eslint thing 😅 ) so it didn't get caught

it doesn't affect the execution of the rule to my knowledge (except for the edge case where `importSources = []`), but your suggestion does reduce confusion

.changeset/purple-squids-switch.md
dddlr: whoops, good catch

JakeLane (approved):

PR #1566, opened 2023-11-24
[SPIKE] Optional types API

itsdouges:
TBD

PR #1623, opened 2024-02-07
Add importSources to @compiled/webpack-loader

dddlr:
Follow-up to https://github.com/atlassian-labs/compiled/pull/1566

Make `importSources` option available through `@compiled/webpack-loader` configuration as well.

### DIFF
--- a/packages/webpack-loader/src/compiled-loader.ts
+++ b/packages/webpack-loader/src/compiled-loader.ts
@@ -34,6 +34,7 @@ function getLoaderOptions(context: LoaderContext<CompiledLoaderOptions>) {
     classNameCompressionMap = undefined,
     extractStylesToDirectory = undefined,
     resolver = undefined,
+    importSources = undefined,
   }: CompiledLoaderOptions = typeof context.getOptions === 'undefined'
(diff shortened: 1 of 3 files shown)


packages/webpack-loader/src/compiled-loader.ts
```diff
@@ -105,6 +109,7 @@ function getLoaderOptions(context: LoaderContext<CompiledLoaderOptions>) {
     classNameCompressionMap,
     extractStylesToDirectory,
     resolver,
+    importSources,
```
kylorhall-atlassian, line 112: I don't know anything about this, I presume this whole object `getLoaderOptions` gets passed into Babel?

JakeLane (approved):

packages/webpack-loader/src/compiled-loader.ts
dddlr: yep! `@compiled/webpack-loader` is acts as a kind of wrapper around `@compiled/babel-plugin` and `@compiled/babel-plugin-strip-runtime`

kylorhall-atlassian (approved):

PR #1630, opened 2024-02-16
Fix Compiled props being incorrectly forwarded when default values used

### LINKED ISSUE #1629: Properties fall-through to wrapped DOM element when there are default values in the property declaration
**Describe the bug**

If a component is declared as

```
const Box = styled.div({
  marginTop: ({ marginTop }) => marginTop,
});
```

When used it ends-up having a `--variable-for-margin-top: ...` value set to the property value, assuming it is used like `<Box marginTop={...} />`.

When the property parameters object destructuring expression has a property default however, such as:

```
const Box = styled.div({
  marginTop: ({ marginTop = 0 }) => marginTop,
});
```

When used it ends-up having a `--variable-for-margin-top: ...` value set to the property value, but **also having the `margintop=...` attribute set on the DOM element**. This leads to react warnings on unexpected attribute values.

dddlr:
Fixes #1629 

https://github.com/atlassian-labs/compiled/blob/9860df3850ac5023ffa93201f5648fd7c81807c5/packages/babel-plugin/src/utils/build-styled-component.ts#L80 expects a MemberExpression in order to detect what props shouldn't be forwarded to React. All other props are forwarded to React. (see this PR for more context: https://github.com/atlassian-labs/compiled/pull/1273)

However, when we handle default values in situations like these...

```
const styled.div({
    margin: ({ marginTop = 12 }) => marginTop,
});
```

`@compiled/babel-plugin` used the `t.identifier` function to generate the `__cmplp.marginTop` expression:

```
styled.div<{
  marginTop?: string;
}>({
  marginTop: __cmplp => __cmplp.marginTop ?? '24px'
})
```

The above appears correct, because the build output is identical and Babel doesn't stop us from passing invalid expressions to the `t.identifier` function... however, the use of `t.identifier` and `t.memberExpression` meant that it would never get detected as a MemberExpression, and thus the prop would never _not_ be forwarded to React :(

This  PR fixes this by replacing `t.identifier` with `t.memberExpression`.

### DIFF
--- a/packages/babel-plugin/src/utils/normalize-props-usage.ts
+++ b/packages/babel-plugin/src/utils/normalize-props-usage.ts
@@ -116,17 +116,20 @@ const normalizeDestructuredString = (
  * @returns A member expression.
  */
 const createNestedMemberExpression = (objectChain: string[]): t.Identifier | t.MemberExpression => {
-  if (objectChain.length === 1) return t.identifier(objectChain[0]);
-  if (objectChain.length > 1) {
-    const [init, last] = [
-      objectChain.slice(0, objectChain.length - 1),
-      objectChain[objectChain.length - 1],
-    ];
-    return t.memberExpression(createNestedMemberExpression(init), t.identifier(last));
+  if (objectChain.length === 0) {
+    throw new Error(
+      'Could not build a Compiled component, due to objectChain being empty when generating a member expression. This is likely a bug with Compiled - please file a bug report.'
+    );
   }
-  throw new Error(
-    'Could not build a Compiled component, due to objectChain being empty when generating a member expression. This is likely a bug with Compiled - please file a bug report.'
-  );
+  if (objectChain.length === 1) {
+    return t.identifier(objectChain[0]);
+  }
+
+  const [initial, last] = [
+    objectChain.slice(0, objectChain.length - 1),
+    objectChain[objectChain.length - 1],
+  ];
+  return t.memberExpression(createNestedMemberExpression(initial), t.identifier(last));
 };
 
 const normalizeDestructuredObject = (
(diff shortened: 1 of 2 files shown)


packages/babel-plugin/src/utils/normalize-props-usage.ts
```diff
@@ -108,6 +108,27 @@ const normalizeDestructuredString = (
...
+ * Create a member expression from a list of strings. For example, if
+ * `objectChain = ['a', 'b', 'c']`, the generated member expression will be
+ * `a.b.c`.
+ * @param objectChain List of strings
+ * @returns A member expression.
+ */
+const createNestedMemberExpression = (objectChain: string[]): t.Identifier | t.MemberExpression => {
+  if (objectChain.length === 1) return t.identifier(objectChain[0]);
+  if (objectChain.length > 1) {
+    const [init, last] = [
```
dddlr, line 121: is there a better name for all the items of a list except the last one? i looked it up and init (from haskell) was the only one that i could easily find

kylorhall-atlassian: In lodash it's `initial` as well IIRC; no clue why, but that feels the FP naming..

packages/babel-plugin/src/utils/normalize-props-usage.ts
```diff
@@ -108,6 +108,27 @@ const normalizeDestructuredString = (
...
+      objectChain.slice(0, objectChain.length - 1),
+      objectChain[objectChain.length - 1],
+    ];
+    return t.memberExpression(createNestedMemberExpression(init), t.identifier(last));
+  }
+  throw new Error(
```
JakeLane, line 127: Bit of a preference thing but it's a bit confusing to know when the error will throw. It might be explicit to throw an exception if the length === 0

JakeLane (dismissed):

packages/babel-plugin/src/utils/normalize-props-usage.ts
dddlr: no worries, happy to add it for clarity

JakeLane (approved):

PR #1631, opened 2024-02-29
Fix source map behaviour in Parcel and move node_modules compiled extraction to new parcel transformer

JakeLane:
The current behaviour for the compiled parcel transformer does not handle source maps correctly. This PR resolves these issues by using the source maps correctly and introducing a new parcel transformer to handle external compiled deps correctly.


marcins (commented):
I've gone through this PR and conceptually it seems fine to me - there's a fair bit of change, and I don't have a huge amount of context, so it wasn't necessarily easily to mentally parse the changes.

The only comment I had when looking through was whether there are is any opportunity to pass existing sourcemaps into some of the Babel methods (`generate` or `transform`) that might save some of the manual merging / manipulation after, but you likely already considered that and there were reasons it wouldn't work.

packages/parcel-transformer-external/src/__tests__/transformer.parceltest.ts
```diff
@@ -0,0 +1,89 @@
...
+
+const workerFarm = createWorkerFarm();
+
+afterAll(() => {
+  workerFarm.end();
+});
+
+const outputFS = new MemoryFS(workerFarm);
+
+const getParcelInstance = (workingDir: string, sourceMap = false) => {
```
dddlr, line 17: i don't have enough context on the Parcel stuff to properly review it beyond "seems fine to me", but i'll do what i can to review the parts that are Compiled related

packages/parcel-transformer/src/__tests__/transformer.parceltest.ts
```diff
@@ -15,15 +18,35 @@ afterAll(() => {
...
 
-const getParcelInstance = (workingDir: string) => {
+function findTargetSourcePosition(source: string, regex: RegExp): { line: number; column: number } {
+  const lines = source.split('\n');
+
+  for (let i = 0; i < lines.length; i++) {
+    const line = lines[i];
+    const match = line.match(regex);
+
+    if (match && match.index != null) {
```
dddlr, line 30: is there a reason loose equality is used here and not strict equality?

dddlr:
>  introducing a new parcel transformer to handle external compiled deps correctly

i'm still reading through and parsing through the PR 😅 but what was the reason behind taking some of the functionality of `@compiled/parcel-transformer` and putting them into a new package `@compiled/parcel-transformer-external`? is it to avoid bogging down `@compiled/parcel-transformer` with too much complexity?

package.json
```diff
@@ -45,7 +45,7 @@
     "test": "jest --no-cache",
     "test:cover": "yarn test --collectCoverage",
     "test:imports": "node test/test-imports",
-    "test:parcel": "jest --testMatch '**/src/**/*.parceltest.{ts,tsx}'",
+    "test:parcel": "jest --testMatch '**/src/**/*.parceltest.{ts,tsx}' --testEnvironment=node",
```
dddlr, line 48: does it not work without this change?

packages/parcel-transformer/src/index.ts
```diff
@@ -148,13 +129,16 @@ export default new Transformer<ParcelTransformerOpts>({
     const includedFiles: string[] = [];
     const code = asset.isASTDirty() ? undefined : await asset.getCode();
 
+    const shouldUseSourceMaps = asset.env.sourceMap != null;
```
dddlr, line 132: wouldn't `!!shouldUseSourceMaps` be clearer?

JakeLane:
> I've gone through this PR and conceptually it seems fine to me - there's a fair bit of change, and I don't have a huge amount of context, so it wasn't necessarily easily to mentally parse the changes.
> 
> The only comment I had when looking through was whether there are is any opportunity to pass existing sourcemaps into some of the Babel methods (`generate` or `transform`) that might save some of the manual merging / manipulation after, but you likely already considered that and there were reasons it wouldn't work.

The main goal for the new transformer is to avoid running the compiled babel plugin for external node modules when not required. Bringing in babel context would be a concern for that goal.

I think we can reuse the AST in other babel contexts. I think that's a perf win on the table but I didn't want to make this change more complicated than it already is. I'll create a ticket on the backlog to action when we can.

packages/parcel-transformer/src/__tests__/transformer.parceltest.ts
JakeLane: These are borrowed from a different implementation. I'll refactor to remove the sketchy null comparison

package.json
JakeLane: yep this is required

dddlr (approved):
looks good!

dddlr (approved):

liamqma (approved):

PR #1634, opened 2024-03-04
Add media query typing to XCSS prop / createStrictAPI

itsdouges:
This pull request introduces media query support to XCSS prop, and in turn changes some internals to make it function. Read on for the individual changes to each API.

---

**cssMap changes**

The CSS map API now allows defining top level media queries. Previously you had to define them inside a `@media` object, this restriction has now been removed bringing it inline with the CSS function API.

```jsx

const styles = cssMap({
  topLevel: { '@media (min-width: 30rem)': { color: 'green' } },
  nested: { '@media': { '(min-width: 30rem)': { color: 'red' } } },
});
```

Both APIs are still supported however only the top level definition will pass type checking with XCSS prop. In the future we'll remove support for the nested object.

For the strict API the default media query type is `@media ${never}`. Meaning, by default, no media queries are available. Yo can expand it to specific string values passing in a string literal union as the second generic arg.

> **Note** — Loose media queries can't be passed to strict XCSS prop! This is because we can't ensure the type safety of that object.

**XCSS prop changes**

The XCSS prop now takes top level media queries. Nested media queries is not allowed.

```jsx
import { cssMap, css } from '@compiled/react';

const styles = cssMap({
  valid: { '@media (min-width: 30rem)': { color: 'green' } },
  invalid: { '@media': { '(min-width: 30rem)': { color: 'red' } } },
});

<Component xcss={styles.valid} />;
```

**createStrictAPI changes**

Now takes an optional second generic to define what media queries are supported:

```diff
createStrictAPI<
  { color: 'var(--text)' }
+  { media: '(min-width: 30rem)' | '(min-width: 48rem)' }
>();
```

Which is then flushed to all output APIs.

### DIFF
--- a/packages/babel-plugin/src/utils/css-map.ts
+++ b/packages/babel-plugin/src/utils/css-map.ts
@@ -17,8 +17,18 @@ export const getKeyValue = (key: ObjectKeyWithLiteralValue): string => {
   throw new Error(`Expected an identifier or a string literal, got type ${(key as any).type}`);
 };
 
-export const isAtRule = (key: ObjectKeyWithLiteralValue): key is ObjectKeyWithLiteralValue =>
-  getKeyValue(key).startsWith('@');
+export const isAtRuleObject = (
+  key: ObjectKeyWithLiteralValue
+): key is ObjectKeyWithLiteralValue => {
+  const keyValue = getKeyValue(key);
+  if (keyValue.includes(' ')) {
+    // Anything that includes a space character won't have an at rule "object".
+    // So we can skip it here.
+    return false;
+  }
+
+  return keyValue.startsWith('@');
+};
 
 export const isPlainSelector = (selector: string): boolean => selector.startsWith(':');
 
--- a/packages/babel-plugin/src/utils/css-map.ts
+++ b/packages/babel-plugin/src/utils/css-map.ts
@@ -1,9 +1,29 @@
 import * as t from '@babel/types';
+import type { AtRules } from 'csstype';
 
 import type { Metadata } from '../types';
 import { buildCodeFrameError } from '../utils/ast';
 
 export const EXTENDED_SELECTORS_KEY = 'selectors';
+
+const atRules: Record<AtRules, boolean> = {
+  '@charset': true,
+  '@counter-style': true,
+  '@document': true,
+  '@font-face': true,
+  '@font-feature-values': true,
+  '@font-palette-values': true,
+  '@import': true,
+  '@keyframes': true,
+  '@layer': true,
+  '@media': true,
+  '@namespace': true,
+  '@page': true,
+  '@property': true,
+  '@scroll-timeline': true,
+  '@supports': true,
+  '@viewport': true,
+};
 
 type ObjectKeyWithLiteralValue = t.Identifier | t.StringLiteral;
 
@@ -21,13 +41,7 @@ export const isAtRuleObject = (
   key: ObjectKeyWithLiteralValue
 ): key is ObjectKeyWithLiteralValue => {
   const keyValue = getKeyValue(key);
-  if (keyValue.includes(' ')) {
-    // Anything that includes a space character won't have an at rule "object".
-    // So we can skip it here.
-    return false;
-  }
-
-  return keyValue.startsWith('@');
+  return keyValue in atRules;
 };
 
 export const isPlainSelector = (selector: string): boolean => selector.startsWith(':');
--- a/packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api.ts
+++ b/packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api.ts
@@ -19,6 +19,6 @@ type MediaQuery =
   | '(prefers-color-scheme: dark)'
   | '(prefers-color-scheme: light)';
 
-const { css, XCSSProp, cssMap, cx } = createStrictAPI<CSSPropertiesSchema, MediaQuery>();
+const { css, XCSSProp, cssMap, cx } = createStrictAPI<CSSPropertiesSchema, { media: MediaQuery }>();
 
 export { css, XCSSProp, cssMap, cx };
--- a/packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api.ts
+++ b/packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api.ts
@@ -10,6 +10,15 @@ interface CSSPropertiesSchema {
   bkgrnd: 'red' | 'green';
 }
 
-const { css, XCSSProp, cssMap, cx } = createStrictAPI<CSSPropertiesSchema>();
+type MediaQuery =
+  | '(min-width: 30rem)'
+  | '(min-width: 48rem)'
+  | '(min-width: 64rem)'
+  | '(min-width: 90rem)'
+  | '(min-width: 110rem)'
+  | '(prefers-color-scheme: dark)'
+  | '(prefers-color-scheme: light)';
+
+const { css, XCSSProp, cssMap, cx } = createStrictAPI<CSSPropertiesSchema, MediaQuery>();
 
 export { css, XCSSProp, cssMap, cx };
--- a/packages/react/src/create-strict-api/index.ts
+++ b/packages/react/src/create-strict-api/index.ts
@@ -145,7 +145,8 @@ export interface CompiledAPI<
     TAllowedPseudos,
     TAllowedMediaQueries,
     TSchema,
-    TRequiredProperties
+    TRequiredProperties,
+    'strict'
   >;
 }
 
--- a/packages/react/src/create-strict-api/index.ts
+++ b/packages/react/src/create-strict-api/index.ts
@@ -4,6 +4,10 @@ import { type CompiledStyles, cx, type Internal$XCSSProp } from '../xcss-prop';
 
 import type { AllowedStyles, ApplySchema, ApplySchemaMap, CompiledSchemaShape } from './types';
 
+export interface StrictOptions {
+  media: string;
+}
+
 export interface CompiledAPI<
   TSchema extends CompiledSchemaShape,
   TAllowedMediaQueries extends string
@@ -164,17 +168,19 @@ export interface CompiledAPI<
  *
  * To set up:
  *
- * 1. Declare the API in a module (either local or in a package):
+ * 1. Declare the API in a module (either local or in a package), optionally declaring accepted media queries.
  *
  * @example
  * ```tsx
  * // ./foo.ts
- * const { css } = createStrictAPI<{
+ * interface Schema {
  *   color: 'var(--ds-text)',
  *   '&:hover': { color: 'var(--ds-text-hover)' }
- * }>();
+ * }
  *
- * export { css };
+ * const { css, cssMap, XCSSProp, cx } = createStrictAPI<Schema, { media: '(min-width: 30rem)' }>();
+ *
+ * export { css, cssMap, XCSSProp, cx };
  * ```
  *
  * 2. Configure Compiled to pick up this module:
@@ -200,8 +206,8 @@ export interface CompiledAPI<
  */
 export function createStrictAPI<
   TSchema extends CompiledSchemaShape,
-  TAllowedMediaQueries extends string = never
->(): CompiledAPI<TSchema, TAllowedMediaQueries> {
+  TCreateStrictAPIOptions extends StrictOptions = never
+>(): CompiledAPI<TSchema, TCreateStrictAPIOptions['media']> {
   return {
     css() {
       throw createStrictSetupError();
--- a/packages/react/src/create-strict-api/index.ts
+++ b/packages/react/src/create-strict-api/index.ts
@@ -199,7 +199,7 @@ export interface CompiledAPI<
  */
 export function createStrictAPI<
   TSchema extends CompiledSchemaShape,
-  TAllowedMediaQueries extends string = string
+  TAllowedMediaQueries extends string = never
 >(): CompiledAPI<TSchema, TAllowedMediaQueries> {
   return {
     css() {
--- a/packages/react/src/create-strict-api/index.ts
+++ b/packages/react/src/create-strict-api/index.ts
@@ -4,7 +4,10 @@ import { type CompiledStyles, cx, type Internal$XCSSProp } from '../xcss-prop';
 
 import type { AllowedStyles, ApplySchema, ApplySchemaMap, CompiledSchemaShape } from './types';
 
-export interface CompiledAPI<TSchema extends CompiledSchemaShape> {
+export interface CompiledAPI<
+  TSchema extends CompiledSchemaShape,
+  TAllowedMediaQueries extends string
+> {
   /**
    * ## CSS
    *
@@ -23,7 +26,7 @@ export interface CompiledAPI<TSchema extends CompiledSchemaShape> {
    * ```
    */
   css<TStyles extends ApplySchema<TStyles, TSchema>>(
-    styles: AllowedStyles & TStyles
+    styles: AllowedStyles<TAllowedMediaQueries> & TStyles
     // NOTE: This return type is deliberately not using ReadOnly<CompiledStyles<TStyles>>
     // So it type errors when used with XCSS prop. When we update the compiler to work with
     // it we can update the return type so it stops being a type violation.
@@ -45,10 +48,10 @@ export interface CompiledAPI<TSchema extends CompiledSchemaShape> {
    * ```
    */
   cssMap<
-    TObject extends Record<string, AllowedStyles>,
+    TObject extends Record<string, AllowedStyles<TAllowedMediaQueries>>,
     TStylesMap extends ApplySchemaMap<TObject, TSchema>
   >(
-    styles: Record<string, AllowedStyles> & TStylesMap
+    styles: Record<string, AllowedStyles<TAllowedMediaQueries>> & TStylesMap
   ): {
     readonly [P in keyof TStylesMap]: CompiledStyles<TStylesMap[P]>;
   };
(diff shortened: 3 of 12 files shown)


packages/react/src/xcss-prop/index.ts
```diff
@@ -27,19 +25,29 @@ type XCSSPseudos<
...
  */
 type BlockedRules = {
   selectors?: never;
-} & {
-  /**
-   * We currently block all at rules from xcss prop.
-   * This needs us to decide on what the final API is across Compiled to be able to set.
-   */
-  [Q in CSS.AtRules]?: never;
+  '@media'?: never;
```
dddlr, line 50: do we want to block other at rules?

itsdouges: Maybe. What's your thoughts?

itsdouges: Hi it's me again: Yes we should.

dddlr: yeah e.g. we want to make sure people don't start using `@keyframes` cuz we have a specialised `keyframes` function for that, and we don't want people using `@import` and `@font-face` in unexpected places... so a whitelist system for at-rules would make sense

packages/babel-plugin/src/utils/css-map.ts
```diff
@@ -17,8 +17,18 @@ export const getKeyValue = (key: ObjectKeyWithLiteralValue): string => {
...
 };
 
-export const isAtRule = (key: ObjectKeyWithLiteralValue): key is ObjectKeyWithLiteralValue =>
-  getKeyValue(key).startsWith('@');
+export const isAtRuleObject = (
+  key: ObjectKeyWithLiteralValue
+): key is ObjectKeyWithLiteralValue => {
+  const keyValue = getKeyValue(key);
+  if (keyValue.includes(' ')) {
+    // Anything that includes a space character won't have an at rule "object".
```
dddlr, line 25: not quite sure i'm understanding this correctly... what is the definition for an "at rule object" here?

`'@media (max-width: 256px)': { ... }` has a few space characters inside the media query - we would want it to be true for that right?

itsdouges: Maybe the internal code could use a refactor, but basically for cssMap anything that starts with an `@` is assumed to be a nested at rule declaration, which then has its styles flattened prior to being resolved.

```
const styles = cssMap({
  nested: { '@media': { '(min-width: 30rem)': { color: 'red' } } },
});
```

This is no longer true however as we want top level declarations to be valid.

```
const styles = cssMap({
  topLevel: { '@media (min-width: 30rem)': { color: 'green' } },
});
```

An 'easy' way to make that happen is: if it has a space in it it's not something we're interested in flattening. We could flip it and only flatten it if it's a specific value (like at media, at supports, and so on).

packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api.ts
```diff
@@ -1,13 +1,25 @@
...
+  {
+    '&:hover': {
+      color: 'var(--ds-text-hover)';
+      background: 'var(--ds-surface-hover)' | 'var(--ds-surface-sunken-hover)';
+    };
+    color: 'var(--ds-text)';
+    background: 'var(--ds-surface)' | 'var(--ds-surface-sunken)';
+    bkgrnd: 'red' | 'green';
+  },
+  MediaQuery
```
itsdouges, line 22: The new "optional" generic

packages/babel-plugin/src/utils/css-map.ts
kylorhall-atlassian: Unfortunately I'd say `@media(…)`, eg. `@supports(display:flex)` are valid at rules without spaces.

Perhaps `return /^@[a-z]+$/i.test(keyValue);`

itsdouges: Yes but that wouldn't be supported by cssMap API (before and after this PR).

packages/react/src/create-strict-api/index.ts
```diff
@@ -146,7 +154,13 @@ interface CompiledAPI<TSchema extends CompiledSchema> {
       requiredProperties: TAllowedProperties;
       requiredPseudos: TAllowedPseudos;
     } = never
-  >(): Internal$XCSSProp<TAllowedProperties, TAllowedPseudos, TSchema, TRequiredProperties>;
+  >(): Internal$XCSSProp<
+    TAllowedProperties,
+    TAllowedPseudos,
+    TMediaQueries,
```
kylorhall-atlassian, line 160: Did we want to support any other at-rules, eg. `@supports`?  Can easily wait for an ask—easy to slot in later if necessary, nothing here really blocks it (aside from the one type).

packages/react/src/create-strict-api/index.ts
```diff
@@ -146,7 +154,13 @@ interface CompiledAPI<TSchema extends CompiledSchema> {
       requiredProperties: TAllowedProperties;
       requiredPseudos: TAllowedPseudos;
     } = never
-  >(): Internal$XCSSProp<TAllowedProperties, TAllowedPseudos, TSchema, TRequiredProperties>;
+  >(): Internal$XCSSProp<
```
kylorhall-atlassian, line 157: Would have to touch `StrictXCSSProp` 👀

packages/react/src/xcss-prop/__tests__/media-query-strict.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+  },
+});
+
+describe('xcss prop', () => {
+  it('should allow valid media queries in inline xcss prop', () => {
+    const { getByText } = render(
+      <CSSPropComponent
+        xcss={{
+          '@media (min-width: 110rem)': {
+            // TODO: This should be a type error as color should be one of the base styles.
```
kylorhall-atlassian, line 38: TODO 👀 in the future?

packages/react/src/xcss-prop/index.ts
```diff
@@ -27,19 +27,31 @@ type XCSSPseudos<
     : never;
 };
 
+type XCSSMediaQuery<
+  TAllowedProperties extends keyof CSSProperties,
+  TAllowedPseudos extends CSSPseudos,
+  TAllowedMediaQueries extends string,
+  TRequiredProperties extends { requiredProperties: TAllowedProperties },
```
kylorhall-atlassian, line 34: Do you want a `requiredAtRules`?  (joking, but could actually be realistic one day!)

packages/react/src/xcss-prop/index.ts
```diff
@@ -27,19 +27,31 @@ type XCSSPseudos<
...
 type BlockedRules = {
   selectors?: never;
+  // We also block all type level at rule "objects" that are present on cssMap.
 } & {
-  /**
-   * We currently block all at rules from xcss prop.
-   * This needs us to decide on what the final API is across Compiled to be able to set.
-   */
-  [Q in CSS.AtRules]?: never;
+  [Q in AtRules]?: never;
```
kylorhall-atlassian, line 54: How does this work `XCSSMediaQuery & BlockedRules`? 🤔

packages/react/src/xcss-prop/index.ts
```diff
@@ -27,19 +27,31 @@ type XCSSPseudos<
...
 
+type XCSSMediaQuery<
+  TAllowedProperties extends keyof CSSProperties,
+  TAllowedPseudos extends CSSPseudos,
+  TAllowedMediaQueries extends string,
+  TRequiredProperties extends { requiredProperties: TAllowedProperties },
+  TSchema
+> = {
+  [Q in `@media ${TAllowedMediaQueries}`]?:
+    | MarkAsRequired<
```
kylorhall-atlassian, line 38: Given this is a union, will it actually be required?  Doing my head in!

Just more or less worth a test, eg. you had a TODO on a test, so perhaps there?

packages/react/src/create-strict-api/index.ts
itsdouges: I'm not sure. For XCSS prop I don't think so. For strict API probably, but it would be an internal change.

packages/react/src/create-strict-api/index.ts
itsdouges: Would you though?

packages/react/src/xcss-prop/__tests__/media-query-strict.test.tsx
itsdouges: TODO right now. I'm currently working on another PR to merge before this one that fixes up the types.

packages/react/src/xcss-prop/index.ts
itsdouges: Hehe. Yeah right now it's just assumed media queries are always available. And nothing else is (for XCSS prop at least).

packages/react/src/xcss-prop/index.ts
itsdouges: Last one wins.

packages/react/src/xcss-prop/index.ts
itsdouges: Which one are we talking about?

packages/react/src/create-strict-api/index.ts
kylorhall-atlassian: Nope, sorry, I didn't even look… `XCSSProp<TAllowedProperties, TAllowedPseudos, TRequiredProperties>` is untouched.

packages/react/src/xcss-prop/index.ts
kylorhall-atlassian: Do we require all properties in `MarkAsRequired` if you have `XCSSPsuedos`?

Let's say `requiredProperties: 'color'`, I think this would pass:
```tsx
css({
  color: 'red',
  '@media (whatever)': {
    '&:hover': { color: 'red' },
  },
});
```
_or_ this would pass:
```tsx
css({
  color: 'red',
  '@media (whatever)': {
    color: 'red',
  },
});
```

Could of course load the repo up to validate, sorry, I didn't do that…

itsdouges: Good q. I'll test when I swing back.

kylorhall-atlassian (changes requested):
I think you rebased in, nothing really new to review, I think tests were fine, but two requested changes from before (and/or just to review):
 - [[link](https://github.com/atlassian-labs/compiled/pull/1634/files#r1510582295)] `isAtRuleObject` logic
 - [[link](https://github.com/atlassian-labs/compiled/pull/1634/files#r1510593158)] Do `requiredProperties`, `requiredPseudos` work in nested `@media` rules—should they?  Maybe they shouldn't actually…that'd be weird to require deep property overrides on _every_ media rule…

itsdouges:
@kylorhall-atlassian check out the PR now thanks mate

packages/babel-plugin/src/utils/css-map.ts
```diff
@@ -17,8 +37,12 @@ export const getKeyValue = (key: ObjectKeyWithLiteralValue): string => {
   throw new Error(`Expected an identifier or a string literal, got type ${(key as any).type}`);
 };
 
-export const isAtRule = (key: ObjectKeyWithLiteralValue): key is ObjectKeyWithLiteralValue =>
-  getKeyValue(key).startsWith('@');
+export const isAtRuleObject = (
+  key: ObjectKeyWithLiteralValue
+): key is ObjectKeyWithLiteralValue => {
+  const keyValue = getKeyValue(key);
```
kylorhall-atlassian, line 43: So this value must exactly match as this is always explicitly like `cssMap({ primary: { '@media': { '(min-width: 400rem)': { … } } } });` right?  LGTM…

packages/react/src/xcss-prop/__tests__/media-query-strict.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+  return <div className={xcss}>foo</div>;
+}
+
+const styles = cssMap({
+  invalid: {
+    // @ts-expect-error — @media at rule object is not allowed in strict cssMap
+    '@media': {
+      screen: { color: 'red' },
+    },
+  },
```
kylorhall-atlassian, line 24: Can these (both) be split into two?
```suggestion
  invalidObject: {
    // @ts-expect-error — @media at rule object is not allowed in strict cssMap
    '@media': {
      screen: { color: 'red' },
    },
  },
  invalidQuery: {
    // @ts-expect-error — @media is not in our allowed types
    '@media (min-width: 100px)': {
      color: 'red',
    },
  },
```

packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api.ts
```diff
@@ -10,6 +10,15 @@ interface CSSPropertiesSchema {
   bkgrnd: 'red' | 'green';
 }
 
-const { css, XCSSProp, cssMap, cx } = createStrictAPI<CSSPropertiesSchema>();
+type MediaQuery =
```
kylorhall-atlassian, line 13: We don't really test this limits other queries, right?

kylorhall-atlassian (dismissed):
Ah, just had a comment on the test, we don't really check that our `MediaQuery` does _restrict_ anything (other than the `@media: { screen: { … } }` syntax.

packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api.ts
itsdouges: Good to add, thanks

packages/react/src/xcss-prop/__tests__/media-query-strict.test.tsx
itsdouges: Thanks

packages/babel-plugin/src/utils/css-map.ts
itsdouges: Yes exactly.

packages/react/src/create-strict-api/index.ts
```diff
@@ -188,7 +204,10 @@ export interface CompiledAPI<TSchema extends CompiledSchemaShape> {
  * <div css={styles} />
  * ```
  */
-export function createStrictAPI<TSchema extends CompiledSchemaShape>(): CompiledAPI<TSchema> {
+export function createStrictAPI<
+  TSchema extends CompiledSchemaShape,
+  TCreateStrictAPIOptions extends StrictOptions = never
```
itsdouges, line 209: This defaults to `never` else the union can't be effectively represented by TypeScript and it causes a type violation.

packages/react/src/xcss-prop/index.ts
```diff
@@ -30,18 +30,29 @@ type XCSSPseudo<
...
 /**
  * These APIs we don't want to allow to be passed through the `xcss` prop but we also
  * must declare them so the (lack-of a) excess property check doesn't bite us and allow
  * unexpected values through.
  */
-type BlockedRules = {
+type BlockedRules<TMode extends 'loose' | 'strict'> = {
+  // To ensure styles that aren't allowed through XCSS prop strict APIs we block any
+  // loose media queries from being passed through as we can't ensure they're correct.
+  '@media [loose]'?: TMode extends 'loose' ? any : never;
```
itsdouges, line 52: Remember this is for the input to xcss prop, so for loose we accept anything.

kylorhall-atlassian: You could use something like `@media not all` if you wanted it to be a bit more _semantic_ 🤣

packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api.ts
```diff
@@ -10,6 +10,15 @@ interface CSSPropertiesSchema {
...
+  | '(min-width: 30rem)'
+  | '(min-width: 48rem)'
+  | '(min-width: 64rem)'
+  | '(min-width: 90rem)'
+  | '(min-width: 110rem)'
+  | '(prefers-color-scheme: dark)'
+  | '(prefers-color-scheme: light)';
+
+const { css, XCSSProp, cssMap, cx } = createStrictAPI<CSSPropertiesSchema, { media: MediaQuery }>();
```
kylorhall-atlassian, line 22: Ah, I suppose we could restrict our pseudos like this too…I guess that's kinda the issue; we're restricting keys via a passed-around generic…while pseudos are never _extended._

packages/react/src/xcss-prop/__tests__/media-query-strict.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+  invalidMediaQuery: {
+    // @ts-expect-error — this specific @media is not in our allowed types
+    '@media (min-width: 100px)': {
+      color: 'red',
+    },
+  },
+  valid: {
```
kylorhall-atlassian, line 37: I renamed this to `validMediaQueryInvalidProperty` and added a _valid_ scenario, but there's no difference as `valid` and `validMediaQueryInvalidProperty` are both not erroring, eg. `validMediaQueryInvalidProperty` would error _here_, but not in the `xcss={styles.validMediaQueryInvalidProperty}`.

kylorhall-atlassian (approved):
Not perfect, but I think it will us to about 99%, only failing scenario I can see is that a valid media query with an invalid value does not fail when used in `xcss={styles.validMediaQueryInvalidProperty}`, but it does fail when referenced in `cssMap`.  The hardest part I think is it's pretty much impossible to decipher the errors 🤣 but at least there's an error!

packages/react/src/xcss-prop/index.ts
itsdouges: I didn't want it to be semantic haha, didn't want it to be valid!

packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api.ts
itsdouges: Yeah. I suppose one argument for restricting pseudos would be the unsafe "not my ownership" ones.

PR #1635, opened 2024-03-04
Fix CSS func types

itsdouges:
This adds basic intellisense back to the CSS function. It's not exhaustive, e.g. doesn't include pseudos / media. But it's a step forward.

### DIFF
--- a/packages/react/src/css/index.ts
+++ b/packages/react/src/css/index.ts
@@ -36,7 +36,9 @@ export default function css<TProps = unknown>(
   ...interpolations: CssFunction<TProps>[]
 ): CSSProps<TProps>;
 
-export default function css<T = unknown>(styles: CssObject<T> | CssObject<T>[]): CSSProps<T>;
+export default function css<T = unknown>(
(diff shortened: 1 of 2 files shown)


dddlr (approved):

packages/react/src/css/index.ts
```diff
@@ -36,7 +36,9 @@ export default function css<TProps = unknown>(
   ...interpolations: CssFunction<TProps>[]
 ): CSSProps<TProps>;
 
-export default function css<T = unknown>(styles: CssObject<T> | CssObject<T>[]): CSSProps<T>;
+export default function css<T = unknown>(
+  styles: CssObject<T> | CssObject<T>[] | CSSProps<T> | CSSProps<T>[]
```
kylorhall-atlassian, line 40: Will this allow something like this?  Or maybe it looks like it already allows it because `CSSProps<T> extends CssObject<T>` (effectively).
```tsx
const oneStyles = css({ color: 'red' });
const twoStyles = css(oneStyles);
```

itsdouges: Yeah 😬.

PR #1636, opened 2024-03-04
Refactor strict API types

itsdouges:
This pull request refactors the types in and around the `createStrictAPI()` to fix a few bugs and underlying architectural issues. To prevent scope creeping the PR I haven't implemented nested styles for the strict API (e.g. hover inside an after pseudo) which may cause for another refactor. We can tackle that at a later date.

This is to be merged before https://github.com/atlassian-labs/compiled/pull/1634 which then will build ontop of this one.

**Background**

Previously defining the schema came with a lot of redundant work. For every pseudo that you wanted to explicitly type you would have to define it, and then all of the base types again, like so:

```tsx
interface Schema {
  background: 'var(--bg)';
  color: 'var(--color)';
  '&:hover': {
    background: 'var(--bg)';
    color: 'var(--color-hovered)';
  }
}

createStrictAPI<Schema>();
```

If you missed a value / didn't type every pseudo it would fallback to the CSSProperties value (from `csstype`). This was mostly unexpected. So for example right now `:hover` has been typed, but nothing else has. So it wouldn't benefit from the schema types.

With this refactor it now always falls back to the top level types if not defined, meaning it's more expected and safe for everyone and you only need to type the values you want to explicitly support.

```diff
interface Schema {
  background: 'var(--bg)';
  color: 'var(--color)';
  '&:hover': {
-    background: 'var(--bg)';
    color: 'var(--color-hovered)';
  }
}

createStrictAPI<Schema>();
```

Dive into the pull request and check out my annotations for more information.

**Extra notes**

- I've raised this bug for some odd type behaviour noticed during this PR https://github.com/atlassian-labs/compiled/issues/1639

---

- [x] Test for type edge cases
- [x] Get a review for @kylorhall-atlassian
- [x] Resolve https://github.com/atlassian-labs/compiled/pull/1636#discussion_r1512014348
- [x] Resolve https://github.com/atlassian-labs/compiled/pull/1636#discussion_r1511969832

### DIFF
--- a/packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api-recursive.ts
+++ b/packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api-recursive.ts
@@ -1,3 +1,5 @@
+/* eslint-disable prefer-const */
+import type { CompiledAPI } from '../../index';
 import { createStrictAPI } from '../../index';
 
 type Color = 'var(--ds-text)';
@@ -29,6 +31,11 @@ interface StrictAPI extends Properties {
   '&:active': PressedProperties;
 }
 
-const { css, XCSSProp, cssMap, cx } = createStrictAPI<StrictAPI>();
+let css: CompiledAPI<StrictAPI>['css'];
+let cssMap: CompiledAPI<StrictAPI>['cssMap'];
+let cx: CompiledAPI<StrictAPI>['cx'];
+let XCSSProp: CompiledAPI<StrictAPI>['XCSSProp'];
+
+({ css, cssMap, cx, XCSSProp } = createStrictAPI<StrictAPI>());
 
 export { css, XCSSProp, cssMap, cx };
--- a/packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api-recursive.ts
+++ b/packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api-recursive.ts
@@ -1,5 +1,3 @@
-/* eslint-disable prefer-const */
-import type { CompiledAPI } from '../../index';
 import { createStrictAPI } from '../../index';
 
 type Color = 'var(--ds-text)';
@@ -31,11 +29,6 @@ interface StrictAPI extends Properties {
   '&:active': PressedProperties;
 }
 
-let css: CompiledAPI<StrictAPI>['css'];
-let cssMap: CompiledAPI<StrictAPI>['cssMap'];
-let cx: CompiledAPI<StrictAPI>['cx'];
-let XCSSProp: CompiledAPI<StrictAPI>['XCSSProp'];
-
-({ css, cssMap, cx, XCSSProp } = createStrictAPI<StrictAPI>());
+const { css, cssMap, cx, XCSSProp } = createStrictAPI<StrictAPI>();
 
 export { css, XCSSProp, cssMap, cx };
--- a/packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api-recursive.ts
+++ b/packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api-recursive.ts
@@ -24,11 +24,11 @@ interface PressedProperties {
   backgroundColor: BackgroundPressed;
 }
 
-interface StrictAPI extends Properties {
+interface CSSPropertiesSchema extends Properties {
   '&:hover': HoveredProperties;
   '&:active': PressedProperties;
 }
 
-const { css, cssMap, cx, XCSSProp } = createStrictAPI<StrictAPI>();
+const { css, cssMap, cx, XCSSProp } = createStrictAPI<CSSPropertiesSchema>();
 
 export { css, XCSSProp, cssMap, cx };
--- a/packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api-recursive.ts
+++ b/packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api-recursive.ts
@@ -6,18 +6,20 @@ type ColorPressed = 'var(--ds-text-pressed)';
 type Background = 'var(--ds-bold)' | 'var(--ds-success)';
 type BackgroundHovered = 'var(--ds-bold-hovered)' | 'var(--ds-success-hovered)';
 type BackgroundPressed = 'var(--ds-bold-pressed)' | 'var(--ds-success-pressed)';
+type Space = 'var(--ds-space-050)' | 'var(--ds-space-0)';
 
 interface Properties {
   color: Color;
   backgroundColor: Background;
+  padding: Space;
 }
 
-interface HoveredProperties extends Omit<Properties, 'backgroundColor' | 'color'> {
+interface HoveredProperties {
   color: ColorHovered;
   backgroundColor: BackgroundHovered;
 }
 
-interface PressedProperties extends Omit<Properties, 'backgroundColor' | 'color'> {
+interface PressedProperties {
   color: ColorPressed;
   backgroundColor: BackgroundPressed;
 }
@@ -25,8 +27,6 @@ interface PressedProperties extends Omit<Properties, 'backgroundColor' | 'color'
 interface StrictAPI extends Properties {
   '&:hover': HoveredProperties;
   '&:active': PressedProperties;
-  '&::before': Properties;
-  '&::after': Properties;
 }
 
 const { css, XCSSProp, cssMap, cx } = createStrictAPI<StrictAPI>();
--- a/packages/react/src/create-strict-api/__tests__/generics.test.tsx
+++ b/packages/react/src/create-strict-api/__tests__/generics.test.tsx
@@ -59,13 +59,6 @@ describe('createStrictAPI()', () => {
 
   describe('type violations', () => {
     it('should violate types for css()', () => {
-      css({
-        color: 'var(--ds-text)',
-        '&:hover': {
-          color: 'var(--ds-text-hovered)',
-        },
-      });
-
       const styles = css({
         // @ts-expect-error — Type '""' is not assignable to type ...
         color: '',
--- a/packages/react/src/create-strict-api/__tests__/generics.test.tsx
+++ b/packages/react/src/create-strict-api/__tests__/generics.test.tsx
@@ -162,15 +162,15 @@ describe('createStrictAPI()', () => {
             backgroundColor: '',
             '&:hover': {
               // @ts-expect-error — Type '""' is not assignable to type ...
-              color: '',
+              color: 'var(--ds-text)',
               // @ts-expect-error — Type '""' is not assignable to type ...
-              backgroundColor: '',
+              backgroundColor: 'var(--ds-success)',
             },
             '&:active': {
               // @ts-expect-error — Type '""' is not assignable to type ...
-              color: '',
+              color: 'var(--ds-text)',
               // @ts-expect-error — Type '""' is not assignable to type ...
-              backgroundColor: '',
+              backgroundColor: 'var(--ds-success)',
             },
             '&::before': {
               // @ts-expect-error — Type '""' is not assignable to type ...
--- a/packages/react/src/create-strict-api/__tests__/generics.test.tsx
+++ b/packages/react/src/create-strict-api/__tests__/generics.test.tsx
@@ -59,6 +59,13 @@ describe('createStrictAPI()', () => {
 
   describe('type violations', () => {
     it('should violate types for css()', () => {
+      css({
+        color: 'var(--ds-text)',
+        '&:hover': {
+          color: 'var(--ds-text-hovered)',
+        },
+      });
+
       const styles = css({
         // @ts-expect-error — Type '""' is not assignable to type ...
         color: '',
@@ -99,7 +106,7 @@ describe('createStrictAPI()', () => {
       const styles = cssMap({
         primary: {
           // @ts-expect-error — Type '""' is not assignable to type ...
-          color: 's',
+          color: '',
           // @ts-expect-error — Type '""' is not assignable to type ...
           backgroundColor: '',
           '&:hover': {
@@ -188,21 +195,31 @@ describe('createStrictAPI()', () => {
   describe('type success', () => {
     it('should pass type check for css()', () => {
       const styles = css({
+        // @ts-expect-error — should be a value from the schema
+        padding: '10px',
         color: 'var(--ds-text)',
         backgroundColor: 'var(--ds-bold)',
         '&:hover': {
+          // @ts-expect-error — should be a value from the schema
+          padding: '10px',
           color: 'var(--ds-text-hovered)',
           backgroundColor: 'var(--ds-bold-hovered)',
         },
         '&:active': {
+          // @ts-expect-error — should be a value from the schema
+          padding: '10px',
           color: 'var(--ds-text-pressed)',
           backgroundColor: 'var(--ds-bold-pressed)',
         },
         '&::before': {
+          // @ts-expect-error — should be a value from the schema
+          padding: '10px',
           color: 'var(--ds-text)',
           backgroundColor: 'var(--ds-bold)',
         },
         '&::after': {
+          // @ts-expect-error — should be a value from the schema
+          padding: '10px',
           color: 'var(--ds-text)',
           backgroundColor: 'var(--ds-bold)',
         },
@@ -218,19 +235,30 @@ describe('createStrictAPI()', () => {
         primary: {
           color: 'var(--ds-text)',
           backgroundColor: 'var(--ds-bold)',
+          // @ts-expect-error — should be a value from the schema
+          padding: '10px',
           '&:hover': {
+            accentColor: 'red',
+            // @ts-expect-error — should be a value from the schema
+            padding: '10px',
             color: 'var(--ds-text-hovered)',
             backgroundColor: 'var(--ds-bold-hovered)',
           },
           '&:active': {
+            // @ts-expect-error — should be a value from the schema
+            padding: '10px',
             color: 'var(--ds-text-pressed)',
             backgroundColor: 'var(--ds-bold-pressed)',
           },
           '&::before': {
+            // @ts-expect-error — should be a value from the schema
+            padding: '10px',
             color: 'var(--ds-text)',
             backgroundColor: 'var(--ds-bold)',
           },
           '&::after': {
+            // @ts-expect-error — should be a value from the schema
+            padding: '10px',
             color: 'var(--ds-text)',
             backgroundColor: 'var(--ds-bold)',
           },
--- a/packages/react/src/create-strict-api/__tests__/index.test.tsx
+++ b/packages/react/src/create-strict-api/__tests__/index.test.tsx
@@ -146,7 +146,7 @@ describe('createStrictAPI()', () => {
           accentColor: 'red',
           all: 'inherit',
           '&:hover': { color: 'var(--ds-text-hover)' },
-          '&:invalid': { color: 'orange' },
+          '&:invalid': { color: 'var(--ds-text)' },
         },
       });
 
@@ -275,8 +275,8 @@ describe('createStrictAPI()', () => {
       function Button({ xcss }: { xcss: ReturnType<typeof XCSSProp<'background', never>> }) {
         return <button data-testid="button" className={xcss} />;
       }
-
       const styles = cssMap({ bg: { background: 'var(--ds-surface)' } });
+
       const { getByTestId } = render(<Button xcss={styles.bg} />);
 
       expect(getByTestId('button')).toHaveCompiledCss('background', 'var(--ds-surface)');
@@ -302,7 +302,6 @@ describe('createStrictAPI()', () => {
       function Button({ xcss }: { xcss: ReturnType<typeof XCSSProp<'background', '&:hover'>> }) {
         return <button data-testid="button" className={xcss} />;
       }
-
       const styles = cssMap({
         primary: {
           background: 'var(--ds-surface)',
--- a/packages/react/src/create-strict-api/__tests__/index.test.tsx
+++ b/packages/react/src/create-strict-api/__tests__/index.test.tsx
@@ -280,6 +280,76 @@ describe('createStrictAPI()', () => {
   });
 
   describe('XCSSProp', () => {
+    it('should error with values not in the strict `CompiledStrictSchema`', () => {
+      function Button({
+        xcss,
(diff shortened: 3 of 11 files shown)


packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api-recursive.ts
```diff
@@ -6,29 +6,29 @@ type ColorPressed = 'var(--ds-text-pressed)';
...
 type BackgroundPressed = 'var(--ds-bold-pressed)' | 'var(--ds-success-pressed)';
+type Space = 'var(--ds-space-050)' | 'var(--ds-space-0)';
 
 interface Properties {
   color: Color;
   backgroundColor: Background;
+  padding: Space;
 }
 
-interface HoveredProperties extends Omit<Properties, 'backgroundColor' | 'color'> {
```
itsdouges, line 15: Ensuring every pseudo has "all available properties" is not longer needed as it falls back to the top level values.

packages/react/src/create-strict-api/__tests__/generics.test.tsx
```diff
@@ -155,15 +162,15 @@ describe('createStrictAPI()', () => {
             backgroundColor: '',
             '&:hover': {
               // @ts-expect-error — Type '""' is not assignable to type ...
-              color: '',
+              color: 'var(--ds-text)',
```
itsdouges, line 165: This is typed to be the hovered value so using the base value is a type error.

packages/react/src/create-strict-api/__tests__/generics.test.tsx
```diff
@@ -188,21 +195,31 @@ describe('createStrictAPI()', () => {
   describe('type success', () => {
     it('should pass type check for css()', () => {
       const styles = css({
+        // @ts-expect-error — should be a value from the schema
+        padding: '10px',
```
itsdouges, line 199: Padding is now typed in the example schema above https://github.com/atlassian-labs/compiled/pull/1636/files#diff-2b3750c8f7eefa0e3b3f59fd1e2c8b4f84d0ce0997ba9c24a60287424fa6d89eR14

packages/react/src/create-strict-api/__tests__/index.test.tsx
```diff
@@ -1,24 +1,31 @@
...
 import { render } from '@testing-library/react';
 
-import { css, cssMap, XCSSProp } from './__fixtures__/strict-api';
+import { css, cssMap, XCSSProp, cx } from './__fixtures__/strict-api';
 
 describe('createStrictAPI()', () => {
   describe('css()', () => {
     it('should type error when circumventing the excess property check', () => {
-      const styles = css({
+      const stylesOne = css({
```
itsdouges, line 9: I've split this into two because the entire object is type checked now, meaning it hits the first type error and then stops trying to typecheck the rest. Type safety has not been lost but because of the type refactor behaves slightly differently.

packages/react/src/create-strict-api/__tests__/index.test.tsx
```diff
@@ -79,7 +86,7 @@ describe('createStrictAPI()', () => {
         color: 'var(--ds-text)',
         all: 'inherit',
         '&:hover': { color: 'var(--ds-text-hover)' },
-        '&:invalid': { color: 'orange' },
+        '&:invalid': { color: 'var(--ds-text)' },
```
itsdouges, line 89: All psuedos are now enforced to use the "base" values from the schema. This fixes the type error.

packages/react/src/create-strict-api/__tests__/index.test.tsx
```diff
@@ -357,13 +365,19 @@ describe('createStrictAPI()', () => {
       });
 
       const { getByTestId } = render(
-        <Button
-          // @ts-expect-error -- Errors because `color` conflicts with the `XCSSProp` schema–`color` should be a css variable.
-          xcss={styles.primary}
```
itsdouges, line 362: This no longer errors which was unexpected. I'm not sure why. But it is a type error in the `cssMap` call above, and it is also a type error inline (below). So I'm confident to let it slide.

packages/react/src/create-strict-api/__tests__/index.test.tsx
```diff
@@ -392,21 +406,22 @@ describe('createStrictAPI()', () => {
...
-
       const styles = cssMap({
         primary: {
           // @ts-expect-error -- Fails because `foo` is not assignable to our CSSProperties whatsoever.
           foo: 'bar',
           background: 'var(--ds-surface)',
+        },
+        hover: {
           '&:hover': {
-            // This does not fail, but would if the above was removed; this should be tested in raw `cssMap` fully.
```
itsdouges, line 402: This does type error now because of the type refactor.

packages/react/src/create-strict-api/index.ts
```diff
@@ -2,25 +2,9 @@ import type { StrictCSSProperties, CSSPseudos } from '../types';
 import { createStrictSetupError } from '../utils/error';
 import { type CompiledStyles, cx, type Internal$XCSSProp } from '../xcss-prop';
 
-type PseudosDeclarations = {
-  [Q in CSSPseudos]?: StrictCSSProperties;
-};
+import type { AllowedStyles, ApplySchema, ApplySchemaMap, CompiledSchemaShape } from './types';
 
-type EnforceSchema<TSchema> = {
```
itsdouges, line 9: These types have been deleted and re-written from scratch inside the `./types` module.

packages/react/src/create-strict-api/index.ts
```diff
@@ -36,7 +20,9 @@ interface CompiledAPI<TSchema extends CompiledSchema> {
    * <div css={redText} />
    * ```
    */
-  css(styles: CSSStyles<TSchema>): StrictCSSProperties;
+  css<TStyles extends ApplySchema<TStyles, TSchema>>(
```
itsdouges, line 23: To get the types working I built up `css` and then extrapolated it into `cssMap`. I should probably change the return type so it doesn't work with `xcss` prop...

packages/react/src/create-strict-api/index.ts
```diff
@@ -149,8 +132,6 @@ interface CompiledAPI<TSchema extends CompiledSchema> {
   >(): Internal$XCSSProp<TAllowedProperties, TAllowedPseudos, TSchema, TRequiredProperties>;
 }
 
-type CompiledSchema = StrictCSSProperties & PseudosDeclarations;
```
itsdouges, line 152: Moved into `./types`

packages/react/src/create-strict-api/types.ts
```diff
@@ -0,0 +1,40 @@
...
+  : StrictCSSProperties[TKey];
+
+export type ApplySchema<TObject, TSchema, TPseudoKey extends CSSPseudoClasses | '' = ''> = {
+  [TKey in keyof TObject]?: TKey extends keyof StrictCSSProperties
+    ? ApplySchemaValue<TSchema, TKey, TPseudoKey>
+    : TKey extends CSSPseudoClasses
+    ? ApplySchema<TObject[TKey], TSchema, TKey>
+    : TKey extends `@${string}` | CSSPseudoElements
+    ? ApplySchema<TObject[TKey], TSchema>
+    : never;
```
itsdouges, line 35: This causes properties that are unknown to be a type error.

packages/react/src/create-strict-api/__tests__/index.test.tsx
```diff
@@ -79,7 +86,7 @@ describe('createStrictAPI()', () => {
         color: 'var(--ds-text)',
         all: 'inherit',
         '&:hover': { color: 'var(--ds-text-hover)' },
-        '&:invalid': { color: 'orange' },
```
dddlr, line 82: Inside the `&:invalid { }` block, if I had a property or value that wasn't one of these

```
  color: 'var(--ds-text)';
  background: 'var(--ds-surface)' | 'var(--ds-surface-sunken)';
  bkgrnd: 'red' | 'green';
```
(according to the source code in https://github.com/atlassian-labs/compiled/blob/fe384d5fc2d1f52cd2dba93937b5c85e013fdf53/packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api.ts#L3 ) , would we have a type error?

I noticed that `&:invalid` isn't explicitly specified in `packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api.ts`, so not sure how the typechecking should work inside `&:invalid { ... }`

packages/react/src/create-strict-api/types.ts
```diff
@@ -0,0 +1,40 @@
...
+      ? TSchema[TPseudoKey][TKey]
+      : TSchema[TKey]
+    : TSchema[TKey]
+  : StrictCSSProperties[TKey];
+
+export type ApplySchema<TObject, TSchema, TPseudoKey extends CSSPseudoClasses | '' = ''> = {
+  [TKey in keyof TObject]?: TKey extends keyof StrictCSSProperties
+    ? ApplySchemaValue<TSchema, TKey, TPseudoKey>
+    : TKey extends CSSPseudoClasses
+    ? ApplySchema<TObject[TKey], TSchema, TKey>
```
itsdouges, line 32: This passes on the pseudo property key so we can use it to get the actual value. We'll most likely need to change this type when implementing nested pseudos like:

```
css({
  ':hover': {
    '::after': {}
  }
});
```

Or we could potentially keep them flat. Open to ideas — but for another PR not this one.

packages/react/src/types.ts
```diff
@@ -35,14 +35,7 @@ export type CssFunction<TProps = unknown> =
...
 
-/*
- * This list of pseudo-classes and pseudo-elements are from csstype
- * but with & added to the front. Compiled supports both &-ful
- * and &-less forms and both will target the current element
- * (`&:hover` <==> `:hover`), however we force the use of the
- * &-ful form for consistency with the nested spec for new APIs.
- */
-export type CSSPseudos =
+export type CSSPseudoElements =
```
itsdouges, line 38: I separated the pseudo types so we can differentiate between them. Pseudo _elements_ are not supported in the Compiled schema so we skip over them.

packages/react/src/create-strict-api/types.ts
```diff
@@ -0,0 +1,40 @@
...
+    : TKey extends `@${string}` | CSSPseudoElements
+    ? ApplySchema<TObject[TKey], TSchema>
```
itsdouges, line 34: For pseudo elements and at rules we skip over them and don't pass their key down (as they aren't supported in the compiled schema).

packages/react/src/xcss-prop/index.ts
```diff
@@ -1,27 +1,30 @@
...
-    ?
-        | CompiledPropertyDeclarationReference
-        | (Q extends keyof TSchema ? TSchema[Q] : CSSProperties[Q])
+type XCSSValue<
+  TStyleDecl extends keyof CSSProperties,
+  TSchema,
+  TPseudoKey extends CSSPseudoClasses | ''
+> = {
+  [Q in keyof StrictCSSProperties]: Q extends TStyleDecl
+    ? CompiledPropertyDeclarationReference | ApplySchemaValue<TSchema, Q, TPseudoKey>
```
itsdouges, line 15: We re-use the `ApplySchemaValue` here from the strict types, using the same type behaviour to get the value.

packages/react/src/xcss-prop/index.ts
```diff
@@ -1,27 +1,30 @@
...
     : never;
 };
 
-type XCSSPseudos<
-  TAllowedProperties extends keyof CSSProperties,
+type XCSSPseudo<
+  TAllowedProperties extends keyof StrictCSSProperties,
```
itsdouges, line 20: Internally we use `StrictCSSProperties` now as valid values. Anything outside that will resolve to `never`.

packages/react/src/create-strict-api/__tests__/index.test.tsx
itsdouges: This is the new behaviour — values now resolve to the top level ones unless there is a more specific one defined (inside an object in the schema e.g. `&:invalid: {}`).

packages/react/src/create-strict-api/__tests__/index.test.tsx
```diff
@@ -392,21 +406,22 @@ describe('createStrictAPI()', () => {
...
       const styles = cssMap({
         primary: {
           // @ts-expect-error -- Fails because `foo` is not assignable to our CSSProperties whatsoever.
           foo: 'bar',
           background: 'var(--ds-surface)',
+        },
+        hover: {
           '&:hover': {
-            // This does not fail, but would if the above was removed; this should be tested in raw `cssMap` fully.
+            // @ts-expect-error -- Fails because `foo` is not assignable to our CSSProperties whatsoever.
```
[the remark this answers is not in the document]
itsdouges: Only because I split the declaration into two.

packages/react/src/create-strict-api/types.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export type PseudosDeclarations = { [Q in CSSPseudos]?: StrictCSSProperties };
+
+export type AllowedStyles = StrictCSSProperties & PseudosDeclarations;
+
+export type ApplySchemaValue<
+  TSchema,
+  TKey extends keyof StrictCSSProperties,
+  TPseudoKey extends CSSPseudoClasses | ''
+> = TKey extends keyof TSchema
```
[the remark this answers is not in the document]
itsdouges: I'll have a look at commenting + seeing if it can be abstracted into util types.

packages/react/src/create-strict-api/__tests__/index.test.tsx
dddlr: ahh awesome 👍

packages/react/src/create-strict-api/__tests__/index.test.tsx
kylorhall-atlassian: So luckily this doesn't allow a `styles.primary` with more keys on it than desired (eg. if `createStrictAPI<'color' | 'backgroundColor'>` and `XCSSProp<'color'>` are mismatched, but it doesn't deeply check the values.

If there was ever a need for say `<Button xcss>` to not allow a specific background/color via the `XCSSProp<…>` generic, it's probably all going to have to be rewritten—but maybe that's a good line in the sand to draw for design, we don't care about shipping design opinions, the component is the design opinion, we care about shipping boundaries to not cross so we can evolve the system over time, which may include things you don't like…

packages/react/src/create-strict-api/__tests__/index.test.tsx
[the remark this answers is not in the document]
kylorhall-atlassian: Could be worth another test to validate that any valid `CSS.Properties` can be set, but we already do this with `accentColor`, so I think it's fine…?  Did note other concerns on that via DM.

packages/react/src/create-strict-api/types.ts
```diff
@@ -0,0 +1,63 @@
...
+    TPseudoKey extends keyof TSchema
+    ? TKey extends keyof TSchema[TPseudoKey]
+      ? // We found a more specific value under TPseudoKey.
+        TSchema[TPseudoKey][TKey]
+      : // Did not found anything specific, use the top level TSchema value.
+        TSchema[TKey]
+    : // Did not found anything specific, use the top level TSchema value.
+      TSchema[TKey]
+  : // TKey wasn't found on the schema, fallback to the CSS property value
+    StrictCSSProperties[TKey];
```
kylorhall-atlassian, line 36: @itsdouges On second thought (from DMs), we could feasibly even `never` this until we're confident in our defensive API…

packages/react/src/create-strict-api/types.ts
```diff
@@ -0,0 +1,63 @@
...
+      // while passing down the parent pseudo key to resolve any specific schema types.
+      ApplySchema<TObject[TKey], TSchema, TKey>
+    : TKey extends `@${string}` | CSSPseudoElements
+    ? // TKey is either an at rule or a pseudo element, either way we don't care about
+      // passing down the key so we recursively resolve its child properties starting at
+      // the base schema, treating it as if it's not inside an object.
+      ApplySchema<TObject[TKey], TSchema>
+    : // Fallback case, did not find a valid CSS property, at rule, or pseudo.
+      // Resolve the value to `never` which will end up being a type violation.
+      never;
```
kylorhall-atlassian, line 58: So this `never` applies to `css({ foo: 'bar' })` ✅

packages/react/src/xcss-prop/index.ts
kylorhall-atlassian: I feel like something in here isn't working as I expected given `xcss={styles.primary}` passes even when the type does not match `TSchema` and the `cssMap` fails 😕 but I can't place my finger on it…

packages/react/src/create-strict-api/types.ts
kylorhall-atlassian: I gather it's just another ternary nesting, eg. `Q in CSSPseudoClasses ? …` vs. `Q in CSSPseudoElements ? …`.  Doing my head in a bit, but same as doing all of this `TSchema[TPseudoKey][TKey]`…

It's fortunately only a maximum of 5 layers deep 😨 inside `CssMap`, but I think they're all explicitly different layers?  Something like?
```js
MapKey // primary
  CSSKey // display
  PseudoElementKey // &::after
    PseudoElementKey // &:hover
      CSSKey // display
  PseudoElementKey // &:hover
    CSSKey // display
    PseudoElementKey // &::after
      CSSKey // display
  AtRuleKey // @media
    PseudoElementKey // &:hover
      CSSKey // display
      PseudoElementKey // &::after
        CSSKey // display
    PseudoElementKey // &::after
      PseudoElementKey // &:hover
        CSSKey // display
    PseudoElementKey // &:hover
      CSSKey // display
```

kylorhall-atlassian (changes requested):

packages/react/src/create-strict-api/__tests__/index.test.tsx
itsdouges: There may be a need to restrict values in the future, potentially. But at the same time maybe it doesn't matter as, at least, createStrictAPI constrains the values. Also noting that the inline object was fine 🤔.

packages/react/src/create-strict-api/index.ts
itsdouges: Yeah. Will make a task!

packages/react/src/create-strict-api/__tests__/index.test.tsx
itsdouges: @kylorhall-atlassian I've marked this as a breaking change. Gut feel was it's breaking a bunch of behaviour. Or at least you'll get some type errors. I could go both ways, wdyt?

packages/react/src/create-strict-api/types.ts
itsdouges: Yeah for sure. Something we can build on!

packages/react/src/xcss-prop/index.ts
itsdouges: Yeah I didn't spend that much time inside xcss prop types. I got the type violations fixed and then moved on... What's your (deeper) thoughts?

packages/react/src/create-strict-api/__tests__/index.test.tsx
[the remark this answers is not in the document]
itsdouges: Okay.

packages/react/src/create-strict-api/types.ts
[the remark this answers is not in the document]
itsdouges: Does the pushed changes work for you Grant?

packages/react/src/create-strict-api/__tests__/index.test.tsx
kylorhall-atlassian: I don't think it hurts anyone to call it a breaking change, as for us, if this was used, it's a change that would be noisy or difficult to adopt.  Again, I think it's a breaking change that will impact no one.

packages/react/src/xcss-prop/index.ts
kylorhall-atlassian: Oh, I think this introduces a bug 😢 using `import { cssMap } from '@compiled/react'` I can now pass anything I want.

Screenshot 2024-03-05 at 2 35 11 PM

Can't find a solution for it, I'll try to look a bit later today if I can..

packages/react/src/create-strict-api/types.ts
kylorhall-atlassian: Yeah, more of a question of how much we're worried about the unknown.  Either way, I gather we just make a 100% strict `CSS.Properties`-like interface ourselves to be honest, so if our `keyof TSchema` === `keyof StrictCSSProperties` then this doesn't matter.

packages/react/src/create-strict-api/__tests__/index.test.tsx
```diff
@@ -271,12 +280,82 @@ describe('createStrictAPI()', () => {
...
+            // @ts-expect-error — TODO: This should conflict, but when `cssMap` conflicts, it gets a different type (this has `ApplySchema`, not the raw object), so this doesn't error?  Weird…
+            xcss={stylesInvalid.primary}
+          />
+          <Button
+            testId="button-invalid-strict-cx"
+            // @ts-expect-error — TODO: This should conflict, but when `cssMap` conflicts, it gets a different type (this has `ApplySchema`, not the raw object), so this doesn't error?  Weird…
+            xcss={cx(stylesInvalid.primary, stylesValid.primary)}
+          />
+          <Button testId="button-valid" xcss={stylesValid.primary} />
+          <Button testId="button-valid-cx" xcss={cx(stylesValid.primary, stylesValid.primary)} />
```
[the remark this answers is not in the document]
kylorhall-atlassian: [sent via DM] https://github.com/atlassian-labs/compiled/commit/be2fbbced159fe8a728ab1b16df99ad2b536f299

kylorhall-atlassian (approved):
I think this is fine…just had that suggestion on the tests in https://github.com/atlassian-labs/compiled/commit/be2fbbced159fe8a728ab1b16df99ad2b536f299

packages/react/src/create-strict-api/__tests__/index.test.tsx
itsdouges: Fuck yeah.

liamqma (approved):

PR #1637, opened 2024-03-05
EXAMPLE: Allow explicit value-blocking in `xcss` prop.

kylorhall-atlassian:
_[example for @itsdouges to review into https://github.com/atlassian-labs/compiled/pull/1636]_

We don't actually need the generic value for this to be valuable, this should block it so an `@compiled/react.cssMap` with conflicting values is disallowed, but also can allow prop-level restrictions (if we want).

This is by moving `CompiledPropertyDeclarationReference` to not be optional as that object seemingly messed something up as `| CompiledPropertyDeclarationReference` was always "truthy" (if that makes sense).

However, this removes the functionality for inline `xcss` and requires a `cssMap` or future `css`(?) to exist..which is probably fine.

---

This has one remaining scenario not erroring (if I fixed the other inline `xcss={{ … }}` object tests), but it does error somewhere in the stack, and this is because a cssMap-level type error is typed differently than a non-erroring cssMap?

### DIFF
--- a/packages/react/src/create-strict-api/__tests__/index.test.tsx
+++ b/packages/react/src/create-strict-api/__tests__/index.test.tsx
@@ -146,7 +146,7 @@ describe('createStrictAPI()', () => {
           accentColor: 'red',
           all: 'inherit',
           '&:hover': { color: 'var(--ds-text-hover)' },
-          '&:invalid': { color: 'orange' },
+          '&:invalid': { color: 'var(--ds-text)' },
         },
       });
 
@@ -275,8 +275,8 @@ describe('createStrictAPI()', () => {
       function Button({ xcss }: { xcss: ReturnType<typeof XCSSProp<'background', never>> }) {
         return <button data-testid="button" className={xcss} />;
       }
-
       const styles = cssMap({ bg: { background: 'var(--ds-surface)' } });
+
       const { getByTestId } = render(<Button xcss={styles.bg} />);
 
       expect(getByTestId('button')).toHaveCompiledCss('background', 'var(--ds-surface)');
@@ -302,7 +302,6 @@ describe('createStrictAPI()', () => {
       function Button({ xcss }: { xcss: ReturnType<typeof XCSSProp<'background', '&:hover'>> }) {
         return <button data-testid="button" className={xcss} />;
       }
-
       const styles = cssMap({
         primary: {
           background: 'var(--ds-surface)',
--- a/packages/react/src/create-strict-api/__tests__/index.test.tsx
+++ b/packages/react/src/create-strict-api/__tests__/index.test.tsx
@@ -1,6 +1,9 @@
 /** @jsxImportSource @compiled/react */
 import { render } from '@testing-library/react';
 
+import baseCssMap from '../../css-map';
+
+import type { CompiledStrictAPI } from './__fixtures__/strict-api';
 import { css, cssMap, XCSSProp, cx } from './__fixtures__/strict-api';
 
 describe('createStrictAPI()', () => {
@@ -344,7 +347,7 @@ describe('createStrictAPI()', () => {
       expect(getByTestId('button')).toHaveCompiledCss('color', 'var(--ds-text)');
     });
 
-    it('should error with values not in the strict `CompiledAPI`', () => {
+    it('should error with values not in the strict `CompiledStrictSchema`', () => {
       function Button({
         xcss,
         testId,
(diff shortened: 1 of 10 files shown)


packages/react/src/create-strict-api/__tests__/index.test.tsx
```diff
@@ -355,20 +358,54 @@ describe('createStrictAPI()', () => {
...
+          />
+          <Button
+            testId="button-invalid-root-cx"
+            // @ts-expect-error -- This conflicts with the custom API, should be a different bg color
+            xcss={cx(stylesInvalidRoot.primary, stylesValid.primary)}
+          />
+          <Button
+            testId="button-invalid-strict"
+            // @ts-expect-error -- TODO: This should conflict, but when `cssMap` conflicts, it gets a different type (this has `ApplySchema`, not the raw object), so this doesn't error?  Weird…
+            xcss={stylesInvalid.primary}
```
kylorhall-atlassian, line 398: This one's a bit odd, I can't figure it out, only "problem" in this PR, minus the `xcss={{ … }}` removal.

When `cssMap()` has an error in it (this test), the type is different 😕?  Why does this have `ApplySchema` and not the type??
Screenshot 2024-03-06 at 9 32 15 AM

Vs. below, where the `cssMap()` type does not error, but the `xcss` errors instead:
Screenshot 2024-03-06 at 9 32 25 AM

packages/react/src/jsx/jsx-local-namespace.ts
```diff
@@ -80,7 +80,11 @@ export namespace CompiledJSX {
        * The class name prop now can be given the output of xcss prop from `@compiled/react`.
        */
       className?: string | Record<string, any> | null | false;
-      css?: CssFunction<void> | CssFunction<void>[];
+      css?:
+        | CssFunction<void>
+        | CssFunction<void>[]
+        | (CssFunction<void> & CompiledPropertyDeclarationReference)
```
itsdouges, line 86: Ah adding to css prop because it was type erroring?

packages/react/src/xcss-prop/index.ts
```diff
@@ -23,10 +29,11 @@ type XCSSPseudo<
   TSchema
 > = {
   [Q in CSSPseudos]?: Q extends TAllowedPseudos
-    ? MarkAsRequired<
-        XCSSValue<TAllowedProperties, TSchema, Q extends CSSPseudoClasses ? Q : ''>,
-        TRequiredProperties['requiredProperties']
-      >
+    ? XCSSValue<TAllowedProperties, TSchema, Q extends CSSPseudoClasses ? Q : ''> &
```
itsdouges, line 32: Ah yes, nice pick up

packages/react/src/xcss-prop/index.ts
```diff
@@ -57,8 +60,8 @@ type CompiledPropertyDeclarationReference = {
 export type CompiledStyles<TObject> = {
   [Q in keyof TObject]: TObject[Q] extends Record<string, unknown>
     ? CompiledStyles<TObject[Q]>
-    : CompiledPropertyDeclarationReference;
-};
+    : TObject[Q];
+} & CompiledPropertyDeclarationReference;
```
itsdouges, line 64: I actually made this change in my PR when I messaged you last night but made it also a union of the object values so it could still be used in inline Xcss

packages/react/src/create-strict-api/index.ts
```diff
@@ -128,8 +128,9 @@ export interface CompiledAPI<TSchema extends CompiledSchemaShape> {
     TRequiredProperties extends {
       requiredProperties: TAllowedProperties;
       requiredPseudos: TAllowedPseudos;
-    } = never
-  >(): Internal$XCSSProp<TAllowedProperties, TAllowedPseudos, TSchema, TRequiredProperties>;
+    } = never,
+    CustomTSchema extends CompiledSchemaShape = TSchema
```
itsdouges, line 132: Oh! Is this enabling custom values?

packages/react/src/jsx/jsx-local-namespace.ts
kylorhall-atlassian: Yeah, exactly, because our `css|cssMap` returns `CompiledStyles = { … } & CompiledPropertyDeclarationReference` (top-level).

I wasn't quite certain if there's any sweeping implications of this, to be frank…LGTM 🤣

packages/react/src/create-strict-api/index.ts
kylorhall-atlassian: Yeah, so in this example we only allow a specific background color: https://github.com/atlassian-labs/compiled/blob/a03690befd404f6d5c78e3b067be3309ddf78b3a/packages/react/src/create-strict-api/__tests__/index.test.tsx#L431

packages/react/src/xcss-prop/index.ts
itsdouges: Yeah sorry my head was in the media query PR. This might still be needed?

PR #1640, opened 2024-03-06
Better cssMap types: fix inference of complex types and adds descriptions to the type.

### LINKED ISSUE #1639: CSS Map has its output property types unexpectedly expanded at times
**Describe the bug**

Sometimes properties on the output object are expanded to `string` unexpectedly. This affects passing values to XCSS prop that expect narrowed types.

**To Reproduce**

```tsx

const stylesValidRoot = cssMap({
  primary: {
    color: 'var(--ds-text)',
    '&:hover': { color: 'var(--ds-text-hover)', background: 'var(--ds-surface-hover)' },
  },
});

stylesValidRoot.primary.['&:hover'].background // string
```

However if we define it as `backgroundColor` it stays narrowed:

```tsx

const stylesValidRoot = cssMap({
  primary: {
    color: 'var(--ds-text)',
    '&:hover': { color: 'var(--ds-text-hover)', backgroundColor: 'var(--ds-surface-hover)' },
  },
});

stylesValidRoot.primary.['&:hover'].backgroundColor // 'var(--ds-surface-hover)'
```

**Expected behavior**

The property remains narrowed to `'var(--ds-surface-hover)'`.

**Additional context**

This was found during the strict typing refactor, see: 

https://github.com/atlassian-labs/compiled/pull/1636/files#diff-1d4b97255cb11e9b25307047d51c7f2c1d94052d4626cc8e19c410a8b0444788R299

kylorhall-atlassian:
:one: Fixes inference of complex CSS properties like `background`
Fixes #1639
Instead of `background: string`, it will be inferred as the value, `background: 'red'` — I could not figure out how to infer numbers, however.
Screenshot 2024-03-07 at 10 41 23 AM

:two: Adds `CSS.Properties` descriptions into the cssMap object types (rather than showing the description of the inferred object)
Screenshot 2024-03-07 at 10 42 04 AM


packages/react/src/types.ts
```diff
@@ -100,7 +100,7 @@ export type CSSPseudos =
  * excess property check doesn't enable makers to circumvent the
  * system and pass in values they shouldn't.
  */
-export type CSSProperties = Readonly<CSS.Properties<string | number>>;
```
itsdouges, line 103: Ah, yep. That'd do it. Surprised it didn't affect more!

itsdouges (approved):
LGTM

packages/react/src/types.ts
kylorhall-atlassian: I assume it touched a lot of properties to be honest, just until our strict mode, we never needed the inference to work.

liamqma (approved):

PR #1643, opened 2024-03-10
Remove required pseudo from XCSS prop

itsdouges:
During the implementation of media query types for strict / XCSS apis we came to the realization that required pseudos isn't actually useful, only ensuring the base case properties are required is. 

This pull request removes the type property.

### DIFF
--- a/packages/react/src/create-strict-api/__tests__/index.test.tsx
+++ b/packages/react/src/create-strict-api/__tests__/index.test.tsx
@@ -720,7 +720,7 @@ describe('createStrictAPI()', () => {
       expect(getByTestId('button-valid')).toHaveCompiledCss('color', 'var(--ds-text-hover)', {
         target: ':hover',
       });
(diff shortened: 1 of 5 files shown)


packages/react/src/create-strict-api/__tests__/index.test.tsx
```diff
@@ -711,7 +703,7 @@ describe('createStrictAPI()', () => {
         primary: { '&:hover': { color: 'var(--ds-text-hover)' } },
       });
       const stylesInvalid = cssMap({
-        primary: { '&:focus': { color: 'var(--ds-text)' } },
+        primary: { '&:focus': { background: 'var(--ds-surface)' } },
```
itsdouges, line 706: Keep the error but for `background` instead of the missing `hover` pseudo

kylorhall-atlassian (approved):
Straightforward and makes sense to me…but you'll have to get tests passing!

kylorhall-atlassian (approved):

PR #1645, opened 2024-03-12
Check for missing AST in parcel transformer

JakeLane:
A simple fix to check for `program` being defined, as well as some more specific checks. This appears to occur in some edge cases where the AST object is available but the value is not set.

### DIFF
--- a/packages/parcel-transformer/src/index.ts
+++ b/packages/parcel-transformer/src/index.ts
@@ -108,17 +108,21 @@ export default new Transformer<ParcelTransformerOpts>({
       plugins: config.transformerBabelPlugins ?? undefined,
     });
 
-    return {
-      type: 'babel',
-      version: '7.0.0',
-      program,
-    };
+    if (program) {
+      return {
+        type: 'babel',
+        version: '7.0.0',
+        program,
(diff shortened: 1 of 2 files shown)


packages/parcel-transformer/src/index.ts
```diff
@@ -108,17 +108,21 @@ export default new Transformer<ParcelTransformerOpts>({
       plugins: config.transformerBabelPlugins ?? undefined,
     });
 
-    return {
```
itsdouges, line 111: What happens when skipping over this? Do things still get transformed / compiled / extracted?

JakeLane: This shouldn't happen in practice, because the AST will generate if there's some code. This is just making the behaviour more explicit, where the result will be ignored if there's no AST.

itsdouges: Alright. Does Parcel need some types updated now since program can be undefined?

JakeLane: probably! These values are `any` :/

itsdouges (approved):

PR #1648, opened 2024-03-19
Write a few unit-tests for babel plugin

yamadapc:


packages/babel-plugin/src/__tests__/css-builder.test.ts
```diff
@@ -49,4 +49,150 @@ describe('css builder', () => {
...
+        An unhandled exception was raised when parsing your CSS, this is probably a bug!
+        Raise an issue here: https://github.com/atlassian-labs/compiled/issues/new?assignees=&labels=&template=bug_report.md&title=CSS%20Parsing%20Exception:%20
+
+        Input CSS: {
+          red
+        }
+
+        Exception: <css input>:1:1: Unknown word
+      "
+    `);
```
dddlr, line 196: personally i'd prefer our snapshots not depend on the exact formatting of an error message - suggest changing to this:

```suggestion
    expect(actual).toThrow();
```

or `expect(actual).toThrowError(/Exception: <css input>:1:1: Unknown word/);` if you want to keep the unique parts of the error (though "Unknown word" doesn't tell us anything useful either)

packages/babel-plugin/src/utils/__tests__/css-builders.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    const meta: Metadata = {
+      parentPath: path!.parentPath,
+      state: {
+        cssMap: {},
+        filename: '',
+      },
+    } as any;
+
+    const { css, variables } = buildCss(path!.node, meta);
+    // TODO: This should not happen
```
dddlr, line 112: isn't this a bug? i'm not sure why we're adding a test case to enforce buggy behaviour here - perhaps creating a github issue would be better than having this test case here, as the former is more easily discoverable (we don't really check back on test cases unless they get broken)

the other test cases in this file look great, thank you for adding them 🙌

packages/babel-plugin/src/utils/__tests__/evaluate-expression.test.ts
```diff
@@ -0,0 +1,63 @@
+import { parse } from '@babel/parser';
```
dddlr, line 1: this file looks good!

packages/babel-plugin/src/utils/__tests__/css-builders.test.ts
yamadapc: Sure, let me remove some of these; generally I'm only trying to reproduce bugs with these.

It's a bit of a structural issue regarding what `css` + `cssMap` APIs do.

JakeLane (approved):

Issue #1531, opened 2023-10-19
Duplicated CSS rule in devloop if using SSR


liamqma:
**Describe the bug**
Duplicated CSS rule in devloop if using SSR

**To Reproduce**

1. Checkout the master branch
2. run `yarn run start:ssr`
3. Go `localhost:3000`
4. See `._1wyb1osq { font-size:100% }` is defined twice

![Screenshot 2023-10-19 at 11 24 20 am](https://github.com/atlassian-labs/compiled/assets/4413219/aa27e503-d0d3-4e2e-a66a-f21646a2e2b7)

**Expected behavior**
CSS rule should be defined once

**Additional context**
Compiled moves all found server-side rendered style elements to the head before : https://github.com/atlassian-labs/compiled/blob/master/packages/react/src/runtime/style-cache.tsx#L23
Also, insert style again https://github.com/atlassian-labs/compiled/blob/master/packages/react/src/runtime/style.tsx#L63
It causes same CSS being inserted to DOM twice.

PR #1649, opened 2024-03-20
feat: add title attribute to Compiled style tag

pancaspe87:
TBH, I am not sure if this is a valuable change considering [these devlop issues ](https://github.com/atlassian-labs/compiled/issues/1531). Also, we know that classes starting with `_` are usually Compiled; hence, we can just look for it.
However, it may save some keystrokes when debugging :)

----

**Dev mode**
It seems that we are having multiple style tags when extraction is not enabled, only one is getting the `title` attribute
Screenshot 2024-03-20 at 3 22 42 pm

**VS**

**Prod mode**
We can easily discern what style is for Compiled. 
Screenshot 2024-03-20 at 3 22 01 pm

### DIFF
--- a/packages/parcel-optimizer/src/index.ts
+++ b/packages/parcel-optimizer/src/index.ts
@@ -71,7 +71,7 @@ export default new Optimizer<ParcelOptimizerOpts, unknown>({
           .use(
             insertAt({
               selector: 'head',
-              append: '<style>' + stylesheet + '</style>',
+              append: '<style title="compiled">' + stylesheet + '</style>',
               behavior: 'inside',
             })
           )
--- a/.changeset/fluffy-lies-shake.md
+++ b/.changeset/fluffy-lies-shake.md
@@ -1,0 +1,5 @@
+---
+'@compiled/parcel-optimizer': minor
+---
+
+Add title attribute to Compiled style tag to discern with other style within the page
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,7 +99,7 @@ yarn test <filter> --watch
 
 Looking at tests first is generally the best way to get started.
 
-- Run a specific test file: `yarn test <filename> --watch`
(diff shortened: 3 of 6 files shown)


packages/parcel-optimizer/src/index.ts
```diff
@@ -71,7 +71,7 @@ export default new Optimizer<ParcelOptimizerOpts, unknown>({
           .use(
             insertAt({
               selector: 'head',
-              append: '<style>' + stylesheet + '</style>',
+              append: '<style title="compiled">' + stylesheet + '</style>',
```
pancaspe87, line 74: Btw, do we inline CSS also in webpack? I don't think I have seen `inlineCss` for it

liamqma (approved):

dddlr:
ooh good idea 👍 do you reckon it would be useful to also add this to the SSR styles that have the `data-cmpld="true"` attribute?

JakeLane:
We already have some similar behaviour for the runtime, maybe we should match that? https://github.com/atlassian-labs/compiled/blob/44b13234d50889c1f610102ee2751c9839d2ae35/packages/react/src/runtime/style.tsx#L48

dddlr:
> We already have some similar behaviour for the runtime, maybe we should match that?
> 
> https://github.com/atlassian-labs/compiled/blob/44b13234d50889c1f610102ee2751c9839d2ae35/packages/react/src/runtime/style.tsx#L48

hmm yeah maybe we could keep `data-cmpld` for the SSR stuff and have `data-compiled-parcel` for the `@compiled/parcel-optimizer` output? will let @pancaspe87 make the final call on the best name for this

pancaspe87:
Yeahh, I did see `data-cmpld` and thought it was specific to runtime. 

TBH, I am happy to reuse `data-cmpld` or apply the same `title='compiled' to everything` . Whatever works best for you @JakeLane @dddlr @liamqma

JakeLane (approved):

PR #1652, opened 2024-03-29
Remove the invariant to fail on an empty cssMap object for easier iterative development.

kylorhall-atlassian:
I personally noticed this wasn't breaking in local development, but was blocking in CI, and given we don't have any type checking around it, I wonder if the error is of any value?

Just figured I'd drop a PR to validate if there's a reason.

### DIFF
--- a/.changeset/dry-ghosts-brake.md
+++ b/.changeset/dry-ghosts-brake.md
@@ -1,0 +1,5 @@
+---
+'@compiled/babel-plugin': patch
+---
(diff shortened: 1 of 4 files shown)


itsdouges:
I'm a fan. If we want to prevent dangling empty objects a lint rule would be more appropriate.

itsdouges (approved):

liamqma (approved):

PR #1655, opened 2024-04-08
Add additional type exports

mitch1995:
Currently when using moduleResolutions such as `node16`, `nodenext` or `bundler` with `declaration: true` there are quite a few typescript errors present, as shown below

image

I created a minimum reproduction repo to demonstrate this [here](https://github.com/mitch1995/compiled-react-module-resolution-bundler/tree/master). 

**Steps to reproduce:**
1. `npx create-react-app my-app --template typescript`
2. add dependency for `@compiled/react` (0.17.0)
3. Update `tsconfig.json` to have `moduleResolution: Bundler` and `delcaration: true`
4. Create a file using imports from `@compiled/react` like `styled` and `createStrictAPI` and you will experience type errors

### DIFF
--- a/.changeset/empty-bags-remember.md
+++ b/.changeset/empty-bags-remember.md
@@ -1,0 +1,5 @@
+---
+'@compiled/react': patch
+---
+
+added additional type exports to support other ts moduleResolution methods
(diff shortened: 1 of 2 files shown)


dddlr (approved):
looks great! thanks mitchell 👍

PR #1656, opened 2024-04-08
Allow `@compiled/babel-plugin`'s `increaseSpecificity` to work with `@compiled/jest`

kylorhall-atlassian:
tl;dr: if you have `increaseSpecificity: true` set, you get styles such as `._18u0u2gc:not(#\\9){margin-left:var(--ds-space-100,8px)}` which will never match `._18u0u2gc`

Could use advice on either:
 - I could not think of a nice way to sneak a test in for this without trying to overhaul jest config entirely, so I did not.
 - This magic variable comes from https://github.com/atlassian-labs/compiled/blob/e69f5339a670bc4244b5d2b3e293c0609ef23dae/packages/css/src/plugins/increase-specificity.ts#L4C1-L5C1 and could similarly break at any time.

### DIFF
--- a/packages/jest/src/matchers.ts
+++ b/packages/jest/src/matchers.ts
@@ -5,6 +5,8 @@ import type { MatchFilter } from './types';
 
 type Arg = [{ [key: string]: string }, MatchFilter?];
 
+/** With the babel plugin in `increaseSpecificity:true`, classes will be appended with this. */
+const INCREASE_SPECIFICITY_SELECTOR = ':not(#\\9)';
 const DEFAULT_MATCH_FILTER: MatchFilter = { media: undefined, target: undefined };
 
 const kebabCase = (str: string) =>
@@ -61,7 +63,13 @@ const getRules = (ast: CSS.Stylesheet, filter: MatchFilter, className: string) =
   const klass = target ? `.${className}${target}` : `.${className}`;
   return allRules?.filter((r) => {
     if ('selectors' in r) {
-      return r.selectors?.find((s) => removeSpaces(s) === removeSpaces(klass));
+      return r.selectors?.find((s) => {
+        const sTrimmed = removeSpaces(s);
+        return (
+          sTrimmed === removeSpaces(klass) ||
+          sTrimmed === removeSpaces(klass + INCREASE_SPECIFICITY_SELECTOR)
+        );
+      });
     }
     return;
   });
--- a/packages/jest/src/matchers.ts
+++ b/packages/jest/src/matchers.ts
@@ -1,4 +1,3 @@
-import { INCREASE_SPECIFICITY_SELECTOR } from '@compiled/utils';
 import type { Media, StyleRules } from 'css';
 import CSS from 'css';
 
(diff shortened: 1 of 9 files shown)


liamqma (approved):

dddlr (dismissed):

packages/jest/src/matchers.ts
```diff
@@ -5,6 +5,8 @@ import type { MatchFilter } from './types';
 
 type Arg = [{ [key: string]: string }, MatchFilter?];
 
+/** With the babel plugin in `increaseSpecificity:true`, classes will be appended with this. */
+const INCREASE_SPECIFICITY_SELECTOR = ':not(#\\9)';
```
dddlr, line 9: nit: maybe this could be moved to `@compiled/utils`, as well as the `INCREASE_SPECIFICITY_SELECTOR` definition in `packages/css/src/plugins/increase-specificity.ts`, considering that this `:not(#\\9)` string is pretty esoteric

not too important though

dddlr:
> I could not think of a nice way to sneak a test in for this without trying to overhaul jest config entirely, so I did not.

hmm not sure either... i personally wouldn't worry about it (curious to see what @JakeLane thinks)

> This magic variable comes from https://github.com/atlassian-labs/compiled/blob/e69f5339a670bc4244b5d2b3e293c0609ef23dae/packages/css/src/plugins/increase-specificity.ts#L4C1-L5C1 and could similarly break at any time.

my other comment might help with this https://github.com/atlassian-labs/compiled/pull/1656#discussion_r1555199370 it's not as good as a real test, but a bit less prone to human error at least

packages/jest/src/matchers.ts
kylorhall-atlassian: Ah, I didn't know we had a utils; done in 87df1354

packages/utils/src/increase-specificity.ts
```diff
@@ -0,0 +1,2 @@
+/** With the babel plugin in `increaseSpecificity:t rue`, classes will be appended with this. */
```
kylorhall-atlassian, line 1:
```suggestion
/**
 * Configuring the babel plugin with `increaseSpecificity: true` will result in this being appended to the end of generated classes.
 */
```

packages/jest/tsconfig.json
```diff
@@ -5,5 +5,5 @@
     "outDir": "dist",
     "module": "commonjs"
   },
-  "references": []
+  "references": [{ "path": "../utils" }]
```
kylorhall-atlassian, line 8: Ah, this broke things as `utils/tsconfig.json` references this this, so:
> error TS6202: Project references may not form a circular graph. Cycle detected: /home/runner/work/compiled/compiled/packages/tsconfig.browser.json

kylorhall-atlassian: Copy & pasted to just work around this…

packages/jest/src/matchers.ts
kylorhall-atlassian: Nevermind, hit a TS circular reference issue, so I did end up copy and pasting: https://github.com/atlassian-labs/compiled/pull/1656#discussion_r1555243279 😅

JakeLane (approved):

PR #1658, opened 2024-04-09
Fix handling of CSS `content` property

### LINKED ISSUE #1650: Object syntax creates unnecessary quotes when using properties like `content: ''`, resulting in style not being applied
**Describe the bug**

**To Reproduce**
Update any of the existing stories to use [CSS counters](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_counter_styles/Using_CSS_counters). For example:

If using Tagged Template Expression, this works.

```
const Component = styled.div`
  width: calc(100% - ${spacing * 2}px);
  position: relative;
  counter-increment: step;
  :before {
    content: counter(step) ". ";
    position: absolute;
    top: 0;
    left: 0;
    width: calc(100% - ${spacing * 2}px);
    height: 100%;
    background-color: red;
}`;
```

If using Object syntax, with both `styled` or `CSS` APIs, this won't work. For example:
```
  color: 'white',
  position: 'relative',
  counterIncrement: 'step',
  width: '100%',
  '&::before' : {
    content: 'counter(step) ". "',
    backgroundColor: 'red',
    position: 'absolute',
    top: 0,
    left: 0,
    width: '100%',
    height: '100%',
    }
});
```

**Expected behaviour**
Make sure that additional quotes are not added.

Possibly handled in [here](https://github.com/atlassian-labs/compiled/blob/master/packages/babel-plugin/src/utils/css-builders.ts#L63)

**Workaround**
Use Tagged Template Expression instead - styled.div``

dddlr:
Fixes #1650, where quotes are incorrectly added around the value in situations like `content: url(...)` and `content: counter(...)`

Thanks to @mattcompiles for suggesting this approach to fixing the bug.

Based on https://github.com/vanilla-extract-css/vanilla-extract/blob/a623c1c65a543afcedb9feb30a7fe20452b99a95/packages/css/src/transformCss.ts#L266

For the curious:

* here is how Emotion handles the `content` property: https://github.com/emotion-js/emotion/blob/f4640f6a7cbe77f790e07706b287cb29342fd0ea/packages/serialize/src/index.js#L83
  * Emotion throws an error if a value does not have extra quotes surrounding it and it's not a recognised CSS value (e.g. url(...), content(...), initial, none...)
* styled-components v6 *never* adds quotes around the `content` property (and I think styled-components v3 too)

---

* [x] Add tests
* [ ] Check for potential regressions in atlassian codebases
* [x] Check how this handling differs from styled-components
* [x] Add changeset
* [x] Check why vanilla-extract checks whether the value is an array or not??

### DIFF
--- a/packages/babel-plugin/src/styled/__tests__/content-property.test.ts
+++ b/packages/babel-plugin/src/styled/__tests__/content-property.test.ts
@@ -1,0 +1,177 @@
+import type { TransformOptions } from '../../test-utils';
+import { transform as transformCode } from '../../test-utils';
+
+describe('handling of values for CSS `content` property', () => {
+  beforeAll(() => {
+    process.env.AUTOPREFIXER = 'off';
+  });
+
+  afterAll(() => {
+    delete process.env.AUTOPREFIXER;
+  });
+
+  const transform = (code: string, opts: TransformOptions = {}) =>
+    transformCode(code, { pretty: false, ...opts });
+
+  // Tests are based on those covered by vanilla-extract, credit goes to them :)
+  //
+  // https://github.com/vanilla-extract-css/vanilla-extract/blob/a623c1c65a543afcedb9feb30a7fe20452b99a95/packages/css/src/transformCss.test.ts#L639
+
+  it('should handle blank content', () => {
+    const code = `
+      import { styled } from '@compiled/react';
+
+      const ListItem = styled.div({
+        content: '',
+      });
+    `;
+
+    const actual = transform(code, { pretty: true });
+    expect(actual).toContain('._1sb2b3bt{content:""}');
+  });
+
+  it('should handle blank content (variant #1)', () => {
+    const code = `
+      import { styled } from '@compiled/react';
+
+      const ListItem = styled.div({
+        content: \`\`,
+      });
+    `;
+
+    const actual = transform(code, { pretty: true });
+    expect(actual).toContain('._1sb2b3bt{content:""}');
+  });
+
+  it('should handle blank content (variant #2)', () => {
+    const code = `
+      import { styled } from '@compiled/react';
+
+      const hello = '';
+      const ListItem = styled.div({
+        content: \`\${hello}\`,
+      });
+    `;
+
+    const actual = transform(code, { pretty: true });
+    expect(actual).toContain('._1sb2b3bt{content:""}');
+  });
+
+  it('should handle blank content (variant #2)', () => {
+    const code = `
+      import { styled } from '@compiled/react';
+
+      const hello = 'this is a string';
+      const ListItem = styled.div({
+        content: hello,
+      });
+    `;
+
+    const actual = transform(code, { pretty: true });
+    expect(actual).toContain('._1sb2b3bt{content:""}');
+  });
+
+  it('should add quotes to custom content values', () => {
+    const code = `
+      import { styled } from '@compiled/react';
+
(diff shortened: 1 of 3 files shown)


mattcompiles:
Nice one @dddlr 
I'd consider porting the test across here as well. We based most of ours off the [examples on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/content).

packages/babel-plugin/src/styled/__tests__/content-property.test.ts
```diff
@@ -0,0 +1,177 @@
...
+        '._17 &': { content: 'inherit' },
+        '._18 &': { content: 'initial' },
+        '._19 &': { content: 'revert' },
+        '._20 &': { content: 'unset' },
+      });
+    `;
+
+    const actual = transform(code, { pretty: true });
+    const expectedStrings = [
+      '._20 ._dqocn7od{content:unset}',
```
kylorhall-atlassian, line 151: I was like "dude, these are completely wrong" oh, it's just ordered 🤣

packages/babel-plugin/src/styled/__tests__/content-property.test.ts
```diff
@@ -0,0 +1,177 @@
...
+      '._14 ._1gw4qmeg{content:no-open-quote}',
+      '._13 ._1erv1lzr{content:close-quote}',
+      '._12 ._yz5n12u0{content:open-quote}',
+      '._11 ._1sv1f7m1{content:attr(value string)}',
+      '._10 ._11jd1td3{content:counters(section_counter,".",decimal-leading-zero)}',
+      '._09 ._1vjjgvpy{content:counters(section_counter,".")}',
+      '._08 ._tc0r17sh{content:counter(chapter_counter,upper-roman)}',
+      '._07 ._1g4mlfyt{content:counter(chapter_counter)}',
+      '._06 ._4g5v1dlo{content:"prefix"}',
+      '._05 ._10o677hy{content:url("http://www.example.com/test.png") /"This is the alt text"}',
```
kylorhall-atlassian, line 166: TIL

packages/babel-plugin/src/utils/css-builders.ts
```diff
@@ -56,16 +56,25 @@ const findBindingIdentifier = (
...
+ * but only when we reasonably think that they were intended. For example, url(...) and counter(...)
+ * will NOT have quotes added around them.
  */
 const normalizeContentValue = (value: string) => {
-  if (value.charAt(0) !== '"' && value.charAt(0) !== "'") {
-    return `"${value}"`;
+  // Adapted from vanilla-extract's handling of the `content` key
+  // https://github.com/vanilla-extract-css/vanilla-extract/blob/a623c1c65a543afcedb9feb30a7fe20452b99a95/packages/css/src/transformCss.ts#L266
+  const contentValuePattern =
+    /^([A-Za-z\-]+\([^]*|[^]*-quote|inherit|initial|none|normal|revert|unset)(\s|$)/;
```
kylorhall-atlassian, line 67: I'll trust them!

kylorhall-atlassian (approved):
Looks good, thank you to vanila-extract!

packages/babel-plugin/src/styled/__tests__/content-property.test.ts
dddlr: ah right yeah they show up in the snapshot in reverse order (20 to 1), not sure why

PR #1660, opened 2024-04-10
Add new default import source (@atlaskit/css)

itsdouges:
To make it easier to use the design system import source we're setting it as a default import source in Compiled. This means consumers don't need to configure it, they just need to turn Compiled on.

### DIFF
--- a/packages/babel-plugin/src/babel-plugin.ts
+++ b/packages/babel-plugin/src/babel-plugin.ts
@@ -31,7 +31,8 @@ import { visitXcssPropPath } from './xcss-prop';
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const packageJson = require('../package.json');
 const JSX_SOURCE_ANNOTATION_REGEX = /\*?\s*@jsxImportSource\s+([^\s]+)/;
-const DEFAULT_IMPORT_SOURCE = '@compiled/react';
+const COMPILED_IMPORT_SOURCE = '@compiled/react';
+const DEFAULT_IMPORT_SOURCES = [COMPILED_IMPORT_SOURCE, '@atlaskit/css'];
 
 let globalCache: Cache | undefined;
 
@@ -41,7 +42,7 @@ const findClassicJsxPragmaImport: Visitor<State> = {
 
     t.assertImportDeclaration(path.parent);
     // We don't care about other libraries
-    if (path.parent.source.value !== '@compiled/react') return;
+    if (path.parent.source.value !== COMPILED_IMPORT_SOURCE) return;
 
     if (
       (specifier.imported.type === 'StringLiteral' && specifier.imported.value === 'jsx') ||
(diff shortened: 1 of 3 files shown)


packages/babel-plugin/src/babel-plugin.ts
```diff
@@ -31,7 +31,8 @@ import { visitXcssPropPath } from './xcss-prop';
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const packageJson = require('../package.json');
 const JSX_SOURCE_ANNOTATION_REGEX = /\*?\s*@jsxImportSource\s+([^\s]+)/;
-const DEFAULT_IMPORT_SOURCE = '@compiled/react';
+const COMPILED_IMPORT_SOURCE = '@compiled/react';
+const DEFAULT_IMPORT_SOURCES = [COMPILED_IMPORT_SOURCE, '@atlaskit/css'];
```
kylorhall-atlassian, line 35: I guess it really barely adds much performance or bundle impact—if call it the Atlassian-internal package and it makes sense.  Would just be weird if we changed this over time or didn't actually get `@atlaskit/css` out there.

packages/babel-plugin/src/babel-plugin.ts
```diff
@@ -236,7 +237,7 @@ export default declare<State>((api) => {
         const userLandModule = path.node.source.value;
 
         const isCompiledModule = this.importSources.some((compiledModuleOrigin) => {
-          if (userLandModule === DEFAULT_IMPORT_SOURCE || compiledModuleOrigin === userLandModule) {
```
kylorhall-atlassian, line 239: Makes sense..

kylorhall-atlassian (commented):
Makes sense to me, I think an Atlassian-specific inclusion is fine.

packages/babel-plugin/src/babel-plugin.ts
itsdouges: Yeah just redundant code. Removed it.

packages/babel-plugin/src/babel-plugin.ts
itsdouges: > Would just be weird if we changed this over time or didn't actually get @atlaskit/css out there.

And we can update it if it changes. Also it will get out there ;).

packages/babel-plugin/src/babel-plugin.ts
dddlr: i'm confused why `userLandModule === DEFAULT_IMPORT_SOURCE` was checked at all, since `this.importSources` appears to always include `@compiled/react`... am i missing something?

dddlr: ahh just saw your comment @itsdouges yeah looks good

dddlr (approved):

JakeLane (approved):

packages/babel-plugin/src/babel-plugin.ts
alexreardon: Will this impact all consumers, or just Atlassian consumers?

itsdouges: Great question! It impacts everyone, by design. So imagine any Ecosystem folks wanting to use the new xcss prop capabilities – they don't have to configure the custom import source to use styles from atlaskit/css (which is the primary API we'll be documenting for usage on ADS site).

Outside of that it's just a no-op, none would be the wiser.

PR #1663, opened 2024-04-11
Fix CSS variables casing

mitchgavan:
Fixes the parsing of custom properties (CSS variables) names in object syntax. The casing is now preserved instead of being converted to kebab case.

### DIFF
--- a/packages/babel-plugin/src/__tests__/css-builder.test.ts
+++ b/packages/babel-plugin/src/__tests__/css-builder.test.ts
@@ -1,6 +1,24 @@
 import { transform } from '../test-utils';
 
 describe('css builder', () => {
+  it('should convert CSS properties to kebab-case', () => {
+    const actual = transform(`
+      import '@compiled/react';
+      <div css={{ backgroundColor: 'red' }} />
+    `);
+
+    expect(actual).toInclude('background-color:red');
+  });
+
+  it('should preserve custom property value casing', () => {
+    const actual = transform(`
+      import '@compiled/react';
+      <div css={{ '--panelColor': 'red', '--PANEL_WIDTH': 280 }} />
+    `);
+
+    expect(actual).toIncludeMultiple(['--panelColor:red', '--PANEL_WIDTH:280px']);
+  });
+
   it('should keep nested unconditional css together', () => {
     const actual = transform(`
       import '@compiled/react';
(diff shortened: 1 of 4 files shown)


packages/babel-plugin/src/__tests__/css-builder.test.ts
```diff
@@ -1,6 +1,24 @@
...
+      import '@compiled/react';
+      <div css={{ backgroundColor: 'red' }} />
+    `);
+
+    expect(actual).toInclude('background-color:red');
+  });
+
+  it('should preserve custom property value casing', () => {
+    const actual = transform(`
+      import '@compiled/react';
```
itsdouges, line 15: Do we need a test for kebab-case vars?

packages/babel-plugin/src/__tests__/css-builder.test.ts
```diff
@@ -1,6 +1,24 @@
 import { transform } from '../test-utils';
 
 describe('css builder', () => {
+  it('should convert CSS properties to kebab-case', () => {
+    const actual = transform(`
+      import '@compiled/react';
+      <div css={{ backgroundColor: 'red' }} />
```
itsdouges, line 7: Can you write a test with cssMap + css funcs please 🙏

itsdouges (commented):
Looks great mate! Left a few comments.

itsdouges:
Can you add a patch changeset for all bundler packages as well please so they immediately get the bug fix? 🙏 (will ensure their dependency of babel is brought forward).

packages/babel-plugin/src/__tests__/css-builder.test.ts
dddlr: and styled too

(perhaps xcss? not sure if worth it)

packages/babel-plugin/src/__tests__/css-builder.test.ts
dddlr: i would add it yeah

dddlr:
thank you for the PR @mitchgavan :heart: LGTM once extra tests are added (douges' comments)

packages/babel-plugin/src/__tests__/css-builder.test.ts
mitchgavan: Sure, I've added the tests

mitchgavan:
Thanks folks, I've added the tests and changesets for the other packages.

JakeLane (approved):

itsdouges (approved):

liamqma (approved):

PR #1665, opened 2024-04-18
Add custom property values to loose XCSSProp

AndrewOCC:
Currently, the loose `XCSSProp` allows makers to type an `xcss` prop that only accepts certain CSS properties (for example, limiting a text component to only accept `color`. And `create-strict-api` allows libraries to create a restricted `XCSSProp` with limited values.

However, there's no lightweight approach for a component to customize `XCSSProp` to accept both a restricted set of properties, and restricted values. After pairing on the issue with @itsdouges , he suggested the following change!

This PR augments `XCSSProp` to accept an object as the first generic `TAllowedProperties`, in addition to the existing string union type. When TAllowedProperties is set to an object, the required properties for XCSSProp are set automatically, and so `requiredProperties` is set to `never`.

### DIFF
--- a/packages/react/src/xcss-prop/__tests__/xcss-prop.test.tsx
+++ b/packages/react/src/xcss-prop/__tests__/xcss-prop.test.tsx
@@ -267,13 +267,12 @@ describe('xcss prop', () => {
     }: {
       xcss: XCSSProp<
         {
-          color: 'color.red' | 'color.blue';
+          color: 'color.text.red' | 'color.text.blue';
         },
-        never,
-        { requiredProperties: 'color' }
+        never
       >;
       // As with standard XCSS, invalid XCSS attributes don't error
-      xcssWrong?: XCSSProp<{ color: 'color.red'; invalidCSSAttr: 'color.red' }, never>;
+      xcssWrong?: XCSSProp<{ color: 'color.text.red'; invalidCSSAttr: 'invalid' }, never>;
     }) {
       return <div className={xcss}>foo</div>;
     }
@@ -282,7 +281,7 @@ describe('xcss prop', () => {
       <>
         <CSSPropComponent
           xcss={{
-            color: 'color.red',
+            color: 'color.text.red',
           }}
         />
         <CSSPropComponent
@@ -293,10 +292,6 @@ describe('xcss prop', () => {
             backgroundColor: 'color.background',
           }}
         />
-        <CSSPropComponent
-          // @ts-expect-error - Property 'color' is missing in type '{}'.
-          xcss={{}}
-        />
       </>
     ).toBeObject();
   });
@@ -305,16 +300,27 @@ describe('xcss prop', () => {
     function CSSPropComponent({
       xcss,
     }: {
-      xcss: XCSSProp<{ color: 'color.red' | 'color.blue' }, never, { requiredProperties: 'color' }>;
+      xcss: XCSSProp<
+        { color: 'color.text.red' | 'color.text.blue'; borderColor?: 'color.border' },
+        never
+      >;
     }) {
       return <div className={xcss}>foo</div>;
     }
 
     expectTypeOf(
-      <CSSPropComponent
-        // @ts-expect-error — Property 'color' is missing in type '{}' but required in type {color: 'color.red' | 'color.blue'}.
-        xcss={{}}
-      />
+      <>
+        <CSSPropComponent
+          // @ts-expect-error — Property 'color' is missing in type '{}' but required in type {color: 'color.red' | 'color.blue'}.
+          xcss={{}}
+        />
+        <CSSPropComponent
+          xcss={{
+            color: 'color.text.red',
+            // Border color is optional
+          }}
+        />
+      </>
     ).toBeObject();
   });
 });
--- a/packages/react/src/xcss-prop/__tests__/xcss-prop.test.tsx
+++ b/packages/react/src/xcss-prop/__tests__/xcss-prop.test.tsx
@@ -328,18 +328,21 @@ describe('xcss prop', () => {
     function CSSPropComponent({
       xcss,
     }: {
-      xcss: XCSSProp<
-        {
-          color: 'color.text.red' | 'color.text.blue';
-          borderColor?: 'color.border';
-          '&:hover': {
-            color: 'color.text.green';
-          };
-          // All pseudos are optional; optional pseudos don't work as expected currently
-          '&:active'?: object;
-        },
-        never
-      >;
+      xcss: XCSSProp<{
+        color: 'color.text.red' | 'color.text.blue';
+        borderColor?: 'color.border';
+        '&:hover': {
+          // overrides valid color values above
+          color: 'color.text.green';
+          // borderColor can still be set, restricted to values above
+        };
+        '&:focus': {
+          // setting color to `never` should prevent it from being set
+          color: never;
+        };
+        // Pseudos are already optional; optional pseudos don't work as expected
+        '&:active'?: object;
+      }>;
       xcssInvalid?: XCSSProp<
         {
           color: 'color.text.red' | 'color.text.blue';
@@ -389,6 +392,13 @@ describe('xcss prop', () => {
           color: 'invalid color',
         },
       },
+      validPseudoInvalidProperty: {
+        color: 'color.text.red',
+
+        '&:focus': {
+          color: 'color.text.red',
+        },
+      },
       invalidPseudo: {
         color: 'color.text.red',
         '&::after': {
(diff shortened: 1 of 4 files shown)


packages/react/src/xcss-prop/index.ts
```diff
@@ -145,23 +145,31 @@ export type XCSSAllPseudos = CSSPseudos;
  * <Component xcss={styles.text} />
  * ```
  *
- * To concatenate and conditonally apply styles use the {@link cssMap} {@link cx} functions.
+ * To concatenate and conditionally apply styles use the {@link cssMap} {@link cx} functions.
  */
 export type XCSSProp<
-  TAllowedProperties extends keyof StrictCSSProperties,
+  TAllowedProperties extends keyof StrictCSSProperties | StrictCSSProperties, // extend StrictCSSProperties with CSS vars.
```
kylorhall-atlassian, line 151: This won't work deeply though, you cannot restrict pseudos or anything, I don't think they will ever be allowed.

This is how I did it before, to swap out the entire `TSchema` thing: https://github.com/atlassian-labs/compiled/pull/1637/files#diff-1d4b97255cb11e9b25307047d51c7f2c1d94052d4626cc8e19c410a8b0444788R431 and unfortunately that _might_ be the only way to make that work without a major refactor.

I do like the idea of just taking a schema here upfront, _maybe_ there's a world in which the type becomes
```tsx
import type { StrictXCSSProp, DesignTokenSchema } from '@atlaskit/css';

type IconSchema = Omit<DesignTokenSchema, …> & { color: … | 'currentColor' };

function Icon (props: {
  xcss?: StrictXCSSProp<IconSchema, { requiredProperties: 'color' }>
}) { 
```

kylorhall-atlassian (changes requested):
This doesn't look to work with deep objects (eg. pseudos, media queries), and it's got a bit left to touch either way…a quick comment.

I'd recommend this test, it could guide this: https://github.com/atlassian-labs/compiled/pull/1637/files#diff-1d4b97255cb11e9b25307047d51c7f2c1d94052d4626cc8e19c410a8b0444788R420

packages/react/src/xcss-prop/index.ts
kylorhall-atlassian: …but that still doesn't work in the test I referenced, so not sure, didn't dive in terribly deep.

AndrewOCC: @kylorhall-atlassian good catch on the filtering issue; had another look over, added the filtering logic for pseudos, auto-inferred the value for the second generic, and added more tests around pseudos.

Re: working deeply, what behaviour are you looking for? Pseudos seem to be working:

 ```tsx
 interface MyComponentProps {
   xcss: XCSSProp<{
     // Properties can be optional or required, and can have strict types
     color: 'color.text.red' | 'color.text.blue';
     borderColor?: 'color.border';
     // Pseudos are always optional, but you can define what pseudos are valid
     '&:hover': {
       // Setting a value in a pseudo lets you further refine the valid values
       color: 'color.text.green';
     };
     '&:focus': {
       // setting color to `never` prevents it from being set
       color: never;
     };
   }>
 }
 ```

With some limitations:
- You can't mark a pseudo as required (this isn't possible with XCSSProp as I understand, it just seems intuitive to be able to do this when you're passing in a schema as the first generic)
- If a pseudo _is_ optional, you can't customise the properties it accepts (not sure why)  
- If you want to prevent pseudos from accepting certain values, you need to explicitly re-type them, or set them to `never`

>You will have to actually change the underlying implementation, just the loose one probably isn't sufficient

Could you clarify this a bit? Currently this PR is just intended for the loose variant of XCSSProp - are you requesting that the functionality be extended to the strict version for consistency?

packages/react/src/xcss-prop/index.ts
```diff
@@ -145,23 +171,35 @@ export type XCSSAllPseudos = CSSPseudos;
...
+  TAllowedPseudos extends TAllowedProperties extends object ? never : CSSPseudos,
   TRequiredProperties extends {
-    requiredProperties: TAllowedProperties;
+    requiredProperties: TAllowedProperties extends object ? never : TAllowedProperties;
   } = never
 > = Internal$XCSSProp<
-  TAllowedProperties,
-  TAllowedPseudos,
+  TAllowedProperties extends object ? FilterPseudos<keyof TAllowedProperties> : TAllowedProperties,
```
kylorhall-atlassian, line 183: On second thought, wouldn't this work?  We already use this syntax and have the keys to pick this side of it 🤦 my bad, anyways, still helps to filter/extract here.
```suggestion
  TAllowedProperties extends object
    ? Extract<keyof TAllowedProperties, keyof StrictCSSProperties>
    : TAllowedProperties,
```

packages/react/src/xcss-prop/__tests__/xcss-prop.test.tsx
```diff
@@ -260,4 +260,187 @@ describe('xcss prop', () => {
...
+          color: 'color.text.red' | 'color.text.blue';
+        },
+        // @ts-expect-error - Type 'string' is not assignable to type 'never'.
+        '&:hover' | '&:active'
+      >;
+    }) {
+      return <div className={xcss}>foo</div>;
+    }
+
+    const validStyles = cssMap({
```
kylorhall-atlassian, line 360: Problem is here I can't set `&:focus` at all, eg.
```tsx
'&:focus': {
  borderColor?: 'color.border';
}
```

Results in this: 
Screenshot 2024-04-19 at 8 51 15 PM

It expects `'&:focus'` to have `{ readonly color: never; }` because `color: 'red'` becomes a `requiredProperties` and passed down into `XCSSPseudo<…>`, so that doesn't quite work.  We almost have to remove `requiredProperties` entirely and just replace it with an object…which, frankly, might even just be better entirely.

---

Maybe it's just better to say you can't even have, in both loose, and strict, a non-object-based `XCSSProp<T>`—but I feel like we tried this and it's a rabbit hole…I could be wrong 😅

packages/react/src/xcss-prop/index.ts
kylorhall-atlassian: Got it, maybe this `Filter<…>` fixed my initial issue with deep pseudos blowing up—now it _mostly works_.

Originally, I figured we'd have to touch `Internal$XCSSProp` (and I still think so if we want to touch on any edge-cases)—that's the main reason for the comment.  A few minor deviations isn't ideal as the APIs are already incongruent, but if we go so deep (which I guess I expected), it feels painful.

> If you want to prevent pseudos from accepting certain values
Testing your above doesn't work though—it expects `'&:focus'` to have `{ readonly color: never; }` because `color: 'red'` becomes a `requiredProperties` and passed down into `XCSSPseudo<…>`, so that doesn't quite work.
```tsx
<CSSPropComponent
  xcss={{
    color: 'color.text.red',
    '&:hover': { color: 'color.text.green', borderColor: 'color.border' },
    '&:focus': { borderColor: 'color.border' }, // errors
  }}
/>
```
I think that actually just works fine in media queries, because we don't deeply require them, IIRC…

AndrewOCC:
Closing this PR; after discussion with @kylorhall-atlassian we found that although this feature is desired, a more robust implementation through a separate API may be necessary. In addition, there are implications for other parts of the compiled API works - for example, how do you support tight types in CSSMap when the types are different per-component?

kylorhall-atlassian:
> For example, how do you support tight types in CSSMap when the types are different per-component?

Yeah, I think this change is required for us to move the needle in the future—but what does this look like?  How do we allow `color: Text | IconFill` color schemes in a shared `cssMap` where we only want `color: Text` by default?

A generic override and on the common css source?
```tsx
import { cssMap } from '@atlaskit/css';
import { Icon, type IconSchema } from '@atlaskit/icon';
const styles = cssMap<IconSchema>({ primary: { color: 'var(--ds-icon-color)' } });
export default () => <Icon xcss={styles.primary} />;
```

Per-package css exports from every package:
```tsx
import { Icon, cssMap } from '@atlaskit/icon';
const styles = cssMap({ primary: { color: 'var(--ds-icon-color)' } });
export default () => <Icon xcss={styles.primary} />;
```

Very loose interfaces with strict component-level interfaces
```tsx
import { cssMap } from '@atlaskit/css';
import { Button } from '@atlaskit/Button';
import { Icon } from '@atlaskit/Icon';

// This `color` interface allows for both icon and text colors.
const styles = cssMap({ primary: { color: 'var(--ds-icon-color)' } });

export default () => <>
  {/* This is fine because it allows icon-specific colors */}
  <Icon xcss={styles.primary} />
  {/* This errors because it doesn't allow icon colors */}
  <Button xcss={styles.primary} />
</>
```

Issue #1306, opened 2022-10-31
Parcel doesn't pick up new CSS change on local development when stylesheet extraction is enabled.


liamqma:
**Describe the bug**
Parcel doesn't pick up new CSS change on local development when stylesheet extraction is enabled.

**To Reproduce**

```
git clone git@github.com:atlassian-labs/compiled.git
yarn
yarn run start:parcel
```

- Wait for the web page to load
- Make a CSS change 
- The new change doesn't appear even if refreshing the browser

**Expected behavior**
The newly made CSS change should appear on local development

**Additional context**
This bug is introduced from https://github.com/atlassian-labs/compiled/pull/1303. parcel-optimizer is used to get stylesRules from asset.metadata and insert it to HTML. When a CSS change is made, Parcel won't know a need to re-run parcel-optimizer because there isn't a dependency between HTML and css change in the js file.

The workaround is to disable stylesheet extraction on local but the proper solution is to tell Parcel that the HTML has a dependency on the CSS.

PR #1666, opened 2024-04-23
Add media query sorting

dddlr:
Add a deterministic sorting to media queries and other at-rules in Compiled. We use a simplified version of what the [`sort-css-media-queries`](https://github.com/OlehDutchenko/sort-css-media-queries?tab=readme-ov-file#mobile-first) package does - sorting `min-width` and `min-height` from smallest to largest, then `max-width` and `max-height` from largest to smallest

We also merge the `sort-at-rule-pseudos` and `sort-atomic-style-sheet` postcss plugins into one, as they have a lot of overlap. Previously:

* `sort-atomic-style-sheet` was used in stylesheet extraction (creating the `compiled-css.css` stylesheet for the project) -- it sorted pseudo-selectors and now also sorts media queries.
* `sort-at-rule-pseudos` only sorted pseudo-selectors that appeared inside at-rules (like `@media` queries). It was used in conjunction with `@compiled/react/runtime` which [organises the pseudo-selectors into buckets](https://github.com/atlassian-labs/compiled/blob/master/packages/react/src/runtime/sheet.ts) when stylesheet extraction is not turned on... each bucket being a separate `<style>` element, with one bucket for each pseudo-selector.

Now the two are one and the same. Technically when stylesheet extraction is not turned on, pseudo-selectors will get sorted twice: (1) pseudo-selectors are sorted as part of the aforementioned postcss plugin in the build stage, then (2) the pseudo-selector styles will get collated and organised into buckets (separate `<style>` elements) at runtime. This is necessary because step (1) only sorts the styles for a single component, not all the styles for the current page.

### Limitations

* PostCSS plugins only act on the styles for a single component at a time. Media queries will lose their sorting when the styles of multiple components in the same page are concatenated together by `@compiled/react/runtime`.
  * This is not an issue when using Webpack, and `extract: true` is turned on for `@compiled/webpack-loader`. However, this is an issue for Parcel when in development mode, because stylesheet extraction is disabled in development mode there: https://github.com/atlassian-labs/compiled/issues/1306
  * ~Should we add a sorting step to `@compiled/react/runtime` too?~ I tested this with a toy example and looks like the postcss processing makes `@compiled/react/runtime` 100x slower... this doesn't seem to be worth it to me, so I didn't put any effort looking into it further. Happy to reconsider if y'all think it's worth the performance degradation.

### Checklist

* [x] Complete tests
* [x] Update `packages/css/src/plugins/sort-at-rule-pseudos.ts` so that media query sorting isn't dependent on stylesheet extraction being turned on lol
* [x] Determine whether it is worth it to have this running in `@compiled/react/runtime` too?
  * My answer is: nah
* [x] Update snapshots
* [x] Better distinguis `packages/css/src/plugins/__tests__/sort-atomic-style-sheet.test.ts` from `packages/css/src/plugins/__tests__/sort-at-rules.test.ts`
* [x] Double-check that this works in an example app importing several `.compiled.css` stylesheets
* [x] Improve docstrings for functions
* [x] Ensure that other at-rules are handled too (`@container`)
* [ ] Make PR to update Compiled website
  * [ ] Detail the limitations around media query sorting when importing `.compiled.css` stylesheets, and when using Parcel in development mode (no stylesheet extraction)
  * [ ] Detail the limitations that atomic CSS and `@layer` should not be used together, because our sorting can wreck `@layer` behaviour
* [x] Add support for device-width, device-height
* [x] Add changeset
* [x] Verify that this doesn't crash for Atlassian products (platform, jira)


packages/css/src/plugins/at-rules/parse-at-rule.ts
```diff
@@ -0,0 +1,106 @@
...
+} from './parsers';
+import type { ParsedAtRule, Situations } from './types';
+
+const comparisonOperators = /(?<operator>(?:<=?)|(?:>=?)|=)\s*/;
+const comparisonOperators2 = /(?<operator2>(?:<=?)|(?:>=?)|=)\s*/;
+const property = /(?:(?<property>(?:(?:min|max|device)-)?(?:width|height))\s*)/;
+const widthOrHeight = /(?<property>width|height)\s*/;
+const colon = /(?<colon>:\s*)/;
+
+const length_ = /(?<length>-?\d*\.?\d+)(?<lengthUnit>ch|em|ex|px|rem)?\s*/;
```
dddlr, line 15: `length` caused some typescript issues, so going with `length_` instead...

packages/css/src/plugins/__tests__/sort-at-rules.test.ts
```diff
@@ -0,0 +1,849 @@
+import postcss from 'postcss';
```
dddlr, line 1: let me know if you'd like me to split this file in two

packages/css/src/plugins/at-rules/parse-at-rule.ts
```diff
@@ -0,0 +1,111 @@
...
+  parseSituationTwo,
+} from './parsers';
+import type { ParsedAtRule, Situations } from './types';
+
+const comparisonOperators = /(?<operator>(?:<=?)|(?:>=?)|=)\s*/;
+const comparisonOperators2 = /(?<operator2>(?:<=?)|(?:>=?)|=)\s*/;
+// Matches min-width, min-height, max-width, max-height,
+// min-device-width, min-device-height, max-device-width, max-device-height,
+// width, height, device-width, and device-height.
+const property = /(?:(?<property>((?:min|max)-)?(?:device-)?(?:width|height))\s*)/;
```
dddlr, line 14: ok i know regex is hard to read 😓 

you can treat this as being equivalent to

```
/((((min|max)-)?(device-)?(width|height))\s*)/
```

ignore the `?:` stuff, this just makes a [non-capturing group](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Non-capturing_group)

the `(?<text>` stuff is a [named capturing group](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Named_capturing_group) which lets me "store" parts of the match as values in the `match.groups` object

https://regexr.com/ is also helpful here

packages/css/src/plugins/at-rules/parse-at-rule.ts
```diff
@@ -0,0 +1,111 @@
+import {
+  parseSituationFour,
+  parseSituationOne,
+  parseSituationThree,
+  parseSituationTwo,
+} from './parsers';
+import type { ParsedAtRule, Situations } from './types';
+
+const comparisonOperators = /(?<operator>(?:<=?)|(?:>=?)|=)\s*/;
```
dddlr, line 9: equivalent to

```
/((<=?)|(>=?)|=)\s*/
```

packages/css/src/plugins/sort-at-rule-pseudos.ts
dddlr, line 1: merged into `sort-atomic-style-sheet.ts`

liamqma (approved):

.changeset/two-carrots-poke.md
```diff
@@ -0,0 +1,7 @@
+---
+'@compiled/css': minor
+---
+
+[BREAKING] Add a deterministic sorting to media queries and other at-rules in Compiled. We use a simplified version of what the [`sort-css-media-queries`](https://github.com/OlehDutchenko/sort-css-media-queries?tab=readme-ov-file#mobile-first) package does - sorting `min-width` and `min-height` from smallest to largest, then `max-width` and `max-height` from largest to smallest.
```
itsdouges, line 5: Do we need to release a breaking version of the bundler / babel plugins as well?

packages/css/src/plugins/__tests__/sort-at-rules.test.ts
```diff
@@ -0,0 +1,845 @@
...
+      `);
+
+      expect(actual).toMatchInlineSnapshot(`
+        "
+                @layer module {
+                  p {
+                    color: pink;
+                  }
+                }
+                @layer module, state;
```
itsdouges, line 812: What's the rule for sorting layers?

packages/css/src/plugins/at-rules/parse-at-rule.ts
itsdouges: Oh :-)

packages/css/src/plugins/at-rules/parse-at-rule.ts
```diff
@@ -0,0 +1,111 @@
...
+        length_.source +
+        comparisonOperators.source +
+        property.source +
+        comparisonOperators2.source +
+        length2.source,
+      parser: parseSituationFour,
+    },
+  ];
+
+  for (const { regex, parser } of situations) {
```
itsdouges, line 95: Were there any AST based solutions available?

packages/css/src/plugins/at-rules/parsers.ts
```diff
@@ -0,0 +1,231 @@
...
+    return;
+  }
+
+  if (isPropertyValid(property)) {
+    return {
+      property,
+    };
+  }
+
+  throw new SyntaxError(`Unexpected property '${property}' found when sorting media queries.`);
```
itsdouges, line 106: Should this category of invariant (here and below) be mentioned in the changeset?

packages/css/src/plugins/at-rules/parsers.ts
```diff
@@ -0,0 +1,231 @@
...
+
+const getBasicMatchInfo = (match: RegExpMatchArray): BasicMatchInfo | undefined => {
+  if (!match.index) {
+    return;
+  }
+
+  return { index: match.index, match: match[0] };
+};
+
+export const parseSituationOne = (match: RegExpMatchArray): ParsedAtRule | undefined => {
```
itsdouges, line 152: Having these names as what they're working with would be more ideal. None of us will remember what each situation is later on (yes it's somewhere above in a comment but still).

packages/css/src/plugins/at-rules/parse-at-rule.ts
```diff
@@ -0,0 +1,111 @@
...
+    for (const match of matches) {
+      const parsedMatch = parser(match);
+      if (parsedMatch) {
+        parsedMatches.push(parsedMatch);
+      }
+    }
+  }
+
+  // Sort matches from first to last, as they appear in the at-rule / media query
+  parsedMatches.sort((a, b) => {
```
itsdouges, line 106: What is this sort needed for? Vs. the sort we do in the plugin below.

packages/css/src/plugins/sort-atomic-style-sheet.ts
```diff
@@ -1,20 +1,40 @@
...
- * @returns
- */
-const pseudoSelectorScore = (selector: string) => {
-  const index = styleOrder.findIndex((pseudoClass) => selector.trim().endsWith(pseudoClass));
-  return index + 1;
+import { parseAtRule } from './at-rules/parse-at-rule';
+import { sortAtRules } from './at-rules/sort-at-rules';
+import type { AtRuleInfo } from './at-rules/types';
+
+const sortAtRulePseudoSelectors = (atRule: AtRule) => {
```
itsdouges, line 9: Brought this in to reduce duplicate behaviour?

packages/react/src/__tests__/browser.test.tsx
```diff
@@ -100,7 +100,7 @@ describe('browser', () => {
       <style nonce="k0Mp1lEd">._f8pjbf54:focus{color:green}</style>
       <style nonce="k0Mp1lEd">._30l31gy6:hover{color:yellow}</style>
       <style nonce="k0Mp1lEd">._9h8h13q2:active{color:blue}</style>
-      <style nonce="k0Mp1lEd">@supports (display:grid){._1df61gy6:focus{color:yellow}._7okp11x8:active{color:black}}@media (max-width:800px){._1o8z1gy6:focus{color:yellow}._jbabtwqo:focus-visible, ._6146twqo:hover{color:grey}._1cld11x8:active{color:black}}</style>
```
itsdouges, line 103: Where'd the supports go?

.changeset/two-carrots-poke.md
dddlr: hmm that's a good point 🤔 yeah i'll add the bundler plugins and babel plugin to this

packages/react/src/__tests__/browser.test.tsx
dddlr: wdym, its still there
originally it was `@supports` then `@media`, but because the at-rule sorting kicks in, we get `@media` then `@supports` now

packages/css/src/plugins/at-rules/parse-at-rule.ts
dddlr: there are a couple

* https://github.com/tbjgolden/media-query-parser
* https://github.com/ericf/css-mediaquery

but i didn't feel comfortable importing them, as they had very few stars (9 and 111) which made me not confident from a maintainability perspective

packages/css/src/plugins/sort-atomic-style-sheet.ts
dddlr: ye

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.ts
```diff
@@ -449,8 +449,8 @@ describe('css prop behaviour', () => {
     `);
 
     expect(actual).toIncludeMultiple([
-      'const _2="._y44vk4ag{animation:fadeOut 2s ease-in-out}"',
-      'const _="@keyframes fadeOut{0%{opacity:1}50%{opacity:0.5}to{opacity:0}}"',
+      'const _2="@keyframes fadeOut{0%{opacity:1}50%{opacity:0.5}to{opacity:0}}"',
```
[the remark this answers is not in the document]
dddlr: i believe this is because at rules are sorted so that they always come after non at-rules now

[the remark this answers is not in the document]
dddlr: > The order here wouldn't matter right

yeah `@keyframes` can go anywhere

packages/css/src/plugins/at-rules/parse-at-rule.ts
dddlr: this isn't really a sorting step 😅 it just makes sure that if we get cases like

```css
@media (width > 400px) or (width < 200px) { /* ... */ }
```

we want `parsedMatches` to preserve the order of the `width` conditions... in other words, we want to return something like `['width > 400px', 'width < 200px']`, and not `['width < 200px', 'width > 400px']`

this is so that when we compare this against another media query, e.g.

```css
@media (max-width: 200px) and (min-width: 100px) { /* ... */ }
```

we'll compare `width > 400px` with `max-width: 200px`, and `width < 200px` with `min-width: 100px` (after the two are parsed)


i'll add this to the comment for clarity

packages/css/src/plugins/at-rules/parsers.ts
dddlr: hmm

assuming i didn't screw up somewhere, the only situation where this error would be triggered if someone used `max`/`min` in a place where it wasn't valid CSS syntax, such as `@media (max-width < 500px)`. I suppose we would go from silently accepting invalid CSS to now giving a build-time error.

the comparison operator error would only be triggered if there is a bug in the code 😅

i suppose we *could* add this to the changeset, but not sure how much value this would provide

packages/css/src/plugins/at-rules/parse-at-rule.ts
itsdouges: Could this have been a `.reverse()`?

packages/react/src/__tests__/browser.test.tsx
itsdouges: Oops. Okay!

itsdouges:
Over arching question before sending it: Should this be under a config option so we can turn it off just in case things go wrong?

As a follow up: What happens during runtime / non-extracted mode? If nothing, what about increasing specificity just during dev?

packages/css/src/plugins/at-rules/types.ts
```diff
@@ -0,0 +1,84 @@
+import type { AtRule, Rule } from 'postcss';
+
+export type Property = 'width' | 'height' | 'device-width' | 'device-height';
+export type ComparisonOperator = '<=' | '=' | '>=' | '<' | '>';
+export type LengthUnit = 'ch' | 'em' | 'ex' | 'px' | 'rem';
```
kylorhall-atlassian, line 5: I could be missing it, but if we see say `min-width: 32px` and `min-width: 2rem`, are these sorted as equals±?  Technically we can't make assumptions on some of these, so not even sure how we would.

Just not sure if it's said anywhere, we make assumptions about the sorting, but it's deterministic…I think at the end of the day, that's it, it's deterministic and should be fit for our purpose—could have some hard bugs as this code is a **_lot_** 😮 (good job), but should fix more bugs than it has!

packages/css/src/plugins/at-rules/parsers.ts
kylorhall-atlassian: We would probably know real quick in our primary codebases, right?  Could be worth a major if it has a semi-reasonable chance to throw odd errors here for existing setups, but up to you.

packages/css/src/plugins/__tests__/sort-at-rules.test.ts
kylorhall-atlassian: Oh, interesting… it's worth noting that sorting `@layer` would be antithetical to the w3 spec as the definition-based order is heavily important…but does this ever fit into atomic css, or should layers require some `@keyframes` type hoisting where they must be globally ordered by another means, possibly manually by the user.

Does not impact us, just might be worth a note in a changeset or some page that atomic css and `@layer` conflict.

…that being said, again, going from possibly non-deterministic to deterministic, and the solution before and after this sorting would always be to define `@layer` upfront—if anyone's actually using Compiled with layers, they must do this, or they probably don't know what they're doing.  LGTM.

.changeset/two-carrots-poke.md
kylorhall-atlassian: Does `[BREAKING]` do anything special?  To me `breaking === major`, but I won't nitpick anything here, some of these might even be `0.x` where `minor === major` anyways…just thinking aloud.

kylorhall-atlassian (commented):
Sorry to not get around to a full review, love the work, tons of complexity, and I think you already got pretty good coverage and deeper thought than I have in me right now.

I didn't dig in a ton into most of the code—I tried to have a real think about your comments and description, and I think you've addressed the problem statements I had quite well.  It's definitely a beta type feature, I think the determinism will solve far, far more than it breaks, most of the misuse today would already be broken anyways, so I'm very happy with the output from what I looked at and diving into the tests.

.changeset/two-carrots-poke.md
dddlr: not really, it's just flagging that something _really_ can break

packages/css/src/plugins/at-rules/types.ts
dddlr: > Technically we can't make assumptions on some of these, so not even sure how we would. Might just be worth documenting, we only sort against same unit'd values and expect all media queries to be the same unit…? (if I'm understanding the code well)

oh yeah it would be pretty easy to normalise different units

do you anticipate any developers mixing units like this? i wasn't sure about how much value it would provide

kylorhall-atlassian: I do anticipate mixing `rem` and `px` as ADS ships `rem` and I see `px` in use a lot actually, I think even within our products or WAC you might find this dozens of times.

_Perhaps_ having some like `mediaQueryRemRatio: 16` defaulted/set somewhere to convert `em` and `rem` to `px` and normalize around `px` (configurable or not, I don't care) could work?

I can't tell you what `ch` or `ex` to `px` is though, and this might just be scope creeping, maybe worth testing what mixing `rem` and `px` looks like is sufficient…

dddlr: ahh right 👍  yeah i'll add it in

packages/css/src/plugins/at-rules/parsers.ts
dddlr: ah minor is what we use as a de-facto "major release" because we still don't have a v1.0 release yet

we should definitely release a v1.0 eventually but haven't had the chance to do it yet

dddlr:
> Over arching question before sending it: Should this be under a config option so we can turn it off just in case things go wrong?

good point yeah @itsdouges i'll add that in a commit

> As a follow up: What happens during runtime / non-extracted mode? If nothing, what about increasing specificity just during dev?

ah this is covered in the PR description, let me know if anything is unclear (i'm not aware of any practical way to increase specificity for media queries - `@layer` would *decrease* specificity for Compiled styles, and we can't know in advance how many `:not(#\#)` to add)

> * PostCSS plugins only act on the styles for a single component at a time. Media queries will lose their sorting when the styles of multiple components in the same page are concatenated together by `@compiled/react/runtime`.
>   * This is not an issue when using Webpack, and `extract: true` is turned on for `@compiled/webpack-loader`. However, this is an issue for Parcel when in development mode, because stylesheet extraction is disabled in development mode there: https://github.com/atlassian-labs/compiled/issues/1306
>   * ~Should we add a sorting step to `@compiled/react/runtime` too?~ I tested this with a toy example and looks like the postcss processing makes `@compiled/react/runtime` 100x slower... this doesn't seem to be worth it to me, so I didn't put any effort looking into it further. Happy to reconsider if y'all think it's worth the performance degradation.

packages/css/src/plugins/at-rules/parse-at-rule.ts
dddlr: ah sorry, i left off a bit of important context in my previous message

the `for (const { regex, parser } of situations) { ... }` loop above this means that any parts of the media query that match situation one (e.g. `min-width: 300px`) will appear in `parsedMatches` before any features that match situation two (e.g. `200px >= height`), which will appear before any features that match situations three and four (e.g. `0 <= width <= 200px`)

so if we have `@media (0 <= width <= 200px) and (200px >= height) or (min-width: 300px)`, then the three matches we found in the media query will go out of order - so we need to re-sort them to restore the correct order

.changeset/two-carrots-poke.md
```diff
@@ -0,0 +1,19 @@
...
+'@compiled/webpack-loader': minor
+'@compiled/parcel-config': minor
+---
+
+[BREAKING] Add a deterministic sorting to media queries and other at-rules in Compiled. We use a simplified version of what the [`sort-css-media-queries`](https://github.com/OlehDutchenko/sort-css-media-queries?tab=readme-ov-file#mobile-first) package does - sorting `min-width` and `min-height` from smallest to largest, then `max-width` and `max-height` from largest to smallest. If ranges or features involving `height` and `width` are not present in the at-rule, the at-rule will be sorted lexicographically / alphabetically.
+
+Situations you may need to be careful of:
+
+- In situations where two at-rules with the same property apply at the same time, this may break your styles, as the order in which the styles are applied will change. (For example, overlapping `@media` queries)
+- Because all at-rules will now be sorted lexicographically / alphabetically, `@layer` blocks you pass to Compiled APIs may not be outputted in the same order, causing different CSS than expected.
```
kylorhall-atlassian, line 13: Realistically with atomic css, this was always the case with `@layer` I expect.  But good to have it noted

packages/css/src/plugins/__tests__/sort-at-rules.test.ts
```diff
@@ -0,0 +1,886 @@
...
+                  color: red;
+                }
+                @supports (display: flex) {
+                  color: blue;
+                }
+              "
+      `);
+    });
+
+    it('should normalise lengths of different units', () => {
```
kylorhall-atlassian, line 845: 👀 thank you—it's deterministic at least!  Now to find the magic…

packages/css/src/plugins/at-rules/parsers.ts
```diff
@@ -0,0 +1,220 @@
...
+  ComparisonOperator,
+  LengthInfo,
+  LengthUnit,
+  OperatorInfo,
+  ParsedAtRule,
+  Property,
+  PropertyInfo,
+} from './types';
+
+const REM_SIZE = 16;
```
kylorhall-atlassian, line 12: We know this assumption is probably wrong, even within Atlassian it's wrong in one place, but it's a good start—configurable later if it's really a problem…

packages/css/src/plugins/at-rules/parsers.ts
```diff
@@ -0,0 +1,220 @@
...
+  if (typeof length_ !== 'string' || typeof lengthUnit !== 'string') {
+    return undefined;
+  }
+
+  switch (lengthUnit as LengthUnit) {
+    case 'ch':
+    case 'ex':
+      // https://drafts.csswg.org/css-values/#font-relative-lengths
+      // We assume 1ch and 1ex are 0.5em, as we cannot rely on more specific information here.
+      return { length: +length_ * 0.5 * REM_SIZE };
```
kylorhall-atlassian, line 150: Honestly, if they're using `ch` or `ex` for a media query I'm a bit worried…could've thrown an error here instead in my books 🤣, but again, it's deterministic!

packages/css/src/plugins/at-rules/types.ts
kylorhall-atlassian: This was added, LGTM: https://github.com/atlassian-labs/compiled/pull/1666/files#diff-e6c66e3c597ed6c80303321a9daac2eaabb0f46a431faea945997f8c4eb1a23bR145

kylorhall-atlassian (commented):
No further comments from before, looking great from me, thank you!  Consider mine approval, I just didn't do a thorough review before, just looking at high-level and diff (and I'm happy).

kylorhall-atlassian (approved):

PR #1674, opened 2024-05-23
Fix jest matcher kebab-casing CSS custom properties

itsdouges:
Previously the jest matcher would always kebab case all properties. This fixes it so it bails out if the property is a custom property.

### DIFF
--- a/packages/jest/src/matchers.ts
+++ b/packages/jest/src/matchers.ts
@@ -21,7 +21,10 @@ const kebabCase = (str: string) =>
 const removeSpaces = (str?: string) => str && str.replace(/\s/g, '');
 
 const mapProperties = (properties: Record<string, any>) =>
-  Object.keys(properties).map((property) => `${kebabCase(property)}:${properties[property]}`);
+  Object.keys(properties).map((property) => {
+    const key = property.startsWith('--') ? property : kebabCase(property);
+    return `${key}:${properties[property]}`;
+  });
 
 const onlyRules = (rules?: StyleRules['rules']) => rules?.filter((r) => r.type === 'rule');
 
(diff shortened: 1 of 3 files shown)


liamqma (approved):

packages/react/src/__tests__/jest-matcher.test.tsx
```diff
@@ -124,6 +124,14 @@ describe('toHaveCompliedCss', () => {
     expect(el).toHaveCompiledCss('color', 'blue', { target: ':active' });
   });
 
+  it('should match camelCase css vars', () => {
```
kylorhall-atlassian, line 127: Did you want a test for `{ '--my-snake-var': 'red' }` as well…?

packages/jest/src/matchers.ts
```diff
@@ -21,7 +21,10 @@ const kebabCase = (str: string) =>
 const removeSpaces = (str?: string) => str && str.replace(/\s/g, '');
 
 const mapProperties = (properties: Record<string, any>) =>
-  Object.keys(properties).map((property) => `${kebabCase(property)}:${properties[property]}`);
+  Object.keys(properties).map((property) => {
+    const key = property.startsWith('--') ? property : kebabCase(property);
```
kylorhall-atlassian, line 25: To be the devil's advocate here, it could technically break something, right?

kylorhall-atlassian (commented):
I'm late in my message, but a few comments.

packages/react/src/__tests__/jest-matcher.test.tsx
itsdouges: https://github.com/atlassian-labs/compiled/pull/1676

packages/jest/src/matchers.ts
itsdouges: Maybe! I'll upgrade Jira and find out. It needs to match on both ends so I'm not too worried. If it was always kebab case - no break. If it was not kebab case, well, you just wouldn't be able to test anything because the matcher would never pass!

PR #1676, opened 2024-05-23
Test kebab case var

itsdouges:


kylorhall-atlassian (approved):
🥪 there's no kebab emoji?

PR #1677, opened 2024-05-30
Merge Compiled website into Compiled monorepo

dddlr:
We would like to merge the Compiled website repository (which is itself a monorepo-like setup) into the remainder of the Compiled monorepo.

Benefits:

* We will be able to add merge checks for Compiled contributions that enforce updating the documentation for any changes made.
  * This is a long-term solution for our documentation gap, where our API documentation is out-of-date and/or is incomplete.
  *  Since documentation updates can be in the same commit as the code changes that necessitate them, this will make finding the surrounding context for Compiled changes easier in the future.
* Make it easier for us to maintain dependencies / Renovate PRs all in one place.

In scope for this PR:

* Moving `compiled-website` into the Compiled repository, and applying the bare minimum amount of changes so that the website doesn't break.
  * I have deliberately not fixed the many instances of outdated code practices and syntaxes in the code, as this would be too much work without providing any value :(

To be added in follow-up PRs:

* [ ] Add a custom merge check that ensures that Compiled contributions also update the Compiled website (implementation method TBD)
* [ ] Remove the need for the extraneous `website/package.json`, etc with the rest of the Compiled repo, as long as it doesn't cause errors.
  * Not sure if this is possible...
* [ ] Merge `renovate.json`, `.prettierrc`, `.nvmrc` with rest of Compiled repo
* [ ] Set up Netlify on Compiled repo
* [ ] Add Compiled website's merge checks to Compiled repo
* [ ] Blog post to share benefits of this change (might be better to do this after we update the docs)

Out of scope in this PR - can do in a follow-up PR if we get time:

* Fix eslint disables and ts-expect-error comments that I needed to add to make ESLint / TypeScript pass
* Update website to use current Compiled best practices


package.json
```diff
@@ -47,7 +47,9 @@
     "test:imports": "node test/test-imports",
     "test:parcel": "jest --testMatch '**/src/**/*.parceltest.{ts,tsx}' --testEnvironment=node",
     "test:vr": "yarn loki --host host.docker.internal",
-    "test:watch": "jest --no-cache --watch"
+    "test:watch": "jest --no-cache --watch",
+    "website:build": "cd website/ && NODE_ENV=production yarn build",
```
liamqma, line 51: Can you use workspace rather than changing the directory?

```
yarn works @compiled/website build
```

website/LICENCE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
liamqma, line 1: Do we need it here since we already have one on the root?

liamqma:
Do you think if we should lock the write access to https://github.com/atlassian-labs/compiled-website/ during the transition period to prevent out of sync?

website/LICENCE
dddlr: oh nah we can remove it

dddlr:
> Do you think if we should lock the write access to https://github.com/atlassian-labs/compiled-website/ during the transition period to prevent out of sync?

hmm looks like the closest option for us would be archiving the compiled-website repo, would u be alright with this? (we can unarchive it if we need to later)

liamqma:
We can lock branch.
image

dddlr:
done, ty @liamqma

JakeLane (approved):

PR #1678, opened 2024-06-03
Add `ElementType` to the global compiled JSX namespace

jordanoverbye:
Follow up PR to PR #1617. 

This PR adds `ElementType` the global compiled JSX namespace, as right now it's only available in the automatic jsx namespace.


itsdouges (approved):
LGTM

dddlr (approved):

PR #1680, opened 2024-06-03
Add `babelrc: false` to all Babel calls, and set `parserBabelPlugins` to `typescript` and `jsx` by default

dddlr:
`babelrc: false` was passed to `transformFromAstAsync`, but not passed to `parseAsync`. Might as well make this consistent.

I also updated `caniuse` because it was giving console warnings.

---

Edit:

I've expanded the scope of the PR to set parserBabelPlugins to ['typescript', 'jsx'] by default as well, as several internal developers in the past ran into build issues due to not setting this config option in their codebases.

### DIFF
--- a/packages/parcel-transformer/src/index.ts
+++ b/packages/parcel-transformer/src/index.ts
@@ -100,6 +100,8 @@ export default new Transformer<ParcelTransformerOpts>({
 
     const program = await parseAsync(code, {
       filename: asset.filePath,
+      babelrc: false,
+      configFile: false,
       caller: { name: 'compiled' },
       rootMode: 'upward-optional',
       parserOpts: {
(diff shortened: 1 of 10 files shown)


packages/parcel-transformer/src/index.ts
```diff
@@ -100,6 +100,8 @@ export default new Transformer<ParcelTransformerOpts>({
 
     const program = await parseAsync(code, {
       filename: asset.filePath,
+      babelrc: false,
```
liamqma, line 103: To be consistant, can you do the same for webpac-loader?

dddlr: sure

dddlr:
oops let me fix some stuff

JakeLane (approved):

PR #1694, opened 2024-07-16
Add custom GitHub action and pull request template to enforce that documentation is updated

dddlr:
Added a pull request template with a checklist that asks that documentation be updated for every PR.

This is enforced by a custom GitHub action that checks whether any tasks in the checklist are marked as incomplete. If any are incomplete, the GitHub action will give an error.

You can exempt parts the PR description from the incomplete task check by adding `<!-- task-checklist-ignore-start -->` and `<!-- task-checklist-ignore-end -->` around the sections where you want to disable the check.

---

Test it out here!

* [x] Item in the checklist
* [x] Another item in the checklist
* [x] A third item in the checklist

Screenshot 2024-07-16 at 11 32 59

Screenshot 2024-07-16 at 11 34 14

---

Alternative methods considered for finding incomplete tasks in the PR description:

* https://github.com/kentaro-m/task-completed-checker-action no longer works
* https://github.com/venkatsarvesh/pr-tasks-completed-action doesn't give any way to skip the check


.github/workflows/task-checklist.yml
```diff
@@ -0,0 +1,15 @@
...
+  pull_request:
+    types: [opened, edited]
+
+jobs:
+  task-checklist:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install dependencies
+        run: cd .github/actions/task-checklist && yarn install
```
kylorhall-atlassian, line 14: I feel like you can just put this into `.github/actions/task-checklist/action.yml`, but … it's been 2 years since I build Github Actions a lot and I don't have any examples in repos I have access to nowadays; doesn't really matter.

.github/workflows/task-checklist.yml
```diff
@@ -0,0 +1,15 @@
+name: Check all tasks are completed
+
+on:
+  check_run:
```
kylorhall-atlassian, line 4: Do you want `check_run`…?  It doesn't guarantee a PR exists, you might get errors because `payload.pull_request` won't exist.

.github/workflows/task-checklist.yml
```diff
@@ -0,0 +1,15 @@
...
+  pull_request:
+    types: [opened, edited]
```
kylorhall-atlassian, line 6: These types make sense to me!  Good job here.

.github/actions/task-checklist/index.js
```diff
@@ -0,0 +1,63 @@
...
+  for (const match of matches) {
+    console.error(`- ${match}`);
+  }
+
+  console.log('---');
+  console.log(
+    'False positive? Insert <!-- task-checklist-ignore-start --> and <!-- task-checklist-ignore-end --> in the section(s) of your PR where you want to skip the check.'
+  );
+  console.log('However, with great power comes great responsibility...');
+  console.log('---');
```
kylorhall-atlassian, line 54: I think the suggestion is to use `core.info` instead of `console.log` and `core.debug` for anything you want hidden behind the debug flag.

I'd suggest changing `console.log` => `core.info`, but to be honest I can't tell you the difference.

.github/actions/task-checklist/index.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  const plural = matches.length > 1 ? 's' : '';
+  console.error(`Found incomplete task${plural}:`);
+  for (const match of matches) {
+    console.error(`- ${match}`);
+  }
+
+  console.log('---');
+  console.log(
```
kylorhall-atlassian, line 50: Could be a `core.warning` if you care at all.

.github/actions/task-checklist/index.js
```diff
@@ -0,0 +1,63 @@
...
+// - s flag required so that ".*?" matches over multiple lines
+// - "?" in ".*?" ensures that we match the smallest possible string
+// - note that m flag is not required.
+const DISABLE_COMMENT_REGEX =
+  /<!--\s*task-checklist-ignore-start\s*-->.*?<!--\s*task-checklist-ignore-end\s*-->/gs;
+
+const run = () => {
+  const body = github.context.payload.pull_request?.body;
+  if (!body) {
+    console.log('PR description empty, skipping this check.');
```
kylorhall-atlassian, line 22: Could be a `core.warning` or even `core.notice` (to annotate) if you care at all.

kylorhall-atlassian (commented):
I didn't exactly validate the logic of the regex's and all that, but generally looks good to me!  Some minor, very minor suggestions to freely ignore around logging mostly.

.github/actions/task-checklist/index.js
dddlr: ah looks like the `core` functions https://github.com/actions/toolkit/tree/main/packages/core#annotations have better integrations with GitHub's UI, good idea

kylorhall-atlassian: Yeah, I just don't exactly know what `core.info` does, the rest is helpful!

.github/workflows/task-checklist.yml
dddlr: hmm yeah you're right, i was misunderstanding what `check_run` actually did

.github/workflows/task-checklist.yml
dddlr: had a quick look into this - doesn't look like [javascript actions](https://docs.github.com/en/actions/creating-actions/creating-a-javascript-action) support calling shell commands out of the box

[composite actions](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) seems like a potential way to make this possible, will try it out

.github/actions/task-checklist/yarn.lock
JakeLane, line 1: Atlassian NPM mirror

.github/workflows/task-checklist.yml
kylorhall-atlassian: Yeah, that was it, composite actions…you do you though, no dramas here, it's clear what you're doing.

.github/actions/task-checklist/index.js
dddlr: Screenshot 2024-07-16 at 14 19 57
Screenshot 2024-07-16 at 14 20 42

nifty

.github/actions/task-checklist/.npmrc
```diff
@@ -0,0 +1 @@
+registry="https://registry.yarnpkg.com"
```
dddlr, line 1: ensure we don't accidentally use atlassian NPM mirror

kylorhall-atlassian (approved):
Just looking at changes since last review, I think you covered it (and the NPM issue), LGTM.

PR #1696, opened 2024-07-18
Allow `ax()` to handle non-string classnames when collected.

kylorhall-atlassian:
I want to allow our `ax()` collection to better handle arbitrary values.  The example being passing using our pass-through styles with `cssMap` like `<Component xcss={cx(…)} />` or `<Component xcss={[…]} />` results in `ax` capturing things like this:
```tsx
[
  '_1bah1yb4',
  undefined,
  ['_16jlkb7n', undefined],
  AtomicGroups {
    values: Map(2) { '_1bsb' => '_1bsb1osq', '_16jl' => '_16jlkb7n' }
  }
]
```

I don't know if it's the best way to handle this, but internally we've just been invoking `toString()` via a wrapper, and we're trying to get to native Compiled.  Makes sense (and works) to me!

### PR checklist

I have...

- [x] Chosen to not document this.

### DIFF
--- a/packages/react/src/runtime/ax.ts
+++ b/packages/react/src/runtime/ax.ts
@@ -1,4 +1,4 @@
-import { AtomicGroups } from './ac';
+import type { AtomicGroups } from './ac';
 
 const UNDERSCORE_UNICODE = 95;
 
@@ -52,11 +52,10 @@ export default function ax(
     if (typeof cls === 'string') {
       groups = cls.split(' ');
     } else if (Array.isArray(cls)) {
-      groups = cls;
-    } else if (cls instanceof AtomicGroups) {
-      groups = cls.toString().split(' ');
+      groups = cls; // NOTE: Not really a valid scenario, so we don't deeply recurse or handle this properly
     } else {
-      // NOTE: Could throw an error here to help understand that this is collecting invalid styles.
+      // NOTE: Could throw an error here if this isn't our `AtomicGroups` to help understand that this is collecting invalid styles.
+      groups = String(cls).split(' ');
     }
 
     for (let x = 0; x < groups.length; x++) {
--- a/packages/react/src/runtime/ax.ts
+++ b/packages/react/src/runtime/ax.ts
@@ -1,3 +1,5 @@
+import { AtomicGroups } from './ac';
+
 const UNDERSCORE_UNICODE = 95;
 
 /**
@@ -27,8 +29,13 @@ const ATOMIC_GROUP_LENGTH = 5;
  *
  * @param classes
  */
-export default function ax(classNames: (string | undefined | null | false)[]): string | undefined {
-  if (classNames.length <= 1 && (!classNames[0] || classNames[0].indexOf(' ') === -1)) {
+export default function ax(
+  classNames: (string | undefined | null | false | AtomicGroups)[]
+): string | undefined {
+  if (
+    classNames.length <= 1 &&
+    (!classNames[0] || (typeof classNames[0] === 'string' && classNames[0].indexOf(' ') === -1))
+  ) {
     // short circuit if there's no custom class names.
     return classNames[0] || undefined;
   }
@@ -41,7 +48,16 @@ export default function ax(classNames: (string | undefined | null | false)[]): s
       continue;
     }
 
-    const groups = cls.split(' ');
+    let groups: string[] = [];
+    if (typeof cls === 'string') {
+      groups = cls.split(' ');
+    } else if (Array.isArray(cls)) {
+      groups = cls;
+    } else if (cls instanceof AtomicGroups) {
+      groups = cls.toString().split(' ');
+    } else {
+      // NOTE: Could throw an error here to help understand that this is collecting invalid styles.
+    }
 
     for (let x = 0; x < groups.length; x++) {
       const atomic = groups[x];
--- a/.changeset/modern-olives-chew.md
+++ b/.changeset/modern-olives-chew.md
@@ -1,0 +1,5 @@
+---
+'@compiled/react': patch
+---
+
+Allows `ax()` class collection to handle outputs from `cx()` functions.
--- a/package.json
+++ b/package.json
@@ -133,7 +133,7 @@
     },
     {
       "path": "./packages/react/dist/browser/runtime/ax.js",
-      "limit": "195B",
+      "limit": "240B",
       "import": "ax"
     },
     {
--- a/packages/react/src/runtime/__tests__/ax.test.ts
+++ b/packages/react/src/runtime/__tests__/ax.test.ts
@@ -1,4 +1,9 @@
+import { AtomicGroups } from '../ac';
 import ax from '../ax';
+
+const atomicMap = new Map();
+atomicMap.set('_aaaa', '_aaaabbbb');
+atomicMap.set('_bbbb', '_bbbbcccc');
 
 describe('ax', () => {
   const isEnabled: boolean = (() => false)();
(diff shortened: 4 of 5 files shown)


packages/react/src/runtime/ax.ts
```diff
@@ -27,8 +29,13 @@ const ATOMIC_GROUP_LENGTH = 5;
  *
  * @param classes
  */
-export default function ax(classNames: (string | undefined | null | false)[]): string | undefined {
-  if (classNames.length <= 1 && (!classNames[0] || classNames[0].indexOf(' ') === -1)) {
+export default function ax(
+  classNames: (string | undefined | null | false | AtomicGroups)[]
```
kylorhall-atlassian, line 33: With the introduction of `cx`, I believe this is the proper type as this does happen (or `cx` should change in a way I don't know how).

Realistically this could also be infinitely recursive, I don't think anything flattens here…but I won't solve a problem that doesn't really exist.

packages/react/src/runtime/ax.ts
```diff
@@ -41,7 +48,16 @@ export default function ax(classNames: (string | undefined | null | false)[]): s
       continue;
     }
 
-    const groups = cls.split(' ');
```
itsdouges, line 44: What if:

```
const groups = (cls + '').split(' ');
```

packages/react/src/runtime/ax.ts
kylorhall-atlassian, line 1: Could use help actually figuring out how to test this change as it's apparent the scenario doesn't work, or I'm doing something wrong.  I'm not sure how our `<Button xcss={cx(…)} />` tests are passing (I figured this wasn't tested, but [it is](https://github.com/atlassian-labs/compiled/blob/39d9a02c8718a439085e40e6fe8eca1bf5274377/packages/react/src/create-strict-api/__tests__/index.test.tsx#L333)?  I expected those to error…)

packages/react/src/runtime/ax.ts
kylorhall-atlassian: Yeah actually, doing `const groups = String(cls).split(' ');` does work actually, just feels like it doesn't tell you what it's doing is all.

itsdouges: I'd do whatever results in better performance.

kylorhall-atlassian: Well, not sure about the array syntax `<Component xcss={[style, style]} />`, but whatever, that fails types anyways.

kylorhall-atlassian: Doing arbitrary online benchmarks which mean little, protecting via a typeof like:
```tsx
if (typeof c1 === 'string') {
	c1.split(' ')
} else {
	String(c1).split(' ')
}
```
is faster by ~4% than just blindly calling `String(c1)` first.  I gather `String()` doesn't handle this (but could/should).

Worst-case, I'd expect it to be 1:1, I don't see how v8 could do anything super special here—basically the negative impact is just from doing `typeof` checks, which are pretty minimal.

packages/react/src/runtime/ax.ts
kylorhall-atlassian: At least I tested this in `ax.test.ts` now.

packages/react/src/runtime/ax.ts
```diff
@@ -41,7 +48,15 @@ export default function ax(classNames: (string | undefined | null | false)[]): s
...
+    let groups: string[] = [];
+    if (typeof cls === 'string') {
+      groups = cls.split(' ');
+    } else if (Array.isArray(cls)) {
+      groups = cls; // NOTE: Not really a valid scenario, so we don't deeply recurse or handle this properly
+    } else {
+      // NOTE: Could throw an error here if this isn't our `AtomicGroups` to help understand that this is collecting invalid styles.
```
kylorhall-atlassian, line 57: I found doing a `cls instanceof AtomicGroups` wasn't consistently working, might as well try to blanket support, though…it does mean previous things that would throw `'cls.split is not a function'` now more quietly fail.

packages/react/src/runtime/ax.ts
itsdouges: Sweet ok. I think we can lean on type assumptions and safely use the typeof check tbh.

packages/react/src/runtime/ax.ts
itsdouges: Yeah that's fine. We can lean on the typesystem for everything else.

dddlr: do you guys use `ac` and `AtomicGroups` for anything? i thought that was part of the [class name compression work](https://github.com/atlassian-labs/compiled/pull/1437) that we've now shelved (probably safe to just yeet)

kylorhall-atlassian: So the root example comes from using `cx()` which uses `ac()`.  Frankly, I do not know _anything_ here, I'm just hacking it together.

itsdouges: I think we used `ac` based on @liamqma's guidance. If you think it's not useful we can switch it out for whatever.

dddlr: [ahh i see](https://github.com/atlassian-labs/compiled/pull/1533/files#diff-2821c097864f8f4f268973d4fafa1cab339af4bd3805b489c4777a0cb8ff0294R144)

i forgot we were using it there haha, thank you for clarifying

itsdouges: Yeah. So should we just use ax and side step the whole issue?

packages/react/src/runtime/ax.ts
```diff
@@ -41,7 +48,15 @@ export default function ax(classNames: (string | undefined | null | false)[]): s
       continue;
     }
 
-    const groups = cls.split(' ');
+    let groups: string[] = [];
+    if (typeof cls === 'string') {
+      groups = cls.split(' ');
+    } else if (Array.isArray(cls)) {
+      groups = cls; // NOTE: Not really a valid scenario, so we don't deeply recurse or handle this properly
```
dddlr, line 55: this is an area in the codebase i haven't looked at in a while, so i'm not too familiar with the runtime stuff 😅

from the comment and the tests, i take it only APIs outside of Compiled (and never Compiled APIs) would ever create a situation where `Array.isArray(cls)` is true right? (e.g. `className = [['_abcd1234', '_efgh1234'], '_ijkl1234 mnop1234', ...]`)

i don't have a preference on whether to keep the `else if (Array.isArray(cls)) { ... }` branch in or not, just aiming to understand the PR fully

packages/react/src/runtime/ax.ts
liamqma: Yeah, we've parked classname compression. In theory, no Compiled users should use `ac`. 
I suggest switching it out for the old good `ax`. 🙂

packages/react/src/runtime/ax.ts
kylorhall-atlassian: Just to drop the scenario, right now let's say if you do like this:
```tsx
const Component = (props) => (
  <div className={xcss} css={{ margin: 0 }} />
);

export default (props) => <Component xcss={[{ margin: '12px' }, props.xcss]} />
`````

You will get a type error because our `xcss` implementation requires `xcss={cx({ margin: '12px' }, props.xcss)}` and arrays won't work, but in theory I could see people copy + pasting `css={[…]}` array into `xcss={[…]}` and ignoring it and it throwing an error (because I kind of did that 🤣).

I don't mind much either way to be honest, it does have a type error.

I would prefer to either `throw new Error(…)` in here or keep it, because if it does get passed, doing `String(array).split(' ');` is quite unperformant.

packages/react/src/runtime/ax.ts
dddlr: oops i was going to write a big comment but it got deleted 😢 

but yeah depends on whether you want to support nesting like `cx([cx('_aaaa_a'), '_aaaa_b'])`... if not, i guess there's no value that an `ac`-based approach would provide

dddlr: (and maybe delete the `ac`/class name compression stuff altogether too? but that's probably more suited for a follow-up PR)

packages/react/src/runtime/ax.ts
dddlr: that makes sense, thank you for the additional context

packages/react/src/runtime/ax.ts
kylorhall-atlassian: Makes sense to me, is it literally as simple as swapping out `ac` for `ax`?  It kinda looks like it locally as I _think_ I've got all my tests passing 🙈

If that's a drop-in, I can pop that out tomorrow, but I won't make the whole removal (bit over my head).

kylorhall-atlassian: I _think_ this works: https://github.com/atlassian-labs/compiled/pull/1697

kylorhall-atlassian:
Closing as I think https://github.com/atlassian-labs/compiled/pull/1697 is just a better approach.

PR #1697, opened 2024-07-18
Fix the `cx()` function's class collection at runtime as it generated class instances rather than strings.

kylorhall-atlassian:
Alternative to https://github.com/atlassian-labs/compiled/pull/1696

Migrate from `ac()` which generates `AtomicGroups` classes to `ax()` which strictly generates strings (I think that's what this all does).

This is because passing `<Component xcss={cx(style, style)}>` around resulted in errors, `'cls.split is not a function'` trying to iterate over non-strings.

Previously this generated classes collected into `ac` like:
```
[
  '_1bah1yb4',
  undefined,
  ['_16jlkb7n', undefined], // from `xcss={[…]}`, not accounted for in this PR, however, not real
  AtomicGroups { // from `xcss={cx(…)}`
    values: Map(2) { '_1bsb' => '_1bsb1osq', '_16jl' => '_16jlkb7n' }
  }
]
```

### PR checklist

I have...

- [x] Chosen not to document this in `website/`

### DIFF
--- a/packages/react/src/xcss-prop/index.ts
+++ b/packages/react/src/xcss-prop/index.ts
@@ -1,7 +1,7 @@
 import type * as CSS from 'csstype';
 
 import type { ApplySchemaValue } from '../create-strict-api/types';
-import { ac } from '../runtime';
+import { ax } from '../runtime';
 import type { CSSPseudos, CSSPseudoClasses, CSSProperties, StrictCSSProperties } from '../types';
 
 type MarkAsRequired<T, K extends keyof T> = T & { [P in K]-?: T[P] };
@@ -208,5 +208,5 @@ export const cx = <TStyles extends [...XCSSProp<any, any>[]]>(
 
   // The output should be a union type of passed in styles. This ensures the call
   // site of xcss prop can raise violations when disallowed styles have been passed.
-  return ac(actualStyles) as TStyles[number];
+  return ax(actualStyles) as TStyles[number];
 };
--- a/.changeset/friendly-eels-cover.md
+++ b/.changeset/friendly-eels-cover.md
@@ -1,0 +1,5 @@
+---
+'@compiled/react': patch
+---
+
+Fix the `cx()` function's class collection at runtime as it generated class instances rather than strings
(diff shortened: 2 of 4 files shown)


packages/react/src/xcss-prop/index.ts
```diff
@@ -208,5 +208,5 @@ export const cx = <TStyles extends [...XCSSProp<any, any>[]]>(
 
   // The output should be a union type of passed in styles. This ensures the call
   // site of xcss prop can raise violations when disallowed styles have been passed.
-  return ac(actualStyles) as TStyles[number];
+  return ax(actualStyles) as TStyles[number];
```
kylorhall-atlassian, line 211: I'm worried this doesn't fail tests, not sure how to test this to be honest.  I don't think anything is testing this properly (else it would've been failing before).

liamqma (dismissed):

packages/react/src/xcss-prop/index.ts
dddlr: a setup like that in `packages/react/src/create-strict-api/__tests__/css-func.test.tsx` could work for testing it right?

dddlr (dismissed):
lgtm as long as you're confident that `cx` will continue to work

packages/react/src/xcss-prop/index.ts
kylorhall-atlassian: I thought this would've failed in this test before to be frank: https://github.com/atlassian-labs/compiled/blob/39d9a02c8718a439085e40e6fe8eca1bf5274377/packages/react/src/create-strict-api/__tests__/index.test.tsx#L526, but I think because it was passing to the root `<div className={xcss}>` it wasn't a problem, it's when it's wrapped in `<Component xcss={xcss}>` it becomes a problem…?

Added tests in f0f18993 and can confirm, without this `ax` diff it breaks:
Screenshot 2024-07-19 at 1 18 44 PM

itsdouges (approved):

dddlr (approved):

PR #1700, opened 2024-07-22
Sort shorthand vs. longhand declarations — breaking change behind config

kylorhall-atlassian:
## What is this change?

Adds sorting of shorthand vs. longhand to ensure longhand deterministically will override shorthand.

⚠️ This could result in regressions as this will no longer work, so it's behind config, but this may mean this change should never go through (if we added border to this sorting):
```tsx
const styles = css({ borderTop: '1px solid blue' });
const noBorderStyles = css({ border: 'none' });

export default ({ noBorder }) => <div
  css={[styles, noBorder && noBorderStyles]}
/>
```

This is because quite commonly this code may be applied in either direction meaning sometimes `font-weight: bold` is applied and sometimes it's not, depending on the order that the stylesheet was generated in.
```tsx
const styles = css({ font: '12px / 24px' });
const boldStyles = css({ fontWeight: 'bold' });

export default ({ isBold }) => <div
  css={[styles, isBold && boldStyles]}
/>
```

Example in a Storybook:
Screenshot 2024-07-22 at 3 47 41 PM

Where we already sort atomic styles, we now sort shorthand vs. longhand in that same area.

---

### PR checklist

- [x] Validate this works consistently as it's unclear if this `sortAtomicStyleSheet` iterates over _all declarations_ or just _one by one_…
- [x] Added an applicable test suite
- [x] Expand the list of shorthand properties, right now this is just `outline` and `font`, there are many more!
- [x] Investigate if there's further tests to update
- [x] Update the documentation in `website/` with this new config


packages/react/src/runtime/__tests__/style-ssr.test.tsx
```diff
@@ -43,12 +43,19 @@ describe('<Style />', () => {
...
+          `._j1234567{ border-top: 1px solid transparent; }`,
+          `._j1234567{ border-top-color: red; }`,
+          `._j1234567:focus{ border-top-color: transparent; }`,
+          `._j1234567:focus{ border-color: blue; }`,
+          `._j1234567:focus{ border: 1px solid; }`,
         ]}
       </Style>
     );
 
+    // WARNING: This is wrong, the `focus` border properties are different than without focus, borders would be indeterministic.
```
kylorhall-atlassian, line 56: If this is required, all "bucket sorting" needs to be drastically rewritten to be nested.

Right now it's flat:
```css
.abc { border: none }
.abc:link { border: none }
.abc:visited { border: none }
…
.abc:active { border: none }
@media (…) { .abc { border: none } }
```

But this doesn't support `@media (…)` being sorted internally or `.abc:focus { … }` being sorted against each other.

This would need deeper nested bucketing:
```css
.abc { border: none }
.abc{ border-top: none }
.abc:link { border: none }
.abc:link { border-top: none }
@media (…) {
  .abc { border: none }
  .abc { border-top: none }
  .abc:link { border: none }
  .abc:link { border-top: none }
}
```

And I'm not exactly sure if that's even feasible for `@media` queries either…

packages/react/src/runtime/shorthand.ts
```diff
@@ -0,0 +1,108 @@
+import type { ShorthandProperties } from '@compiled/utils';
+
+const shorthandFor: Record<ShorthandProperties, true> = {
```
kylorhall-atlassian, line 3: Could change to `1` instead of `true`, but I'd expect compilation might do this anyways.

I like your approach because types protects it 👌

dddlr: ah that's a good idea

it doesn't look like the current compilation does that, so changing `true` to 1 would definitely help with byte size

i plan to make another commit to make overlapping properties with the same "depth" (e.g. `borderTop` and `borderColor`) deterministically sorted too, so this file will change some more 😅

packages/utils/src/shorthand.ts
```diff
@@ -0,0 +1,519 @@
...
+    'border-bottom-style',
+    'border-bottom-width',
+    'border-block-start-color',
+    'border-block-start-style',
+    'border-block-start-width',
+    'border-block-end-color',
+    'border-block-end-style',
+    'border-block-end-width',
+  ],
+  'border-color': [
```
kylorhall-atlassian, line 193: Example: Should this sort _above_ `border-top` as well, not just the children?

packages/utils/src/shorthand.ts
```diff
@@ -0,0 +1,519 @@
...
+    'border-top-color',
+  ],
+  'border-image': [
+    'border-image-outset',
+    'border-image-repeat',
+    'border-image-slice',
+    'border-image-source',
+    'border-image-width',
+  ],
+  'border-inline': [
```
kylorhall-atlassian, line 212: Example: Should this sort _above_ `border-right` entirely, not just the children?

dddlr:
More todos

* [ ] Add changesets

packages/react/src/runtime/shorthand.ts
```diff
@@ -0,0 +1,88 @@
...
+  'border-image': 1,
+  'border-inline': 7,
+  'border-inline-end': 8,
+  'border-inline-start': 9,
+  'border-left': 10,
+  'border-radius': 1,
+  'border-right': 10,
+  'border-style': 11,
+  'border-top': 12,
+  'border-width': 13,
```
kylorhall-atlassian, line 29: I'm not quite seeing this, I think this would be more like this:
```suggestion
  border: 1,
  'border-block': 3,
  'border-block-end': 5,
  'border-block-start': 5,
  'border-bottom': 4,
  'border-color': 2,
  'border-image': 2,
  'border-inline': 3,
  'border-inline-end': 5,
  'border-inline-start': 5,
  'border-left': 4,
  'border-radius': 1,
  'border-right': 4,
  'border-style': 2,
  'border-top': 4,
  'border-width': 2,
```

 - `border` > everything else (`border-radius` is different)
 - `border-color|width|style` > everything else
 - `border-inline|block` > `border-left|right|top|bottom`
 - `border-top` > `border-inline-start` — technically could leave this out if you want and drop this layer (make them even, but not sure how we'd fix them later)
 - `border-top-color` or `border-inline-start-color` always loses ⚠️ _maybe_ we should sort these consistently against each other, but I don't know…"logical property sorting" is kind of different, let's leave it out — could add and move things into a 6th layer later on, if we care

packages/utils/src/shorthand.ts
```diff
@@ -0,0 +1,616 @@
...
+  ],
+  'view-timeline': ['view-timeline-name', 'view-timeline-axis'],
+};
+
+// Make sure to update the `shorthandBuckets` variable in `packages/react/src/runtime/shorthand.ts`
+// after making changes to this object.
+//
+// You may want to use `packages/scripts/src/generate-shorthand.ts` to help
+// you generate this object.
+export const shorthandBuckets: Record<ShorthandProperties, Depths> = {
```
kylorhall-atlassian, line 535: Could you import just the type of this to make sure they're 1:1?

dddlr: it doesn't look like it would achieve anything unfortunately... 

the two copies of `shorthandBuckets` are already type annotated as `Record<ShorthandProperties, Depths>`, so the keys are guaranteed to be the same (but the values aren't)

if i were to remove the type annotation from one of the `shorthandBuckets`, and set the type of the other `shorthandBuckets` to `typeof importedShorthandBuckets` (where `importedShorthandBuckets` points to the first `shorthandBuckets`), the typing would be achieve almost the same result as if we just annotated the two variables as `Record<ShorthandProperties, Depths>`
Screenshot 2024-09-26 at 10 56 08

maybe an automated PR check could enforce that they be 1:1

packages/react/src/runtime/shorthand.ts
dddlr: ah good idea 👍 

manually configuring the bucket numbers > writing custom scripts to calculate the bucket numbers after all 🏆

packages/utils/src/shorthand.ts
kylorhall-atlassian: This might be a bit of magic, but it should work:
```tsx
export const shorthandBuckets = {
  all: 0,
} as const satisfies Record<ShorthandProperties, Depths>;
```

```tsx
import type {
  Depths,
  ShorthandProperties,
  shorthandBuckets as ShorthandBucketsType,
} from '@compiled/utils';

const shorthandBuckets: typeof ShorthandBucketsType = {
```

Screenshot 2024-09-26 at 1 11 02 PM

Screenshot 2024-09-26 at 1 10 58 PM

dddlr: you're a genius!! 😄

dddlr:
docs deploy preview

https://deploy-preview-1700--compiled-css-in-js.netlify.app/docs/shorthand

based on my best understanding of how it works, feel free to make any corrections

packages/utils/src/source-maps.ts
dddlr, line 1: this appeared to be unused everywhere and was causing me grief when building

examples/webpack/webpack.config.js
```diff
@@ -11,6 +11,8 @@ const classNameCompressionMap = require('./class-name-compression-map.json');
 
 const extractCSS = process.env.EXTRACT_TO_CSS === 'true';
 
+console.log('Stylesheet extraction enabled?', extractCSS);
```
liamqma, line 14: Is this console.log intentional?

liamqma (dismissed):

packages/react/src/runtime/__tests__/style.test.tsx
```diff
@@ -66,7 +66,7 @@ describe('<Style />', () => {
     });
   });
 
-  it('should warn in dev when using a dangerous pseduo selector', () => {
+  it('should warn in dev when using a dangerous pseudo selector', () => {
```
kylorhall-atlassian, line 69: I never get this word right myself—either way I spell it, both somehow fail spellcheck 🤣

packages/react/src/runtime/__tests__/style.test.tsx
```diff
@@ -124,6 +124,49 @@ describe('<Style />', () => {
...
+          ]}
+        </Style>
+      );
+      expect(document.head.innerHTML.split('</style>').join('</style>\n')).toMatchInlineSnapshot(`
+        "<style>._b1234567{ border: solid 1px blue; }._j1234567{ padding: 5px; }</style>
+        <style>._k1234567{ padding-block: 6px; }._l1234567{ padding-inline: 7px; }</style>
+        <style>._c1234567{ border-block: solid 2px blue; }._g1234567{ border-inline: solid 5px blue; }</style>
+        <style>._e1234567{ border-bottom: solid 4px blue; }._h1234567{ border-top: solid 6px blue; }</style>
+        <style>._d1234567{ border-block-end: solid 3px blue; }</style>
+        <style>._a1234567{ all: unset; }._i1234567{ border-top-color: pink; }._m1234567{ padding-top: 8px; }</style>
```
kylorhall-atlassian, line 162: ⚠️ Wait, shouldn't `all` be hoisted above everything else?  It's in `s-0`…I can guess the bug, let me review onwards!

packages/react/src/runtime/sheet.ts
```diff
@@ -114,15 +127,25 @@ export const getStyleBucketName = (sheet: string): Bucket => {
...
     // Return a mapped pseudo else the default catch all bucket.
-    return pseudosMap[sheet.slice(14, sheet.indexOf('{'))] || '';
+    const mapped = pseudosMap[sheet.slice(14, firstBracket)];
+    if (mapped) return mapped;
+  }
+
+  const property = sheet.slice(firstBracket + 1, sheet.indexOf(':', firstBracket)).trim();
+
+  const shorthandDepth = getShorthandDepth(property);
+  if (shorthandDepth) {
```
kylorhall-atlassian, line 147: `getShorthandDepth('all') === 0` which fails this logical step

⚠️ This must be changed to fix `all` bucketing (I'm pretty sure I had this bug originally as well), I just noticed in the test `packages/react/src/runtime/__tests__/style.test.tsx`
```suggestion
  if (shorthandDepth !== null) {
```
or
```suggestion
  if (typeof shorthandDepth === 'number') {
```
Sanity check: `shorthandDepth` is null | Number` right?  Both would work, but typeof would be a bit more defensive.

packages/react/src/runtime/shorthand.ts
```diff
@@ -0,0 +1,105 @@
+import type {
+  Depths,
+  ShorthandProperties,
+  shorthandBuckets as ShorthandBucketsType,
```
kylorhall-atlassian, line 4: Sanity check: this import doesn't bloat bundle sizes, right?  We have that protection, so CI would fail if this increased from 500bytes => 4kb again or whatnot.  `type` imports should do this, just making sure this "magic" works as I expect.

packages/utils/src/shorthand.ts
kylorhall-atlassian: @dddlr I'm not decided enough on what "perfect" looks like for these, so I think it's fine to have some bugs for very rarely used properties…let's circle back to this and call it a gap unless it's obvious.

…in a way, these are "siblings", but to me it feels like there's an explicit order?  MDN doesn't really provide anything and I think there's no explicit browser setting—it's a first-defined-basis:
 - https://developer.mozilla.org/en-US/docs/Web/CSS/border-width
 - https://developer.mozilla.org/en-US/docs/Web/CSS/border-top

But if this makes sense, in theory I think for all of `border-color`, `border-style`, `border-width`, this should be:
```suggestion
  'border-color': [
    'border-top',
    'border-block',
    'border-inline',
    'border-left',
    'border-right',
    'border-bottom',
```

packages/utils/src/shorthand.ts
```diff
@@ -0,0 +1,610 @@
...
+    'border-inline-start-style',
+    'border-inline-start-width',
+    'border-inline-end-color',
+    'border-inline-end-style',
+    'border-inline-end-width',
+    'border-right-color',
+    'border-right-style',
+    'border-right-width',
+  ],
+  'border-style': [
```
kylorhall-atlassian, line 285: Same as the other comment, if you fully agree:
```suggestion
  'border-style': [
    'border-top',
    'border-block',
    'border-inline',
    'border-left',
    'border-right',
    'border-bottom',
```

packages/utils/src/shorthand.ts
```diff
@@ -0,0 +1,610 @@
...
+    'border-block-start-style',
+    'border-block-start-width',
+    'border-block-end-color',
+    'border-block-end-style',
+    'border-block-end-width',
+    'border-top-color',
+    'border-top-style',
+    'border-top-width',
+  ],
+  'border-width': [
```
kylorhall-atlassian, line 308: Same as the other comment, if you fully agree:
```suggestion
  'border-width': [
    'border-top',
    'border-block',
    'border-inline',
    'border-left',
    'border-right',
    'border-bottom',
```

packages/utils/src/shorthand.ts
```diff
@@ -0,0 +1,610 @@
...
+  'scroll-padding-inline': 2,
+
+  'scroll-timeline': 1,
+  'text-decoration': 1,
+  'text-emphasis': 1,
+  'text-wrap': 1,
+  transition: 1,
+  'view-timeline': 1,
+} as const satisfies Record<ShorthandProperties, Depths>;
+// ^^ this type lets us enforce that the copy of shorthandBuckets
```
kylorhall-atlassian, line 608: To question aloud, _why_ do we even have this in this file?

…what if we imported this into the other file, doesn't that just work?  I'd expect `shorthandFor` to be tree shaken, am I wrong?

It's fine, it feels maintainable, just looking at it again, I'm not sure why it's in both places…

website/packages/docs/src/pages/pkg-babel-plugin-strip-runtime.mdx
kylorhall-atlassian, line 1: Would this be on the `babel-plugin-strip-runtime` too, or just `babel-plugin`?  I don't even know what `babel-plugin-strip-runtime` is to be honest.

website/packages/docs/src/pages/shorthand.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+# Shorthand properties
+
+Shorthand properties are properties that are used to set the values of several other properties ("longhand properties"). Examples include the shorthand property `font`, which can be used to set the longhand properties `fontWeight`, `fontStyle`, `fontFamily`, and so on.
+
+We **do not** recommend mixing shorthand properties and longhand properties in the same component, as the precedence of the properties may not be what you expect.
+
+## How does Compiled handle shorthand and longhand properties?
+
+If shorthand sorting is enabled, then all shorthand properties in Compiled will come before their longhand forms. For example, `font` will come before `fontWeight` and `fontStyle`. The order your CSS properties appear in your source code is **ignored**.
```
kylorhall-atlassian, line 14: Is it worth calling out expansion…?  I don't see we do it on the site at all, can't find it, just the Media Query sorting page.
```suggestion
By default, Compiled expands a few common properties, like `margin` and `padding`, but this is complex and not done for all properties.

If shorthand sorting is enabled, then all shorthand properties in Compiled will come before their longhand forms. For example, `font` will come before `fontWeight` and `fontStyle`. The order your CSS properties appear in your source code is **ignored**.
```

kylorhall-atlassian (commented):
@dddlr This is my own PR so I can't request changes, but could use a sanity check in a few places and minor suggestions, nothing too blocking.

packages/react/src/runtime/sheet.ts
dddlr: nice catch! i had forgotten to update this after i changed `root` to `0`

packages/react/src/runtime/shorthand.ts
dddlr: the `import type ...` statements don't get included in `dist/` yeah

packages/utils/src/shorthand.ts
dddlr: ah the ordering of the longhand properties in `shorthandFor` no longer matters - we use `shorthandBuckets` to enforce that `border-color/width/style` comes before `border-top`, in both `packages/css/src/plugins/sort-shorthand-declarations.ts` and `packages/react/src/runtime/sheet.ts`

Screenshot 2024-09-27 at 10 13 52


i'm happy to reset the order of the longhand properties in `shorthandFor` to alphabetical order to reduce potential confusion

dddlr: ah the order of properties in `shorthandFor` no longer matters - i should probably just reset that to alphabetical order to be honest ...

`shorthandBuckets` is now used in `packages/css/src/plugins/sort-shorthand-declarations.ts` (previously was just in `packages/react/src/runtime/sheet.ts`) to enforce that `border-{color,style,width}` comes before `border-{top,right,bottom,left}`

example https://github.com/atlassian-labs/compiled/pull/1700/files#diff-3b4ae519f265680a755d14f6de289c4d7507a936e4911e95b9b3889fa1ac579bR187

Screenshot 2024-09-27 at 10 22 54

let me know if this aligns with what you have in mind

packages/utils/src/shorthand.ts
dddlr: ah i made two copies of `shorthandBuckets` because in `@compiled/react/runtime`, importing `@compiled/utils` bloats the bundle size and causes `npx size-limit` to fail yeah

re: tree shaking, can we assume that tree-shaking always runs in realistic Compiled usage? if so, i can see if i can update the `size-limit` config to account for tree-shaking maybe

packages/utils/src/shorthand.ts
kylorhall-atlassian: Got it, the test LGTM, that's what this `aShorthandBucket - bShorthandBucket` does 👌

packages/utils/src/shorthand.ts
kylorhall-atlassian: I have no clue 😸

examples/webpack/webpack.config.js
dddlr: oh yeah, i added it as a debugging measure but then decided to keep it as it seemed to be useful

happy to remove it if it's not valuable

website/packages/docs/src/pages/pkg-babel-plugin-strip-runtime.mdx
dddlr: oh yup, let me update the docs

JakeLane:
Can we add linting for this area?

This change makes the behaviour deterministic but it's still confusing to developers. Given the example in the description:
```
const styles = css({ font: '12px / 24px' });
const boldStyles = css({ fontWeight: 'bold' });

export default ({ isBold }) => <div
  css={[styles, isBold && boldStyles]}
/>
```

The inverse could happen:
```
const styles = css({ font: '12px / 24px' });
const boldStyles = css({ fontWeight: 'bold' });

export default ({ isFont }) => <div
  css={[boldStyles, isFont && styles]}
/>
```
which would have deterministic behaviour opposite to what the developer intended.

Generally, we should make the linting sort the code with autofixers (and warn when not possible) so that the source represents the production code.

liamqma (approved):

JakeLane (approved):

PR #1706, opened 2024-09-10
Update `@compiled/jsx-pragma` to properly support `importSources` 

kylorhall-atlassian:
### What is this change?

This allows for more scenarios where `options.importSources` is used in the `@compiled/jsx-pragma` rule so we don'd need to import `@atlaskit/css` (our typed wrapper) _and_ `@compiled/react`.

This just refactors some parts of the AST to search for all compiled-like libraries, assuming `importSources` is that.  I think this is what the original option intended to do, but it falls short in some circumstances.

---

### PR checklist

- [x] Updated the documentation in `website/` (n/a — links to the code)

### DIFF
--- a/packages/eslint-plugin/src/rules/jsx-pragma/__tests__/rule.test.ts
+++ b/packages/eslint-plugin/src/rules/jsx-pragma/__tests__/rule.test.ts
@@ -1,3 +1,5 @@
+import { outdent } from 'outdent';
+
 import { tester } from '../../../test-utils';
 import { jsxPragmaRule } from '../index';
 
@@ -5,40 +7,48 @@ tester.run('jsx-pragma', jsxPragmaRule, {
   valid: [
     {
       name: 'should be valid if automatic jsx pragma is used',
-      code: `
+      code: outdent`
         /** @jsxImportSource @compiled/react */
         <div css={{ display: 'block' }} />
       `,
     },
     {
+      name: 'should be valid if automatic jsx pragma is with a Compiled import',
+      code: outdent`
+        /** @jsxImportSource @compiled/react */
+        import { css, jsx } from '@compiled/react';
+        <div css={css({ display: 'block' })} />
+      `,
+    },
+    {
       name: 'should be valid if using xcss on a React component',
-      code: `
+      code: outdent`
         <AnotherComponent className={xcss} />
       `,
     },
     {
       name: 'should be valid if not using xcss on a JSX element',
-      code: `
+      code: outdent`
         <div className={anythingElse} />
       `,
     },
     {
       name: 'should be valid if using xcss on a JSX element with automatic jsx pragma',
-      code: `
+      code: outdent`
         /** @jsxImportSource @compiled/react */
         <div className={xcss} />
       `,
     },
     {
       name: 'should be valid if using xcss under a different name with automatic jsx pragma',
-      code: `
+      code: outdent`
         /** @jsxImportSource @compiled/react */
         <div className={innerXcss} />
       `,
     },
     {
       name: 'should be valid if runtime = automatic and automatic jsx pragma is used',
-      code: `
+      code: outdent`
         /** @jsxImportSource @compiled/react */
         <div css={{ display: 'block' }} />
       `,
@@ -46,7 +56,7 @@ tester.run('jsx-pragma', jsxPragmaRule, {
     },
     {
       name: 'should be valid if runtime = classic and classic jsx pragma is used',
-      code: `
+      code: outdent`
         /** @jsx jsx */
         import { jsx } from '@compiled/react';
         <div css={{ display: 'block' }} />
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/src/rules/jsx-pragma/__tests__/rule.test.ts
```diff
@@ -64,21 +74,30 @@ tester.run('jsx-pragma', jsxPragmaRule, {
...
         import { css, jsx } from '@compiled/react';
         <div css={css({ display: 'block' })} />
       `,
       options: [{ runtime: 'classic' }],
     },
+    {
+      name: 'should be valid if runtime = classic and classic jsx pragma is used with css function call found in importSources',
+      code: outdent`
+        /** @jsx jsx */
+        import { css, jsx } from '@atlaskit/css';
```
kylorhall-atlassian, line 88: This is the bulk of the change _we need_, but I go on to fix the automated variants as well, works as far as I'm aware..

packages/eslint-plugin/src/rules/jsx-pragma/__tests__/rule.test.ts
```diff
@@ -581,15 +739,15 @@ import * as React from 'react';
...
         import { css } from '@atlaskit/css';
 
         const styles = css({ display: 'block' });
         <div css={styles} />
       `,
-      output: `
-        /** @jsxImportSource @compiled/react */
-import { css } from '@atlaskit/css';
+      output: outdent`
+        /** @jsxImportSource @atlaskit/css */
```
kylorhall-atlassian, line 749: This one is a **change**, but I'm _pretty certain_ the name of the test implied this should work?

liamqma (dismissed):

JakeLane (dismissed):

packages/eslint-plugin/src/rules/jsx-pragma/__tests__/rule.test.ts
```diff
@@ -90,78 +109,131 @@ tester.run('jsx-pragma', jsxPragmaRule, {
...
     {
       name: 'should add jsx pragma and Compiled import if neither are there',
-      code: `<div css={{ display: 'block' }} />`,
-      output: `/** @jsx jsx */
-import { jsx } from '@compiled/react';
-<div css={{ display: 'block' }} />`,
+      code: outdent`
+        <div css={{ display: 'block' }} />
+      `,
+      output: outdent`
```
dddlr, line 196: thank you for adding the `outdent` calls 🙏

packages/eslint-plugin/src/rules/jsx-pragma/index.ts
```diff
@@ -62,10 +65,16 @@ const findReactDeclarationWithDefaultImport = (
 };
 
 function createFixer(context: Rule.RuleContext, source: SourceCode, options: Options) {
-  const compiledImports = findLibraryImportDeclarations(context, [COMPILED_IMPORT]);
+  const compiledImports = [
+    // TODO: Should this just be one call, eg. `[COMPILED_IMPORT, ...options.importSources]`?
```
dddlr, line 69: is this a TODO for us? 😅

packages/eslint-plugin/src/rules/jsx-pragma/__tests__/rule.test.ts
dddlr: that looks reasonable to me

assuming that you've done a quick check that `/** @jsxImportSource @atlaskit/css */` does work

packages/eslint-plugin/src/rules/jsx-pragma/index.ts
kylorhall-atlassian: Nope, I forgot this.  Fixed in 5534012e

Doing `findLibraryImportDeclarations(context, [COMPILED_IMPORT, ...options.importSources])` instead of this spread and join should be more performant than two separate calls.  I'm not quite too concerned about performance enough to cache this across the main ESLint context and this `createFixer` context, but best to consolidate better.

dddlr: imo `findLibraryImportDeclarations(context, [COMPILED_IMPORT, ...options.importSources])` is less confusing yeah


i don't like the person below who made the second parameter of `findLibraryImportDeclarations` optional, that confused me a lot
Screenshot 2024-09-11 at 11 24 10

dddlr (approved):

packages/eslint-plugin/src/rules/jsx-pragma/__tests__/rule.test.ts
kylorhall-atlassian: Actually, no, we use `classic` mostly—any good place to test (feel free to DM)?

PR #1707, opened 2024-09-11
Allow an import of `jsx` from our internal `@atlaskit/css` package for babel plugin pragma resolutions.

kylorhall-atlassian:
### What is this change?

With https://github.com/atlassian-labs/compiled/pull/1706, I missed the fact that `@compiled/babel-plugin` requires jsx to import from strictly `@compiled/react` and doesn't fully make use of `options.importSources` here either—that pull request passed in tests and one of my dev envs, but not in a real devloop.

Additionally, this configuration had to be added to `@compiled/webpack-plugin` which wraps the Babel Plugin and had a hardcoded early exit.  For context, `@compiled/parcel-plugin` already appears to have this functionality: https://github.com/atlassian-labs/compiled/blob/master/packages/parcel-transformer/src/index.ts#L94 and I believe I've manually reviewed all instances of `'@compiled/react'` being used as a string 🤞

To enable JSX pragmas from any imported `options.importSources` through Babel and Webpack plugins (and previously ESLint).

---

### PR checklist

- [x] Updated the documentation in `website/` (this is already documented)

### DIFF
--- a/packages/babel-plugin/src/babel-plugin.ts
+++ b/packages/babel-plugin/src/babel-plugin.ts
@@ -31,8 +31,7 @@ import { visitXcssPropPath } from './xcss-prop';
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const packageJson = require('../package.json');
 const JSX_SOURCE_ANNOTATION_REGEX = /\*?\s*@jsxImportSource\s+([^\s]+)/;
-const COMPILED_IMPORT_SOURCE = '@compiled/react';
-const DEFAULT_IMPORT_SOURCES = [COMPILED_IMPORT_SOURCE, '@atlaskit/css'];
+const DEFAULT_IMPORT_SOURCES = ['@compiled/react', '@atlaskit/css'];
 
 let globalCache: Cache | undefined;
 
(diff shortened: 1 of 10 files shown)


packages/babel-plugin/src/babel-plugin.ts
```diff
@@ -131,7 +130,7 @@ export default declare<State>((api) => {
 
             // jsxPragmas currently only run on the top-level compiled module,
             // hence we don't interrogate this.importSources.
-            if (jsxSourceMatches && jsxSourceMatches[1] === COMPILED_IMPORT_SOURCE) {
+            if (jsxSourceMatches && this.importSources.includes(jsxSourceMatches[1])) {
```
kylorhall-atlassian, line 133: I believe this could have a side-effect, we're saying "transform `@atlaskit/css` to be `@compiled/react`" which if what _I expect_, but is there any scenario where this isn't intended…?  I'm not exactly sure what `options.importSources` intends to do otherwise—like the `ImportDeclaration` walker works, but omits the jsx pragma.  I'll assume this is just extending that logic to include `jsx` (while before it's just `['styled', 'ClassNames', 'css', 'keyframes', 'cssMap']`).

JakeLane (approved):

PR #1712, opened 2024-09-27
See whether Parcel stylesheet extraction works with XCSS

dddlr:
Spike whether https://github.com/atlassian-labs/compiled/issues/1306 is actually an issue or not...

From what I can tell, the following situation doesn't have good dev ex when stylesheet extraction is turned on and Parcel is used:

```tsx
// otherFile.tsx

export const someStyles = css({ color: 'yellow' });
```
```tsx
// index.tsx

import { someStyles } from './otherFile';

const App = () => (
  <>
    <div css={someStyles}>Some styles</div>
  </>
);
```

Updating the styles `someStyles` doesn't result in the new styles being applied to `App` unless you restart the whole app.

But [the UI Styling Standard forbids exporting `css`](https://atlassian.design/components/eslint-plugin-ui-styling-standard/no-exported-styles/usage), so I thought I'd check another situation where styles are exported: XCSS prop / `createStrictAPI`. Looks like there's no dev ex issues with that.


packages/parcel-transformer/src/index.ts
dddlr, line 1: oh i'm dumb, i forgot to update `transform` function

PR #1714, opened 2024-09-30
AFB-777 create eslint to enforce shorthand sorting

zesmith2:
### PR checklist

Don't delete me!

I have...


packages/eslint-plugin/src/rules/shorthand-property-sorting/utils.ts
```diff
@@ -0,0 +1,68 @@
+export const shorthandFor: Record<string, number> = {
```
dddlr, line 1: replace this with `shorthandBuckets` from `packages/utils/src/shorthand.ts`

https://github.com/atlassian-labs/compiled/blob/4ed15ebec15830eb3b67069a9f8f54986f211594/packages/utils/src/shorthand.ts

(you'll need to rebase from master)

packages/eslint-plugin/src/rules/shorthand-property-sorting/README.md
```diff
@@ -0,0 +1,31 @@
+# `shorthand-property-sorting`
+
+Enforces css property sorting for packages `css`, `cssMap`, `styled` that originates from `@compiled/react`, and `xcss` from `@atlaskit/primitives`.
+
+When using both shorthand and longhand properties, compiled scrambles the properties and it is unintuative to understand which properties override which others. Ordering the shorthand first ensures that longhand always overrides shorthand.
```
dddlr, line 5:
```suggestion
At build time, Compiled automatically sorts shorthand properties (like `font` and `border`) so that they come before any longhand properties (like `fontSize` and `borderTopColor`) defined on the component. This means that longhand properties will always override shorthand properties.

This rule enforces that the order in which the properties appear in a component's source code matches the actual ordering the properties will have at build time and runtime.
```

or something like this

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,124 @@
...
+    },
+    messages: {
+      'shorthand-first':
+        'When using both shorthand and longhand properties, the shorthand property should be first.',
+    },
+    type: 'problem',
+    fixable: 'code',
+  },
+  create(context) {
+    const selectorString: string =
```
dddlr, line 32: we probably don't want to enforce shorthand properties coming before longhand properties for non-Compiled libraries (unless @kylorhall-atlassian has a different opinion)

you can use the `isCss` and `isCssMap` functions in `packages/eslint-plugin/src/utils/is-compiled-import.ts`, and the `isStyled` function from `packages/eslint-plugin/src/rules/no-styled-tagged-template-expression/utils.ts` and `packages/eslint-plugin/src/utils/styled-import.ts`

you'll need to

* create an `isXcss` function
* add `@atlaskit/css` to `COMPILED_IMPORT` in `packages/eslint-plugin/src/utils/constants.ts` (this will involve changing all usages of `COMPILED_IMPORT` so they can handle an array instead of a string)
* maybe move the contents of `packages/eslint-plugin/src/rules/no-styled-tagged-template-expression/utils.ts` to `packages/eslint-plugin/src/utils/is-compiled-import.ts` and potentially refactor to make more consistent


(in an ideal world, we'd probably re-use some of the code that's in the UI Styling Standard ESLint plugin https://atlaskit.atlassian.com/packages/design-system/eslint-utils but oh well)

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,124 @@
+import type { Rule } from 'eslint';
+
+import { shorthandFor } from './utils';
+
+const findProgramRoot = (node: Rule.Node) => {
```
dddlr, line 5: ah looks like this type error is because typescript version in compiled is outdated, let me fix that

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
zesmith2: I'm dealing with this from line 47 - 73 with my own checks if I understand what you're saying? I'l check out those functions if that's an easier way to do it though

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
dddlr: never mind, it's a bit of a pain to upgrade typescript 😓

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,124 @@
...
+              node.parent.callee.name === 'xcss';
+
+            if (program && (logicCssAndCssMap || logicStyled)) {
+              importedFromCorrectPackage = program.body.find(
+                (n) => n.type === 'ImportDeclaration' && n.source.value === '@compiled/react'
+              );
+            }
+            if (program && logicXcc) {
+              importedFromCorrectPackage = program.body.find(
+                (n) => n.type === 'ImportDeclaration' && n.source.value === '@atlaskit/primitives'
```
dddlr, line 71: when `xcss` is migrated to Compiled, will it be through `@atlaskit/primitives`? or `@atlaskit/css`? (or some other library) @kylorhall-atlassian

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
dddlr: yeah you shouldn't need to implement the logic from scratch, you can extend the existing util functions to do what you need

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,124 @@
...
+              );
+            }
+
+            // loop through the css properties of a ObjectExpression object
+            node.properties.some((property) => {
```
dddlr, line 76: add something like `if (importedFromCorrectPackage) { return; }` above the `node.properties.some((property) => { ... }` loop, so we don't needlessly loop over every non-Compiled style usage

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,124 @@
...
+              }
+            });
+
+            if (fixRequired && importedFromCorrectPackage) {
+              context.report({
+                node: node,
+                messageId: 'shorthand-first',
+                fix: (fixer) => {
+                  // sort the properties by depth
+                  const sortedProperties = node.properties.slice().sort((a, b) => {
```
dddlr, line 96: can we add some tests that include the spread operator ([we lint against this](https://atlassian.design/components/eslint-plugin-design-system/consistent-css-prop-usage/usage) but it is technically possible)

```
const styles = { /* ... */ };
    ...styles,
    fontColor: 'blue',
});
```

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,124 @@
...
+                      b.key.type === 'Identifier'
+                    ) {
+                      return shorthandFor[a.key.name] - shorthandFor[b.key.name];
+                    }
+                    return 0;
+                  });
+
+                  const sourceCode = context.getSourceCode();
+                  const sortedCode = sortedProperties
+                    .map((property) => sourceCode.getText(property))
```
dddlr, line 110: can we have a test that includes pseudo-selectors (e.g. `&:hover`), pseudo-elements (e.g. `&::after`), and media queries? e.g.

```tsx
const styles = css({
    '&:hover': {
        color: 'blue',
        fontWeight: 'bold',
        font: '...',
    },
    fontWeight: 'bold',
    font: '...',
    color: 'blue',

})
```

note that eslint does not apply overlapping fixes, so you'll get wonky test results if you try to sort the stuff inside `&:hover` and the stuff outside `&:hover` at the same time (that's ok - outside of tests, eslint will re-run the eslint rule over and over until the rule no longer suggests a fix)

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,124 @@
...
+      recommended: true,
+      description:
+        'When using both shorthand and longhand properties, compiled scrambles the properties and it is unintuative to understand which properties override which others. Ordering the shorthand first ensures that longhand overrides shorthand.',
+      url: 'https://github.com/atlassian-labs/compiled/tree/master/packages/eslint-plugin/src/rules/shorthand-property-sorting',
+    },
+    messages: {
+      'shorthand-first':
+        'When using both shorthand and longhand properties, the shorthand property should be first.',
+    },
+    type: 'problem',
```
dddlr, line 28: if we can't preserve comments, maybe safer to change this rule from a `problem` to a `suggestion`

see here https://eslint.org/docs/latest/extend/custom-rules#providing-suggestions for what you'd need to change

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,124 @@
...
+                  });
+
+                  const sourceCode = context.getSourceCode();
+                  const sortedCode = sortedProperties
+                    .map((property) => sourceCode.getText(property))
+                    .join(', ');
+
+                  // Replace the old object expression with the new sorted one
+                  const newObjectExpression = `{ ${sortedCode} }`;
+                  return [fixer.replaceText(node, newObjectExpression)];
```
dddlr, line 115: does this need to be in an array?

packages/eslint-plugin/src/rules/shorthand-property-sorting/__tests__/rule.test.ts
```diff
@@ -0,0 +1,173 @@
+import { outdent } from 'outdent';
+
+import { typeScriptTester as tester } from '../../../test-utils';
+import { shorthandFirst } from '../index';
+
+const packages = ['css', 'styled', 'cssMap', 'xcss'];
+
+tester.run('shorthand-property-sorting', shorthandFirst, {
+  valid: [
+    ...packages.map((pkg) => ({
```
dddlr, line 10: looks a bit messy...

i suggest generating all the combinations in advance

```tsx
const packages = ['css', 'styled', 'cssMap', 'xcss'] as const;
const imports = ['@atlaskit/css', '@compiled/react'] as const;

const packages_and_imports = packages.flatMap((pkg) => imports.map((imp) => [pkg, imp]));

tester.run('shorthand-property-sorting', shorthandFirst, {
  valid: [
    ...packages_and_imports.map(([pkg, imp]) => ({
      name: `correct property ordering (${pkg})`,
      code: outdent`
      import {${pkg}} from '${imp}';
      // and so on...
```

`@atlaskit/css` is a planned wrapper for `@compiled/react` that we'll need to treat as equivalent to `@compiled/react`

zesmith2: in AFM I only ever saw `xcss` imported from `@atlaskit/primitives`. Can we treat it like the others like you suggested?

kylorhall-atlassian (commented):
Top-level comments cc: @dddlr 
1. Do we have CSS values sorted already internally by chance, eg. does anyone use an alphabetical sorting rule this might conflict with?
2. Feel free to drop `xcss` as you'll touch up a few places, easiest to just drop it entirely.

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,140 @@
...
+      isStyled(node.callee as Rule.Node, references))
+  );
+};
+
+export const shorthandFirst: Rule.RuleModule = {
+  meta: {
+    docs: {
+      recommended: true,
+      description:
+        "At build time, Compiled automatically sorts shorthand properties (like `font` and `border`) so that they come before any longhand properties (like `fontSize` and `borderTopColor`) defined on the component. This means that longhand properties will always override shorthand properties. This rule enforces that the order in which the properties appear in a component's source code matches the actual ordering the properties will have at build time and runtime.",
```
pancaspe87, line 57: I may be missing something, but I think this rule ensures that redundant CSS won't cause unwanted side effects

pancaspe87: As discussed internally on a Confluence page, I think we should change the wording in both rule description and README. 

Below my understanding; up to you on how to re-write it 

Having a longhand before a shorthand is not valid CSS. While this is usually not a problem, it becomes a problem in Compiled where classes are generated for each CSS declaration and rules are then sorted at build time during stylesheet extraction. As a result, this could cause unwanted side effects when CSS longhands are redundant.

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,140 @@
...
+      url: 'https://github.com/atlassian-labs/compiled/tree/master/packages/eslint-plugin/src/rules/shorthand-property-sorting',
+    },
+    messages: {
+      'shorthand-first':
+        'When using both shorthand and longhand properties, the shorthand property should be first.',
```
pancaspe87, line 62: I would change it to something like:

If the intention is to override a shorthand property with a longhand, the longhand should come after otherwise it is redundant and may cause unwanted side effects with stylesheet extraction. Please remove the longhand if it is not your intention to override the shorthand

packages/eslint-plugin/src/rules/shorthand-property-sorting/README.md
```diff
@@ -0,0 +1,31 @@
+# `shorthand-property-sorting`
+
+Enforces css property sorting for packages `css`, `cssMap`, `styled` that originates from `@compiled/react`, and `xcss` from `@atlaskit/primitives`.
```
pancaspe87, line 3: Same as https://github.com/atlassian-labs/compiled/pull/1714/files#r1783752036

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,140 @@
...
+  },
+  create(context) {
+    const selectorString: string =
+      'CallExpression[callee.type="Identifier"][callee.name="css"] > ObjectExpression,' +
+      'CallExpression[callee.type="Identifier"][callee.name="cssMap"] > ObjectExpression,' +
+      'CallExpression[callee.type="Identifier"][callee.name="xcss"] > ObjectExpression,' +
+      'CallExpression[callee.type="MemberExpression"][callee.object.name="styled"] > ObjectExpression';
+
+    return {
+      [selectorString]: (node: Rule.Node) => {
```
liamqma, line 76: Great job! This enforcement includes several cases to consider, such as:

Styles Composition
```
const foo = css({ margin: ... });
const bar = css({ marginTop: ... });
<div css={[bar, foo]} />;
```
Style Inheritance
```
const Foo = styled.div({ marginTop: ... });
const Bar = styled(Foo)({ margin: ... });
```
I don't think we need to address all these cases in this pull request, I'm mentioning them here so we can improve the linting rule in future iterations.

packages/eslint-plugin/src/utils/constants.ts
```diff
@@ -1 +1,2 @@
 export const COMPILED_IMPORT = '@compiled/react';
+export const ATLASKIT_IMPORT = '@atlaskit/css';
```
pancaspe87, line 2: This looks odd. 
While `@atlaskit` is OSS too, I am not sure whether we should mix the two.

Why is this required?

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,140 @@
...
+        if (node.type === 'ObjectExpression' && node.properties.length > 0) {
+          let lowestDepth = 0;
+          let fixRequired = false;
+
+          const references = context.getScope().references;
+
+          // loop through the css properties of a ObjectExpression object
+          node.properties.some((property) => {
```
pancaspe87, line 84: I haven't worked with Compiled in a while, but rather than an ESLint rule, I wonder whether we could have leverage the existing logic and throw an error at compilation time instead. Is doing this with ESLint more efficient/performant and less disruptive, since linting can be optional and it won't affect build time?

cc @dddlr

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,140 @@
+import { shorthandBuckets, kebabCase, type ShorthandProperties } from '@compiled/utils';
+import type { Rule, Scope } from 'eslint';
+
+import { isStyled, isCss, isCssMap } from '../../index';
+
+const fixProperties = (context: Rule.RuleContext, node: Rule.Node) => {
```
pancaspe87, line 6: I am still not sold on the autofixer:
- it may cause unwanted side effects in existing code. Do we have an estimate of the current violations?
- a more descriptive error message should be enough

cc @dddlr @kylorhall-atlassian

packages/eslint-plugin/src/index.ts
```diff
@@ -29,6 +31,8 @@ export const rules = {
   'no-empty-styled-expression': noEmptyStyledExpressionRule,
```
dddlr, line 31: need to add rule here

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
dddlr: we could yeah, and it would definitely be more robust (ESLint doesn't let us detect violations across components, but a compile time approach might)

but a compile time approach would be harder (and we'd need to make exceptions for the few times that DST deliberately puts longhand properties before shorthand properties), and considering that it's hard to justify impact, I'm hesitant to give ourselves more work there

dddlr: curious to hear what you think - i could be wrong or missed something

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
dddlr: with the current approach, running the eslint rule in jira gives 4395 violations

though looking at the specific violations that were detected, it looks like this is far more violations than are actually necessary 🤔

the current approach [checks `shorthandBuckets`](https://github.com/atlassian-labs/compiled/blob/master/packages/utils/src/shorthand.ts#L526), which is a mapping of what "buckets" 
each property should fall into, such that more broad shorthand properties (`border`) always come before more specific shorthand properties (`borderColor`, `borderTop`), and more specific shorthand properties always come before longhand properties (`borderTopColor`).

for example, `border` is in bucket 1, while `borderColor` is in bucket 4, meaning that `border` should come before `borderColor`. so for the following code:

```tsx
const styles = css({
  borderColor: '...' // <-- bucket 4
  border: '...', // <-- bucket 1
});
```

the rule will correctly mark this as a lint violation, because a bucket 4 property erroneously comes before a bucket 1 property.

but if we have something like this:

```tsx
const styles = css({
  borderBottom: '...' // <-- bucket 4
  font: '...' // <-- bucket 1
});
```

the rule will still mark this as a lint violation, even though the properties have nothing to do with each other.

maybe an improvement we could do for the ESLint rule is to also check `shorthandFor` (or a custom variable based on `shorthandFor`) to ensure that

* properties that are shorthands/longhands of each other are in the right order (e.g. `border`, `borderColor`, `borderBottom`, `borderTopColor`), while
* properties that are not shorthands/longhands of each other do not trigger a lint violation?

the more lint violations, the more our team or jira developers have to fix... which is not ideal

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
dddlr: i wouldn't consider this to be a dependency 🤔 

* `@atlaskit/css` is explicitly designed to just be a wrapper around `@compiled/react` so we can treat them as equivalent in this case

* neither we nor DST have shied away from adding [ADS-specific logic](https://atlassian.design/) to Compiled (e.g. the two are [already treated as equivalent in `@compiled/babel-plugin`](https://github.com/atlassian-labs/compiled/blob/b4864f9936d3c0cfbba874cf34ae8bd3b736b816/packages/babel-plugin/src/babel-plugin.ts#L34), all the XCSS stuff)

packages/eslint-plugin/src/utils/constants.ts
kylorhall-atlassian: We already do in a few other places, and I even plan to remove the `importSources: ['@atlaskit/css']` config requirement in the 50+ configs we have to add it to internally and make that a defaulted alias 😄 so I approve…given Compiled is nearly exclusively used internally.

Alternatively, we'd want to configure things better internally, but this is easier.

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
pancaspe87: > the rule will still mark this as a lint violation, even though the properties have nothing to do with each other.
This doesn't seem right

> the more lint violations, the more our team or jira developers have to fix
agreed. However, I still believe we shouldn't have an autofixer. A suppression should be enough for existing cases so that developers can be educated and fix their own mistakes; a clear error message should be enough for instant feedback while introducing changes

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
pancaspe87: I am hesitant too because I haven't worked with Compiled in a while - however, I got the feeling that we could just throwing an error when generating the classes. Also, I don't know what implications could cause and I am hesitant on adding changes that could impact build times

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,177 @@
...
+  node: ObjectExpression,
+  propertyA: Property,
+  i: number
+): boolean => {
+  if (node.properties.length > 0) {
+    if (propertyA.key.type === 'Identifier') {
+      const propA = kebabCase(propertyA.key.name) as ShorthandProperties;
+      const depthA = shorthandBuckets[propA];
+
+      for (let j = i + 1; j < node.properties.length; j++) {
```
dddlr, line 59:
```suggestion
      for (const propertyB of node.properties) {
```

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,177 @@
...
+        const propertyB = node.properties[j];
+
+        if (propertyB.type === 'Property') {
```
dddlr, line 62: nit: can we combine the two `if` statements

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,177 @@
...
+          if (propertyB.key.type === 'Identifier') {
+            const propB = kebabCase(propertyB.key.name) as ShorthandProperties;
+            const depthB = shorthandBuckets[propB];
+
+            if (propA === propB) {
+              node.properties.splice(node.properties.indexOf(propertyB), 1);
```
zesmith2, line 68: old code, will delete next commit

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,177 @@
...
+              continue;
+            }
+
+            const shorthandForResA = shorthandFor[propA];
+            const shorthandForResB = shorthandFor[propB];
+
+            if (
+              shorthandForResA !== true &&
```
dddlr, line 76: change to Array.isArray(shorthandForResA) && Array.isArray(shorthandForResB)

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,177 @@
...
+              shorthandForResA !== undefined &&
+              shorthandForResB !== true &&
+              shorthandForResB !== undefined
+            ) {
+              // find intersection between objects
+              const intersectionAB = shorthandForResA.filter((x) => shorthandForResB.includes(x));
```
dddlr, line 82: i wonder if converting both `shorthandForResA` and `shorthandForResB` to sets, and [then running `shorthandForResA.intersection(shorthandForResB)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/intersection) would be faster?

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,177 @@
...
+              if (intersectionAB.length > 0) {
+                if (depthB < depthA) {
+                  return true;
+                }
+              }
+            }
+          }
+        }
+      }
+      return false;
```
dddlr, line 93: you only need one `return false`

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,177 @@
...
+            if (
+              shorthandForResA !== true &&
+              shorthandForResA !== undefined &&
+              shorthandForResB !== true &&
+              shorthandForResB !== undefined
+            ) {
+              // find intersection between objects
+              const intersectionAB = shorthandForResA.filter((x) => shorthandForResB.includes(x));
+
+              if (intersectionAB.length > 0) {
```
dddlr, line 84: nit: change the two `if` statements to one

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
zesmith2: I tried this initially and couldn't find the `intersection()` method on the `Set` type? 🤔

dddlr: ah available in node v22 https://caniuse.com/mdn-javascript_builtins_set_intersection , we are still on node v18 😓

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,161 @@
+import {
+  shorthandBuckets,
+  shorthandFor,
+  kebabCase,
+  type ShorthandProperties,
+} from '@compiled/utils';
+import type { Rule, Scope } from 'eslint';
+import type { ObjectExpression, Property } from 'estree';
+
+import { isStyled, isCss, isCssMap } from '../../index';
```
JakeLane, line 10: Seems like a circular dep? Can you move these to a utils file and import from there?

zesmith2: @JakeLane just moved some things around, does that look better to you?

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -0,0 +1,128 @@
...
+      isCssMap(node.callee as Rule.Node, references) ||
+      isStyled(node.callee as Rule.Node, references))
+  );
+};
+
+export const shorthandFirst: Rule.RuleModule = {
+  meta: {
+    docs: {
+      recommended: true,
+      description:
```
pancaspe87, line 60: Can we change the description to reflect what is in the [README](https://github.com/atlassian-labs/compiled/pull/1714/files#diff-9c805949ce856aceb094bd27cd205e3341156dee6f13f17566c9035cfa494c58R5)

packages/eslint-plugin/src/utils/index.ts
```diff
@@ -3,3 +3,4 @@ export { checkIfCompiledExport as validateDefinition } from './create-no-exporte
 export { createNoTaggedTemplateExpressionRule } from './create-no-tagged-template-expression-rule';
 export { CssMapObjectChecker, getCssMapObject } from './css-map';
 export { isCss, isCssMap, isKeyframes, isCxFunction } from './is-compiled-import';
+export { isStyled } from '../rules/no-styled-tagged-template-expression/utils';
```
dddlr, line 6: why not move `isStyled` to `is-compiled-import`? importing from a parent directory looks like a code smell imo

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
zesmith2: Is the description necessary? I have the url which links directly to the README. I don't want to keep squeezing things into one string it looks messy

packages/eslint-plugin/src/utils/index.ts
zesmith2: Hmm that's a good suggestion actually

packages/eslint-plugin/src/rules/no-styled-tagged-template-expression/index.ts
```diff
@@ -1,8 +1,7 @@
 import type { Rule } from 'eslint';
 
 import { createNoTaggedTemplateExpressionRule } from '../../utils';
```
dddlr, line 3: nit
```suggestion
import { createNoTaggedTemplateExpressionRule, isStyled } from '../../utils';
```

packages/eslint-plugin/src/utils/is-compiled-import.ts
```diff
@@ -13,10 +14,10 @@ const isImportSpecifierWrapper = (name: string) => {
     def.node.imported.type === 'Identifier' &&
     def.node.imported.name === name &&
     def.parent?.type === 'ImportDeclaration' &&
-    def.parent?.source.value === COMPILED_IMPORT;
+    (def.parent.source.value === ATLASKIT_IMPORT || def.parent.source.value === COMPILED_IMPORT);
```
dddlr, line 17: do the types allow us to remove the `?`

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
pancaspe87: If that's the case, I would shorten the description to something more descriptive and link the README, for example

Prevent unwanted side-effect by ensuring shorthand properties are always defined before their related longhands. See more in the README.

So you have only 1 thing to maintain

packages/eslint-plugin/src/rules/shorthand-property-sorting/__tests__/rule.test.ts
```diff
@@ -0,0 +1,240 @@
...
+      const styles = ${call}({
+        borderTop: '1px solid #00b8d9',
+        border: '#00b8d9',
+      });
+      export const EmphasisText = ({ children }) => <span css={styles}>{children}</span>;
+    `,
+      errors: [{ messageId: 'shorthand-first' }],
+    })),
+    ...packages_calls_and_imports.map(([pkg, call, imp]) => ({
+      name: `incorrect property ordering -> 3 properties (${pkg}: '${imp}'`,
```
pancaspe87, line 97: out of curiosity, what information does `3 properties` provide?

packages/eslint-plugin/src/rules/shorthand-property-sorting/__tests__/rule.test.ts
```diff
@@ -0,0 +1,240 @@
...
+        },
+        borderTop: '1px solid #00b8d9', // 4
+        borderColor: 'red', // 2
+        border: '1px solid #00b8d9', // 1
+      })
+      `,
+      errors: [{ messageId: 'shorthand-first' }],
+    })),
+
+    /* fixer can't deal with nested fixing in one go. I've split this test into:
```
pancaspe87, line 206: Is this comment still required?

packages/eslint-plugin/src/utils/constants.ts
kylorhall-atlassian: Plan actioned: https://github.com/atlassian-labs/compiled/pull/1720
(I will fix my PR or this PR, whichever one lands first)

packages/eslint-plugin/src/utils/is-compiled-import.ts
zesmith2: Yeah unfortunately not

packages/eslint-plugin/src/rules/shorthand-property-sorting/__tests__/rule.test.ts
zesmith2: Yes, I that test can't deal with the whole case the first time round and I had to split it in 2, it looks confusing without that context

packages/eslint-plugin/src/rules/shorthand-property-sorting/__tests__/rule.test.ts
zesmith2: Not a whole lot since the rest of the tests also deal with this, can take it out while I make the other changes.

packages/eslint-plugin/src/rules/shorthand-property-sorting/__tests__/rule.test.ts
pancaspe87: I thought we remove the autofixer

pancaspe87 (approved):

PR #1717, opened 2024-10-09
Add the option to provide a prefix to generated class names 

guilhermehto:
### What is this change?

This PR introduces a new configuration prop to `@compiled/babel-plugin` named `classHashPrefix` (feel free to suggest a better alternative) that allows users to define a prefix to be added to when generating class name hashes. 

We have an internal problem in a consumer of a remote module where it has its styles messed up due to the loss of specificity once the remote module is loaded. This is because both the wrapping application and the remote module use `@compiled` and due to the way we generate the class names, both end up generating a few with the same name. 

This is the first time I ever work with a Babel plugin so I might be missing something, all feedback is appreciated. I tried to follow the code and tests to figure out where we were generating the hashes and hooked it up with the new config prop. 

Lastly, I only worked off of tests on this, guidance on how to better test it are welcomed 😅 

~_edit:_ I assume I'm probably missing a changeset on this pr~ done

---

### PR checklist

Don't delete me!

I have...

### DIFF
--- a/packages/babel-plugin/src/__tests__/index.test.ts
+++ b/packages/babel-plugin/src/__tests__/index.test.ts
@@ -205,7 +205,7 @@ describe('babel plugin', () => {
     expect(actual).toInclude('c_MyDiv');
   });
 
-  it.only('should add a prefix to style hash hashPrefix is present', () => {
+  it('should add a prefix to style hash hashPrefix is present', () => {
     const actual = transform(
       `
       import { styled } from '@compiled/react';
@@ -220,7 +220,7 @@ describe('babel plugin', () => {
     expect(actual).toInclude('myprefix');
   });
 
-  it.only('should throw if a given hashPrefix is not a valid css identifier', () => {
+  it('should throw if a given hashPrefix is not a valid css identifier', () => {
     expect(() => {
       transform(
         `
--- a/packages/babel-plugin/src/__tests__/index.test.ts
+++ b/packages/babel-plugin/src/__tests__/index.test.ts
@@ -205,6 +205,36 @@ describe('babel plugin', () => {
     expect(actual).toInclude('c_MyDiv');
   });
 
+  it.only('should add a prefix to style hash hashPrefix is present', () => {
+    const actual = transform(
+      `
+      import { styled } from '@compiled/react';
+
+      const MyDiv = styled.div\`
+        font-size: 12px;
+      \`;
+    `,
+      { hashPrefix: 'myprefix' }
+    );
+
+    expect(actual).toInclude('myprefix');
+  });
+
+  it.only('should throw if a given hashPrefix is not a valid css identifier', () => {
+    expect(() => {
+      transform(
+        `
+        import { styled } from '@compiled/react';
+
+        const MyDiv = styled.div\`
+          font-size: 12px;
+        \`;
+        `,
+        { hashPrefix: '$invalid%' }
+      );
+    }).toThrow();
+  });
+
   it('should compress class name for styled component', () => {
     const actual = transform(
       `
--- a/packages/babel-plugin/src/__tests__/index.test.ts
+++ b/packages/babel-plugin/src/__tests__/index.test.ts
@@ -205,7 +205,7 @@ describe('babel plugin', () => {
     expect(actual).toInclude('c_MyDiv');
   });
 
-  it.only('should add a prefix to style hash hashPrefix is present', () => {
+  it('should add a prefix to style hash classHashPrefix is present', () => {
     // changes to css/src/plugins/atomicify-rules can break this due to how the class name is hashed
     const hashedClassName = '_1lv61fwx';
     const actual = transform(
@@ -216,13 +216,13 @@ describe('babel plugin', () => {
         font-size: 12px;
       \`;
     `,
-      { hashPrefix: 'myprefix' }
+      { classHashPrefix: 'myprefix' }
     );
 
     expect(actual).toInclude(hashedClassName);
   });
 
-  it.only('should throw if a given hashPrefix is not a valid css identifier', () => {
+  it('should throw if a given classHashPrefix is not a valid css identifier', () => {
     expect(() => {
       transform(
         `
@@ -232,7 +232,7 @@ describe('babel plugin', () => {
           font-size: 12px;
         \`;
         `,
-        { hashPrefix: '$invalid%' }
+        { classHashPrefix: '$invalid%' }
       );
     }).toThrow();
   });
--- a/packages/babel-plugin/src/__tests__/index.test.ts
+++ b/packages/babel-plugin/src/__tests__/index.test.ts
@@ -205,7 +205,9 @@ describe('babel plugin', () => {
     expect(actual).toInclude('c_MyDiv');
   });
 
-  it('should add a prefix to style hash hashPrefix is present', () => {
+  it.only('should add a prefix to style hash hashPrefix is present', () => {
+    // changes to css/src/plugins/atomicify-rules can break this due to how the class name is hashed
+    const hashedClassName = '_1lv61fwx';
     const actual = transform(
       `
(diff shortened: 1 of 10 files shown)


packages/babel-plugin/src/__tests__/index.test.ts
```diff
@@ -205,6 +205,36 @@ describe('babel plugin', () => {
     expect(actual).toInclude('c_MyDiv');
   });
 
+  it.only('should add a prefix to style hash hashPrefix is present', () => {
```
JakeLane, line 208:
```suggestion
  it('should add a prefix to style hash hashPrefix is present', () => {
```

JakeLane (changes requested):
How does this work with `ax`? That parses the classname and I don't think it'll work without changes.

Have a read of https://github.com/atlassian-labs/compiled/blob/master/packages/react/src/runtime/ax.ts#L30 and probably add some tests there

guilhermehto:
hey @JakeLane I was playing around with the code but I couldn't find any clear solutions besides adding the hash prefix into the hash of the group. This way groups would wind up with the same hash and `ax` should work as expected, I believe 🤔 

Essentially

```typescript
const group = hash(`${opts.atRule}${selectors}${node.prop}`).slice(0, 4);
// becomes
const group = hash(`${opts.hashPrefix}${opts.atRule}${selectors}${node.prop}`).slice(0, 4);
```

Seems too easy to be true though, and I potentially missing something?

packages/babel-plugin/src/types.ts
```diff
@@ -103,6 +103,12 @@ export interface PluginOptions {
    * Defaults to `true`.
    */
   sortAtRules?: boolean;
+
+  /**
+   * Adds a defined prefix to the generated classes' hashes.
+   * Useful in micro frontend environments to avoid clashing/specificity issues.
+   */
+  hashPrefix?: string;
```
pancaspe87, line 111: nit

what about `addClassPrefix`? cc @dddlr @JakeLane

JakeLane: yeah more descriptive naming is helpful

JakeLane:
> hey @JakeLane I was playing around with the code but I couldn't find any clear solutions besides adding the hash prefix into the hash of the group. This way groups would wind up with the same hash and `ax` should work as expected, I believe 🤔
> 
> Essentially
> 
> ```ts
> const group = hash(`${opts.atRule}${selectors}${node.prop}`).slice(0, 4);
> // becomes
> const group = hash(`${opts.hashPrefix}${opts.atRule}${selectors}${node.prop}`).slice(0, 4);
> ```
> 
> Seems too easy to be true though, am I potentially missing something?
> 
> _edit:_ Obviously we'd lose the ability to check for the prefix when looking through the class names as they'd be hashed but that's not really why we're making this change to begin with so I think it's an acceptable tradeoff.

I think this is fine. It might be good to validate VR test correctness manually; I don't think we have a way to change the babel config on VR tests right now.

packages/babel-plugin/src/types.ts
guilhermehto: Imo `add` suggests the expected value is a boolean and not a string that we're setting. Would `classHashPrefix` be better in this case?

dddlr: ye that's fine

website/packages/docs/src/pages/pkg-babel-plugin.mdx
```diff
@@ -93,6 +94,15 @@ Extensions that we should consider code. We use these to identify if a file shou
 - Type: `string[]`
 - Default: `['.js', '.jsx', '.ts', '.tsx']`
 
+#### hashPrefix
+
+Adds a prefix to the generated hashed css rule names.
```
dddlr, line 99: nit: mention what format `hashPrefix` is required to be

e.g. how it needs to match the `/^[a-zA-Z\-][a-zA-Z\-_0-9]*$/` regex

packages/css/src/plugins/atomicify-rules.ts
```diff
@@ -8,8 +8,21 @@ interface PluginOpts {
...
 
+/**
+ * Returns true if a given string is a valid CSS identifier
+ *
+ * @param value the value to test
+ * @returns `true` if given value is valid, `false` if not
+ *
+ */
+const isCssIdentifierValid = (value: string): boolean => {
+  const validCssIdentifierRegex = /^[a-zA-Z\-][a-zA-Z\-_0-9]*$/;
```
dddlr, line 22: where did you find this regex? it looks reasonable but it doesn't match with what i could easily find online e.g. https://stackoverflow.com/a/449000

website/packages/docs/src/pages/pkg-babel-plugin.mdx
guilhermehto: ah yep good idea!

packages/css/src/plugins/atomicify-rules.ts
guilhermehto: I made it myself as it was supposed to be in the middle of an already existing identifier. It's actually not 100% now that I look at it; 

That said, since it's now being used in the hash and not as a standalone string it could be pretty much anything but I think it makes sense to continue to only accept identifiers to prevent any potential unwanted side effects. I'll refactor it

packages/css/src/plugins/atomicify-rules.ts
```diff
@@ -21,14 +34,22 @@ interface PluginOpts {
  *
  * @param node CSS declaration
  * @param opts AtomicifyOpts
+ *
+ * @throws Throws an error if `opts.hashPrefix` contains invalid css class/id characters
  */
 const atomicClassName = (node: Declaration, opts: PluginOpts) => {
+  if (opts.hashPrefix && !isCssIdentifierValid(opts.hashPrefix)) {
```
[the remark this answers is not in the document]
guilhermehto: can do! Probably a better idea, we most likely don't wanna be running regex matches all willy nilly

packages/css/src/plugins/atomicify-rules.ts
```diff
@@ -8,8 +8,21 @@ interface PluginOpts {
...
+/**
+ * Returns true if a given string is a valid CSS identifier
+ *
+ * @param value the value to test
+ * @returns `true` if given value is valid, `false` if not
+ *
+ */
+const isCssIdentifierValid = (value: string): boolean => {
+  const validCssIdentifierRegex = /^[a-zA-Z\-_]+[a-zA-Z\-_0-9]*$/;
+  return Boolean(value.match(validCssIdentifierRegex));
```
dddlr, line 23: oops sorry i just realised, can this be simplified to `validCssIdentifierRegex.test(value)` instead?

packages/css/src/plugins/atomicify-rules.ts
```diff
@@ -21,10 +34,14 @@ interface PluginOpts {
  *
  * @param node CSS declaration
  * @param opts AtomicifyOpts
+ *
+ * @throws Throws an error if `opts.classHashPrefix` contains invalid css class/id characters
```
dddlr, line 38: this `@throws` should be removed

packages/babel-plugin/src/__tests__/index.test.ts
```diff
@@ -205,6 +205,38 @@ describe('babel plugin', () => {
...
+      import { styled } from '@compiled/react';
+
+      const MyDiv = styled.div\`
+        font-size: 12px;
+      \`;
+    `,
+      { classHashPrefix: 'myprefix' }
+    );
+
+    expect(actual).toInclude(hashedClassName);
```
liamqma, line 222: should we check `actual` include `classHashPrefix` too?

liamqma (approved):

dddlr (approved):

JakeLane (approved):

PR #1720, opened 2024-10-11
Make support for `@atlaskit/css` as a first-class import consistently default.

kylorhall-atlassian:
### What is this change?

This means the same functionality of parsing JSX pragmas, linting specific imports, and extracting styles should all work from `@compiled/react` and `@atlaskit/css` equally without the `importSources: ['@atlaskit/css']` config we use internally.

This was already the default in about 1/3rd of the code, but not consistent. Now it's consistent and I've cleaned up duplicated import patterns.

With the public launch of the Atlassian Design System using Compiled soon, this configuration must be actioned by everyone using our plugins, we might as well make it default.  This would also cleanup dozens of configs internally.

### PR checklist

- [x] Updated or added applicable tests — might be room for additional tests, but I can't find one.

---

NOTE: https://github.com/atlassian-labs/compiled/pull/1714/files#r1783837443 has related changes, depending on which merges first, I'll handle this.


JakeLane (dismissed):

packages/eslint-plugin/src/rules/jsx-pragma/index.ts
```diff
@@ -66,10 +66,7 @@ const findReactDeclarationWithDefaultImport = (
 };
 
 function createFixer(context: Rule.RuleContext, source: SourceCode, options: Options) {
-  const compiledImports = findLibraryImportDeclarations(context, [
-    COMPILED_IMPORT,
-    ...options.importSources,
-  ]);
+  const compiledImports = findLibraryImportDeclarations(context, options.importSources);
```
dddlr, line 69: should we update `findLibraryImportDeclarations` too?

packages/parcel-transformer/src/index.ts
```diff
@@ -90,8 +90,10 @@ export default new Transformer<ParcelTransformerOpts>({
 
     const code = await asset.getCode();
     if (
-      code.indexOf('@compiled/react') === -1 &&
-      (config.importSources?.every((importSource) => !code.includes(importSource)) ?? true)
+      // If Compiled (default) and noone of the additional import sources are not found in the code, we bail out.
```
dddlr, line 93:
```suggestion
      // If neither Compiled (default) nor any of the additional import sources are found in the code, we bail out.
```

dddlr (approved):
lgtm 👍  as long as you apply this nit fix:
https://github.com/atlassian-labs/compiled/pull/1720/files#r1796343239

as for https://github.com/atlassian-labs/compiled/pull/1720/files#r1796334209 , i'm not sure whether `findLibraryImportDeclarations` needs to be updated to include `@atlaskit/css` by default -- if you find that adding `@atlaskit/css` there would break anything, feel free not to add it

packages/eslint-plugin/src/rules/jsx-pragma/index.ts
kylorhall-atlassian: Yeah, you got it, this `createFixer(…, options)` comes from:
```tsx
const options: Options = {  
  // …
  importSources: [...DEFAULT_IMPORT_SOURCES, ...(optionsRaw.importSources ?? [])],
};
```

Updated `sources = [COMPILED_IMPORT]` to be `sources = DEFAULT_IMPORT_SOURCES` in 6b49bde6

Only other remaining `COMPILED_IMPORT` used in _isolation_ is either for codemod or autofixes or for `styled` imports, which don't exist in `DEFAULT_IMPORT_SOURCES` (might be odd, but alas). 🙏

packages/eslint-plugin/src/utils/is-compiled-import.ts
```diff
@@ -14,7 +14,7 @@ const isImportSpecifierWrapper = (name: string) => {
     def.node.imported.type === 'Identifier' &&
     def.node.imported.name === name &&
     def.parent?.type === 'ImportDeclaration' &&
-    (def.parent.source.value === ATLASKIT_IMPORT || def.parent.source.value === COMPILED_IMPORT);
+    DEFAULT_IMPORT_SOURCES.includes(String(def.parent.source.value));
 };
```
kylorhall-atlassian, line 18: Fixed when I rebased in #1714 cc: @zesmith2 @dddlr I got this right, right?

packages/eslint-plugin/src/rules/jsx-pragma/index.ts
dddlr: ahh cool, yeah that makes sense

dddlr (approved):

PR #1724, opened 2024-10-15
Throw an error when mixing `extract: true` and `classHashPrefix` configuration options to avoid unsupported usage and bundle size bloat.

kylorhall-atlassian:
### What is this change?

Following https://github.com/atlassian-labs/compiled/pull/1717, throws an error when Parcel attempts to transform or Webpack's `CompiledExtractPlugin` is initialized with both `extract: true` and `classHashPrefix` being defined.

To avoid unsupported usage and bundle size creep as adding 5 bytes to every class is not something we want to generally promote and instead `classHashPrefix` is only desired as a temporary workaround to avoid conflicts in scenarios where non-extraction is used.

Into the Parcel Transformer and Webpack Loader, the entrypoints where this configuration bleeds in.  Technically this wasn't documented properly for `@compiled/webpack-loader`, so I adjusted it briefly, but kept the `throw new Error` in place to avoid this (as it would still be passed through, I'd gather).

---

### PR checklist

- [x] NOT ~updated or added applicable tests~ — I don't think this is feasible without making a brand new fixture for each scenario.
- [x] Updated the documentation in `website/` — Modified briefly, but I'm not sure it's worth documenting in full (this already isn't documented in full).
- [x] Review @guilhermehto's internal PR to validate it will work with this


kylorhall-atlassian:
> No Changeset found
Oh, do I need a changeset for the website…?

packages/babel-plugin/src/types.ts
```diff
@@ -107,6 +107,9 @@ export interface PluginOptions {
   /**
    * Adds a defined prefix to the generated classes' hashes.
    * Useful in micro frontend environments to avoid clashing/specificity issues.
+   *
+   * Avoid mixing this with extraction as this may throw an error if combined with extraction
+   * or `extract: true` in Webpack loaders or Parcel tranformers.
```
kylorhall-atlassian, line 112: This documentation isn't relevant, I don't think you can mix Babel + extraction elsewhere, but just covering my bases…

packages/webpack-loader/src/extract-plugin.ts
```diff
@@ -108,6 +108,13 @@ export class CompiledExtractPlugin {
 
   constructor(options: CompiledExtractPluginOptions = {}) {
     this.#options = options;
+
+    // @ts-expect-error -- Make sure this config doesn't bleed in as it's passed through
+    if (options.classHashPrefix) {
+      throw new Error(
+        '`@compiled/webpack-loader.CompiledExtractPlugin` is mixing `extract: true` and `classHashPrefix` options, which will not supported and will result in bundle size bloat.'
+      );
+    }
```
kylorhall-atlassian, line 117: I've removed the config actually, so you can't do `CompiledExtractPlugin({ classHashPrefix: … })`, but kept the `@ts-expect-error` to document it in full (because it'd work without TS blocking you, I believe.

website/packages/docs/src/pages/pkg-webpack-loader.mdx
```diff
@@ -211,7 +211,6 @@ module.exports = {
 `@compiled/webpack-loader` also accepts the following options. These are not used in the loader itself, but instead they are passed directly to the underlying `@compiled/babel-plugin`.
 
 - `addComponentName`
-- `classHashPrefix`
```
kylorhall-atlassian, line 214: This was actually on the wrong place, this says it works with `@compiled/webpack-loader`, but I'm _pretty sure_ it only works with `new CompiledExtractPlugin({…})`.  Could be wrong!

liamqma (approved):

packages/parcel-transformer/src/index.ts
```diff
@@ -124,6 +124,12 @@ export default new Transformer<ParcelTransformerOpts>({
   },
 
   async transform({ asset, config, options }) {
+    if (config.extract && config.classHashPrefix) {
```
dddlr, line 127: note that `config.extract` is currently ignored in non-production environments when using Parcel

packages/webpack-loader/src/extract-plugin.ts
```diff
@@ -108,6 +108,13 @@ export class CompiledExtractPlugin {
 
   constructor(options: CompiledExtractPluginOptions = {}) {
     this.#options = options;
+
+    // @ts-expect-error -- Make sure this config doesn't bleed in as it's passed through
+    if (options.classHashPrefix) {
+      throw new Error(
+        '`@compiled/webpack-loader.CompiledExtractPlugin` is mixing `extract: true` and `classHashPrefix` options, which will not supported and will result in bundle size bloat.'
```
dddlr, line 115:
```suggestion
        '`@compiled/webpack-loader.CompiledExtractPlugin` is mixing the `extract: true` and `classHashPrefix` options, which is not supported and will result in bundle size bloat.'
```

packages/parcel-transformer/src/index.ts
kylorhall-atlassian: Got it, this kind of might be the wrong thing in that case if there's no different configs…I'm not sure exactly, might wait to see the PR from @guilhermehto to see if this would blow up or not.

packages/webpack-loader/src/extract-plugin.ts
```diff
@@ -108,6 +108,13 @@ export class CompiledExtractPlugin {
 
   constructor(options: CompiledExtractPluginOptions = {}) {
     this.#options = options;
+
+    // @ts-expect-error -- Make sure this config doesn't bleed in as it's passed through
+    if (options.classHashPrefix) {
+      throw new Error(
+        '`@compiled/webpack-loader.CompiledExtractPlugin` is mixing the `extract: true` and `classHashPrefix` options, which is not supported and will result in bundle size bloat.'
+      );
+    }
```
kylorhall-atlassian, line 117: This may not work given the `classHashPrefix` is passed to the babel plugin isn't it?  I'm a bit uncertain how this works to be frank!

liamqma (approved):

PR #1726, opened 2024-10-16
chore: added Zara as reviewer

pancaspe87:


stories/binary-expression.tsx
```diff
@@ -65,11 +65,34 @@ export const OrderOfOperations = (): JSX.Element => {
   return <Component>Order of operations - padding: 5 * (spacing + 2)</Component>;
 };
 
+const RightArrow = styled.div({
```
pancaspe87, line 68: unintentional

PR #1727, opened 2024-10-16
Add flat config preset for compiled & fix docs

### LINKED ISSUE #1691: Update eslint-plugin to include full list of eslint rules
From https://github.com/atlassian-labs/compiled/pull/1689

https://github.com/atlassian-labs/compiled/tree/master/packages/eslint-plugin does not currently contain a full list of ESLint rules defined by `@compiled/eslint-plugin`. We should fix this so that every ESLint rule is listed.

AllySummers:
### What is this change?

This PR:
- Adds a flat config preset for the ESLint plugin
- Corrects the readme for the eslint plugin to match the actual code

- So that people using `@compiled/eslint-plugin` can use this with flat config without the legacy compat utils
- To make our documentation more accurate

### PR checklist

Don't delete me!

I have...

Closes #1691


packages/eslint-plugin/src/rules/no-keyframes-tagged-template-expression/index.ts
```diff
@@ -5,6 +5,8 @@ import { createNoTaggedTemplateExpressionRule, isKeyframes } from '../../utils';
 export const noKeyframesTaggedTemplateExpressionRule: Rule.RuleModule = {
   meta: {
     docs: {
+      recommended: true,
```
kylorhall-atlassian, line 8: Adding a few `recommended: true` _could_ be a problem, but this is a minor bump on a `0.x` version which means major to me, LGTM if Eurybia is happy!

packages/eslint-plugin/src/index.ts
```diff
@@ -29,8 +34,25 @@ export const rules = {
...
+  version,
+  rules,
+  configs: {
+    recommended,
+    'flat/recommended': {
+      ...flatRecommended,
+      plugins: {
+        get '@compiled'() {
+          return plugin;
+        },
```
kylorhall-atlassian, line 50: This is a bit more coherent than `plugin.configs['flat/recommended'].plugins['@compiled'] = plugin` which is what you'd have to do otherwise, right?

Only real concern is could this ever backfire with some Node support?  I feel like it's been around _forever_ but also a syntax I never use (because when it was introduced it wasn't ready to be used)…it's good now, right?

Also, would you ever want to merge this though?
```suggestion
        ...flatRecommended.plugins,
        get '@compiled'() {
          return plugin;
        },
```

kylorhall-atlassian (dismissed):
From what I can see you really uplifted things a bit!  I've got no concerns with this having interacted on the other plugins you've uplifted!

Worth an FYI review from the owners still though!

liamqma (approved):

liamqma (dismissed):

packages/eslint-plugin/README.md
```diff
@@ -10,13 +10,42 @@ npm install @compiled/eslint-plugin --save-dev
 
 ## Usage
 
+### Flat Config
+
+Import the `@compiled/eslint-plugin` and add it to your plugins like so, then configure the ruels you want to use under the rules section.
```
dddlr, line 15: nit:
```suggestion
Import the `@compiled/eslint-plugin` and add it to your plugins like so, then configure the rules you want to use under the rules section.
```

feel free to skip tho, not important

packages/eslint-plugin/README.md
```diff
@@ -32,15 +61,22 @@ You can also enable the recommended rules for compiled by adding `plugin:@compil
...
-| [@compiled/no-exported-css](./src/rules/no-exported-css)                                                 | Disallows `css` usages from being exported           | :white_check_mark: |          |
-| [@compiled/no-exported-keyframes](./src/rules/no-exported-keyframes)                                     | Disallows `keyframes` usages from being exported     | :white_check_mark: |          |
-| [@compiled/no-keyframes-tagged-template-expression](./src/rules/no-keyframes-tagged-template-expression) | Disallows the `keyframes` tagged template expression | :white_check_mark: | :wrench: |
-| [@compiled/no-styled-tagged-template-expression](./src/rules/no-styled-tagged-template-expression)       | Disallows the `styled` tagged template expression    | :white_check_mark: | :wrench: |
-| [@compiled/no-css-prop-without-css-function](./src/rules/no-css-prop-without-css-function)               | Disallows css prop without the css function          | :white_check_mark: | :wrench: |
+✅ Included in the recommended configuration.\
```
dddlr, line 64: thank you ❤️ ❤️ ❤️

dddlr (approved):

liamqma (approved):

dddlr (approved):

PR #1729, opened 2024-10-22
cloneElement support

danieldelcore:
### What is this change?

Compiled now treats `react.cloneElement` as a stylable element.

this can be extended to support `react.createElement` fairly easily.

To allow cases where a CSS prop is used like so:

```tsx
import { cloneElement } from 'react';

const MyDiv = ({ children }) => cloneElement(children, { css: css({ fontSize: 12 }) });
```

- Adds logic to the babel plugin to make it aware of `cloneElement` `CallExpression`
- Looks at the second argument (props) for an inline `css` prop 
- (this is fragile when indirection is introduced, open to feedback on how to address this properly)
- Rename the object key to `className`
- Wrap call in compiled runtime API

This might not be a direction the library wants to go in, which is understandable.
Alternatively, we could address this via `<ClassNames>` however Compiled currently does not pick up non-jsx elements as children of the `<ClassNames>` API. 

---

### PR checklist

Don't delete me!

I have...


packages/babel-plugin/src/clone-element/index.ts
```diff
@@ -0,0 +1,183 @@
...
+
+      if (!ancestorPath) {
+        return;
+      }
+
+      ancestorPath.replaceWith(buildCompiledCloneElement(ancestorPath.node, builtCss, meta));
+    },
+  });
+
+  // // Second pass to replace all usages of `style`.
```
danieldelcore, line 144: TODO, makes sense to hook the `style` prop up as well

packages/babel-plugin/src/__tests__/index.test.ts
```diff
@@ -114,6 +114,36 @@ describe('babel plugin', () => {
     `);
   });
 
+  it('should transform React.cloneElement css prop', () => {
+    const actual = transform(`
+      import { cloneElement } from 'react';
+      import { css } from '@compiled/react';
+
+      const MyDiv = ({ children }) => {
+        return cloneElement(children, { css: css({ fontSize: 12 }) });
```
kylorhall-atlassian, line 123: This is would be the full example of what we need this for—do you think it would work?
```tsx
const styles = css({ fontSize: 12 });
const MyDiv = ({ children, className }) => {
  return cloneElement(children, { css: cx(styles, props.className) });
}
```

packages/babel-plugin/src/clone-element/index.ts
kylorhall-atlassian: Does anything else ever use `style` in Babel?  `style={css(…)}` isn't valid, right?

packages/babel-plugin/src/__tests__/index.test.ts
kylorhall-atlassian: The other question I've got, probably for Compiled, is should this be `props.css` or `props.className` exclusively?

kylorhall-atlassian (commented):
Without pulling it up locally, I'm not of much help reviewing the AST, but great like spike into Compiled, I hope it's not too far off and they'll like—it will drastically reduce some technical debt as the alternative is using `<ClassNames>{({ css }) => cloneElement(…, { className: ax([css({…}), props.className]) })}</ClassNames>` or something.

packages/babel-plugin/src/utils/build-compiled-component.ts
```diff
@@ -142,3 +142,58 @@ export const buildCompiledComponent = (
...
+  meta: Metadata
+): t.Node => {
+  const { sheets, classNames } = transformCssItems(cssOutput.css, meta);
+
+  const props = node.arguments[1];
+
+  // TODO: This is a temporary fix to prevent the plugin from crashing when the second argument of cloneElement is not an object expression.
+  if (!t.isObjectExpression(props)) {
+    throw new Error('Second argument of cloneElement must be an object expression.');
+  }
```
danieldelcore, line 165: If someone is using a variable/spread/etc it'll probably be impossible to reliably fish out and modify the CSS prop. 
Are there any standard approaches to this problem in use in the repo already? Do we throw/Log an err/try really hard to locate the variable?

packages/babel-plugin/src/babel-plugin.ts
```diff
@@ -295,6 +296,19 @@ export default declare<State>((api) => {
         path: NodePath<t.TaggedTemplateExpression> | NodePath<t.CallExpression>,
         state: State
       ) {
+        if (
```
dddlr, line 299: should this check that `cloneElement` is imported from `react` too?

packages/babel-plugin/src/__tests__/index.test.ts
danieldelcore: I had the same question. 

I'll add a test case for the example you provided!

packages/babel-plugin/src/babel-plugin.ts
danieldelcore: Yes definitely and also should check for aliased imports

```
import { cloneElement as foo } from 'react';
```

packages/babel-plugin/src/clone-element/index.ts
kylorhall-atlassian: That's only when it's in runtime mode, but yeah, maybe right…really glad if you can learn this part of it because I have no clue how extraction or runtime _actually work_ 🤣

danieldelcore: Neither hahah

packages/babel-plugin/src/babel-plugin.ts
danieldelcore: Edit: I added support for alias and member expressions in the latest commit

PR #1730, opened 2024-10-22
Fix shorthand sorting buggy

dddlr:
### What is this change?

Fix the shorthand property sorting being buggy when not using runtime mode (e.g. when using stylesheet extraction).

@kylorhall-atlassian and I noticed that `sortShorthand` wasn't actually sorting the properties as it should.

It turns out that a lot of the time in the sorting algorithm, the two properties being compared are

* not related to each other (e.g. `padding` is related to `paddingLeft`, but `padding` is not related to `borderColor`), and
* one of the properties is a longhand property that doesn't exist in `shorthandBuckets`

As a result, there are cases where the stylesheet would not be sorted _at all_, simply because the wrong nodes were being compared and so the `sortNodes` function returns `0` every time.

Fix and simplify the `sortNodes` function:

* Remove the logic of checking whether one property is a shorthand of another. It is safe to assume that this logic evaluates to `false` most of the time anyway.
* In `aShorthandBucket` and `bShorthandBucket`, set the default value to `Infinity`. This way, if one property is a shorthand property and the other property is not a shorthand property, we guarantee that the shorthand property will come in front. For example, `margin` will come before `paddingRight` (even though they're not related).

---

### PR checklist

Don't delete me!

I have...

- [x] Verify the changes to existing tests are correct
- [x] Verify that this PR fixes the lack of shorthand sorting in Jira
- [x] Address this PR comment from Kylor: https://github.com/atlassian-labs/compiled/pull/1730#discussion_r1809966320
- [x] Add changeset

### DIFF
--- a/packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
+++ b/packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
@@ -36,8 +36,8 @@ describe('sort shorthand vs. longhand declarations', () => {
 
     expect(actual).toMatchInlineSnapshot(`
       ".a {
-        outline-width: 1px;
         font: 16px;
+        outline-width: 1px;
       }
       .b {
         font: 24px;
@@ -69,8 +69,8 @@ describe('sort shorthand vs. longhand declarations', () => {
     expect(actual).toMatchInlineSnapshot(`
       ".a {
         outline: none;
-        outline-width: 1px;
         font: 16px normal;
+        outline-width: 1px;
         font-weight: bold;
       }
       .b {
@@ -147,8 +147,8 @@ describe('sort shorthand vs. longhand declarations', () => {
 
       .a {
         outline: none;
-        outline-width: 1px;
         font: 16px normal;
+        outline-width: 1px;
         font-weight: bold;
       }
       .b {
@@ -162,14 +162,14 @@ describe('sort shorthand vs. longhand declarations', () => {
 
       .a:focus {
         outline: none;
-        outline-width: 1px;
         font: 16px normal;
+        outline-width: 1px;
         font-weight: bold;
       }@media all {
         .a {
           outline: none;
-          outline-width: 1px;
           font: 16px normal;
+          outline-width: 1px;
           font-weight: bold;
         }
         .b {
@@ -284,10 +284,10 @@ describe('sort shorthand vs. longhand declarations', () => {
 
     expect(actual).toMatchInlineSnapshot(`
       "
-      .a {
+      .a > .external {
         all: unset;
       }
-      .a > .external {
+      .a {
         all: unset;
       }
       .b[disabled] {
@@ -308,18 +308,21 @@ describe('sort shorthand vs. longhand declarations', () => {
       .c[data-foo='bar'] {
         border-top: none;
       }
+      .d {
+        border-top: none;
+      }
+      .c {
+        border-block-start: none;
+      }
+      .d:active {
+        border-block-start: none;
+      }
 
       .j {
         margin-bottom: 6px;
       }
       .f {
         display: block;
-      }
-      .d {
-        border-top: none;
-      }
-      .c {
-        border-block-start: none;
       }
       .e {
         border-block-start-color: transparent;
@@ -328,17 +331,11 @@ describe('sort shorthand vs. longhand declarations', () => {
       .f:focus {
         display: block;
       }
-      .d:active {
-        border-block-start: none;
-      }
       .e:hover {
         border-block-start-color: transparent;
       }@media all {
         .a {
           all: unset;
-        }
-        .f {
-          display: block;
         }
         .b {
           border: none;
@@ -348,6 +345,9 @@ describe('sort shorthand vs. longhand declarations', () => {
         }
         .c {
           border-block-start: none;
+        }
+        .f {
+          display: block;
         }
         .e {
           border-block-start-color: transparent;
@@ -386,10 +386,58 @@ describe('sort shorthand vs. longhand declarations', () => {
         border-block-start: 1px solid;
         border-block-start-color: transparent;
       }
-      .b { all: unset; }
       .c { all: unset; }
+      .b { all: unset; }
       .d { border: none; }
       .f { border-block-start-color: transparent; }"
+    `);
+  });
+
+  it('sorts a subset of atomic classes taken from a real stylesheet', () => {
+    const actual = transform(outdent`
+      ._oqicidpf{padding-top:0}
+      ._1rmjidpf{padding-right:0}
+      ._cjbtidpf{padding-bottom:0}
+      ._pnmbidpf{padding-left:0}
+      ._glte7vkz{width:1pc}
+      ._165t7vkz{height:1pc}
+      ._ue5g1408{margin:0 var(--ds-space-800,4pc)}
+      ._1yag1dzv{padding:var(--ds-space-100) var(--ds-space-150)}
+      ._dbjg12x7{margin-top:var(--ds-space-075,6px)}
+
+      @media (min-width:1200px){
+        ._jvpg11p5{display:grid}
+        ._szna1wug{margin-top:auto}
+        ._13on1wug{margin-right:auto}
+        ._1f3k1wug{margin-bottom:auto}
+        ._inid1wug{margin-left:auto}
+        ._1oqj1epz{padding:var(--ds-space-1000,5pc)}
+        ._12wp9ac1{max-width:1400px}
+        ._jvpgglyw{display:none}
+      }
+    `);
+
+    expect(actual).toMatchInlineSnapshot(`
+      "
+      ._ue5g1408{margin:0 var(--ds-space-800,4pc)}
+      ._1yag1dzv{padding:var(--ds-space-100) var(--ds-space-150)}._oqicidpf{padding-top:0}
+      ._1rmjidpf{padding-right:0}
+      ._cjbtidpf{padding-bottom:0}
+      ._pnmbidpf{padding-left:0}
+      ._glte7vkz{width:1pc}
+      ._165t7vkz{height:1pc}
+      ._dbjg12x7{margin-top:var(--ds-space-075,6px)}
+
+      @media (min-width:1200px){
+        ._1oqj1epz{padding:var(--ds-space-1000,5pc)}
+        ._jvpg11p5{display:grid}
+        ._szna1wug{margin-top:auto}
+        ._13on1wug{margin-right:auto}
+        ._1f3k1wug{margin-bottom:auto}
+        ._inid1wug{margin-left:auto}
+        ._12wp9ac1{max-width:1400px}
+        ._jvpgglyw{display:none}
+      }"
     `);
   });
 
--- a/packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
+++ b/packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
@@ -393,7 +393,7 @@ describe('sort shorthand vs. longhand declarations', () => {
     `);
   });
 
-  it('sorts a subset of atomic classes taken from a real stylesheet', () => {
+  it('sorts a stylesheet that is mainly longhand properties', () => {
     const actual = transform(outdent`
       ._oqicidpf{padding-top:0}
       ._1rmjidpf{padding-right:0}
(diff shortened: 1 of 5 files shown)


packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
```diff
@@ -386,13 +386,61 @@ describe('sort shorthand vs. longhand declarations', () => {
...
       }
-      .b { all: unset; }
       .c { all: unset; }
+      .b { all: unset; }
       .d { border: none; }
       .f { border-block-start-color: transparent; }"
     `);
   });
 
+  it('sorts a subset of atomic classes taken from a real stylesheet', () => {
```
dddlr, line 396: this is a small subset of the jira stylesheet that i found was not being sorted correctly

packages/css/src/plugins/sort-shorthand-declarations.ts
```diff
@@ -37,31 +37,19 @@ const sortNodes = (a: ChildNode, b: ChildNode): number => {
 
   if (!aDecl?.prop || !bDecl?.prop) return 0;
 
-  const aShorthand = shorthandFor[aDecl.prop as ShorthandProperties];
```
dddlr, line 40: remove the previous logic that checked whether the property was a shorthand property of the other (e.g. `padding` and `paddingTop`)

the likelihood that we manage to sort two nodes such that `a === padding && b in [paddingTop, paddingLeft, paddingBottom, paddingRight]` is pretty low, so it's much easier just to lean on the shorthand bucket logic later in the function

packages/css/src/plugins/sort-shorthand-declarations.ts
kylorhall-atlassian, line 1: Yeah, I see.  I think my original intent was to simply sort `padding` above `paddingLeft` for example, and ignore everything else…and then it grew a lot since then.

Now it's more like bucketing, but still sorted, which makes sense to me.  Will suggest a few scenarios to test is all.

packages/css/src/plugins/sort-shorthand-declarations.ts
```diff
@@ -37,31 +37,19 @@ const sortNodes = (a: ChildNode, b: ChildNode): number => {
...
-  const bShorthand = shorthandFor[bDecl.prop as ShorthandProperties];
-  if (bShorthand === true || bShorthand?.includes(aDecl.prop)) {
-    return 1;
-  }
-
-  const aShorthandBucket = shorthandBuckets[aDecl.prop as ShorthandProperties];
-  const bShorthandBucket = shorthandBuckets[bDecl.prop as ShorthandProperties];
+  // Why default to Infinity? Because if the property is not a shorthand property,
+  // we want it to come after all of the other shorthand properties.
+  const aShorthandBucket = shorthandBuckets[aDecl.prop as ShorthandProperties] ?? Infinity;
```
dddlr, line 42: how the bucket system works is that every shorthand property is assigned a number

e.g. `padding` is 1, `padding-block` and `padding-inline` is 2

`padding-{left, top, right, bottom}` is assigned a number and so defaults to `Infinity`

when sorting, we ensure that properties with lower numbers come before properties with higher numbers

this also has the side effect of making the behaviour almost identical to runtime mode

packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
kylorhall-atlassian: Got it, some other thoughts if they're _not_ tested in here should be `border > border-top > border-top-color` right?

Input:
```tsx
._abcd1234 { border-top-color: red }
._abcd1234 { border-top: 1px solid }
._abcd1234 { border: none }
._abcd1234:hover { border-top-color: red }
._abcd1234:hover { border-top: 1px solid }
._abcd1234:hover { border: none }
@media (…) { ._abcd1234 { border-top-color: red } }
@media (…) { ._abcd1234 { border-top: 1px solid } }
@media (…) { ._abcd1234 { border: none } }
```
Output:
```tsx
._abcd1234 { border: none }
._abcd1234 { border-top: 1px solid }
._abcd1234 { border-top-color: red }
._abcd1234:hover { border: none }
._abcd1234:hover { border-top: 1px solid }
._abcd1234:hover { border-top-color: red }
@media (…) { ._abcd1234 { border: none } }
@media (…) { ._abcd1234 { border-top: 1px solid } }
@media (…) { ._abcd1234 { border-top-color: red } }
```

The main question I'd have is do we still maintain this pseudo and media query sorting, the rest seems fine to me…

kylorhall-atlassian (commented):
Just one comment on an additional test I think, LGTM otherwise.

packages/css/src/plugins/sort-shorthand-declarations.ts
kylorhall-atlassian: Yeah, this changed because mine originally was like "A is a shorthand of B", now it's more like "A is shorthand 5", "B is shorthand 3" and this change LGTM.

This doesn't sort `padding-block-start` vs. `padding-left`, but that's still fine, those are "siblings" and that's what we want.  I can't think of any scenario where this isn't what we want…maybe like `border-block-color` and `border-top` are sorted as siblings now and there could be some odd conflicts…?  But I'd say if we have issues with that we have like `sort-logical-properties` or something—that's separate to this and should rarely come up (but likely does exist).

dddlr: > This doesn't sort padding-block-start vs. padding-left, but that's still fine, those are "siblings" and that's what we want…maybe like border-block-color and border-top are sorted as siblings now and there could be some odd conflicts…?

I can add it to `packages/utils/src/shorthand.ts` if you want :) just let me know

kylorhall-atlassian: No, let's not, I don't know what that should even be. It's fine as-is, and if we need to add this later we have a good path forward!

liamqma (dismissed):

packages/css/src/plugins/sort-shorthand-declarations.ts
```diff
@@ -12,14 +12,6 @@ const findDeclaration = (node: ChildNode): Declaration | Declaration[] | undefin
     if (node.nodes.length === 1 && nodeIsDeclaration(node.nodes[0])) {
       return node.nodes[0];
     }
-
-    const declarations = node.nodes.map(findDeclaration).filter(Boolean) as Declaration[];
-
-    if (declarations.length === 1) {
```
dddlr, line 18: I deleted this as it would cause
```
@media (max-width: 100px) { ._abcd1234 { border-top: 1px solid } }
```
to be sorted with
```
@media (max-width: 100px) { ._abcd1234 { border: 1px solid } }
```

which overrides the sorting from the `sort-at-rules` plugin and I don't think is intentional?

packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
```diff
@@ -386,13 +386,115 @@ describe('sort shorthand vs. longhand declarations', () => {
...
+      ._abcd1234 { border: none }
+      ._abcd1234:hover { border: none }
+      ._abcd1234:active { border: none }
+      ._abcd1234 { border-top: 1px solid }
+      ._abcd1234:hover { border-top: 1px solid }
+      ._abcd1234:active { border-top: 1px solid }
+      ._abcd1234 { border-top-color: red }
+      ._abcd1234:hover { border-top-color: red }
+      ._abcd1234:active { border-top-color: red }
+      @media (max-width: 150px) { ._abcd1234 { border: none } }
```
dddlr, line 482: note that the media query sorting is also being run here

packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
```diff
@@ -36,8 +36,8 @@ describe('sort shorthand vs. longhand declarations', () => {
 
     expect(actual).toMatchInlineSnapshot(`
       ".a {
-        outline-width: 1px;
         font: 16px;
+        outline-width: 1px;
```
dddlr, line 40: i don't think this matters, because styles are atomicised in Compiled anyway

packages/css/src/plugins/sort-shorthand-declarations.ts
kylorhall-atlassian: If I had to guess, I think I had issues understanding what sorting even looked at in the beginning, eg. there's a lot of different types of styles:
 - a whole stylesheet with multiple classes, media, etc in it
 - an `@media` query with multiple classes in it
 - a single `.class {property}`
 - A multiple `.class {property, webkitproperty}` — though I don't know if I even solved for that.

…so yeah, treat any of the code (if I wrote this) around here as possibly wrong!

packages/css/src/plugins/sort-atomic-style-sheet.ts
```diff
@@ -101,10 +109,6 @@ export const sortAtomicStyleSheet = (config: {
       }
 
       root.nodes = [...catchAll, ...rules, ...atRules.map((atRule) => atRule.node)];
-
-      if (sortShorthandEnabled) {
-        sortShorthandDeclarations(root.nodes);
-      }
```
kylorhall-atlassian, line 107: Got it, so it was just my bad this was ordered like this.

packages/css/src/sort.ts
```diff
@@ -8,14 +8,22 @@ import { sortAtomicStyleSheet } from './plugins/sort-atomic-style-sheet';
...
- * @returns
+ * @returns the sorted stylesheet
  */
 export function sort(
   stylesheet: string,
   {
     sortAtRulesEnabled,
     sortShorthandEnabled,
-  }: { sortAtRulesEnabled: boolean | undefined; sortShorthandEnabled: boolean | undefined }
+  }: { sortAtRulesEnabled: boolean | undefined; sortShorthandEnabled: boolean | undefined } = {
```
kylorhall-atlassian, line 18: Given this is now a required prop, I actually expect this to fail in TS anywhere `sort()` is called…but yeah I guess we have to find them directly in code if we find them:
```dif
-sort(styles);
+sort(styles, { sortAtRulesEnabled: true, sortShorthandEnabled: true });
```

Not the best configuration experience, but we'll survive for now.

packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
```diff
@@ -386,13 +388,115 @@ describe('sort shorthand vs. longhand declarations', () => {
...
+    `);
+
+    expect(actual).toMatchInlineSnapshot(`
+      "
+      ._abcd1234 { border: none }
+      ._abcd1234 { border-top: 1px solid }
+      ._abcd1234 { border-top-color: red }
+      ._abcd1234:hover { border: none }
+      ._abcd1234:hover { border-top: 1px solid }
+      ._abcd1234:hover { border-top-color: red }
```
kylorhall-atlassian, line 480: Oh is this fixed now? 😮

kylorhall-atlassian (approved):
Love your work, I think this is great and tests cover all my concerns really.

packages/css/src/sort.ts
dddlr: Screenshot 2024-10-23 at 16 54 31

typescript says it's optional 🤔

dddlr: but yeah we would need to turn on `sortShorthandEnabled` manually everywhere in order to roll it out

(how would we make sure we don't forget? 🤔)

packages/css/src/plugins/sort-atomic-style-sheet.ts
dddlr: it was my bad as well 😅 lots of subtle behaviour quirks and bugs i didn't spot even from working on it for hours

packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
dddlr: yeah i fixed it in https://github.com/atlassian-labs/compiled/pull/1730/commits/18adcbe2bd8fe30f3cb1deebf7db9ea6e260e4ee

dddlr: whoops not that commit

dddlr: fixed in this one https://github.com/atlassian-labs/compiled/pull/1730/commits/8b6edba726e3733b36a0c286bd74f0eb31539f50

PR #1732, opened 2024-10-23
Sort shorthand across multi-properties

kylorhall-atlassian:
### What is this change?

Making sure sorting works across mutli-property classes.

Example: this `{-webkit-text-decoration-color:initial;text-decoration-color:initial}` currently blocks all sorting above or below it, effectively creating a new "sorting bucket" because it always returns `0` as `a` or `b`.

This doesn't get sorted:
```css
._y44v1tgl{animation:kbayob8 5s ease infinite}
._syaz1qtk{color:var(--ds-link)}
._1jmq18uv{-webkit-text-decoration-color:initial;text-decoration-color:initial}
._zulph461{gap:var(--ds-space-050)}
```

Added a test to show how this works now!

### PR checklist

I have...

- [x] ~~Not updated the documentation in `website/`~~

### DIFF
--- a/packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
+++ b/packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
@@ -20,6 +20,104 @@ describe('sort shorthand vs. longhand declarations', () => {
     process.env.BROWSERSLIST = 'last 1 version';
   });
 
+  it('sorts against multi-property classes', () => {
+    const actual = transform(outdent`
+      ._zulph461{gap:var(--ds-space-050)}
+      ._y44v1tgl{animation:kbayob8 5s ease infinite}
+      ._j6xt1fef:hover{background:var(--_hcgrh3)}
+      ._bfhk29zg, ._irr329zg:hover{background-color:var(--ds-background-selected,#deebff)}
+      ._bfhk1ayf{background-color:var(--_rgmfhj)}
+      ._irr312tn:hover{background-color:var(--_1hwiu8a)}
+      ._syaz1qtk{color:var(--ds-link)}
+      ._1jmq18uv:hover .linkText{-webkit-text-decoration-color:initial;text-decoration-color:initial}
+      ._1owrotz2:active, ._jzfzotz2:focus, ._1mq6otz2:hover{padding:0 var(--ds-space-050,4px)}
+      ._apju8stv:hover .linkText{-webkit-text-decoration-line:underline;text-decoration-line:underline}
+      ._kkk21kw7{all:inherit}
+      @media (min-width:1024px){
+        ._j6xt1fef:active{background:var(--_hcgrh3)}
+        ._syaz1qtk{color:var(--ds-link)}
+        ._1jmq18uv:hover .linkText{-webkit-text-decoration-color:initial;text-decoration-color:initial}
+        ._1owrotz2:active, ._jzfzotz2:focus, ._1mq6otz2:hover{padding:0 var(--ds-space-050,4px)}
+        ._apju8stv:hover .linkText{-webkit-text-decoration-line:underline;text-decoration-line:underline}
+        ._kkk21kw7{all:inherit}
+        ._ku3unqa1:hover .linkText{-webkit-text-decoration-style:solid;text-decoration-style:solid}
+        ._18s8paju{margin:var(--ds-space-150,9pt) auto 28px auto}
+        ._y44v1syn{animation:kbpspdk 5s ease infinite}
+      }
+      ._ku3unqa1:hover .linkText{-webkit-text-decoration-style:solid;text-decoration-style:solid}
+      ._18s8paju{margin:var(--ds-space-150,9pt) auto 28px auto}
+      ._y44v1syn{animation:kbpspdk 5s ease infinite}
+      @keyframes kbpspdk{0%{stroke-dashoffset:10}}
+      ._y44vbxa2{animation:k1qo9wnt 5s ease infinite}
+      @media (min-width:30rem){
+        ._1letfkly{margin:0 var(--ds-space-400,2pc)}
+        ._1wbfxncg{padding:var(--ds-space-800,4pc)}
+        ._1tn2iyik{font:var(--ds-font-heading-xlarge,normal 600 29px/2pc ui-sans-serif,-apple-system,BlinkMacSystemFont,"Segoe UI",Ubuntu,system-ui,"Helvetica Neue",sans-serif)}
+        ._l7hko0gd [data-ds--text-field--input]{font:var(--ds-font-heading-medium,normal 500 20px/24px ui-sans-serif,-apple-system,BlinkMacSystemFont,"Segoe UI",Ubuntu,system-ui,"Helvetica Neue",sans-serif)}._1odf1h6o{justify-content:center}
+        ._j6xt1fef:active{background:var(--_hcgrh3)}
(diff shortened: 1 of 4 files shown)


yarn.lock
kylorhall-atlassian, line 1: I noticed this file gets regenerated quite often, I'm not really sure why—if this is developer error or our deps are so prone to changing…?  Making CI protect this in another PR.

packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
```diff
@@ -374,13 +493,8 @@ describe('sort shorthand vs. longhand declarations', () => {
...
     `);
 
-    // WARNING: This is technically wrong as `.a { … }` is not sorted as we expect;
-    // it _should_ be 'abcdef' not 'eabcdf'.
-    //
-    // We are ok with this because we expect atomicifyRules to run before this plugin,
-    // so each class will never have more than one property.
     expect(actual).toMatchInlineSnapshot(`
-      ".e { border-top: none; }
+      "
```
kylorhall-atlassian, line 497: 👀 Guess it wasn't what I thought and this was a bug actually!

packages/css/src/plugins/sort-shorthand-declarations.ts
```diff
@@ -9,15 +9,17 @@ const findDeclaration = (node: ChildNode): Declaration | undefined => {
   }
 
   if ('nodes' in node) {
-    if (node.nodes.length === 1 && nodeIsDeclaration(node.nodes[0])) {
-      return node.nodes[0];
-    }
+    // Return the first node that is a declaration, if we find one
```
kylorhall-atlassian, line 12: This may not be correct at all, eg. if Compiled ever tries to sort this, it's kind of uncharted territory…`global` will come second here 🥴
```tsx
.global {padding-left: 8px;all: unset}
._abcd1234{padding: 8px}
```

kylorhall-atlassian: I think we could technically do a deep recursive sort _somehow_, eg. compare the scores of all children vs. another child, but I'm expecting in real-world Compiled, the above is not a legitimate scenario and instead using the first property's "score" is fine.

packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
```diff
@@ -20,6 +20,125 @@ describe('sort shorthand vs. longhand declarations', () => {
...
+    `);
+
+    expect(actual).toMatchInlineSnapshot(`
+      "
+      ._kkk21kw7{all:inherit}._zulph461{gap:var(--ds-space-050)}
+      ._y44v1tgl{animation:kbayob8 5s ease infinite}
+      ._18s8paju{margin:var(--ds-space-150,9pt) auto 28px auto}
+      ._y44v1syn{animation:kbpspdk 5s ease infinite}
+      ._y44vbxa2{animation:k1qo9wnt 5s ease infinite}
+      ._bfhk29zg, ._irr329zg:hover{background-color:var(--ds-background-selected,#deebff)}
```
dddlr, line 100: I suspect sorting pseudos after non pseudos won't (and can't) work for this line? 

(Not sure how we would fix this short of splitting up the declaration...)

packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
```diff
@@ -20,6 +20,125 @@ describe('sort shorthand vs. longhand declarations', () => {
...
+        ._1tn2iyik{font:var(--ds-font-heading-xlarge,normal 600 29px/2pc ui-sans-serif,-apple-system,BlinkMacSystemFont,"Segoe UI",Ubuntu,system-ui,"Helvetica Neue",sans-serif)}
+        ._l7hko0gd [data-ds--text-field--input]{font:var(--ds-font-heading-medium,normal 500 20px/24px ui-sans-serif,-apple-system,BlinkMacSystemFont,"Segoe UI",Ubuntu,system-ui,"Helvetica Neue",sans-serif)}
+        ._18s8paju{margin:var(--ds-space-150,9pt) auto 28px auto}
+        ._y44v1syn{animation:kbpspdk 5s ease infinite}._1odf1h6o{justify-content:center}
+        ._syaz1qtk{color:var(--ds-link)}
+        ._1jmq18uv:hover .linkText{-webkit-text-decoration-color:initial;text-decoration-color:initial}
+        ._apju8stv:hover .linkText{-webkit-text-decoration-line:underline;text-decoration-line:underline}
+        ._ku3unqa1:hover .linkText{-webkit-text-decoration-style:solid;text-decoration-style:solid}
+        ._pwyo12x7{padding-inline-end:var(--ds-space-075,6px)}
+        ._1b7p1epz{padding-left:var(--ds-space-1000,5pc)}
```
dddlr, line 123: why does padding-left and padding-right come before background

packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
```diff
@@ -20,6 +20,125 @@ describe('sort shorthand vs. longhand declarations', () => {
     process.env.BROWSERSLIST = 'last 1 version';
   });
 
+  it('sorts against multi-property classes: simple', () => {
+    // We need to make sure this `{-webkit-text-decoration-color:initial;text-decoration-color:initial}` doesn't "block" sorting
+    const actual = transform(outdent`
+      ._zulph461{gap:3px}
+      ._bfhk1ayf{background-color:red}
+      ._1jmq18uv{-webkit-text-decoration-color:initial;text-decoration-color:initial}
```
dddlr, line 28: ah is this from autoprefixer? I guess the 1 property per class name assumption is incorrect after all

sorry for introducing the bug 😔

packages/css/src/plugins/sort-shorthand-declarations.ts
dddlr: I think if I learnt anything from this, it's that it's safer to ask devs not to mix shorthand and longhand properties at all, than to try to work with it 😅

packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
kylorhall-atlassian: Nah, 99% sure it was there before because I just `return 0`'d arrays (and this is an array of properties)

packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
kylorhall-atlassian: Yeah I'm actually confused what we'd do here…it's _unrelated_, but I guess you would have to split this into two… 😢

packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
kylorhall-atlassian: It's `…:active{bacgkround}` isn't it?

packages/css/src/plugins/__tests__/sort-shorthand-declarations.test.ts
dddlr: yeahhh i'm ok with tackling stuff like `._bfhk29zg, ._irr329zg:hover{background-color:var(--ds-background-selected,#deebff)}` being out of scope, but best to keep in the back of our minds in case we ever figure out _why_ this happens

dddlr (approved):

PR #1733, opened 2024-10-23
Make yarn installs immutable

kylorhall-atlassian:
### What is this change?

Context: https://github.com/atlassian-labs/compiled/pull/1732/files#r1813527167, I've noticed a few times I get `yarn.lock` diffs staged when they should probably be committed.

To make sure we don't leave out important changes.

---

### PR checklist

Don't delete me! 🤣 

- [ ] Testing…this should fail without `yarn.lock` changes as _master_ is currently broken.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,7 +35,7 @@ jobs:
             ${{ runner.os }}-yarn-
 
       - name: Install packages
-        run: yarn
+        run: yarn --immutable --immutable-cache --check-cache
 
       - name: Build source
         run: yarn build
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -35,7 +35,7 @@ jobs:
             ${{ runner.os }}-yarn-
 
       - name: Install packages
-        run: yarn
+        run: yarn --frozen-lockfile
```
kylorhall-atlassian, line 38: Only concern with having this in all the files is the devex could be a bit annoying, but CI runs very quickly here…so I expect it's not worth intentionally omitting it from other workflows.

kylorhall-atlassian:
Nevermind, I'm confused a bit—this doesn't even seem to work in full like I expected; in Yarn v2+ this works, and then `scripts/postinstall.sh` confuses me too…I think it's a good idea, but I forget how to execute it.  I'll ignore for now.

PR #1736, opened 2024-10-25
Add `shorthand-property-sorting` to recommended ESLint rules

dddlr:
### What is this change?

Add `shorthand-property-sorting` to recommended ESLint rules, meaning that it will be turned on by default for all codebases where `@compiled/eslint-plugin` set up correctly.

After https://github.com/atlassian-labs/compiled/pull/1700 was merged in, Compiled will now sort shorthand properties to come before longhand properties at buiild-time and runtime. @zesmith2 [made a lint rule](https://github.com/atlassian-labs/compiled/pull/1714) to enforce that the order of the properties in the source code of Compiled usages matches this ordering. We've resolved all the eslint rule violations in jira and confluence (I think?), so we can just add this as a default recommended eslint rule to remove the need to configure it manually.

### Some asks

Some requests from me before I go:

* If someone could extend `shorthand-property-sorting` so it handles the different forms of style composition, that would be great 🙏  @pancaspe87 has a link to the internal jira ticket for tracking this. @liamqma gave some examples in https://github.com/atlassian-labs/compiled/pull/1714#discussion_r1783816451

* If someone could quickly verify that the rule does indeed get applied in our products after the Compiled eslint plugin is version bumped, that would be great 🙏 

---

### PR checklist

Don't delete me!

I have...

### DIFF
--- a/.changeset/four-beds-own.md
+++ b/.changeset/four-beds-own.md
@@ -1,0 +1,5 @@
+---
+'@compiled/eslint-plugin': patch
+---
+
+Add `shorthand-property-sorting` to recommended ESLint rules
--- a/.changeset/four-beds-own.md
+++ b/.changeset/four-beds-own.md
@@ -1,5 +1,5 @@
 ---
-'@compiled/eslint-plugin': patch
+'@compiled/eslint-plugin': minor
 ---
 
 Add `shorthand-property-sorting` to recommended ESLint rules
--- a/packages/eslint-plugin/src/configs/flat-recommended.ts
+++ b/packages/eslint-plugin/src/configs/flat-recommended.ts
@@ -12,5 +12,6 @@ export const flatRecommended = {
     '@compiled/no-keyframes-tagged-template-expression': 'error',
     '@compiled/no-styled-tagged-template-expression': 'error',
     '@compiled/no-suppress-xcss': 'error',
+    '@compiled/shorthand-property-sorting': 'error',
   },
 } as const;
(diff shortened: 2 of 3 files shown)


zesmith2 (dismissed):

AllySummers:
not sure how you like to handle this in compiled eslint plugin since it's still in 0.x.x (where basically anything goes), but just a note from a (very) opinionated eslint perspective.

this kind of change, if it were following the same semver policy as eslint core (and what many big eslint plugins use) and it wasn't in 0.x.x, it would be considered a breaking change. considering that, it may be appropriate to do at least a minor bump instead of patch?

feel free to ignore if you like to handle this in a different way for this plugin :)

from here: https://github.com/eslint/eslint?tab=readme-ov-file#semantic-versioning-policy
> - Major release (likely to break your lint build)
>   - eslint:recommended is updated and may result in new linting errors (e.g., rule additions, most rule option updates).

AllySummers:
also i just realised, we should probably update the readme to reflect that this is now in the recommended config?

dddlr:
> also i just realised, we should probably update the readme to reflect that this is now in the recommended config?

ah you've already added it to the readme :p

dddlr:
Screenshot 2024-10-25 at 16 03 58

dddlr:
> not sure how you like to handle this in compiled eslint plugin since it's still in 0.x.x (where basically anything goes), but just a note from a (very) opinionated eslint perspective.

sure thing, will change to minor

AllySummers (approved):

PR #1738, opened 2024-10-30
Update the website to document the end-to-end lifecycle of Compiled

kylorhall-atlassian:
### What is this change?

I'm updating the documentation on https://compiledcssinjs.com from internal feedback around the Compiled work as people are questioning and need to understand the differences between runtime and extraction.

**Very open to feedback!**  It just seemed like the content on the homepage was quite important, but even I've never read it!  Moving it to `/docs/how-it-works` might help that.

I did remove a few scenarios around dynamic styles, eg. the `style={{ '--abcde': props.color }}` pattern, but I think it's documented elsewhere and it's best to keep it simple.

---

### PR checklist


website/packages/examples/src/css-prop-conditional-rules.tsx
kylorhall-atlassian, line 1: This file was used on https://compiledcssinjs.com/docs/api-css-prop, I feel it was fine to update it to be a bit more succinct and share it elsewhere.

website/packages/landing/src/pages/landing-content.mdx
```diff
@@ -1,105 +1,36 @@
 import { ExampleContainer, ExampleRoot } from '@compiled/website-ui';
 import { LargeText } from '../components/example';
 
-## Don't change a thing
```
kylorhall-atlassian, line 4: We do kinda want people to change a few things now 😄

website/packages/docs/src/pages/how-it-works.mdx
```diff
@@ -0,0 +1,163 @@
...
+section: 1-Getting started
+---
+
+# How it Works
+
+Once you've [installed and configured](/installation) everything, you can start using it,
+but it's best to explain what happens under the hood.
+
+Compiled uses Babel (with pre-configuration options via Webpack and Parcel) to transform styles
+into atomic CSS determined at build time, resulting in resulting in **very performant components**.
```
pancaspe87, line 12: Typo?

website/packages/docs/src/pages/how-it-works.mdx
```diff
@@ -0,0 +1,163 @@
...
+
+```bash
+$ babel src
+Successfully compiled 1 file with Babel (9ms)
+```
+
+You'll see a few things happening in the code below:
+
+1. All styles are converted into [Atomic CSS](/atomic-css) for reuse.
+2. Some styles, such as `padding:8px` will be expanded out to `padding-left`, `padding-right`, etc,
```
pancaspe87, line 55: IIRC there may be some additional optimitation and certain units could be converted to a more performant unit. Eg px to pc?

Not sure if we should mention this as well

liamqma (dismissed):

liamqma:
Thanks @kylorhall-atlassian!

website/packages/docs/src/pages/how-it-works.mdx
kylorhall-atlassian: I swear I do this a lot, I don't understand how…
```suggestion
into atomic CSS determined at build time, resulting in **very performant components**.
```

website/packages/docs/src/pages/how-it-works.mdx
kylorhall-atlassian: Yeah, maybe worth adding just generally:
> 5. Postcss, autoprefixer, and some other optimizations will run over your code to make it more performant and safer for usage.

pancaspe87 (approved):

alexreardon:
A bit late - but I approve! This is a wonderful change. I've messaged you some opportunities to further build on the great work you have done 🫡

alexreardon (commented):
Looks good!

PR #1741, opened 2024-11-05
new eslint rule to expand border shorthand property

zesmith2:
### What is this change?

Adding an eslint rule that expands `border` shorthand property into it's longhand parts `borderStyle`, `borderWidth`, and `borderColor`. 

`border` property supports `borderStyle`, `borderWidth` being in either order, and `borderColor` always coming last. This rule supports this allowance, and only applies to shorthand that has all 3 properties present, 

e.g. `border: '1px',` isn't covered in this rule.

Expanding the `border` property will decrease stylesheet size due to more common usages without all properties matching.

### PR checklist

Don't delete me!

I have...

### DIFF
--- a/packages/eslint-plugin/src/rules/expand-border-shorthand/__tests__/rule.test.ts
+++ b/packages/eslint-plugin/src/rules/expand-border-shorthand/__tests__/rule.test.ts
@@ -1,0 +1,295 @@
+import { outdent } from 'outdent';
+
+import { tester } from '../../../test-utils';
+import { expandBorderShorthand } from '../index';
+
+const packages_calls_and_imports = [
+  ['css', 'css', '@atlaskit/css'],
+  ['css', 'css', '@compiled/react'],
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/src/rules/expand-border-shorthand/__tests__/rule.test.ts
```diff
@@ -0,0 +1,295 @@
...
+            border: '   1px     solid   blue   ',
+          });
+        `,
+    })),
+  ],
+  invalid: [
+    /* borderStyle borderWidth borderColor */
+
+    ...packages_calls_and_imports.map(([pkg, call, imp]) => ({
+      name: `borderStyle:keyword, borderWidth:px,      borderColor:rgb     - single property             (${call}, ${imp})`,
```
zesmith2, line 86: spaces included in test name so it's easier to scan test names upon running

zesmith2:
Moving this to AFM

PR #1747, opened 2024-11-28
[SPIKE] Throw error if shorthand and longhand properties are mixed, or if invalid CSS properties are used

dddlr:
### What is this change?

Spike the following potential feature:

* Throw a build-time error if shorthand properties and longhand properties are mixed in the same component.

As a bonus, I've also spiked the following feature, because it involves making mostly the same changes:

* Throw a build-time error if an invalid CSS property is used in a component.

While shorthand properties and longhand properties (e.g. `margin` and `marginTop`) are sorted deterministically in almost all cases due to https://github.com/atlassian-labs/compiled/pull/1700 , there are still some edge cases where this sorting cannot be preserved, causing shorthand properties to override longhand properties non-deterministically and vice versa.

For example, if an app imports two packages that use Compiled, but the app does not use Compiled's Webpack or Parcel plugin, we cannot guarantee that the styles of the two packages will remain sorted (because we have no control on how the bundler sorts the stylesheets).

There are several solutions to this:

* Expand all shorthand properties to longhand properties, then use the runtime function `ax` to ensure that duplicate properties (e.g. `margin-top: 5px` and `margin-top: 10px`) are removed
  * This is impractical and a _lot_ of effort, as the values developers use for CSS properties in Atlassian codebases can be dynamic or very complex to statically analyse
  * Some CSS shorthand properties have a very flexible syntax, and thus are an absolute pain to expand
* Stop the root problem - ban mixing shorthand and longhand properties
  * Mixing is uncommon and is often done non-intentionally.
  * We take this approach here.

Build time error. We use a build time error, because build time and runtime are the only times where we can see _all_ of the Compiled styles that will be applied to a given component. Doing this check at runtime is feasible (albeit a bit tricky due to the use of hashed class names) by extending the `ax` function, but ideally we'd like to not introduce any runtime impact if possible.

### Limitations

This DOES NOT work when using `styled` composition (which is _deprecated_ anyway):

```
const BaseComponent = styled.div({ marginTop: '5px' });
const Component = styled(BaseComponent)({ margin: '2px' });
```

`BaseComponent` and `Component` are processed completely separately at build time.

We could make some changes to `@compiled/babel-plugin` to try to make Compiled "aware" of `BaseComponent` when processing `Component`, but it would probably be a bit hacky. Something to explore perhaps.

We could potentially extend `ax` to do a runtime check, but this would be even more hacky, as all that's passed to `ax` are the hashed class names. With this approach, we'd need to somehow go from the hashed class names back to the original CSS property names.

---

### PR checklist

Don't delete me!

I have...

- [ ] Changeset

Todos:

- [x] Verify this works with basic `styled` usages
- [x] Verify this works with `css` usages (including `css` composition)
- [ ] Verify this works with complex `styled` usages (props, dynamic styles...)
- [ ] Verify this works with `cssMap`
- [ ] Verify this works with `ClassName`
- [ ] Verify this works with `xcss`
- [ ] See whether it's possible to run this check when using `styled` composition
- [ ] Resolve TODOs in the source code of the PR
- [ ] **Add ability to configure exceptions for certain CSS properties**, e.g. `font` (shorthand and longhand mixing is required for the Atlassian Design System `font` tokens)


dddlr:
closing stale PRs, feel free to copy this over to a new one if needed!

PR #1748, opened 2024-11-29
Export KeyframeSteps from main entrypoint

danieldelcore:
Makes it so that `KeyframeSteps` is available via the main entrypoint so the keyframe object passed via indirection/props can be correctly typed

### What is this change?

Adds type export for `KeyframeSteps`

In some cases keyframe step objects are passed via props, so without this it's unable to be correctly typed.

---

### PR checklist

Don't delete me!

I have...


danieldelcore:
I realised passing keyframe config as props is not how compiled works anyway, i was attempting to do something that was unsupported anyway.

PR #1749, opened 2024-12-02
Extend shorthand-property-sorting ESLint rule to handle style composition

dddlr:
### What is this change?

Follow up to https://github.com/atlassian-labs/compiled/pull/1714, more specifically https://github.com/atlassian-labs/compiled/pull/1714#discussion_r1783816451

We have an ESLint rule `shorthand-property-sorting` that throws an error if longhand properties come after their corresponding shorthand properties, e.g. errors if `paddingTop` comes after `padding`. This is because Compiled always sorts `padding` to come before `paddingTop`, so we want the source code of the styles to match the sorting Compiled uses for these styles.

Extend the `shorthand-property-sorting` ESLint rule to support style composition

```tsx
const foo = css({ margin: ... });
const bar = css({ marginTop: ... });
<div css={[bar, foo]} />;
```

and style inheritance

```tsx
const Foo = styled.div({ marginTop: ... });
const Bar = styled(Foo)({ margin: ... });
```

This unfortunately involved rewriting the entire rule (sorry 🙇 )

I also fixed a bug where `shorthand-property-sorting` doesn't sort shorthand properties like `padding`, `outline`, and `background`.

Note that due to ESLint limitations, the rule will only detect styles that are defined in the same file. This means that _there is no way_ for any violations in the following example to be detected in ESLint... (adding a build time error would be tricky, as Foo and Bar are processed completely separately -- maybe a runtime error would be the only option?)

```tsx
import { Foo } from './util';
const Bar = styled(Foo)({ margin: ... });
```

* Style inheritance with `styled` is a deprecated but nonetheless common pattern.
* Additionally, in the future, we want developers to use `css` style composition.
* Extending this ESLint rule to support both situations will let us detect more situations where longhand properties are defined after shorthand properties.

---

### PR checklist

Don't delete me!

I have...

**TODOs** specific to this PR:

- [x] Finish tests
- [x] Add changeset
- [x] Cover the `all` property
- [x] Run this rule on internal Atlassian codebases
    - 407 errors in Jira
- [x] Check whether this rule is too slow
    - It impacts the total ESLint running time by around 1%, which is fine
- [x] Decide on whether to keep `invalidProperties`
    - Kept.


packages/utils/src/shorthand.ts
```diff
@@ -251,9 +251,6 @@ export const shorthandFor: Record<ShorthandProperties, true | readonly string[]>
     'border-right-width',
   ],
   'border-left': [
```
dddlr, line 253: Ran into an issue where having `borderTop` and `borderBottom` in the same component caused an ESLint violation

This was caused by `borderTop` and `borderBottom` both being listed as being shorthands for `border-block-end-*`... I don't think we have anyone mixing `borderInline*` and `borderLeft`/`borderRight`, or anyone mixing `borderBlock*` and `borderTop`/`borderBottom`, so I think these extra entries are safe to remove

packages/eslint-plugin/src/rules/shorthand-property-sorting/__tests__/rule.test.ts
```diff
@@ -3,238 +3,757 @@ import { outdent } from 'outdent';
...
+      errors: [
+        { messageId: 'shorthand-first' },
+        { messageId: 'shorthand-first' },
+        { messageId: 'shorthand-first' },
+        { messageId: 'shorthand-first' },
+      ],
+    },
+
+    //
+    // false positives for cssMap
```
dddlr, line 702: do we care?

kylorhall-atlassian: This might just end up adding noise because while you _could_ statically resolve this if there's any valid scenario of them being mixed, it's a ton of work…

Would we basically just have to ignore every `cssMap` with this?

tl;dr: on if we care or not, `cssMap` might overtake `css` one day as we tell people to use `@atlaskit/css` for Primitives.

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
kylorhall-atlassian, line 1: I'm not going to try to begin to understand everything here (but I will still review), I _think_ your tests speak for themselves…and if not, our products will speak for themselves.

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -4,123 +4,419 @@ import {
...
+      //         paddingStyles,
+      //         { margin: '...' },
+      //     )
+
+      // For arguments of `styled` function calls, we can re-use some of the same
+      // logic that we use to parse elements in the css prop array:
+      //
+      //            vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
+      // <div css={[baseStyles, someCondition && someMoreStyles]}>Hello</div>
+      properties.push(...parseCssArrayElement(context, argument));
```
kylorhall-atlassian, line 294: I may not exactly be able to follow everything from a PR review, the comments make it feasible that I could follow it running the tests locally and digging into what I need, thank you for them!

packages/utils/src/shorthand.ts
kylorhall-atlassian: That's fair.  I'd confidently say we have no intentional usages of this `border-left` vs. `border-inline-end`.  Maybe if we explored perfect RTL support, this might need more thought, but this works for me!

kylorhall-atlassian (dismissed):
I think I'm happy?  Really just the one comment around `cssMap` is all I'd be concerned with…how many instances are there of that?

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
dddlr: ah i'm happy to add more comments to other parts of the file to help you with review - let me know if any part doesn't immediately make sense

packages/eslint-plugin/src/rules/shorthand-property-sorting/__tests__/rule.test.ts
dddlr: fixed the false positives for when the cssMap key (or whatever we want to call it) can be easily determined at build time (i.e. it's a string)

packages/eslint-plugin/src/rules/shorthand-property-sorting/__tests__/rule.test.ts
```diff
@@ -682,28 +743,23 @@ tester.run('shorthand-property-sorting', shorthandFirst, {
         import { Box } from '@atlaskit/primitives';
 
         const styles = cssMap({
-          root: { paddingTop: '...' },
           success: {
             padding: '...',
           },
+          root: { paddingTop: '...' },
```
dddlr, line 749: moved `paddingTop` to the bottom, so that we can verify that the ESLint violations are caused by the `cx(styles.root, styles.success)` function call, which applies `paddingTop` before `padding`

kylorhall-atlassian (approved):
Yeah seems fair as long as our repos look good, no concerns from me!

PR #1756, opened 2024-12-11
Fix border-inline-start and border-inline-end not having any valid ordering

dddlr:
### What is this change?

Fix how `border-inline-start` and `border-inline-end` does not have any valid ordering in the `shorthand-property-sorting` ESLint rule.

Follow up to https://github.com/atlassian-labs/compiled/pull/1749

Fix one more false positive that slipped through the cracks.

---

### PR checklist

Don't delete me!

I have...


package.json
```diff
@@ -158,5 +158,6 @@
         "react"
       ]
     }
-  ]
+  ],
+  "packageManager": "yarn@1.22.21+sha256.dbed5b7e10c552ba0e1a545c948d5473bc6c5a28ce22a8fd27e493e3e5eb6370"
```
kylorhall-atlassian, line 162: What's this about?

dddlr: gahhh yarn keeps adding that

dddlr: we're supposed to specify what version of yarn the repo should use through `packageManager`

i never got the chance to set this up properly, so it keeps trying to add it whenever i run `yarn install`

maybe i should just set it up 🤔 hold on

dddlr: hmm looks a bit complicated, something for another day...

kylorhall-atlassian (approved):
Seems fair—worst-case it's an edge-case for us we might want to handle better _one day_, but for now it's perfectly acceptable.

package.json
kylorhall-atlassian: Yeah no clue 🫡

PR #1760, opened 2024-12-13
feat(eslint-plugin): add support for eslint v9 

michaelfaith:
This change adds a set of compatibility functions to bridge the gap between the legacy context api and the new context api, allowing projects using this plugin to upgrade their version of eslint to v9.


package.json
```diff
@@ -84,8 +84,9 @@
     "@types/react": "^17.0.69",
     "@types/react-dom": "^17.0.22",
     "@types/svgo": "^2.6.4",
-    "@typescript-eslint/eslint-plugin": "^5.59.8",
-    "@typescript-eslint/parser": "^5.59.8",
+    "@typescript-eslint/eslint-plugin": "^6.21.0",
```
michaelfaith, line 87: These updates were necessary in order to use v8 compatible types.  This doesn't mean that only v8 and up are support, just that internally, we need to have tall of the type information for v8 context objects to fully support v8 (and 9) in our TS-based rules.

packages/eslint-plugin/src/rules/no-css-prop-without-css-function/index.ts
```diff
@@ -220,7 +221,7 @@ export const noCssPropWithoutCssFunctionRule: TSESLint.RuleModule<string> = {
   meta: {
     docs: {
       url: 'https://github.com/atlassian-labs/compiled/tree/master/packages/eslint-plugin/src/rules/no-css-prop-without-css-function',
-      recommended: 'error',
+      recommended: 'strict',
```
michaelfaith, line 224: TS v6 uses 'strict' instead of 'error'.  This goes back to 'error' in later versions of TS-eslint

packages/eslint-plugin/tsconfig.json
```diff
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "rootDir": "src",
     "outDir": "dist",
-    "module": "commonjs",
+    "module": "nodenext",
+    "moduleResolution": "nodenext",
```
michaelfaith, line 7: TS ESLint started using export conditions in v6 and stopped supporting module resolutions less than node16 (https://github.com/typescript-eslint/typescript-eslint/issues/7185#issuecomment-1631374643)

.changeset/five-pigs-explain.md
AllySummers, line 1: Given that node 14 is jo l

michaelfaith: I'm good with that.  But if we do that, I'd recommend that we add an engines declaration at the same time, to formalize that aspect of the public api, as a new commit (in this PR) with a separate changeset entry and have that be the breaking change.  The current PR doesn't explicitly remove support for anything; we'd be majoring because _maybe_ it'll break in node 14 ¯\_(ツ)_/¯, which isn't super confidence-inspiring, you know what I mean?  If we add engines, then we can state that we're officially dropping support for node versions less than 16.  What do you think?

AllySummers: actually, i realised this is in 0.x.x so anything goes technically.

i support adding an `engines` entry regardless, but unless the rest of the team wants this to go into 1.x.x, then just do minor.

michaelfaith: Added engines and an additional changeset entry for it (both minor).

JakeLane (approved):

michaelfaith:
rebased to resolve lock file conflict

AllySummers (dismissed):

.github/workflows/test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x, 16.x, 18.x]
+        node-version: [16.x, 18.x, 20.x, 22.x]
```
dddlr, line 16: i assume this is a safe change as internal codebases are on v20+ already

.changeset/five-dolphins-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+'@compiled/eslint-plugin': minor
+---
+
+add engines to formalize supported node versions - ^16.0.0 || >= 18.0.0
```
dddlr, line 5: nit: the two changesets can be combined into one

.eslintrc.js
```diff
@@ -48,6 +48,8 @@ module.exports = {
           },
         ],
         'import/no-unresolved': 'off',
+        // TypeScript takes care of this for us (https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/namespace.md)
```
dddlr, line 51: is there a need to turn this off?

packages/eslint-plugin/src/rules/no-css-prop-without-css-function/index.ts
dddlr: are there docs i can read about this?

https://typescript-eslint.io/developers/custom-rules/#extra-rule-docs-types implies that `recommended` only takes a boolean value

michaelfaith: @dddlr it changed to a boolean in the latest version (v8).  Here's the shape of that type in v7: https://github.com/typescript-eslint/typescript-eslint/blob/v7.18.0/packages/utils/src/ts-eslint/Rule.ts#L28

The options for `recommended` in that version are `'recommended' | 'strict' | 'stylistic'` (https://github.com/typescript-eslint/typescript-eslint/blob/v7.18.0/packages/utils/src/ts-eslint/Rule.ts#L9)

.github/workflows/test.yml
michaelfaith: Yeah, and with the engines added, it's now formal.

.eslintrc.js
michaelfaith: I could turn it back on, but the import plugin (and specifcally the default node resolver) notoriously isn't able to resolve modern package structures using `export` conditions.  v6 of TS-eslint is the first version where they shift exclusively to using package export conditions for their api.  So that rule is going to have a hard time determining what's valid and what isn't, which is something TS is already doing for us.  I was having issues with it misfiring, and since it's redundant, I thought it was best to just turn it off.  What do you think?

packages/eslint-plugin/src/rules/no-css-prop-without-css-function/index.ts
michaelfaith: Removed

.changeset/five-dolphins-taste.md
michaelfaith: Oh, ok.  I was trying to keep the commits tidy (1st change -> 1st changeset -> 2nd change -> 2nd changeset).  Happy to re-organize it however you prefer.

packages/eslint-plugin/tsconfig.json
dddlr: the `module` option is a bit confusing for me to understand, but from the typescript docs, `nodenext` looks good to me

https://www.typescriptlang.org/docs/handbook/modules/reference.html#node16-nodenext

does this make a difference to the `@compiled/eslint-plugin` build output? (should we mention it in the changeset?)

and how tricky would it be to update `module` / `moduleResolution` for other packages in Compiled?

.eslintrc.js
dddlr: ahh all good, that's a good reason to disable it yeah

.changeset/five-dolphins-taste.md
dddlr: nah no need to re-organise unless you need to make other changes

probs https://github.com/atlassian-labs/compiled/pull/1760#discussion_r1889504901 is the only thing i'm not 100% confident with, otherwise PR looks good to me :)

packages/eslint-plugin/tsconfig.json
michaelfaith: That's a great question.  It certainly had the potential to, depending on how the build is set up (for instance, I generally use rollup for building packages, which can have a separate set of tsconfig options.  But yeah, I did a before and after comparison of the build output, and the only changes between them are the code changes from this PR (the black text means no difference, red means a difference, and blue is new).  Now, if `target` changed, then that would likely result in more differences, since that determines the specific ECMAScript version that the emitted code should be compatible with.

image

As far as what the effort would be to update those props in the other packages: honestly shouldn't be much.  Of those two, `moduleResolution` is the one that will have the most impact on your internal DX, but when you change that, `module` has to come with it.

dddlr: ah cool, yeah that looks fine

no need to do in this PR, but I think it would be good to make the `module` / `moduleResolution` change for the other Compiled packages if practical, just for consistency

dddlr (approved):

PR #1765, opened 2024-12-16
Fix shorthand-property-sorting crashing when styles not initialised

dddlr:
### What is this change?

Fix the rule crashing when an uninitialised variable is passed to the `css` prop.

ESLint's types give an `any` type for `functionCall.defs[0].node` so this bug was not detected by type-checking 😭 

Prevent rule from crashing when run on one of our internal codebases.

---

### PR checklist

Don't delete me!

I have...

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -20,3 +20,4 @@ I have...
 
 - [ ] Updated or added applicable tests
 - [ ] Updated the documentation in `website/`
+- [ ] Added a changeset (if making any changes that affect Compiled's behaviour)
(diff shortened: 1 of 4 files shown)


.github/pull_request_template.md
```diff
@@ -20,3 +20,4 @@ I have...
 
 - [ ] Updated or added applicable tests
 - [ ] Updated the documentation in `website/`
+- [ ] Added a changeset (if making any changes that affect Compiled's behaviour)
```
dddlr, line 23: i keep forgetting to add a changeset before making the PR

kylorhall-atlassian (approved):
Easy!

Issue #1642, opened 2024-03-06
Types on `css={…}` don't block using the top-level object from `cssMap()` 


kylorhall-atlassian:
**Describe the bug**

There's no type protection when consuming `cssMap` incorrectly, meaning TypeScript will allow code that won't compile properly.

**To Reproduce**

For raw `cssMap`, jump into https://github.com/atlassian-labs/compiled/blob/master/packages/react/src/css-map/__tests__/index.test.tsx#L29 and remove `.danger`
```diff
-const Foo = () => <div css={styles.danger}>hello world</div>;
+const Foo = () => <div css={styles}>hello world</div>;
```

Same, for the strict version:
https://github.com/atlassian-labs/compiled/blob/f093635244ca5accecd5093013694084408be2f7/packages/react/src/create-strict-api/__tests__/index.test.tsx#L162

**Expected behavior**

These should both have a type error.

**Screenshots**

No error 😞 
<img width="640" alt="Screenshot 2024-03-07 at 11 53 13 AM" src="https://github.com/atlassian-labs/compiled/assets/136543114/7e708a9d-b55a-45a0-bf48-78b6a6d6e874">

itsdouges:
Thanks for raising this!

PR #1773, opened 2024-12-23
Fix `@compiled/babel-plugin` to not silently skip `cssMap()` extraction in a few common edge-cases.

kylorhall-atlassian:
### What is this change?

Fixes https://github.com/atlassian-labs/compiled/issues/1642

Resolves two issues revolving around `cssMap` not extracting which could lead to production issues as Runtime and Extraction are handled differently with little to no testability in this area.
1. When `cssMap` is called after it's used, eg. `<div css={styles.root}>; const styles  = cssMap({ root: … });`, styles are never extracted
2. When the root `cssMap` object is passed to css, eg. `<div css={styles}>; const styles  = cssMap(…);`, styles are never extracted and this results in no typescript error — built via https://github.com/atlassian-labs/compiled/pull/1774 and merged into this branch already.

Both of these seemingly acceptable code examples passed full CI, TS, Linting, Babel, etc., but styles were simply never displayed to the user.

```tsx
import { cssMap } from "@compiled/react";
export default () => <div css={styles.root} />;
const styles = cssMap({ root: { padding: 8 } });
```

```tsx
import { cssMap } from "@compiled/react";
const styles = cssMap({ root: { padding: 8 } });
export default () => <div css={styles} />;
```

These should both worth now and be tested, with the downside that this _could_ result in breaking changes by "fixing" these currently erroring paths…

---

### PR checklist

I have...

- [x] ~Updated the documentation in `website/`~ n/a

### DIFF
--- a/packages/babel-plugin/src/babel-plugin.ts
+++ b/packages/babel-plugin/src/babel-plugin.ts
@@ -40,63 +40,6 @@ const JSX_SOURCE_ANNOTATION_REGEX = /\*?\s*@jsxImportSource\s+([^\s]+)/;
 
 let globalCache: Cache | undefined;
 
-const buildCompiledImportsCache: Visitor<State> = {
-  ImportDeclaration(path, state) {
-    const userLandModule = path.node.source.value;
-    const isCompiledModule = this.importSources.some((compiledModuleOrigin) => {
-      if (compiledModuleOrigin === userLandModule) {
-        return true;
-      }
-
-      if (
-        state.filename &&
-        userLandModule[0] === '.' &&
-        userLandModule.endsWith(basename(compiledModuleOrigin))
-      ) {
-        // Relative import that might be a match, resolve the relative path and compare.
-        const fullpath = resolve(dirname(state.filename), userLandModule);
-        return fullpath === compiledModuleOrigin;
-      }
-
-      return false;
-    });
-
-    if (!isCompiledModule) {
-      return;
-    }
-
-    // The presence of the module enables CSS prop
-    state.compiledImports = state.compiledImports || {};
-
-    // Go through each import and enable each found API
-    path.get('specifiers').forEach((specifier) => {
-      if (!state.compiledImports || !specifier.isImportSpecifier()) {
-        // Bail out early
-        return;
-      }
-
-      (['styled', 'ClassNames', 'css', 'keyframes', 'cssMap'] as const).forEach((apiName) => {
-        if (
-          state.compiledImports &&
-          t.isIdentifier(specifier.node?.imported) &&
-          specifier.node?.imported.name === apiName
-        ) {
-          // Enable the API with the local name
-          const apiArray = state.compiledImports[apiName] || [];
-          apiArray.push(specifier.node.local.name);
-          state.compiledImports[apiName] = apiArray;
-
-          specifier.remove();
-        }
-      });
-    });
-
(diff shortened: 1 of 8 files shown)


packages/babel-plugin/src/babel-plugin.ts
```diff
@@ -178,6 +244,11 @@ export default declare<State>((api) => {
               );
             }
           }
+
+          // Build our `state.compiledImports[…]` state now that we've traversed JSX pragmas
+          path.traverse<State>(buildCompiledImportsCache, this);
+          // Build our `state.cssMap[…]` before we traverse the JSXElements which consume it
+          path.traverse<State>(buildCssMapState, this);
```
kylorhall-atlassian, line 251: I don't know if this is the best way to guarantee the order of things—just stuffing them into `Program.enter`, but it appears to work.

This could definitely be improved and traversion could be reduced by 1-5x if desired…

packages/babel-plugin/src/css-map/__tests__/index.test.ts
```diff
@@ -36,6 +36,25 @@ describe('css map basic functionality', () => {
     ]);
   });
 
+  it('should transform css map even when the styles are defined below the component', () => {
```
kylorhall-atlassian, line 39: Before, this test would actually throw a recursion error from finding member expressions—which maybe my recent change introduced the recursion error actually, so it helped me debug this.

kylorhall-atlassian:
Hitting an edge-case in the built tests, may circle back.

packages/babel-plugin/src/babel-plugin.ts
kylorhall-atlassian: This is wrong…in tests and other places, these aren't guaranteed to be at the Program scope, so this breaks our own tests.

packages/babel-plugin/src/utils/css-builders.ts
```diff
@@ -688,11 +691,47 @@ function extractMemberExpression(
...
+
+  if (bindingIdentifier) {
+    /**
+     * If we don't yet have a `meta.state.cssMap[…]` built yet, build it and cache it, eg. in this scenario:
+     * ```tsx
+     * const Component = () => <div css={styles.root} />
+     * const styles = cssMap({ root: { padding: 0 } });
+     * ```
+     *
+     * If we don't hit this, we put it into `ignoreMemberExpressions` to ignore on future runs.
```
kylorhall-atlassian, line 703: Not the most performant thing to do, but I couldn't find a better way that works with `cssMap` defined in different scopes, eg. not hoisted within Jest files as `packages/react/src/create-strict-api/__tests__/index.test.tsx` failed if I injected this into `Program` => `path.traverse()` instead.

packages/react/src/create-strict-api/__tests__/generics.test.tsx
kylorhall-atlassian, line 1: Ideally, this wouldn't be required, but the way these tests are built seems like a false-positive where we allow `it()` level scope for `css` and `cssMap`, yet in the real-world, you would never have `cssMap()` that wasn't at the Program scope as it throws errors.

What's happening is `state.cssMap['styles']` is being called and sometimes this is a `css` and other time it's a `cssMap`…this forces them to be unique at least.

liamqma (approved):

PR #1774, opened 2024-12-23
Throw an error when compiling a `cssMap` object where we expect a `css` or nested `cssMap` object.

kylorhall-atlassian:
### What is this change?

Fixes https://github.com/atlassian-labs/compiled/issues/1642

Example of code that silently fails today, using `styles` directly:
```tsx
const styles = cssMap({ root: { padding: 8 } });
export default () => <div css={styles} />;
```

What we expect to see instead, using `styles.root` instead:
```tsx
const styles = cssMap({ root: { padding: 8 } });
export default () => <div css={styles.root} />;
```

---

### PR checklist

- [x] ~Updated the documentation in `website/`~ n/a

### DIFF
--- a/.changeset/silent-geese-wonder.md
+++ b/.changeset/silent-geese-wonder.md
@@ -1,0 +1,21 @@
+---
+'@compiled/babel-plugin': minor
+---
+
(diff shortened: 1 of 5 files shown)


kylorhall-atlassian:
Hitting an edge-case in the built tests, may circle back.

PR #1781, opened 2025-01-08
Fix some false positives with ternary operators and cssMap

dddlr:
### What is this change?

Fix two edge cases:

* When using the _same properties_ across two `css` function calls through ternary operators, we should not get an ESLint violation.

```tsx

const oldStyles = css({
  font: '...',
  fontWeight: '...',
});
const newStyles = css({
  font: '...',
  fontWeight: '...',
});
const someCondition = true;
export const EmphasisText = ({ children }) => <span css={[someCondition ? oldStyles : newStyles]}>{children}</span>;
```

* When using the _same properties_ across two keys in a `cssMap` function call, we should not get an ESLint violation.

```tsx

const styles = cssMap({
  warning: {
    borderColor: '...',
    borderTop: '...',
  },
  debug: {
    borderColor: '...',
    borderTop: '...',
  },
  normal: {
    borderColor: '...',
    borderTop: '...',
  },
});
export const EmphasisText = ({ children, appearance }) => <span css={styles[appearance]}>{children}</span>;
```

Note that there still remain a lot of other (pre-existing) edge cases -- this rule is really hard to make fully rigorous, sorry 🙇  For example, this will still cause a false positive:

```tsx

const oldStyles = css({
  fontWeight: '...',
});
const newStyles = css({
  font: '...',
});

const someCondition = true;
export const EmphasisText = ({ children }) => <span css={[someCondition ? oldStyles : newStyles]}>{children}</span>;
```

We've received an internal query about a false positive with `css`, and it looks like it's pretty easy to fix.

---

### PR checklist

Don't delete me!

I have...

### DIFF
--- a/packages/eslint-plugin/src/rules/shorthand-property-sorting/__tests__/rule.test.ts
+++ b/packages/eslint-plugin/src/rules/shorthand-property-sorting/__tests__/rule.test.ts
@@ -271,6 +271,88 @@ tester.run('shorthand-property-sorting', shorthandFirst, {
     },
 
     //
+    // styles from ternary operators should not cause false positives, if the same properties
+    // are defined in each style object
+    //
+
+    {
+      name: `styles from ternary operators should not cause false positive, if the same properties are defined (css, ${imp})`,
+      code: `
+        import { css } from '${imp}';
+        const oldStyles = css({
+          font: '...',
+          fontWeight: '...',
+        });
+        const newStyles = css({
+          font: '...',
+          fontWeight: '...',
+        });
+        const someCondition = true;
+        export const EmphasisText = ({ children }) => <span css={[someCondition ? oldStyles : newStyles]}>{children}</span>;
+      `,
+    },
+    {
+      name: `pseudo-classes from ternary operators should not cause false positive, if the same properties are defined (css, ${imp})`,
+      code: `
+        import { css } from '${imp}';
+        const oldStyles = css({
+          '&:hover': {
+            font: '...',
+            fontWeight: '...',
+          }
+        });
+        const newStyles = css({
+          '&:hover': {
+            font: '...',
+            fontWeight: '...',
+          },
+        });
+        const someCondition = true;
+        export const EmphasisText = ({ children }) => <span css={[someCondition ? oldStyles : newStyles]}>{children}</span>;
+      `,
+    },
+    {
+      name: `styles from ternary operators should not cause false positive, if the same properties are defined (cssMap, ${imp})`,
+      code: `
+        import { css } from '${imp}';
+        const myMap = cssMap({
+          warning: {
+            font: '...',
+            fontWeight: '...',
+          },
+          normal: {
+            font: '...',
+            fontWeight: '...',
+          }
+        });
+        const someCondition = true;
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/src/rules/shorthand-property-sorting/__tests__/rule.test.ts
```diff
@@ -796,39 +926,42 @@ tester.run('shorthand-property-sorting', shorthandFirst, {
...
     {
-      // this is a false positive, but making an exception for this would involve
-      // some extra logic... maybe we can revisit this if it becomes a common situation.
-      name: `false positive, if depth in correct order for shorthand properties in the same bucket for cssMap AND if key not static (${imp})`,
+      name: `false positive: shorthands in different selectors for cssMap if key not static (${imp})`,
       code: outdent`
         import { cssMap } from '${imp}';
         const styles = cssMap({
           warning: {
-            borderColor: '...', // 2
```
dddlr, line 826: this test case has been moved to the `valid` section, as I've fixed the false positive. It's now called `depth in correct order for shorthand properties in the same bucket for cssMap AND if key not static` on line 423

packages/eslint-plugin/src/rules/shorthand-property-sorting/__tests__/rule.test.ts
```diff
@@ -796,39 +926,42 @@ tester.run('shorthand-property-sorting', shorthandFirst, {
...
+
+    //
+    // known false positives for cssMap
+    //
+
     {
-      // this is a false positive, but making an exception for this would involve
-      // some extra logic... maybe we can revisit this if it becomes a common situation.
-      name: `false positive, if depth in correct order for shorthand properties in the same bucket for cssMap AND if key not static (${imp})`,
+      name: `false positive: shorthands in different selectors for cssMap if key not static (${imp})`,
```
dddlr, line 956: new test case, to cover a pre-existing limitation that my `union` function won't prevent false positives when we have a shorthand property in one cssMap key, and a longhand property in another cssMap key

packages/eslint-plugin/src/rules/shorthand-property-sorting/index.ts
```diff
@@ -303,26 +344,24 @@ const parseCssMap = (
       // If we know what key in the cssMap function call to traverse,
       // we can make sure we only traverse that.
       if (property.key.type === 'Literal' && key === property.key.value) {
-        properties.push(...getObjectCSSProperties(context, property.value));
-        break;
+        return getObjectCSSProperties(context, property.value);
```
dddlr, line 347: slight optimisation

packages/eslint-plugin/src/rules/shorthand-property-sorting/__tests__/rule.test.ts
```diff
@@ -310,6 +392,54 @@ tester.run('shorthand-property-sorting', shorthandFirst, {
         export const EmphasisText = ({ children }) => <span css={styles}>{children}</span>;
       `,
     },
+
```
dddlr, line 395: this false negative already existed in the rule, it's just that we didn't have a test case for it

kylorhall-atlassian (approved):
I won't try to mentally run through the `union` stuff, but tests and the code that's obvious makes sense to me!

PR #1786, opened 2025-01-23
feat: add support for @starting-style at-rule

michaelabrahamian:
### What is this change?

Adding support for the [@starting-style](https://developer.mozilla.org/en-US/docs/Web/CSS/@starting-style) at-rule. 

So it can be used in animations! Compiled currently throws an exception as it isn't aware of the property.

- Mostly just adding `starting-style` to some lists. 

- The `csstype` dev dependency needed to be added so we could satisfy the `Record<AtRules, boolean>` type [here](https://github.com/atlassian-labs/compiled/pull/1786/files#diff-403ab7d737ac4d231acb7e065756c4481b638dc2451835ae7ec89707d8eba226R9).
  - `@starting-style` was only added to the `AtRules` type in version 3.1.3 of `csstype`: https://github.com/frenic/csstype/blame/fb448e21733ac5cb52523d3b678fdbbe1f9b42f2/index.js.flow#L3573
    - This project was previously not depending on `csstype` explicitly - an older version was being transitively used from `@types/react`: https://github.com/atlassian-labs/compiled/blob/master/yarn.lock#L4364
  - This meant I also had to add `@scope`, as it was also added in version 3.1.3
  - Alternatively, we could just locally extend the type with `Record<AtRules | '@starting-style', boolean>`. Let me know what you think 🙏 

---

### PR checklist

Don't delete me!

I have...

- [x] Updated the documentation in `website/` - N/A

TODO:
- [x] confirm about adding `csstype` dependency 
- [x] confirm what's going on with the failing test


packages/babel-plugin/src/utils/css-map.ts
michaelabrahamian, line 1: **context:** mentioned in the PR description, but will repeat here:

The csstype dependency needed to be added so we could satisfy the Record<AtRules, boolean> type [here](https://github.com/atlassian-labs/compiled/pull/1786/files#diff-403ab7d737ac4d231acb7e065756c4481b638dc2451835ae7ec89707d8eba226R9).

@starting-style was only added to the AtRules type in version 3.1.3 of csstype: https://github.com/frenic/csstype/blame/fb448e21733ac5cb52523d3b678fdbbe1f9b42f2/index.js.flow#L3573
This project was previously not depending on csstype explicitly - an older version was being transitively used from @types/react: https://github.com/atlassian-labs/compiled/blob/master/yarn.lock#L4364
This meant I also had to add @scope, as it was also added in version 3.1.3
Alternatively, we could just locally extend the type with Record<AtRules | '@starting-style', boolean>. Let me know what you think 🙏

stories/starting-style.tsx
```diff
@@ -0,0 +1,142 @@
+/**
+ * @jsxRuntime classic
+ * @jsx jsx
+ */
+import { cssMap, jsx } from '@compiled/react';
+import { useState } from 'react';
+
+export default {
+  title: 'atomic/starting style',
```
michaelabrahamian, line 9: **question:** not sure if `atomic` is the right prefix/group to place it in. Probably not the best place, let me know if you have any suggestions.

stories/starting-style.tsx
michaelabrahamian, line 1: **context:** I've added an example that uses `@starting-style` how we want to use in the new navigation system

packages/babel-plugin/src/css-map/__tests__/at-rules-and-selectors.test.ts
```diff
@@ -277,6 +277,31 @@ describe('css map advanced functionality (at rules, selectors object)', () => {
...
+          },
+        });
+
+        ${EXAMPLE_USAGE}
+      `);
+
+    expect(actual).toIncludeMultiple([
+      '._syaz5scu{color:red}',
+      // TODO: work out why this is failing. The below line is actually being output as '@starting-style color: blue{}'
+      '@starting-style {._syaz5scu{color:blue}}',
```
michaelabrahamian, line 299: **question:** this assertion isn't working out as I'd expect... Any ideas what I'm doing wrong? Is the input not valid?

stories/starting-style.tsx
michaelabrahamian: Have decided to remove it. Doesn't seem too relevant to have in the repo.

packages/babel-plugin/src/css-map/__tests__/at-rules-and-selectors.test.ts
michaelabrahamian: Interestingly, if I change the input to nest another property under `@starting-style`, e.g.:

```tsx
const styles = cssMap({
  success: {
    color: 'red',
    '@starting-style': {
      div: {
        color: 'blue',
      }
    },
  },
});
```

It outputs this:

```css
@starting-style div{._1gzq13q2{color:blue}}
```

Which makes me think the plugin is doing something wrong? It should be this:

```css
@starting-style { div{._1gzq13q2{color:blue}} }
```

michaelabrahamian: To sanity check, I tried [`@font-face`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face), which is an at-rule with a similar syntax to `@starting-style`:

```tsx
const styles = cssMap({
  success: {
    color: 'red',
    '@font-face': {
      fontFamily: 'Trickster',
    },
  },
});
```

This outputs:

```css
@font-face font-family: Trickster
```

Which seems like the same problem.

The `@media` test seems to work around this through nesting:

```tsx
const styles = cssMap({
  success: {
    color: 'red',
    '@media': {
      'screen and (min-width: 500px)': {
        color: 'blue',
      },
    },
  },
});
```

Outputs:
```css
@media screen and (min-width:500px){._1qhm13q2{color:blue}}
```

Which is valid syntax for `@media`... But `@starting-style` (and `@font-face`) don't accept args, so doesn't make sense here.

packages/babel-plugin/src/utils/css-map.ts
```diff
@@ -20,7 +20,9 @@ const atRules: Record<AtRules, boolean> = {
   '@namespace': true,
   '@page': true,
   '@property': true,
+  '@scope': true,
```
michaelabrahamian, line 23: **context:** `@scope` was also added as it is a member of the `AtRules` type in the updated version of `csstype`. Needed to satisfy the `Record` type.

JakeLane (approved):

Issue #1789, opened 2025-01-28
Compiled dynamic css variables are not compatible with indirect/non-dependent selectors.


kylorhall-atlassian:
**Describe the bug**

When I write code using string interpolation and mix with a target that produces `"".concat()` syntax, Compiled uses `props.style` dynamic CSS variable styles in expected ways and scenarios for styles that are otherwise static.

When combined with `& + label` type scenarios, this outright breaks, because the `props.style` CSS variable it generates does not work, because they are assigned to `&` while `label` is not that element or a descendent of that element, so it has no access to this CSS variable.

**To Reproduce**

```tsx
const inlineMargin = '16px';
const styles = css({
  '& + label': {
    margin: `8px ${inlineMargin}`,
  },
});

export default () => <div>
  <span css={styles}>Hello</span>
  <label>world</label>
</div>
```

✅ In `es2019` this works as expected (reduced example):
⚠ This could also break in other scenarios…but not in the standard scenario in our codebase.
```js
React.createElement('span', {
  className: (0, _runtime.ax)(["_1vyxftgi _rhviftgi _1wcd7vkz _7m0t7vkz"]),
});
React.createElement("label");
```
```css
._1vyxftgi+label{margin-top:8px}
._rhviftgi+label{margin-bottom:8px}
._1wcd7vkz+label{margin-left:1pc}
._7m0t7vkz+label{margin-right:1pc}
```

❌ In either `cjs` or `esm` this breaks (reduced example):
```js
React.createElement('span', {
  className: (0, _runtime.ax)(["_115h1t9g"]),
  style: { "--_g70wcu": (0, _runtime.ix)("8px ".concat(inlineMargin)) }
});
React.createElement("label");
```
```css
._115h1t9g+label{margin:var(--_g70wcu)}
```

This breaks because `<label>` doesn't have access to `--_g70wcu` as that's set on `<span>` and I don't think this could ever be moved to `<label>` or `<div>` at all, instead **this scenario should never happen**—setting `props.style` with styles not targeting self or otherwise should throw an error.

---

Additionally, this happens without _any_ babel magic, eg.
```jsx
export default (props) => {
  const styles = css({
    '& + label': {
      margin: `8px ${props.margin}`,
    },
  });

  return <div>
    <span css={styles}>Hello</span>
    <label>world</label>
  </div>
}
```

**Expected behavior**

 - I do not expect any `style` tag at all as this code is static and generates as static in most cases.
[...]

PR #1795, opened 2025-02-13
Error on indirect dynamic styles and attempt to statically concatenate where feasible

kylorhall-atlassian:
### What is this change?

Fixes https://github.com/atlassian-labs/compiled/issues/1789 two scenarios found:
- The static syntax of `"8px ".concat("8px")` which is auto-generated by certain Babel targets generates `props.style` outputs when not necessary—we can simplify this to `"8px 8px"` ourselves.
- Mixing "indirect selectors" (eg. `~`, `+`, `|`, `||`) that do not inherit CSS variables with dynamic styles which generates CSS variables and `props.style` will never work in Compiled, this now throws an error.

This is done mostly because the mix of these two is now a possible scenario for some `@atlaskit/*` packages internally, causing consumption bugs without any warning.

**⚠️ This may be hard to rollout** and cause pain (throwing errors on already-broken styles).  I should test with a patch before merging to understand this pain.

---

### PR checklist

- [x] @kylorhall-atlassian to validate this via a patch internally doesn't cause massive migration pain ([see comment](https://github.com/atlassian-labs/compiled/pull/1795#issuecomment-2661736540))


packages/babel-plugin/src/utils/object-property-to-string.ts
```diff
@@ -48,7 +48,87 @@ const binaryExpressionToString = (
...
+const concatToString = (
+  expression: StringConcatExpression,
+  meta: Metadata,
+  expressionToString: ExpressionToString
+): string => {
+  const callee = expression.callee;
+  const args = expression.arguments;
+  if (!args.length) return callee.object.value;
+
+  return args.reduce<string>((acc, arg) => {
```
kylorhall-atlassian, line 121: Just to call it out, this `args.reduce` and `args.some` iteration could be a performance smell, but I wonder how many times we actually are parsing `concat()`…

packages/babel-plugin/src/utils/css-builders.ts
```diff
@@ -1080,3 +1101,42 @@ export const buildCss = (node: t.Expression | t.Expression[], meta: Metadata): C
...
+ *
+ * @param node Node we're interested in extracting CSS from.
+ * @param meta {Metadata} Useful metadata that can be used during the transformation
+ */
+export const buildCss = (node: t.Expression | t.Expression[], meta: Metadata): CSSOutput => {
+  const output = buildCssInternal(node, meta);
+
+  // Check for invalid dynamic selectors
+  if (
+    output.css.some(
```
kylorhall-atlassian, line 1128: This could be a fairly large amount of recursion—not sure it's unperformant, but it's a smell.

I couldn't find a superb other way to add this in and say "when I resolve an `output.variable` and a parent key matches this regex throw an error" because every single `buildCss(…)` callsite is spread out so far.

liamqma (approved):

packages/babel-plugin/src/__tests__/index.test.ts
```diff
@@ -406,4 +406,137 @@ describe('babel plugin', () => {
...
+    expect(actual).not.toMatch(/:\s*var\(--_/);
+    expect(actual).not.toMatch(/style=\{\{/);
+    expect(actual).toIncludeMultiple([
+      '._1yt4ogpx{padding:8px var(--ds-space-050) 0}',
+      '._18u0idpf{margin-left:0}',
+      '._otyrftgi{margin-bottom:8px}',
+      '._19it6gvt{border:1px solid var(--ds-red)}',
+    ]);
+
+    expect(actual).toMatchInlineSnapshot(`
```
dddlr, line 435: `toIncludeMultiple` and `toMatchInlineSnapshot` seem to have a lot of overlap in what they test -- can we maybe remove one of them? (e.g. if we were to remove `toMatchInlineSnapshot`, we can add the remaining CSS properties into the `toIncludeMultiple` call)

packages/babel-plugin/src/__tests__/index.test.ts
```diff
@@ -406,4 +406,137 @@ describe('babel plugin', () => {
...
+            </div>
+          }
+        </CC>
+      );
+      "
+    `);
+  });
+
+  describe('indirect selectors with dynamic styles', () => {
+    it.each([
```
dddlr, line 465: TIL `it.each`, nifty function

packages/babel-plugin/src/__tests__/index.test.ts
kylorhall-atlassian: Yeah, I agree.  Just leftover from debugging, effectively.

kylorhall-atlassian:
cc: @liamqma Just to pull some metrics for a sanity check, in our `@atlaskit/*` (etc) Platform:
1. I wrapped every `extractObjectExpression.newConcatenation` in a timing function, totalled 79ms building every package with most being <.01ms, 14 being >1ms, and a max of `8.2ms`, LGTM…
2. This found 98 strings that could be concatenated and fixed them, eg. instead of `boxShaow: var(--_…)` and `props.style`, it should just get `box-shadow: 0 0 0  0 0 0 var(--ds-border-width-outline, 2px) var(--ds-border-selected, #0C66E4)` directly.
3. There were zero "indirect selectors" errors 👌 (we had one a few months back that was manually fixed, glad there's no more).

May have wider impact outside of Platform, but if that's a good example, I think it's a fair tradeoff for customer-facing performance and safety.

liamqma (approved):

PR #1797, opened 2025-02-13
Fix height usages causing crash in non-media query

dddlr:
### What is this change?

Update the `parseAtRule` function to ensure it only runs for media queries (what the function is designed for) and not other types of at-rules. Also update the name to clarify that it only works for media queries.

Fixes the bug where this causes Compiled to crash at build time:

```
const overlayStyles = css({
	'@supports not (height: 1lh)': {
		color: token('color.text'),
	},
});
```

This is because the presence of `height` causes the at-rule to be treated like a media query, resulting in Compiled crashing when the at-rule doesn't use the expected syntax.

---

### PR checklist

Don't delete me!

I have...

### DIFF
--- a/packages/css/src/plugins/sort-atomic-style-sheet.ts
+++ b/packages/css/src/plugins/sort-atomic-style-sheet.ts
@@ -2,7 +2,7 @@ import type { ChildNode, Rule, Plugin, AtRule } from 'postcss';
 
 import { sortPseudoSelectors } from '../utils/sort-pseudo-selectors';
 
-import { parseAtRule } from './at-rules/parse-at-rule';
+import { parseMediaQuery } from './at-rules/parse-media-query';
 import { sortAtRules } from './at-rules/sort-at-rules';
 import type { AtRuleInfo } from './at-rules/types';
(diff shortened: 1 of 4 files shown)


packages/css/src/plugins/sort-atomic-style-sheet.ts
```diff
@@ -59,7 +59,10 @@ export const sortAtomicStyleSheet = (config: {
           case 'rule': {
             if (node.first?.type === 'atrule') {
               atRules.push({
-                parsed: sortAtRulesEnabled ? parseAtRule(node.first.params) : [],
+                parsed:
+                  sortAtRulesEnabled && node.first.name === 'media'
```
kylorhall-atlassian, line 63: I'll assume `node.first.name` isn't `@media` and/or tests would cover this!

kylorhall-atlassian (approved):
Makes sense and a good test.

packages/css/src/plugins/sort-atomic-style-sheet.ts
dddlr: postcss docs says it's `media` so im going with that :p

PR #1815, opened 2025-03-10
Throw error if css function call used inside object in unsupported way

dddlr:
### What is this change?

The following code fails silently:

```tsx
const someObject = {
    blue: css({ color: 'blue' }),
    red: css({ color: 'red' }),
}

const App = <div css={someObject['red']}>hello</div>
```

We add an error so that it no longer fails silently.

We support this: (see https://github.com/atlassian-labs/compiled/pull/1321 )

```tsx
const someObject = {
    blue: css({ color: 'blue' }),
    red: css({ color: 'red' }),
}

const App = <div css={someObject.red}>hello</div>
```

But we don't support this:

```tsx
const someObject = {
    blue: css({ color: 'blue' }),
    red: css({ color: 'red' }),
}

const App = <div css={someObject['red']}>hello</div>
```

We should either:

* throw an error for `someObject['red']` telling developers either to switch to `someObject.red` or switch to `cssMap`, or
* add support for `someObject['red']`

The first option is easier so I've done that in this PR 😅  but curious to hear your thoughts on whether we should in fact do the latter instead.

---

### PR checklist

Don't delete me!

I have...


packages/babel-plugin/src/utils/is-compiled.ts
```diff
@@ -135,3 +135,22 @@ export const isCompiledStyledTaggedTemplateExpression = (
...
+
+/** Returns true if and only if the current node is one of the following:
+ * - A usage of the `css` API from Compiled
+ * - A usage
+ * - A usage of the `styled` API from Compiled
+ * - IS NOT a usage of the `keyframes` API from Compiled
+ */
+export const isCompiledStyleCall = (node: t.Expression, state: State): boolean => {
+  const checkers = [
+    isCompiledCSSCallExpression,
```
dddlr, line 147: we probably don't need all of these

packages/babel-plugin/src/utils/css-builders.ts
```diff
@@ -647,6 +648,17 @@ const extractObjectExpression = (node: t.ObjectExpression, meta: Metadata): CSSO
         return;
       }
 
+      if (isCompiledStyleCall(propValue, updatedMeta.state)) {
+        throw new Error(
+          "You can't use Compiled APIs like this within an object.\n\n" +
+            `This is the invalid code: ${generate(node).code}\n\n` +
+            "Usually this happens because you're referencing some styles using syntax like `styles['someString']`. We don't support this for `css` function calls. Instead, you should either:\n" +
+            '- change it to not use square brackets, like `styles.hello`\n' +
+            '- if you need to use square brackets (e.g. you are styling something like `styles[someKey]`), use the cssMap API instead -- https://compiledcssinjs.com/docs/api-cssmap\n\n' +
```
dddlr, line 657: the implication of course is that theoretically, you should only need `cssMap` if _which set_ of styles you are using is dynamic, or otherwise not known until runtime. For example, if you had some info styles and some warning styles, and you didn't know whether to apply the info styles or warning styles until someone clicks on a button.

packages/babel-plugin/src/utils/is-compiled.ts
```diff
@@ -135,3 +135,22 @@ export const isCompiledStyledTaggedTemplateExpression = (
   t.isTaggedTemplateExpression(node) &&
   (isCompiledStyledMemberExpression(node.tag, state) ||
     isCompiledStyledCompositionCallExpression(node.tag, state));
+
+/** Returns true if and only if the current node is one of the following:
+ * - A usage of the `css` API from Compiled
+ * - A usage
```
liamqma, line 141:
```suggestion
 * - A usage of the `cssMap` API from Compiled
```

packages/babel-plugin/src/utils/css-builders.ts
```diff
@@ -647,6 +648,17 @@ const extractObjectExpression = (node: t.ObjectExpression, meta: Metadata): CSSO
         return;
       }
 
+      if (isCompiledStyleCall(propValue, updatedMeta.state)) {
```
liamqma, line 651: Can you please write a test case for it?

liamqma (approved):

kylorhall-atlassian (approved):
Same as Liam — a test case or two and we're golden!  If this throws errors, we should almost certainly fix them and let them block us—I think it's "safe" to land.

alexreardon (approved):
I think `throw`ing in an unhandled case is way better than failing silently. I am happy for this change to go in. As suggested elsewhere, I think it would be great to add tests for this behaviour to ensure it's working as expected

dddlr:
closing this PR for now as i haven't had time to get it ready sorry!

feel free to pick it up, it just needs tests and a changeset

PR #1816, opened 2025-03-13
Fix bug where "import * as React from 'react'" is incompatible with @babel/plugin-transform-runtime

dddlr:
### What is this change?

`@compiled/babel-plugin` automatically inserts a React import at the start of the file by default (as long as `importReact` is not turned off), in the form of `import * as React from 'react'`. This PR changes this import to `import React from 'react'`.

Fix bug where "import * as React from 'react'" is incompatible with @babel/plugin-transform-runtime

---

### PR checklist

Don't delete me!

I have...


JakeLane (approved):

packages/babel-plugin/src/babel-plugin.ts
```diff
@@ -198,7 +198,7 @@ export default declare<State>((api) => {
 
           if (!pragma.jsxImportSource && shouldImportReact && !path.scope.getBinding('React')) {
             // React is missing - add it in at the last moment!
-            path.unshiftContainer('body', template.ast(`import * as React from 'react'`));
```
kylorhall-atlassian, line 201: I would just put this behind a configuration, ship it disabled by default, try it internally, and swap the default…?

PR #1829, opened 2025-05-14
Fix the case when the styles are defined below the component and the …

liamqma:
This PR fixes the below case:

```
<Box xcss={styles.main}>{children}</Box>

const styles = cssMap({
	main: {
		paddingTop: "10px",
	},
});

```

### PR checklist

Don't delete me!

I have...

### DIFF
--- a/packages/babel-plugin/src/xcss-prop/index.ts
+++ b/packages/babel-plugin/src/xcss-prop/index.ts
@@ -4,7 +4,7 @@ import * as t from '@babel/types';
 import type { Metadata } from '../types';
 import { buildCodeFrameError, getPathOfNode } from '../utils/ast';
 import { compiledTemplate } from '../utils/build-compiled-component';
-import { buildCss } from '../utils/css-builders';
+import { buildCss, generateCacheForCSSMap } from '../utils/css-builders';
 import { transformCssItems } from '../utils/transform-css-items';
 
 function getJsxAttributeExpressionContainer(path?: NodePath<t.JSXAttribute>) {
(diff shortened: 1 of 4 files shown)


packages/babel-plugin/src/xcss-prop/index.ts
```diff
@@ -112,7 +114,13 @@ export const visitXcssPropPath = (path: NodePath<t.JSXOpeningElement>, meta: Met
     // 1. Dot notation, such as "styles.text"
     // 2. Bracket notation, such as "styles[appearance]"
     const identifiers = collectPathMemberExpressionIdentifiers(propPath);
-    const sheets = collectPassStyles(meta, identifiers);
+    identifiers.forEach((identifier) => {
```
dddlr, line 117: curious how big of a build time impact running `generateCacheForCSSMap` for every identifier in `cssMap` usages here would be

or do you anticipate that it will be negligible because of this line in `generateCacheForCSSMap`, which i assume prevents `generateCacheForCSSMap` from running more than it needs to?

```ts
  if (meta.state.cssMap[node.name] || meta.state.ignoreMemberExpressions[node.name]) {
    return;
  }
```

liamqma: yea, it should be negligible since it's cached.

liamqma: generateCacheForCSSMap is executed for each identifier in the memberExpression.

For instance, consider the following example:

```
<Box xcss={[styles.foo, styles.bar, anotherStyles.foo]} />
```

The function will iterate over the array [styles.foo, styles.bar, anotherStyles.foo], making the following calls:

- generateCacheForCSSMap(styles)
- generateCacheForCSSMap(styles) // it will early return because styles is already parsed
- generateCacheForCSSMap(anotherStyles)

dddlr (approved):

packages/babel-plugin/src/xcss-prop/index.ts
dddlr: thank you for the extra context! yeah looks good 👍

kylorhall-atlassian: Yeah, I think this makes sense.  Plus, isn't the cache on a per-file basis anyways?  Just avoids that `styles.foo`, `styles.bar` from collecting `styles` twice?

kylorhall-atlassian (approved):

PR #1833, opened 2025-05-30
Improve perf by roughly 10% by getting rid of hidden AST parsing

shanshrew:
### What is this change?

Removes `hidden` AST Parsing via template with direct AST calls.

To start working on moving the performance needle without changing behaviour decreasing build times in AFM.

```
initial run x 1,870 ops/sec
```
after
```
initial run x 2,048 ops/sec
```

---

### PR checklist

Don't delete me!

- [ x] Updated or added applicable tests [ NOT NECESSARY ]
- [ x] Updated the documentation in `website/`. [ NOT NECESSARY ]
- [ x] Added a changeset (if making any changes that affect Compiled's behaviour) [ NOT NECESSARY ]

### DIFF
--- a/packages/babel-plugin/src/utils/build-compiled-component.ts
+++ b/packages/babel-plugin/src/utils/build-compiled-component.ts
@@ -1,7 +1,4 @@
-import generate from '@babel/generator';
-import template from '@babel/template';
 import * as t from '@babel/types';
-import { unique } from '@compiled/utils';
 
 import type { Metadata } from '../types';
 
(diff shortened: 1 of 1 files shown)


JakeLane (changes requested):
Looks good! You'll need to add a changeset with `yarn changeset`

kylorhall-atlassian:
@JakeLane @dddlr was this fine to merge? Shanon asked me separately, but I honestly have zero ability to review this 😄 I'll rebase it at least!

dddlr:
> @JakeLane @dddlr was this fine to merge? Shanon asked me separately, but I honestly have zero ability to review this 😄 I'll rebase it at least!

yeah fine to merge

`compiledTemplate` is a deterministic function that generates some Compiled HTML elements (technically they are babel nodes but they will be converted to HTML elements later)

so it's fine as long as the snapshot test output stays the same (and it looks like they do)

dddlr (approved):
nice work 👍

kylorhall-atlassian:
Thanks @shanshrew now getting this bumped into our internal repos won't happen overnight, but after https://github.com/atlassian-labs/compiled/pull/1837 merges, it will be available on NPM shortly thereafter.

PR #1835, opened 2025-06-03
Allow for flattened chained pseudo-selectors, eg. `&:hover::after` in our types

ivanbarral:
### What is this change?

Adding the types to the types file and copying the specific selectors from xcss 
- should we have more types here ? Are there any other flat pseudos we want? 

Allow for flattened chained pseudo-selectors, eg. `&:hover::after` in our type syntaxes. Kylor's PR https://github.com/atlassian-labs/compiled/pull/1834 allows for the nesting to occur, but sometimes a flattened selector can be cleaner for some users and I don't think it's supported with that PR 🤔 
Happy to decline if i'm misunderstanding that PR !

### PR checklist

Don't delete me!

I have...

### DIFF
--- a/packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api-recursive.ts
+++ b/packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api-recursive.ts
@@ -24,9 +24,15 @@ interface PressedProperties {
   backgroundColor: BackgroundPressed;
 }
 
+interface ChainedProperties {
+  color: ColorPressed;
+  backgroundColor: BackgroundPressed;
+}
+
 interface CSSPropertiesSchema extends Properties {
   '&:hover': HoveredProperties;
   '&:active': PressedProperties;
+  '&:hover::after': ChainedProperties;
 }
 
 const { css, cssMap, cx, XCSSProp } = createStrictAPI<CSSPropertiesSchema>();
--- a/packages/react/src/create-strict-api/types.ts
+++ b/packages/react/src/create-strict-api/types.ts
@@ -1,9 +1,8 @@
 import type {
   StrictCSSProperties,
-  CSSPseudoClasses,
   CSSPseudoElements,
   CSSPseudos,
-  FlattenedChainedCSSPseudos,
+  AllCSSPseudoClasses,
 } from '../types';
 
 /**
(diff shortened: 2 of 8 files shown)


packages/react/src/create-strict-api/__tests__/__fixtures__/strict-api-recursive.ts
```diff
@@ -24,9 +24,15 @@ interface PressedProperties {
...
 
+interface ChainedProperties {
+  color: ColorPressed;
+  backgroundColor: BackgroundPressed;
+}
+
 interface CSSPropertiesSchema extends Properties {
   '&:hover': HoveredProperties;
   '&:active': PressedProperties;
+  '&:hover::after': ChainedProperties;
```
kylorhall-atlassian, line 35: Yeah this is an example, should we do anything here in the `@atlaskit/css` type @danieldelcore?

Would `&:hover` + `&:hover::after` both want `HoveredProperties`?

packages/react/src/create-strict-api/types.ts
```diff
@@ -27,7 +28,7 @@ export type AllowedStyles<TMediaQuery extends string> = StrictCSSProperties &
 export type ApplySchemaValue<
   TSchema,
   TKey extends keyof StrictCSSProperties,
-  TPseudoKey extends CSSPseudoClasses | ''
+  TPseudoKey extends CSSPseudoClasses | FlattenedChainedCSSPseudos | ''
```
kylorhall-atlassian: 👌 This allows it so you can specify types for it, eg. `&:hover` gets "hovered tokens" and I think it's fine opening it up to the same for `&:hover::after` if we wanted to. This doesn't force us to.

kylorhall-atlassian (approved):
Maybe we remove this too: https://github.com/atlassian-labs/compiled/blob/4b2e5eeb17e4d0f2359723e484cbe73f26320531/packages/react/src/create-strict-api/types.ts#L57
```diff
-: TKey extends `@${string}` | CSSPseudoElements
+: TKey extends `@${string}`
```

But let's save that for another PR, if at all. I'm not 100% certain how safe that might be.

packages/react/src/types.ts
```diff
@@ -51,6 +51,16 @@ export type CSSPseudoElements =
   | '&::target-text'
   | '&::view-transition';
 
+export type FlattenedChainedCSSPseudosClasses =
```
ivanbarral, line 54: don't think class is totally the right name for this 😞. Its behavior is most similar to classes is my understanding (can't be nested in theory) so that's why I've aligned it closer to them but happy to change the naming/implementation here

ivanbarral:
Guessing this doesn't require: `Updated the documentation in website/` ?

liamqma (approved):

PR #1843, opened 2025-08-04
[SWC] - Adds rust based replica of babel-plugin and accompanying tests

shanshrew:
### What is this change?
- Adds SWC support via combination of manual and vibe coding.
- Adds tests of babel-plugin to swc-plugin and verifies they all pass.
- Meticulously tested the .wasm binary in a `parcel-transform-swc` step that will come in another PR.

- Babel plugin accounts for roughly 40% of the build **(based on my rough testing)** this improves the performance by up to 9x which is a multi-million dollar cost saving.

Change will be rolled out in 3 parts.
1. swc-plugin PR merged
2. parcel-transform-swc PR merged, verified in JIRA which will make sure the output is completely identical to the babel plugin.
3. atlaspack core rollout (potentially) PR

---

### PR checklist

Don't delete me!

I have...

- [ x] Updated or added applicable tests
- [ x] Updated the documentation in `website/` **When it's verified stable**
- [ x] Added a changeset (if making any changes that affect Compiled's behaviour)


packages/swc-plugin/src/utils/css_builder.rs
```diff
@@ -0,0 +1,242 @@
...
+}
+
+pub fn build_atomic_rules_from_object(obj: &ObjectLit) -> Vec<AtomicRule> {
+    let mut out = Vec::new();
+    collect_atomic_rules_from_object(obj, None, None, &mut out);
+    out
+}
+
+fn needs_px_unit(property: &str) -> bool {
+    matches!(property,
```
kylorhall-atlassian, line 199: Where's this list come from?

There's going to be a lot more, eg. `marginBlock`, `marginBlockEnd`, `offset`, `outline`, etc., right?

packages/swc-plugin/src/visitors/css.rs
```diff
@@ -0,0 +1,345 @@
...
+            PropOrSpread::Prop(prop_box) => {
+                match prop_box.as_ref() {
+                    Prop::KeyValue(kv) => {
+                        match &*kv.value {
+                            Expr::Lit(_) => continue, // Literals are allowed
+                            Expr::Object(nested_obj) => {
+                                is_static_object_strict(nested_obj)?;
+                            }
+                            Expr::Ident(_) => {
+                                return Err("Dynamic variables are not supported in strict mode".to_string());
```
kylorhall-atlassian, line 52: I wonder if we should call it like "in the SWC plugin" for now?  Keep the strict mode less formal.

packages/swc-plugin/src/visitors/css.rs
```diff
@@ -0,0 +1,345 @@
...
+            expr: Box::new(Expr::Array(ArrayLit {
+                span: DUMMY_SP,
+                elems: array_elements,
+            })),
+        }],
+        type_args: None,
+    })
+}
+
+fn add_css_sheet_with_deduplication(
```
kylorhall-atlassian, line 329: Does this sort anywhere, is that functionality copied?  That's a critical part of Babel.

packages/swc-plugin/src/visitors/css_map.rs
```diff
@@ -0,0 +1,216 @@
...
+            Callee::Expr(expr) => match expr.as_ref() {
+                Expr::Ident(ident) => ident.sym.as_ref() == "cssMap",
+                _ => false,
+            },
+            _ => false,
+        }
+    }
+}
+
+pub fn is_static_css_map_strict(obj: &ObjectLit) -> Result<(), String> {
```
kylorhall-atlassian, line 38: In contrast to my comment above with "strict mode" => "…in the SWC plugin…", this makes a lot more sense to be frank.  I'm on the fence.

packages/swc-plugin/tests/strict-mode.test.js
kylorhall-atlassian, line 1: This doesn't have extraction, is that right?  That's fine, just making sure I understand, this is an "experimental dev mode only package" am I right?

packages/swc-plugin/tests/strict-mode.test.js
```diff
@@ -0,0 +1,643 @@
...
+/**
+ * Transform code using our SWC plugin
+ */
+async function transform(code, options = {}) {
+  const wasmPath = path.join(__dirname, '..', 'compiled_swc_plugin.wasm');
+
+  const pluginOptions = {
+    processXcss: true,
+    importSources: ['@compiled/react'],
+    strictMode: true, // Enable strict mode
```
kylorhall-atlassian, line 14: What if this is `false`, would it simply just blow up in a lot of places during parsing, or what happens on on vs. off here?

kylorhall-atlassian (changes requested):
To hoist the comment—this is an "experimental dev-only package" right?

The StrictMode syntax mostly looks good, it's a great start.  I think the only thing I'd like is that `const padding = '8px';` local reusability.

Issues:
 - The hashes are wrong!
 - Could use validation of sorting (it's more important than it may seem)
 - What about sorting?  I'm not 100% certain how it works.

PR #1848, opened 2025-10-17
Skip files that are already transformed

JakeLane:
### What is this change?

In the parcel transformer, we are going to start getting cases where code is already transformed (when we rollout the SWC plugin). We need to optimise the perf of this path

Currently we'll evaluate pre-transformed code which means parsing and transforming with babel.

(Optional.)

---

### PR checklist

Don't delete me!

I have...


marcins (approved):

packages/parcel-transformer/src/index.ts
```diff
@@ -99,6 +99,11 @@ export default new Transformer<ParcelTransformerOpts>({
       // For everything else we bail out.
       return undefined;
     }
+    if (code.includes('/* COMPILED_TRANSFORMED_ASSET */')) {
```
kylorhall-atlassian: I missed this, but yeah, makes perfect sense if that path exists!

PR #1851, opened 2025-12-02
Add Vite plugin

OscarCookeAbbott:
### What is this change?

Implements a new Vite plugin for Compiled, mimicking existing plugins for other bundlers.

We have been requested to support Vite.

The Vite plugin is implemented effectively as a wrapper around the Babel transformer, similar to existing other plugins. It hooks into Vite's plugin system and utilises Vite-specific APIs where applicable. It outputs successful Compiled results and is well optimised by Vite's inherent optimisation capabilities.

---

### PR checklist

I have...

### DIFF
--- a/packages/babel-plugin/src/styled/__tests__/behaviour.test.ts
+++ b/packages/babel-plugin/src/styled/__tests__/behaviour.test.ts
@@ -342,7 +342,7 @@ describe('styled component behaviour', () => {
       import * as React from "react";
       import { ax, ix, CC, CS } from "@compiled/react/runtime";
       import colors from "colors";
-      const _10 = "._tzy4kb7n{opacity:1}";
+      const _0 = "._tzy4kb7n{opacity:1}";
       const _9 = "._tzy4idpf{opacity:0}";
       const _8 = "._1e0c1kw7{display:inherit}";
       const _7 = "._1e0cglyw{display:none}";
@@ -360,7 +360,7 @@ describe('styled component behaviour', () => {
           const { isLoading, state, ...__cmpldp } = __cmplp;
           return (
             <CC>
-              <CS>{[_, _2, _3, _4, _5, _6, _7, _8, _9, _10]}</CS>
+              <CS>{[_, _2, _3, _4, _5, _6, _7, _8, _9, _0]}</CS>
               <C
                 {...__cmpldp}
                 style={{


packages/vite-plugin/src/index.ts
```diff
@@ -0,0 +1,237 @@
...
+        // Join all rules and apply CSS sorting
+        const combinedCss = allRules.join('\n');
+        const sortConfig = {
+          sortAtRulesEnabled: options.sortAtRules,
+          sortShorthandEnabled: options.sortShorthand,
+        };
+
+        // Import sort function dynamically to avoid build issues
+        // eslint-disable-next-line @typescript-eslint/no-var-requires
+        const { sort } = require('./sort-css');
```
JakeLane: Is this necessary? Not ideal to defer potential loading failures until runtime

packages/vite-plugin/src/types.ts
```diff
@@ -0,0 +1,113 @@
+import type { PluginItem } from '@babel/core';
+import type { ParserPlugin } from '@babel/parser';
+
+export interface CompiledVitePluginOptions {
```
JakeLane: nit: can you import the types from the babel plugin?

packages/vite-plugin/src/index.ts
```diff
@@ -0,0 +1,237 @@
...
+        const sortedCss = sort(combinedCss, sortConfig);
+
+        // Emit the CSS file
+        const cssFileName = 'compiled.css';
+        this.emitFile({
+          type: 'asset',
+          fileName: cssFileName,
```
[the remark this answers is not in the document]
OscarCookeAbbott: Ah yeah, I've updated this area to be more inline with the other transformers which produces hashed results as expected.

packages/vite-plugin/src/utils.ts
```diff
@@ -0,0 +1,79 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const enhancedResolve = require('enhanced-resolve');
+
+// Handle both ESM and CJS imports
+const { CachedInputFileSystem, ResolverFactory } = enhancedResolve.CachedInputFileSystem
+  ? enhancedResolve
+  : enhancedResolve.default || enhancedResolve;
+
+export function createDefaultResolver(config: CompiledVitePluginOptions): Resolver {
```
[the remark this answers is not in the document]
OscarCookeAbbott: It seems to work fine, though I can't guarantee there's no better alternative

packages/vite-plugin/src/index.ts
```diff
@@ -0,0 +1,276 @@
...
+        // Throw error to be displayed by Vite
+        const error = e as Error;
+        this.error({
+          message: `[@compiled/vite-plugin] Failed to transform: ${error.message}`,
+          stack: error.stack,
+        });
+      }
+    },
+
+    generateBundle(_outputOptions: any, bundle: OutputBundle) {
```
kylorhall-atlassian: General question, how'd you find the docs to work with this?  I had tried once upon a time to take https://www.npmjs.com/package/@nitedani/vite-plugin-compiled-react and do some of this, but really couldn't figure out the API and google and LLM didn't do me any good 😢 but yeah, superb job from what I can tell!

PR #1853, opened 2025-12-15
Upgrades to React 18 and improves ESM support for `@compiled/react/jsx-runtime` and `@compiled/vite-plugin`

kylorhall-atlassian:
### What is this change?

 - Upgrades all packages to using React 18 (even the `jsx-runtime` change should be a noop change and not breaking as far as I can discover)
 - Removes the `export *` which has issues in some bundling for `@compiled/react/jsx-runtime`
 - Adds ESM support to `@compiled/vite-plugin` to avoid calling via `compiled.default({…})`

To better support ESM bundling.

I looked at [Emotion](https://github.com/emotion-js/emotion/blob/main/packages/react/src/jsx-dev-runtime.ts) and let AI guide me a bit.

---

### PR checklist

### DIFF
--- a/.changeset/large-gifts-sip.md
+++ b/.changeset/large-gifts-sip.md
@@ -1,0 +1,5 @@
+---
+'@compiled/vite-plugin': patch
+---
+
+Split CJS and ESM exports for the Vite Plugin
--- a/packages/tsconfig.esm.json
+++ b/packages/tsconfig.esm.json
@@ -1,4 +1,4 @@
 {
   "files": [],
-  "references": [{ "path": "react/tsconfig.json" }]
+  "references": [{ "path": "react/tsconfig.json" }, { "path": "vite-plugin/tsconfig.json" }]
 }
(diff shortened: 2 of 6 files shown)


.changeset/slimy-cobras-reply.md
```diff
@@ -0,0 +1,19 @@
...
+'@compiled/webpack-loader': minor
+'@compiled/vite-plugin': minor
+'@compiled/webpack-app': minor
+'@compiled/parcel-app': minor
+'@compiled/react': minor
+'@compiled/vite-app': minor
+'@compiled/ssr-app': minor
+---
+
+Bump package dependencies to React 18 globally. Replace jsx and jsx-dev runtimes with modern syntax to circumvent `export \*` issues.
```
kylorhall-atlassian: _Possibly_ a breaking change, but I cannot see a reference to this `jsx-runtime` changing between React 16–19, so calling it a `minor`.

PR #1855, opened 2025-12-15
Fix ESM Vite Plugin support now that we distribute multiple variants

kylorhall-atlassian:
### What is this change?

Fix ESM Vite Plugin support now that we distribute multiple variants.

Bumping to `v1.1.0` had this issue:
```sh
const enhancedResolve = require('enhanced-resolve');
                        ^
ReferenceError: require is not defined in ES module scope, you can use import instead
    at <anonymous> (/home/runner/workspace/node_modules/@compiled/vite-plugin/src/utils.ts:9:25)
```

tl;dr: I overlooked this one-off hack and the `dist/esm/*` shouldn't have `require()` in it.

### PR checklist

### DIFF
--- a/packages/vite-plugin/src/utils.ts
+++ b/packages/vite-plugin/src/utils.ts
@@ -2,16 +2,9 @@ import * as fs from 'fs';
 import { dirname } from 'path';
 
 import type { Resolver } from '@compiled/babel-plugin';
+import { CachedInputFileSystem, ResolverFactory } from 'enhanced-resolve';
 
 import type { PluginOptions } from './types';
-
-// eslint-disable-next-line @typescript-eslint/no-var-requires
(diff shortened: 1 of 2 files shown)


packages/vite-plugin/src/utils.ts
```diff
@@ -23,6 +16,7 @@ const { CachedInputFileSystem, ResolverFactory } = enhancedResolve.CachedInputFi
  */
 export function createDefaultResolver(config: PluginOptions): Resolver {
   const resolver = ResolverFactory.createResolver({
+    // @ts-expect-error - enhanced-resolve CachedInputFileSystem types are not compatible with Node.js fs types, but work at runtime
```
kylorhall-atlassian: Pretty certain it didn't complain before because `require()` didn't bring types in…?

PR #1876, opened 2026-03-30
Use Artifactory for publish only

at-nathan:
### What is this change?

Use Artifactory as publish only registry

Currently Artifactory is set as the registry for download and publishing, when it should be for publish only.
  
Updating `.npmrc`

(Optional.)

---

### PR checklist

Don't delete me!

I have...


liamqma (approved):

.github/workflows/release.yml
```diff
@@ -48,8 +48,5 @@ jobs:
           publish: yarn release
         env:
           GITHUB_TOKEN: ${{ secrets.ATLAS_DST_BOT_GITHUB_TOKEN }}
-          # NPM_TOKEN will be in a generated .npmrc-public file, created by atlassian-labs/artifact-publish-token
-          # Setting dummy value so changesets doesn't complain
```
kylorhall-atlassian: Does not not complain?

kylorhall-atlassian (approved):

.github/workflows/release.yml
at-nathan: Looks like it's all good 😌 
https://github.com/atlassian-labs/compiled/actions/runs/23731938620/job/69127481802

kylorhall-atlassian: We did it! Thanks Nathan!

PR #1879, opened 2026-04-02
create style provider to direct style injection

t-davis-atlassian:
### What is this change?
Adds a StyleContainerProvider React context provider that allows consumers to redirect Compiled's <style> element
injection into a custom DOM node (e.g. a Shadow DOM root) instead of document.head.

Also exports a StyleContainerConfig type and adds container and cacheKey fields to StyleSheetOpts.

When rendering inside a Shadow DOM, styles injected into document.head are not visible to the shadow tree. This change
enables consumers to wrap their shadow-root-rendered subtree with <StyleContainerProvider container={shadowRoot} 
cacheKey="..."> so that Compiled injects styles directly into the shadow root instead.

- A new StyleContainerProvider component and useStyleContainer hook are introduced in packages/react/src/runtime/style-container.tsx. The provider sets a module-level singleton (clientStyleContainer) synchronously via useMemo during the render phase so that child CS components can read it immediately, and clears it on unmount via useEffect.
- The Style component (style.tsx) reads the active container via useStyleContainer() and, when present, passes container and cacheKey into insertRule and namespaces the deduplication cache key (${cacheKey}:${sheet}) so the container's inserted records are tracked independently from the main document cache.
- container and cacheKey are added to the StyleSheetOpts type in types.ts, and sheet.ts is updated to target the provided container when inserting rules.
- StyleContainerProvider and StyleContainerConfig are re-exported from packages/react/src/index.ts.

---

### PR checklist

Don't delete me!

I have...

### DIFF
--- a/packages/react/src/runtime/__tests__/style.test.tsx
+++ b/packages/react/src/runtime/__tests__/style.test.tsx
@@ -4,6 +4,7 @@ import type { ComponentType } from 'react';
 
 import StyleWithContainer from '../style';
 import { StyleContainerProvider } from '../style-container';
+import type * as StyleContainerModule from '../style-container';
 
 jest.mock('../is-server-environment', () => ({
   isServerEnvironment: () => false,
@@ -292,6 +293,33 @@ describe('<Style />', () => {
 
       expect(container.innerHTML).toInclude('nonce="abc123"');
       expect(console.error).not.toHaveBeenCalled();
+    });
+
+    it('should throw when used in a server environment', () => {
+      jest.resetModules();
+      jest.doMock('../is-server-environment', () => ({
+        isServerEnvironment: () => true,
+      }));
+
+      // Re-require to pick up the new mock
+      const { StyleContainerProvider: ServerStyleContainerProvider } =
+        jest.requireActual<typeof StyleContainerModule>('../style-container');
+
+      // Suppress React's error boundary console output during this test
+      const spy = jest.spyOn(console, 'error').mockImplementation(() => {});
+
+      expect(() => {
+        render(
+          <ServerStyleContainerProvider container={container} cacheKey="test">
+            <div />
+          </ServerStyleContainerProvider>
+        );
+      }).toThrow(
+        '@compiled/react: StyleContainerProvider is not supported in server environments.'
+      );
+
+      spy.mockRestore();
+      jest.resetModules();
     });
   });
 });
--- a/packages/react/src/runtime/__tests__/style.test.tsx
+++ b/packages/react/src/runtime/__tests__/style.test.tsx
@@ -295,7 +295,7 @@ describe('<Style />', () => {
       expect(console.error).not.toHaveBeenCalled();
     });
 
-    it('should throw when used in a server environment', () => {
+    it('should warn in dev when used in a server environment', () => {
       jest.resetModules();
       jest.doMock('../is-server-environment', () => ({
         isServerEnvironment: () => true,
@@ -305,20 +305,23 @@ describe('<Style />', () => {
       const { StyleContainerProvider: ServerStyleContainerProvider } =
         jest.requireActual<typeof StyleContainerModule>('../style-container');
 
-      // Suppress React's error boundary console output during this test
-      const spy = jest.spyOn(console, 'error').mockImplementation(() => {});
+      const warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {});
 
-      expect(() => {
-        render(
-          <ServerStyleContainerProvider container={container} cacheKey="test">
-            <div />
-          </ServerStyleContainerProvider>
-        );
-      }).toThrow(
-        '@compiled/react: StyleContainerProvider is not supported in server environments.'
+      const { container: renderContainer } = render(
+        <ServerStyleContainerProvider container={container} cacheKey="test">
+          <div />
+        </ServerStyleContainerProvider>
       );
 
-      spy.mockRestore();
+      expect(warnSpy).toHaveBeenCalledWith(
+        expect.stringContaining(
+          '@compiled/react: StyleContainerProvider has no effect in server environments.'
+        )
+      );
+      // Children should still be rendered
+      expect(renderContainer.querySelector('div')).not.toBeNull();
+
+      warnSpy.mockRestore();
       jest.resetModules();
     });
   });
--- a/packages/react/src/runtime/__tests__/style.test.tsx
+++ b/packages/react/src/runtime/__tests__/style.test.tsx
@@ -2,6 +2,9 @@ import { render } from '@testing-library/react';
 import React from 'react';
 import type { ComponentType } from 'react';
 
+import StyleWithContainer from '../style';
+import { StyleContainerProvider } from '../style-container';
+
 jest.mock('../is-server-environment', () => ({
   isServerEnvironment: () => false,
 }));
@@ -19,7 +22,9 @@ describe('<Style />', () => {
   });
 
   // We want to isolate the test to correctly mimic the environment being loaded in once
-  const createIsolatedTest = (callback: (Style: ComponentType<{ children: string[] }>) => void) => {
+  const createIsolatedTest = (
+    callback: (Style: ComponentType<{ children: string[]; nonce?: string }>) => void
+  ) => {
     jest.isolateModules(() => {
       // eslint-disable-next-line @typescript-eslint/no-var-requires
       const Style = require('../style');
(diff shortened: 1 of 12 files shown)


packages/react/src/runtime/style-container.ts
```diff
@@ -0,0 +1,119 @@
...
+  cacheKey,
+  children,
+}: StyleContainerConfig & { children: React.ReactNode }): JSX.Element {
+  if (isServerEnvironment()) {
+    return (
+      <ServerContext.Provider value={{ container, cacheKey }}>
+        {children as JSX.Element}
+      </ServerContext.Provider>
+    );
+  }
```
kylorhall-atlassian, line 91: I'm a bit confused by this part, would we ever have a runtime + server environment?  Maybe it's fine for backwards compatibility, but production should really just be injecting the generated `css` files, otherwise you're going to break something as runtime injection can be prone to issues and isn't safe for production..

liamqma: Yes, +1 for the production extraction mode. We only addresses the runtime mode path here. With extraction, the Style/CS components are stripped entirely, so there's nothing for the provider to intercept. 
To support extraction + shadow DOM, we'd need to gather the extracted CSS at runtime and inject it into the corresponding StyleContainers. That's non-trivial work 
If supporting extraction + shadow DOM is a separate piece of work, worth calling that out as a known limitation in the docs.

kylorhall-atlassian: If this is possible, it makes sense to me. Is this possible:
 - Document shadow DOM + `StyleContainer` is supported for **RUNTIME** only.
 - In `isServerEnvironment()`(?) or `extract: true` it's explicitly disabled, can we even **throw an error** that you must inject the extracted `compiled.css`

t-davis-atlassian: I pushed a change to throw an error if `isServerEnvironment` is true, but I'm a bit confused why runtime + server environment would be an issue. I was mimicking what's done with `useCache`.

```ts
export const useCache: UseCacheHook = () => {
  if (isCacheDisabled()) {
    return {};
  }

  if (isServerEnvironment()) {
    // On the server we use React Context to we don't leak the cache between SSR calls.
    // During runtime this hook isn't conditionally called - it is at build time that the flow gets decided.
    // eslint-disable-next-line react-hooks/rules-of-hooks
    return useContext(Cache) || {};
  }

  // On the client we use the object singleton.
  return Cache;
};
```

https://github.com/t-davis-atlassian/compiled/blob/master/packages/react/src/runtime/style-cache.tsx#L30-L44

liamqma: Reading the code again and chatting with AI, I have a better understanding now.

Thanks for adding the server guard — the intent is right. My suggestion would be to switch from a throw to a dev-mode warning. Let me explain why.
Shadow DOM is inherently client-only. There's no attachShadow() on the server, and React doesn't support declarative shadow DOM. So in practice, the expected usage pattern is always something like:

```
useEffect(() => {
  setShadowRoot(hostRef.current.attachShadow({ mode: 'open' }));
}, []);

return (
  <div ref={hostRef}>
    {shadowRoot && createPortal(
      <StyleContainerProvider container={shadowRoot} cacheKey="widget">
        <MyButton />
      </StyleContainerProvider>,
      shadowRoot
    )}
  </div>
);
```

With this pattern, StyleContainerProvider never renders on the server because shadowRoot is null until the useEffect fires. So the throw would never actually trigger for correctly-written code.
The concern is what happens when someone doesn't use that pattern and accidentally renders the provider during SSR. In that case the Style component's server path kicks in — it renders <style data-cmpld> inline in the HTML, those tags get moved to document.head during hydration, and the dedup cache marks the styles as already inserted. When the provider then activates on the client, the styles are skipped because they're already in the cache. The result: the component is unstyled inside the shadow root, with no error. A silent failure.
So we do want to tell the user early that this won't work. But a throw is too aggressive — it crashes the entire server render instead of just the shadow DOM feature. A dev-mode warning gives us the best of both:

```
if (isServerEnvironment()) {
  if (process.env.NODE_ENV === 'development') {
    console.warn(
      '@compiled/react: StyleContainerProvider has no effect in server environments. ' +
        'Shadow DOM content should be rendered client-side only (e.g. inside a portal ' +
        'guarded by useEffect/useState).'
    );
  }
  return children as ReactElement;
}
```

This way the developer sees a clear message in dev if they've wired it up wrong.

packages/react/src/runtime/style.tsx
```diff
@@ -16,6 +20,7 @@ interface StyleProps extends StyleSheetOpts {
 
 export default function Style(props: StyleProps): JSX.Element | null {
   const inserted = useCache();
+  const styleContainer = useStyleContainer();
```
liamqma, line 23: Thinking about backwards compatibility here — for users who don't use shadow DOM at all, can you confirm the only extra cost they pay is the useStyleContainer() call in the Style component? From my reading it just returns a module-level variable (null when no provider is mounted), so it should be effectively free — but want to make sure there are no other code paths affected.

t-davis-atlassian: Yes, `useStyleContainer` will read a module-level variable that's null by default. So if you're not using
a provider, the condition below that rely on it fall through to the existing code paths with no extra work.

liamqma: Nice! Thanks for confirming.

packages/react/src/runtime/__tests__/style.test.tsx
```diff
@@ -178,4 +184,142 @@ describe('<Style />', () => {
...
+      const spy = jest.spyOn(console, 'error').mockImplementation(() => {});
+
+      expect(() => {
+        render(
+          <ServerStyleContainerProvider container={container} cacheKey="test">
+            <div />
+          </ServerStyleContainerProvider>
+        );
+      }).toThrow(
+        '@compiled/react: StyleContainerProvider is not supported in server environments.'
```
liamqma: We no longer throw here. Why does the test not fail?

website/packages/docs/src/pages/pkg-react-runtime.mdx
```diff
@@ -29,6 +29,54 @@ import { CS } from '@compiled/react/runtime';
...
+
+Redirects Compiled's `<style>` element injection into a custom DOM node instead of `document.head`.
+This is useful when rendering Compiled components inside a [Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM),
+where styles injected into the main document `<head>` are not visible to the shadow tree.
+
+> **Runtime mode only** <br /> > `StyleContainerProvider` only works in runtime mode. When using [CSS extraction](/css-extraction-webpack),
+> `CS`/`CC` components are removed at build time so there is nothing for this provider to intercept.
+> Supporting Shadow DOM with extraction is a known limitation and is planned as future work.
+>
+> It is also not supported in server environments and will throw if rendered server-side.
```
liamqma: No longer throw

packages/react/src/runtime/__tests__/style.test.tsx
t-davis-atlassian: Whoops forgot to update this. I confirmed the updated test fails if the console warning isn't correct

liamqma (dismissed):

t-davis-atlassian:
@liamqma My changes added 63B to runtime/style.js. I increased its the size limit by 100B to pass the Bundle Size pipeline. Is that okay?

liamqma:
> @liamqma My changes added 63B to runtime/style.js. I increased its the size limit by 100B to pass the Bundle Size pipeline. Is that okay?

Oh, I forgot about the runtime size increase. We need to be careful here. 63 B isn't much, so I believe it's acceptable. That said, could you run a performance benchmark to measure the impact of the increased runtime size? It should be fine, but we should verify first, as Jira and Confluence are particularly sensitive to bundle size. There should already be a benchmarking library or tool available in the repo to help with this.

t-davis-atlassian:
> Oh, I forgot about the runtime size increase. We need to be careful here. 63 B isn't much, so I believe it's acceptable. That said, could you run a performance benchmark to measure the impact of the increased runtime size? It should be fine, but we should verify first, as Jira and Confluence are particularly sensitive to bundle size. There should already be a benchmarking library or tool available in the repo to help with this.

@liamqma Ran the benchmark suite on both branches and asked rovo dev to analyze the metrics.

No meaningful performance impact from the 63B increase.
| Test | master | feature branch |
|--|--|--|
| ax() single |76,842,197 ops/sec |~74,000,000 ops/sec |
| ax() array |964,925 ops/sec |925,757 ops/sec |
| ac() array |668,337 ops/sec |674,312 ops/sec |
| ax() string |1,048,574 ops/sec |1,074,173 ops/sec |
| ac() string |828,709 ops/sec |940,346 ops/sec |
| chain ax() string |504,432 ops/sec |497,796 ops/sec |
| chain ac() string |582,201 ops/sec |599,090 ops/sec |

**One side note:** `packages/react/src/runtime/__perf__/sheet.test.ts` also fails on this branch. The test asserts that insertRule is the fastest, but createStyleSheet now wins. Here's why: the test mocks document with no-op stubs — insertBefore, createElement, and appendChild all do nothing and return instantly. This means DOM operations cost zero, so the only measurable cost is the JS logic around them. On this branch, insertRule now does extra work on every call — resolving opts.container ?? document.head and doing a WeakMap lookup — whereas createStyleSheet (which uses its own independent implementation in the test utils) does neither. In a real browser, actual DOM operations dwarf that overhead and insertRule wins as expected. The test's no-op mocks make it artificially sensitive to JS-level overhead that doesn't matter in practice, so this failure is not a concern. How would you like to handle this test? Options I can see are removing the fastest assertion entirely since benchmarks shouldn't really be CI gates on which implementation wins, or leaving it as-is since the performance contract it's asserting still holds in practice.

liamqma:
> packages/react/src/runtime/__perf__/sheet.test.ts

The 63B and benchmark numbers look fine — no concerns there.
For the perf test, I'd lean toward removing the "fastest" assertion. The no-op mocks make it sensitive to JS-level overhead that's irrelevant in a real browser.

liamqma (approved):

liamqma:
The PR looks good to me. I don't think you have merge permissions on this repo — if everything looks good on your end, let me know and I'll hit the merge button.

website/packages/docs/src/pages/pkg-react-runtime.mdx
kylorhall-atlassian, line 1: This is great, thank you, I feel like it's a really succinct feature with enough documentation that it can be maintained or at least aware why it exists over time.

kylorhall-atlassian (approved):

t-davis-atlassian:
> The PR looks good to me. I don't think you have merge permissions on this repo — if everything looks good on your end, let me know and I'll hit the merge button.

@liamqma Looks good on my end 👍

kylorhall-atlassian:
Rebasing and enabling auto-merge…

PR #1885, opened 2026-04-22
WIP add @compiled/vanilla

xtan-atlas:
### What is this change?

Adds a new `@compiled/vanilla` package exposing `cssMap` and `ax` for use in non-React code. Wires up the Babel plugin and `babel-plugin-strip-runtime` to recognise `@compiled/vanilla` imports and emit framework-agnostic output: an `insertSheets([...])` call replaces the `<CC><CS>` JSX wrapper, no React or `forwardRef` import is added, and production extraction routes the rules into the existing `.compiled.css` pipeline.

The Editor migration from Emotion to Compiled needs a path for non-React code (e.g. ProseMirror `toDOM` helpers). Without it, those files stay on Emotion and Editor ships two CSS-in-JS runtimes, defeating most of the bundle-size win from the React-side migration.

The CSS pipeline (`buildCss`, `transformCssItems`, atomicifier, sorter, hasher) is unchanged — vanilla calls the same functions and gets back the same `(sheets, classNames)` data. Only the code-emission layer is new, gated by `state.isVanilla` set when the plugin sees an `@compiled/vanilla` import. `insertRule` is promoted to a public `@compiled/react/runtime` export so vanilla and React rules cohabit the same `<style>` buckets in `document.head`.

The internal `insertSheets` call is removed when extract css into `.compiled.css` via `babel-plugin-strip-runtime`.

### Out of scope

* Smart atomisation heuristic

---

### PR checklist

Don't delete me!

I have...


packages/babel-plugin/src/__tests__/vanilla.test.ts
```diff
@@ -0,0 +1,194 @@
...
+      `,
+      { importReact: false }
+    );
+
+    // Result is a plain className-string map; no JSX wrapper.
+    expect(actual).toInclude('base:');
+    expect(actual).toInclude('accent:');
+    // Sheets are inserted via the vanilla runtime, not the React runtime.
+    expect(actual).toInclude('import { ax, insertSheets } from "@compiled/vanilla/runtime"');
+    expect(actual).toInclude('insertSheets(');
```
kylorhall-atlassian, line 22: I'd expect at least one full snapshot on one of these, I think it's fair to add if it's deterministic.

packages/babel-plugin/src/utils/append-runtime-imports.ts
```diff
@@ -20,26 +20,39 @@ const COMPILED_RUNTIME_IMPORTS_WITH_COMPRESSION = ['ac', 'ix', 'CC', 'CS'];
 const COMPILED_RUNTIME_IMPORTS_WITHOUT_COMPRESSION = ['ax', 'ix', 'CC', 'CS'];
 const COMPILED_RUNTIME_MODULE = '@compiled/react/runtime';
 
+// Vanilla mode is React-free: there is no JSX wrapper, so neither `CC` nor
+// `CS` is needed. `ix` is omitted because vanilla `cssMap` does not support
+// dynamic CSS variable interpolation today (the runtime helper for that lives
+// in `@compiled/react/runtime` and is JSX-shaped). When dynamic interpolation
+// is added to vanilla, `ix` should be re-introduced here.
```
kylorhall-atlassian, line 27: Ah, I'd say that's quite important, but shouldn't block a demo.

Just ensure the TS API or something blows up if used, otherwise it could be really confusing…

packages/babel-plugin/src/babel-plugin.ts
```diff
@@ -262,6 +275,15 @@ export default declare<State>((api) => {
           return;
         }
 
+        // Detect the framework-agnostic vanilla import source. In vanilla mode
+        // we change a handful of code-emission decisions later in the pass:
+        // sheets are inserted via `insertSheets` from
+        // `@compiled/vanilla/runtime`, no React or `forwardRef` import is
+        // added, and no `<CC><CS>` wrapper is generated.
+        if (userLandModule === COMPILED_VANILLA_IMPORT) {
```
kylorhall-atlassian, line 283: Question, could we have `@compiled/react` and `@compiled/vanilla` in the same file?

Would these sorts of things fail here with this `state`, or what level is this `state`?  I feel like it's at a file level.

So we do a pass, hit a `cssMap from vanilla` and then hit a `styled from react` and do what?

I'd suggest a file init health-check if not already somewhere to just fast fail if both imports are found in the same file (by both I mean (`@compiled/react` or `@atlaskit/css`) and `@compiled/vanilla`)

packages/vanilla/README.md
kylorhall-atlassian, line 1: Could use some web documentation, but honestly looking real great. Feeling the vibes.

kylorhall-atlassian (commented):
Honestly, this leaves me pretty surprised and happy with what I see so far.  I'd have to give these things a dry-run, I don't quite feel there's enough test coverage, eg. we have more end-to-end and performance tests with React that are manually spin up, not put into CI, that are worth either creating or trying to run.

But this is enough to test and validate if it's even close before shipping a `0.x` and calling it a pre-release — to that effect, happy to not document it on the web, or mark it as an Alpha early release in both README and web.

PR #1886, opened 2026-04-24
Update jsx runtime types to be compatible with React 18 & 19

flipatlas:
### What is this change?

Updating jsx runtime types to be compatible with React 18 & 19

`JSX.LibraryManagedAttributes` changed its definition in React 19 types and so now @compiled/react breaks inference and any render-prop callback parameters like { fieldProps, error } in <Field> falls back to any.

```
frontend/components/form/permissions/DropdownPermissions.tsx:54:54 - error TS7031: Binding element 'error' implicitly has an 'any' type.

54                     {({ fieldProps: { id, ...rest }, error }) => (
```

I have verified those changes publishing locally the package and using it against type check in affected code base

(Optional.)

---

### PR checklist

Don't delete me!

I have...

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,6 +136,7 @@
       }
     }
   },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e",
   "size-limit": [
     {
       "path": "./packages/react/dist/browser/runtime/css-custom-property.js",
--- a/package.json
+++ b/package.json
@@ -136,7 +136,7 @@
       }
     }
   },
-  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e",
+  "packageManager": "yarn@1.22.22",
   "size-limit": [
     {
       "path": "./packages/react/dist/browser/runtime/css-custom-property.js",
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -71,6 +71,11 @@
     "jsx-runtime",
     "jsx-dev-runtime"
   ],
+  "scripts": {
+    "typecheck": "yarn typecheck:react18 && yarn typecheck:react19",
+    "typecheck:react18": "tsc --noEmit --project tsconfig.typecheck-react18.json",
+    "typecheck:react19": "tsc --noEmit --project tsconfig.typecheck-react19.json"
+  },
   "dependencies": {
     "csstype": "^3.2.3"
   },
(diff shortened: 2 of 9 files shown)


packages/react/package.json
```diff
@@ -80,11 +85,14 @@
...
     "@types/jsdom": "^16.2.15",
+    "@types/react": "^18.3.1",
     "@types/react-dom": "^18.3.1",
     "expect-type": "^0.20.0",
     "jsdom": "^19.0.0",
     "react": "^18.3.1",
-    "react-dom": "^18.3.1"
+    "react-dom": "^18.3.1",
+    "react18-types": "npm:@types/react@^18.3.1",
+    "react19-types": "npm:@types/react@^19.0.0"
```
kylorhall-atlassian, line 95: dev deps, LGTM

package.json
```diff
@@ -136,6 +136,7 @@
       }
     }
   },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e",
```
kylorhall-atlassian: Will this cause any issues with cross-dev compat?

packages/react/src/jsx/jsx-local-namespace.ts
kylorhall-atlassian, line 1: This file shows a ⚠️ on it, is it a conflict? I can't see what it is…

packages/react/src/jsx/jsx-local-namespace.ts
```diff
@@ -67,8 +66,12 @@ export namespace CompiledJSX {
...
   export type ElementAttributesProperty = ReactJSXElementAttributesProperty;
   export type ElementChildrenAttribute = ReactJSXElementChildrenAttribute;
-  export type LibraryManagedAttributes<C, P> = WithConditionalCSSProp<P> &
-    ReactJSXLibraryManagedAttributes<C, P>;
+  // In React 18's @types/react, ReactManagedAttributes delegated to GlobalJSXLibraryManagedAttributes
+  // which aliased back to JSX.LibraryManagedAttributes. When CompiledJSX is used as the JSX namespace
+  // that created a circular reference through our own definition. React 19's @types/react removed that
+  // indirection, breaking the old code. We avoid the issue entirely by not delegating to
+  // ReactJSXLibraryManagedAttributes and instead intersecting P with { key?: React.Key } directly.
+  export type LibraryManagedAttributes<_C, P> = WithConditionalCSSProp<P> & P & { key?: React.Key };
```
kylorhall-atlassian, line 75: Got it, if that's all this `GlobalJSXLibraryManagedAttributes` was doing, makes perfect sense to me, and `key` works across these versions.

kylorhall-atlassian (dismissed):
I feel like the actual change is perfectly safe to make.

I tried testing in Platform by modifying my `node_modules` directly, restarted typechecking and everything, and can't confirm 100% it picked it up properly, but no errors.

packages/react/src/jsx/jsx-local-namespace.ts
flipatlas: It's a warning from the build about unused generic:

Image

package.json
flipatlas: Removed the checksum, we still probably want the same version of the package manager so keeping that

packages/react/src/jsx/jsx-local-namespace.ts
flipatlas: That's now gone, I added eslint ignore comment as I don't have a better way of allowlisting it without changing original type signature

kylorhall-atlassian (approved):

PR #1890, opened 2026-05-07
Issue/confserver 102228 replace legacy css dependency

rjatkins:
### What is this change?

Replace legacy `css` package dependency in `@compiled/jest` with `@adobe/css-tools`

`css@3.0.0` has unresolved redos vulnerabilities that have been fixed in the fork project `@adobe/css-tools`. Migrating should result in a build speedup, as well as fixing these vulnerabilities in the process

---

### PR checklist

Don't delete me!

I have...

- [/] Updated or added applicable tests: this is itself a test component
- [/] Updated the documentation in `website/`: no changes required
- [/] Added a changeset (if making any changes that affect Compiled's behaviour): no expected change in behaviour from this upgrade


yarn.lock
rjatkins, line 1: This lockfile still references `css@3.0.0` via `storybook-addon-performance@0.16.1` pulling in `@testing-library/jest-dom@5.11.10`. It looks as though upgrading `storybook-addon-performance` to a more recent release will remove this dependency, but that's coupled to a `storybook@7` upgrade, which is well beyond the scope of this PR

.nvmrc
rjatkins, line 1: Minor tooling improvement while I was here: upgrading to a currently maintained LTS release of Node

JakeLane (approved):

dddlr (approved):

PR #1896, opened 2026-05-15
feat: add group param to cssMap

ssysoev-atl:
### What is this change?

Adding `{group: true}` option to `cssMap`. 

This option enables grouping of selectors with combinators to avoid producing excessive css.

This is an opt-in per cssMap invocation. And only affects selectors with combinators.

Given these styles:
```
cssMap({
	telepointerStyle: {
		'.ProseMirror .telepointer': {
			'&:hover': {
				'.telepointer-initial': {
					opacity: 0,
					transitionDelay: '150ms',
				},
			},
		},
	},
})
```

Output:

```
// Without grouping

._ay6b6ebc .ProseMirror .telepointer:hover .telepointer-initial{transition-delay:.15s}
._q7vcidpf .ProseMirror .telepointer:hover .telepointer-initial{opacity:0}

// With grouping

._2t1oyj8x .ProseMirror .telepointer:hover .telepointer-initial{opacity:0;transition-delay:.15s}

```

This reduces our produce css size by ~10%.

**Runtime composition trade-off:** Compiled's runtime (`ax`, `ac`) assumes atomic class names — one class per (atRule + selectors + property). A grouped class breaks that invariant. If two components both emit a grouped rule for the same combinator selector but with different declarations, both classes survive `ax()` (they carry different group hashes) and the CSS cascade — not atomic "last wins" — determines the final style. This means per-property overrides are **not** possible for grouped rules.

---

### PR checklist

Don't delete me!

I have...


packages/css/src/plugins/atomicify-rules.ts
```diff
@@ -172,6 +205,99 @@ const atomicifyRule = (node: Rule, opts: PluginOpts): Rule[] => {
...
+ * group of declarations under a single combinator-rule. The key is fed into
+ * the existing `atomicClassName`/`buildAtomicSelector` pipeline so the
+ * generated class name hashes over **all** declarations rather than only the
+ * first one.
+ */
+const buildGroupedHashKey = (decls: Declaration[]): DeclarationKey => {
+  const props = decls.map((d) => d.prop).join(';');
+  const values = decls.map((d) => `${d.value}${d.important ? '!' : ''}`).join(';');
+  return { prop: props, value: values };
+};
```
kylorhall-atlassian, line 219: This one of the more important things — @JakeLane @liamqma @dddlr (just covering my bases, sorry) any option or validation on this.

I don't know what I don't know, but it seems correct to me…

packages/babel-plugin/src/css-map/__tests__/index.test.ts
```diff
@@ -36,6 +36,36 @@ describe('css map basic functionality', () => {
...
+            'div span': {
+              color: 'red',
+              backgroundColor: 'red'
+            }
+        },
+        success: {
+          color: 'green',
+          backgroundColor: 'green'
+        }
+      }, {group: true});
```
kylorhall-atlassian, line 54: Could we add like 1 `group:false` sanity checks to make sure it's not like `Object.hasOwnPropertyKey(options, 'group')` or something silly?

Alternative, should the type only be `?: true`?  I'll add that comment below.

packages/css/src/plugins/atomicify-rules.ts
```diff
@@ -9,6 +10,7 @@ interface PluginOpts {
   atRule?: string;
   parentNode?: Container;
   classHashPrefix?: string;
+  group?: boolean;
```
kylorhall-atlassian, line 13: Should this be `group?: true`? No `false` allowed?

packages/css/src/transform.ts
```diff
@@ -24,6 +24,22 @@ export interface TransformOpts {
...
+export interface LocalTransformOptions {
+  /**
+   * When `true`, rules whose selector contains a CSS combinator (descendant ` `,
+   * child `>`, adjacent sibling `+`, general sibling `~`) are emitted as a
+   * single grouped rule with one class name instead of being split into one
+   * atomic class per declaration.
+   *
+   * @default false
+   */
+  group?: boolean;
```
kylorhall-atlassian, line 40: Should this be `group?: true`? No `false` allowed?

kylorhall-atlassian (approved):
_Assuming CI failures are fine…_

The only indication of a concern I have here is just hash collision, but that might be perfectly fine, only time will tell.  It might be worth when you get these changes into Confluence or Jira to look at the `bundle.css` outputs and manually [with AI] validate there's no collision, but that's it.

I assume this covered all the bases. Very minor nitpicks.

Edit: Nevermind, requires a `changeset`!

kylorhall-atlassian (changes requested):
Forgot, this requires a `changeset`!

packages/babel-plugin/src/css-map/__tests__/index.test.ts
ssysoev-atl: added a test, and added a type at cssMap level to only allow true, internal babel plugin types still allow boolean :)

ssysoev-atl:
I think it should have everything now, added changset and docs, one more tests, and changed cssMap types. When you get a chance @kylorhall-atlassian please have a look again :)

website/packages/docs/src/pages/api-cssmap.mdx
```diff
@@ -196,3 +196,70 @@ const Component = ({ variant }) => <div css={[base, borderStyleMap[variant]]} />
...
+}
+```
+
+> Important: This only affects selectors with combinator, basic selectors still get the same Atomic CSS treatment.
+
+### Runtime composition trade-off:
+
+Grouped rules do not support per-property composition overrides. If two grouped classes target the same selector, both classes remain on the element and normal CSS cascade order decides the result.
+
+Use grouped selectors only when reducing generated CSS for combinator-heavy selectors is more important than atomic per-property overrides.
```
kylorhall-atlassian, line 265: I do think there's value in making this just an ESLint rule and calling it out here.

It's not just a trade-off, it's possibly an incident-inducing configuration if someone mixes it in the wrong way.
 - Never with `xcss` (on either side of pass-through styles)
 - Never mixed with non-grouped styles, eg. `css={[stylesWithGroup.root, stylesWithoutGroup.root, css(…)]}` should fail.

kylorhall-atlassian: I do think this is fine to do iterative, eg. not a singular blocking PR, but the comment is still valid in here.

kylorhall-atlassian (commented):
@ssysoev-atl while I think this is good, happy to approve, I think the internal DACI on how we approach some other things would make me suggest caution in merging:
1. Let's do after we chat w/ the Compiled team next week if at all possible—I don't think there's a rush anyways?
2. Think about the ESLint risks noted: https://github.com/atlassian-labs/compiled/pull/1896/changes#r3322603087

xtan-atlas:
> _Assuming CI failures are fine…_
> 
> The only indication of a concern I have here is just hash collision, but that might be perfectly fine, only time will tell. It might be worth when you get these changes into Confluence or Jira to look at the `bundle.css` outputs and manually [with AI] validate there's no collision, but that's it.
> 
> I assume this covered all the bases. Very minor nitpicks.
> 
> Edit: Nevermind, requires a `changeset`!

@kylorhall-atlassian I used this PR to create a patch in AFM to verify and here's the results: https://hello.atlassian.net/wiki/spaces/EDITOR/pages/7116582953/EditorContentContainer+with+smart+atomisation+for+nested+styles
Basically, there's still 1 harmful collision, but that's due to the existing hash issue.

xtan-atlas:
> Forgot, this requires a `changeset`!

Just pushed the changeset, with major on babel plugin.

@ssysoev-atl @kylorhall-atlassian

xtan-atlas:
> > Forgot, this requires a `changeset`!
> 
> Just pushed the changeset, with major on babel plugin.
> 
> @ssysoev-atl @kylorhall-atlassian

oh, actually it seems that I can't push to this branch. There's already a changeset.

packages/css/src/plugins/atomicify-rules.ts
JakeLane: The issue with this is we're basically breaking the rules of compiled and will have confusing behaviour in practice. `ax` is used to deduplicate the group.

```
const style = cssMap({
  a: { color: 'red' },
  b: {color: 'green' },
}, group: true);

<div css={[style.a, style.b]} />
```

With this change the above becomes non-deterministic and dangerous.  If this is truly required to land editor, I won't block it but it shouldn't be acceptable to use in new code.

kylorhall-atlassian: Yeah, noted elsewhere, this should land with an ESLint rule blocking the mixture of these.

I wasn't even thinking about that scenario though, using two `group:true` — @ssysoev-atl how do you think we'd mix dynamic styles in?
```tsx
const style = cssMap({
  a: { color: 'red', padding: …, margin: …, otherStuff: …, borderColor: 'red' },
  b: { color: 'green', borderColor: 'green' },
}, group: true);

<div css={[style.a, isSuccess && style.b]} />
```

Maybe this `group:true` is even smaller of a solution than I was thinking, I forgot we can't even mix and override with dynamic styles, you could only have one `css={style.a}` for a `group:true` API at all times, and `ax()` would have to **throw an error** when consuming a `group:true` and a `group:false|undefined` hash.

Eg. of `group:true` hashes like `_g_abcd1234` we'd throw if `ax()` resolved to more than one class…and technically `ax()` is runtime, so you can't throw statically.

It's less safe than I imagined.

PR #1897, opened 2026-05-19
fix: clearer error when object property key cannot be statically evaluated

### LINKED ISSUE #1606: Throw clearer error when using barrel files
Imagine we have a project with the following code:

```tsx
////////////////////////
// constants.ts
////////////////////////

export const UNSAFE_container = {
    below: {
        xs: "@container not all and (min-width: 30rem)",
        sm: "@container not all and (min-width: 48rem)",
        md: "@container not all and (min-width: 64rem)",
        lg: "@container not all and (min-width: 90rem)",
        xl: "@container not all and (min-width: 110rem)"
    }
};

////////////////////////
// index.ts
////////////////////////

export * from './constants';

////////////////////////
// some_other_file.ts
////////////////////////

import { UNSAFE_container } from './index';

<div css={
    css({ [UNSAFE_container.below.xs]: ... })
}>
    hello world
</div>
```

Here, index.ts is acting as a [barrel file](https://adrianfaciu.dev/posts/barrel-files/), re-exporting an already exported constant.

This cannot be statically evaluated by `@compiled/babel-plugin`, and thus we get an error (`Cannot statically evaluate the value of "MemberExpression`)

Ideally, we should provide a more descriptive error than this that provides some actionable information. For example:

```
Compiled cannot determine the value of `UNSAFE_container.below.xs` at build time. There are many reasons why this might happen:

* Use of barrel files
* The value of `UNSAFE_container.below.xs` uses a syntax we don't support (e.g. `as const`)
* (etc etc......)
```

Ideally this message should address all the common things we can think of that `@compiled/babel-plugin` cannot statically analyse, not just barrel files.

leeguooooo:
## What

Improves the error thrown by `@compiled/babel-plugin` when an object property key references a value it cannot statically evaluate (most often a value re-exported through a barrel file, as described in #1606).

## Why

Today the error reads:

```
Cannot statically evaluate the value of "MemberExpression
```

Two problems with that:

1. **It loses the closing quote**, so the message is visibly truncated.
2. **It prints the AST node type** (`MemberExpression`, `Identifier`, …) instead of the source-level name. For anything more complex than a bare identifier the developer has no way to know *which* expression Compiled choked on, let alone *why*.

The issue (#1606) asks for a more actionable message that names the expression and enumerates the common causes (barrel files, mutated bindings, runtime values, unresolved modules, TS-only constructs).

## How

In `packages/babel-plugin/src/utils/object-property-to-string.ts`:

- Added a small helper `expressionToReadableName` that reconstructs the source-level name of an expression — handling `Identifier`, `MemberExpression` (dotted and bracket), `PrivateName`, `ThisExpression`, falling back to the AST type for anything unhandled.
- Rewrote the thrown error to use that name, restore the missing closing quote, and list the most common causes a developer can act on without reading Compiled internals.

### Example

Before:

```
Cannot statically evaluate the value of "MemberExpression
```

After:

```
Cannot statically evaluate the value of "UNSAFE_container.below.xs" at build time.
Compiled needs to know the value of object property keys at build time, but could not resolve "UNSAFE_container.below.xs". This commonly happens when:
  - The value is re-exported through a barrel file (e.g. `export * from './foo'`). Try importing directly from the source module.
  - The binding is declared with `let` or `var`, or is reassigned. Use `const` and avoid mutation.
  - The value is produced at runtime (e.g. function calls, dynamic imports, values from `process.env`). Use a literal expression Compiled can read at build time.
  - The value comes from a module Compiled cannot statically resolve. Check the import path and any custom resolver configuration.
  - The expression relies on TypeScript-only constructs (e.g. `as const` widening) that Compiled does not yet evaluate.
```

## Tests

Added four cases to `packages/babel-plugin/src/utils/__tests__/object-property-to-string.test.ts`:

- the identifier name appears in the error,
- a dotted member-expression key is rendered as e.g. `UNSAFE_container.below.xs` instead of `MemberExpression`,
- a computed string key is rendered as `theme["breakpoint-sm"]`,
- the message mentions the common causes (barrel files, `let`, runtime).

Full `yarn jest packages/babel-plugin/` passes (620/620, 0 regressions), `yarn lint` and `yarn prettier --check` clean.

## Changeset

`patch` bump for `@compiled/babel-plugin` — no API change, just a clearer diagnostic.

Closes #1606

### DIFF
--- a/.changeset/clearer-static-eval-error.md
+++ b/.changeset/clearer-static-eval-error.md
@@ -1,0 +1,32 @@
+---
+'@compiled/babel-plugin': patch
+---
+
+Improve the error message thrown when Compiled cannot statically evaluate an object property key.
+
+When an object property key references a value Compiled cannot resolve at build time (for example, a constant re-exported through a barrel file), the babel plugin now:
+
+- Prints the full source-level name of the expression (e.g. `theme.color.primary` or `theme["breakpoint-sm"]`) instead of the AST node type (`MemberExpression`).
+- Lists the most common causes (barrel files, `let`/`var` bindings, runtime values, unresolved module imports, TypeScript-only constructs) so the developer can self-diagnose without digging into Compiled internals.
+
+Also fixes a small bug where the previous message dropped the closing quote (`"...UNSAFE_container` instead of `"...UNSAFE_container"`).
+
+**Before**
+
+```
+Cannot statically evaluate the value of "MemberExpression
+```
+
+**After**
+
+```
+Cannot statically evaluate the value of "UNSAFE_container.below.xs" at build time.
+Compiled needs to know the value of object property keys at build time, but could not resolve "UNSAFE_container.below.xs". This commonly happens when:
+  - The value is re-exported through a barrel file (e.g. `export * from './foo'`). Try importing directly from the source module.
+  - The binding is declared with `let` or `var`, or is reassigned. Use `const` and avoid mutation.
+  - The value is produced at runtime (e.g. function calls, dynamic imports, values from `process.env`). Use a literal expression Compiled can read at build time.
+  - The value comes from a module Compiled cannot statically resolve. Check the import path and any custom resolver configuration.
+  - The expression relies on TypeScript-only constructs (e.g. `as const` widening) that Compiled does not yet evaluate.
+```
+
+Closes #1606.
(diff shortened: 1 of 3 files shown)


leeguooooo:
Closing on my own — see #1899 for context.

PR #1898, opened 2026-05-19
feat(codemods): handle Interpolation and CSSProperties when migrating from emotion

### LINKED ISSUE #898: Account for `Interpolation` and `CSSProperties` in emotion codemod
The exisiting `@emotion` codemod bails out for unsupported imports. This means any change won't compile.

Ideally this would properties would be annotated with a TODO and shifted to `any` or similar.

leeguooooo:
## What

Extends the `emotion-to-compiled` codemod to handle `Interpolation` and `CSSProperties` — two type-only imports from `@emotion/core` / `@emotion/react` that Compiled doesn't provide an equivalent for.

Closes #898.

## Why

The codemod's "Gotchas" section already calls out that some emotion imports are unsupported and get silently left behind in the original `@emotion/*` import. For value imports like `CSSObject` that's fine — they need real, user-driven decisions. For type-only imports like `Interpolation` and `CSSProperties` it's a worse outcome: the migrated file still pulls in `@emotion/*` (just for one type) and the developer has to chase it down manually before the migration is complete.

#898 suggests annotating these with a `TODO` and shifting them to `any` so the file compiles cleanly after the codemod runs, and the developer can come back to pick a real replacement.

## How

In `packages/codemods/src/transforms/emotion-to-compiled/index.ts`:

- Added a constant `emotionUnsupportedTypeNames = ['Interpolation', 'CSSProperties']`.
- Added `replaceUnsupportedEmotionTypes(j, collection, importPath)` which, for each unsupported type:
    1. resolves the local alias from the import declaration (so `import { Interpolation as Int }` still works),
    2. finds every `TSTypeReference` whose `typeName` matches that local alias,
    3. replaces each match with `j.tsAnyKeyword()` and attaches a `TODO(@compiled/react codemod): …` block comment to the enclosing statement using the existing `addCommentBefore` helper,
    4. strips the specifier from the emotion import; removes the import entirely if no specifiers remain.
- Calls it before `convertMixedImportToNamedImport` for both `@emotion/core` and `@emotion/react` so the downstream import-rewriting steps see a clean set of specifiers.

### Mixed value + unsupported type — emotion import shrinks instead of being left behind

```ts
// Before
import { css, Interpolation } from '@emotion/core';

const styles: Interpolation = css`color: red`;
```

```ts
// After

/* TODO(@compiled/react codemod): Compiled does not provide an equivalent for "Interpolation" from "@emotion/core". This type has been replaced with `any` — replace it with a Compiled-compatible alternative when migrating. */
const styles: any = `color: red`;
```

### Type-only emotion import — declaration removed entirely

```ts
// Before
import { CSSProperties } from '@emotion/react';

const sx: CSSProperties = { color: 'red' };
```

```ts
// After
/* TODO(@compiled/react codemod): Compiled does not provide an equivalent for "CSSProperties" from "@emotion/react". This type has been replaced with `any` — replace it with a Compiled-compatible alternative when migrating. */
const sx: any = { color: 'red' };
```

### Aliased import — local alias is followed

```ts
// Before
import { Interpolation as Int } from '@emotion/core';

const styles: Int = { color: 'red' };
```

```ts
// After
/* TODO(@compiled/react codemod): Compiled does not provide an equivalent for "Interpolation" from "@emotion/core". This type has been replaced with `any` — replace it with a Compiled-compatible alternative when migrating. */
const styles: any = { color: 'red' };
```

### Unused unsupported import — declaration dropped silently

```ts
// Before
import { Interpolation } from '@emotion/core';
```

```ts
// After
(empty)
```

## Tests

Four new `defineInlineTest` cases in `packages/codemods/src/transforms/emotion-to-compiled/__tests__/transformer.test.ts` cover the four shapes above.

- `yarn jest packages/codemods/` — 84/84 pass (0 regressions across both emotion-to-compiled and styled-components-to-compiled).
- `yarn build` — clean.
- `yarn prettier --check` — clean.

## Scope notes

- Only the named-import form is handled. Namespace imports (`import * as E from '@emotion/core'; const x: E.Interpolation = …`) are not addressed — they're rarer in real migrations and worth a separate change so the diff stays reviewable.
- `'Interpolation'` and `'CSSProperties'` are the two cases called out in the issue. Easy to extend the constant if more unsupported types come up.
- Changeset is `minor` for `@compiled/codemods` (new codemod behavior, no breaking change to existing migrations).

## Changeset

`.changeset/emotion-codemod-unsupported-types.md` — `minor` bump for `@compiled/codemods`.

### DIFF
--- a/.changeset/emotion-codemod-unsupported-types.md
+++ b/.changeset/emotion-codemod-unsupported-types.md
@@ -1,0 +1,32 @@
+---
+'@compiled/codemods': minor
+---
+
+The `emotion-to-compiled` codemod now handles `Interpolation` and `CSSProperties` imports from `@emotion/core` / `@emotion/react`.
+
+Previously these type-only imports were silently left behind so the migrated file still pulled in `@emotion/*` (and so wouldn't compile cleanly without manual cleanup). The codemod now:
+
+- replaces every usage of the unsupported type with `any`,
+- adds a `TODO(@compiled/react codemod): …` block comment above the enclosing statement so the developer can pick a Compiled-compatible alternative,
+- removes the now-unused import specifier, and removes the emotion import entirely if no other specifiers remain.
+
+Closes #898.
+
+**Example**
+
+```ts
+// Before
+import { css, Interpolation } from '@emotion/core';
+
+const styles: Interpolation = css`
+  color: red;
+`;
+```
+
+```ts
+// After
+import { css } from '@compiled/react';
+
+/* TODO(@compiled/react codemod): Compiled does not provide an equivalent for "Interpolation" from "@emotion/core". This type has been replaced with `any` — replace it with a Compiled-compatible alternative when migrating. */
+const styles: any = `color: red`;
+```
--- a/packages/codemods/src/transforms/emotion-to-compiled/__tests__/transformer.test.ts
+++ b/packages/codemods/src/transforms/emotion-to-compiled/__tests__/transformer.test.ts
@@ -594,4 +594,62 @@ describe('emotion-to-compiled transformer', () => {
     "console.log('Bring back Netscape');\nimport { styled } from '@compiled/react';",
     'it should use the program visitor from the plugin'
   );
+
+  defineInlineTest(
+    { default: transformer, parser: 'tsx' },
+    { plugins: [] },
+    `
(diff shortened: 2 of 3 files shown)


leeguooooo:
Closing on my own — see #1899 for context.

PR #1899, opened 2026-05-19
feat(react): emit dev warning when ax() receives a non-string input

### LINKED ISSUE #1495: Improve error message when using css attribute without css(...) function call
This produces a runtime error:

```css
import { css, styled } from '@compiled/react';

const styles = css({ color: 'yellow' });

const MyComponent = styled.div({});
const MyOtherComponent = ({ someParameter, children }) => <MyComponent css={someParameter || styles}>{children}</MyComponent>;

<MyOtherComponent someParameter={{ color: 'yellow', fontSize: '50px' }}>hello worilnd</MyOtherComponent>
```

```
Uncaught TypeError: classNames[0].indexOf is not a function
```

Screenshot 2023-08-23 at 11 41 07

This is because `{ color: 'yellow', fontSize: '50px' }` should be wrapped in a `css( ... )` function call. The correct(er) syntax, which does work with Compiled, would be something like this:
```tsx
<MyOtherComponent someParameter={css({ color: 'yellow', fontSize: '50px' })}>hello worilnd</MyOtherComponent>
```

We should change the error message so that it should be more helpful and direct them to using the correct syntax. For example, we can tell the user that they are missing a css(...) function call around the `{ color: 'yellow', fontSize: '50px' }` object.

This would probably not be practical to add to `@compiled/eslint-plugin`, because ESLint runs on each file individually and thus cannot detect usages of `MyOtherComponent` across multiple files. Perhaps we can add this to `@compiled/babel-plugin` so that we get a build time error, not a runtime error

leeguooooo:
## What

Adds a development-only check at the top of `ax()` that detects when the runtime is given anything other than the supported `string | undefined | null | false | ''` shape, and emits an actionable warning before the existing code path crashes.

Closes #1495.

## Why

The reproducer in #1495 — passing a raw style object directly to a `css` prop, e.g. `<MyComponent css={someParameter || styles}>` where `someParameter` is `{ color: 'yellow', fontSize: '50px' }` — surfaces today as a cryptic crash:

```
TypeError: classNames[0].includes is not a function
```

That error message doesn't name the bad value, doesn't say where in the JSX tree the value came from, and gives the developer no hint of the actual fix (wrap the object with `css({...})` so Compiled can transform it at build time). The reporter explicitly notes that ESLint can't catch this cross-file, so a runtime/build-time signal is the practical place to fix it.

This PR keeps the existing runtime behavior (the crash still happens) but emits a clear warning *first* so the developer has somewhere to start from:

```
@compiled/react/runtime - DEV WARNING

ax() received an object (`{"color":"red"}`) at index 0.
  Compiled expects precompiled className strings here. If you passed a raw style object to a `css` prop or styled component — for example `<MyComponent css={{ color: 'red' }} />` — wrap it with `css({...})` so Compiled can transform it at build time:
      <MyComponent css={css({ color: 'red' })} />
```

## How

In `packages/react/src/runtime/dev-warnings.ts`:

- Added `analyzeAxInputInDev(classNames)` that walks the input, ignores the supported falsy/string values, and warns via the existing `warn` helper for anything else. Deduped by `typeof value` so consoles don't get spammed on every render.
- Reused the existing `hasWarned` cache + the same Compiled ASCII-art banner as `analyzeCssInDev`.

In `packages/react/src/runtime/ax.ts`:

- Imported `analyzeAxInputInDev` and called it at the very top of `ax()`, gated behind `process.env.NODE_ENV === 'development'`. The gate matches the pattern already used by `analyzeCssInDev` in `style.tsx`, and lets the consumer's bundler tree-shake the entire check + import in production builds.

## Tests

Four new cases in `packages/react/src/runtime/__tests__/ax.test.ts`:

- warns when an object is passed where a className string is expected (and the message contains the actionable hint),
- does not warn for valid inputs (`'_abc'`, `undefined`, `null`, `false`, `''`),
- dedupes — calling `ax()` repeatedly with bad inputs of the same type only warns once,
- does not warn when `process.env.NODE_ENV === 'production'`.

Each test uses `jest.isolateModules` to give `ax` (and therefore the dev-warnings dedup state) a fresh module scope, so tests don't leak `hasWarned` state between each other.

- `yarn jest packages/react/` — 272/272 pass (0 regressions).
- `yarn build` — clean.
- `yarn prettier --check` / `yarn lint` — clean.

## Scope notes

- Only `ax` is checked; `ac` has a similar surface and could get the same treatment in a follow-up if you want, but kept this PR small to make review easier.
- The check fires for any non-string non-supported-falsy value (object, number, function, …) rather than just objects — same actionable hint helps in all of them.
- Changeset: `patch` for `@compiled/react`.

### DIFF
--- a/.changeset/ax-dev-warning-non-string-input.md
+++ b/.changeset/ax-dev-warning-non-string-input.md
@@ -1,0 +1,26 @@
+---
+'@compiled/react': patch
+---
+
+`ax()` now emits a development-only warning when it receives a non-string value (typically an object) instead of a precompiled className.
+
+The previous failure mode was a cryptic runtime crash — `classNames[0].includes is not a function` — that gave the developer no hint about where in their JSX the bad value was coming from. The most common cause is passing a raw style object to a `css` prop or a styled component without wrapping it with `css({...})` first, for example:
+
+```jsx
+// This makes a raw object flow into ax() at runtime.
+<MyComponent css={someCondition && { color: 'red' }} />
+```
+
+The warning now points at the actual problem and the fix:
+
+```
+@compiled/react/runtime - DEV WARNING
+
+ax() received an object (`{"color":"red"}`) at index 0.
+  Compiled expects precompiled className strings here. If you passed a raw style object to a `css` prop or styled component — for example `<MyComponent css={{ color: 'red' }} />` — wrap it with `css({...})` so Compiled can transform it at build time:
+      <MyComponent css={css({ color: 'red' })} />
+```
+
+The warning is dev-only and deduped per type, so production bundles pay no cost (the `process.env.NODE_ENV === 'development'` guard is tree-shaken) and development consoles aren't spammed on every render.
+
+Closes #1495.
--- a/packages/react/src/runtime/__tests__/ax.test.ts
+++ b/packages/react/src/runtime/__tests__/ax.test.ts
@@ -55,4 +55,82 @@ describe('ax', () => {
   ])('%s', (_, params, expected) => {
     expect(ax(params)).toEqual(expected);
   });
+
+  describe('dev warning for non-string inputs', () => {
+    type AxFn = (input: unknown[]) => string | undefined;
+
+    const originalNodeEnv = process.env.NODE_ENV;
+    let consoleErrorSpy: jest.SpyInstance;
+
+    beforeEach(() => {
+      process.env.NODE_ENV = 'development';
+      consoleErrorSpy = jest.spyOn(console, 'error').mockImplementation(() => {});
+    });
+
+    afterEach(() => {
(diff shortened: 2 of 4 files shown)


dddlr:
Hi, please don't spam our repo with automated PRs thanks

leeguooooo:
Apologies, that's fair feedback. The three-in-a-day cadence wasn't appropriate for a first-time contributor here. Closing all three myself.

leeguooooo:
Closing on my own — see comment above.

liamqma:
Hey @leeguooooo ,
Just out of interest, what website do you use Compiled CSS-in-JS for?

leeguooooo:
Hey @liamqma — honestly, not using it on a website, was just contributing out of interest. sorry to you and @dddlr for the noise.

PR #1900, opened 2026-05-27
[SPIKE] dynamic hash increase

xtan-atlas:
### What is this change?

Makes the atomic class group hash length in atomicify-rules adaptive instead of fixed at 4 characters: it starts at 5 chars and grows by one character per class only when the compiler detects a collision with an already-emitted class. The runtime ax() is updated to compute group keys by stripping the (still fixed) value-hash suffix, so variable-length groups work transparently.
Net result: class names grow from _GGGGVVVV (9 chars) to _GGGGGVVVV (10 chars), and group hash collisions are eliminated deterministically.

The current hash(...).slice(0, 4) produces a meaningful collision rate even in modestly-sized CSS bundles. When two different CSS axes (different selector + property combinations) end up with the same 4-char group hash, the runtime ax() function silently drops one of them — causing styles to disappear at runtime with no error.
We measured this on @atlaskit/editor-core's EditorContentContainer-compiled.compiled.css:

•  3,453 atomic rules in this single file
•  46+ harmful group collisions before this fix, across variants like tableSharedStyle, blockMarksStyles, mediaStyles, syncBlockStylesBase, etc.
•  Each collision = a CSS property silently dropped at runtime
The base-36 slice(0, 4) also has a leading-digit bias (the first character is '1' ~50% of the time, because every 7-character base-36 string starts with '1'), which further shrinks the effective hash space. The dynamic-length approach sidesteps the bias entirely and gives a deterministic guarantee rather than a statistical one.

(Optional.)

---

### PR checklist

Don't delete me!

I have...

- [ x ] Updated or added applicable tests

### DIFF
--- a/packages/react/src/runtime/__tests__/ax.test.ts
+++ b/packages/react/src/runtime/__tests__/ax.test.ts
@@ -30,10 +30,52 @@ describe('ax', () => {
       ['_aaaabbbb _aaaacccc _aaaadddd _aaaaeeee'],
       '_aaaaeeee',
     ],
+    // ── tests for the 10-char default class name shape: `_` + 5 group + 4 value ──
     [
-      'should ensure the last atomic declaration of many multi groups with short class name wins',
+      'should dedup 10-char classes with the same 5-char group (single classes)',
+      ['_aaaaabbbb', '_aaaaacccc'],
+      '_aaaaacccc',
+    ],
+    [
+      'should dedup 10-char classes with the same 5-char group (multi class)',
+      ['_aaaaabbbb _aaaaacccc _aaaaadddd _aaaaaeeee'],
+      '_aaaaaeeee',
+    ],
+    [
+      'should keep 10-char classes from different 5-char groups',
+      ['_aaaaabbbb', '_bbbbbcccc'],
+      '_aaaaabbbb _bbbbbcccc',
+    ],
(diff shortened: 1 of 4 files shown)


packages/react/src/runtime/__tests__/ax.test.ts
```diff
@@ -30,10 +30,52 @@ describe('ax', () => {
...
+      ['_aaaaabbbb', '_aaaabbbb', '_aaaaacccc'],
+      '_aaaaacccc _aaaabbbb',
+    ],
+    [
+      // Under the dynamic group-length scheme, class names of different total length
+      // are treated as different groups (the group is everything except the last 4 chars).
+      // Here `_aaaaaaa` (8 chars) has group `_aaaa`, while `_aaaabbbb` (9 chars) has
+      // group `_aaaab` — they intentionally do not deduplicate.
+      'should treat classes of different total length as different groups',
       ['_aaaabbbb', '_aaaaaaa', '_ddddbbb', '_ddddcccc'],
```
xtan-atlas, line 77: There's one test had to be updated. It seems it allows shorter classnames like 8 chars, assuming we have hard coded 4 digit group hash.

Dynamic group hash means we can't use hard coded 4 digit group hash. So for class like `._abc1234` => group: `abc`. In previous implementation, the group would be `abc1`.

PR #1902, opened 2026-05-29
[SPIKE] smart ax with dynamic length and prefix family matching

xtan-atlas:
### What is this change?

Make `ax` supports longer classes like `_ggggggvvvv`, and can detect group hash is `gggggg`.
Support overrides legacy short classnames: `ax('_ggggvvvv', '_ggggttvvvv') => '_ggggttvvvv'`

The 4 char group hash is too short, cause class collision and drop styles silently. 

---

### PR checklist

Don't delete me!

I have...

- [ x ] Updated or added applicable tests

### DIFF
--- a/packages/react/src/runtime/__tests__/ax.test.ts
+++ b/packages/react/src/runtime/__tests__/ax.test.ts
@@ -51,9 +51,9 @@ describe('ax', () => {
         '_aaaaeeee',
       ],
       [
-        'should treat non-standard shorter atomic-looking class names consistently with suffix parsing',
+        'should treat non-standard shorter atomic-looking class names as independent groups',
         ['_aaaabbbb', '_aaaaaaa', '_ddddbbb', '_ddddcccc'],
-        '_aaaaaaa _ddddcccc',
+        '_aaaabbbb _aaaaaaa _ddddbbb _ddddcccc',
       ],
       [
         'should not remove any atomic declarations if there are no duplicate groups',
@@ -81,6 +81,11 @@ describe('ax', () => {
         'should keep variable-length classes with different groups',
         ['_aaaa12bbbb', '_bbbb12cccc'],
         '_aaaa12bbbb _bbbb12cccc',
+      ],
+      [
+        'should keep variable-length groups that share a legacy prefix but are not in the same prefix family',
+        ['_aabbcc1234', '_aabbdd1234'],
+        '_aabbcc1234 _aabbdd1234',
       ],
       [
         'should allow a longer group to override its shorter prefix family',
--- a/packages/react/src/runtime/__tests__/ax.test.ts
+++ b/packages/react/src/runtime/__tests__/ax.test.ts
@@ -53,7 +53,7 @@ describe('ax', () => {
       [
         'should treat non-standard shorter atomic-looking class names consistently with suffix parsing',
         ['_aaaabbbb', '_aaaaaaa', '_ddddbbb', '_ddddcccc'],
-        '_aaaabbbb _aaaaaaa _ddddcccc',
+        '_aaaaaaa _ddddcccc',
       ],
       [
         'should not remove any atomic declarations if there are no duplicate groups',
@@ -87,6 +87,16 @@ describe('ax', () => {
         ['_aaaabbbb', '_aaaa12cccc'],
         '_aaaa12cccc',
       ],
+      [
+        'should allow a shorter group to override its longer prefix family when order is flipped',
+        ['_aaaa12cccc', '_aaaabbbb'],
+        '_aaaabbbb',
+      ],
+      [
+        'should keep the most recent declaration when multiple prefix-family groups are provided',
+        ['_aaaabbbb', '_aaaa12cccc', '_aaaa1234dddd'],
+        '_aaaa1234dddd',
+      ],
     ])('%s', (_, params, expected) => {
       expect(ax(params)).toEqual(expected);
     });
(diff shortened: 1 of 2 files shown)


packages/react/src/runtime/__tests__/ax.test.ts
```diff
@@ -3,56 +3,114 @@ import ax from '../ax';
...
+        ['_aaaabbbb _aaaacccc'],
+        '_aaaacccc',
+      ],
+      [
+        'should ensure the last atomic declaration of many multi groups wins',
+        ['_aaaabbbb _aaaacccc _aaaadddd _aaaaeeee'],
+        '_aaaaeeee',
+      ],
+      [
+        'should treat non-standard shorter atomic-looking class names consistently with suffix parsing',
```
xtan-atlas: This was failing, because original test was accepting 8 char class. It works before due to it treat first 4 char as group, but after we change logic to `slice(className.length - valueLength)`, the group hash for 8 char string would be first 3 char...

packages/react/src/runtime/ax.ts
```diff
@@ -44,6 +48,14 @@ export default function ax(classNames: (string | undefined | null | false)[]): s
   // Would be happy to move to `Map` if it proved to be better under real conditions.
   const map: Record<string, string> = {};
 
+  const removePrefixFamilyEntries = (key: string) => {
+    for (const existingKey in map) {
+      if (key.startsWith(existingKey)) {
+        delete map[existingKey];
+      }
+    }
```
kylorhall-atlassian: I'm a bit confused on what this does. So let's say we have two: `ax(['_abcd1234', '_abcde2345']);` (old and new hashes)
 - `_abcd1234` runs, sets `map = { '_abcd': '_abcd1234' };`
 - `_abcd1234` runs, deletes that map entry, sets `map = { '_abcde': '_abcde2345' };`

That makes sense, you want the second value with that hash…

But then let's invert it: `ax(['_abcde1234', '_abcd2345']);` don't you get `map = { '_abcd': '_abcd1234', '_abcde': '_abcde2345' };`?

Is that a problem?

packages/react/src/runtime/ax.ts
```diff
@@ -68,7 +80,15 @@ export default function ax(classNames: (string | undefined | null | false)[]): s
...
        * */
-      const key = className.startsWith('_') ? className.slice(0, ATOMIC_GROUP_LENGTH) : className;
+      const isAtomic = className.startsWith('_');
+      const key = isAtomic
+        ? className.slice(0, className.length - ATOMIC_VALUE_HASH_LENGTH)
+        : className;
+
+      if (isAtomic) {
+        removePrefixFamilyEntries(key);
+      }
```
kylorhall-atlassian, line 174: I think a comment (maybe in above jsdoc) explaining this part is helpful, especially RE: my comment above as I think this might get more complex…

packages/react/src/runtime/__tests__/ax.test.ts
```diff
@@ -3,56 +3,114 @@ import ax from '../ax';
...
+    it.each([
+      [
+        'should dedup legacy and variable-length classes when they share the same exact group',
+        ['_aaaabbbb', '_aaaacccc'],
+        '_aaaacccc',
+      ],
+      [
+        'should allow a variable-length longer group to override its legacy prefix family',
+        ['_aaaabbbb', '_aaaa12cccc'],
+        '_aaaa12cccc',
```
kylorhall-atlassian, line 120: Make sure we have a flipped version of this test as well, or like all three in various orders, whatever seems sensible.

kylorhall-atlassian (changes requested):

packages/react/src/runtime/__tests__/ax.test.ts
xtan-atlas: added more flipped tests

packages/react/src/runtime/ax.ts
```diff
@@ -43,6 +48,94 @@ export default function ax(classNames: (string | undefined | null | false)[]): s
...
+   *
+   * Same prefix-family semantics as the bucketed version, but implemented as a
+   * direct scan over the full result map. Easier to read, but per-insert cost
+   * grows with the total map size rather than the small same-prefix bucket.
+   *
+   * Useful for comparison benchmarks or as a fallback if the bucketed version
+   * ever needs to be disabled.
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const setAtomicClassInGroupMapSimpleScan = (groupKey: string, className: string) => {
```
xtan-atlas, line 117: this is much easier to understand but a bit slower when ax large amount of classes, like Editor styles...

packages/react/src/runtime/ax.ts
```diff
@@ -43,6 +48,94 @@ export default function ax(classNames: (string | undefined | null | false)[]): s
...
+   *
+   * To support both formats together, we bucket active group keys by their
+   * legacy 5-character prefix. Only keys in the same bucket can possibly be in
+   * the same prefix family, so we only need to scan a small array per insert
+   * instead of the full result map.
+   *
+   * For runtime performance, the bucket is mutated in place using swap-and-pop
+   * so we avoid allocating a fresh array on every insert.
+   */
+  const setAtomicClassInGroupMapBucketed = (groupKey: string, className: string) => {
```
xtan-atlas, line 76: This is hard to understand, but more performant. But still slower than without legacy group hash support. It's like slower 20-40%. So backwards support do come with cost.

PR #1903, opened 2026-06-04
feat: Introduce new hash strategy, and make ax supports longer hash

xtan-atlas:
### What is this change?

Introduce a few hash strategy for compiled css to resolve hash collision issue.

The hash strategy can be opt-in via cssMap config option.

```
cssMap({
  ...
}, {
 // @ts-expect-error, very dangerous, only used internally with extreme caution!
  hashStrategy: 'max'
}
```
The hashStrategy `max`, will use base62 instead of base32, and also use the full hash for classes.
The hashStrategy `enhanced`, will use base62, and also fix slice issue.
The hashStrategy `default`, the default strategy which is currently used. Used base32, and slice(0,4) for group hash.

The hashStrategy option is not officially supported, and will only be used internally, as it's very dangerous and can lead to bugs. The known limitation is that the `ax` won't support classnames created by different hash strategy.

Current hash strategy for classes creates collisions, caused styles silently dropping issue.

(Optional.)

### Notes

won't update website, as the hashStrategy is not for public use.

---

### PR checklist

Don't delete me!

I have...


packages/react/src/css-map/__tests__/types.typecheck.ts
```diff
@@ -0,0 +1,11 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { cssMap } from '@compiled/react';
+
+const styles = cssMap({ danger: { color: 'red' } });
+
+// Verify hashStrategy is NOT part of the public type signature.
+// If this @ts-expect-error ever becomes unnecessary, hashStrategy has been accidentally re-exposed.
+// @ts-expect-error - hashStrategy is experimental and intentionally omitted from the public API.
+const stylesWithHashStrategy = cssMap({ danger: { color: 'red' } }, { hashStrategy: 'max' });
```
xtan-atlas, line 9: If I remove `@ts-expect-error`, users would see type error. It's intentionally prevent external users to use it.

Image

.changeset/grouped-atomic-rules.md
```diff
@@ -0,0 +1,19 @@
+---
+'@compiled/babel-plugin': major
```
kylorhall-atlassian: First major? 😄 should we cut a major for related packages as well @JakeLane…?  Like Webpack, Parcel, etc., that use babel, should they all bump to a major at once?

.changeset/hash-strategy-experimental.md
```diff
@@ -0,0 +1,12 @@
...
+'@compiled/css': patch
+'@compiled/react': minor
+---
+
+add experimental `hashStrategy` option for `cssMap`, only used for internal use.
+make `ax` supports longer group hash.
+
+```
+
+```
```
kylorhall-atlassian: I think this needs more context akin to the other bump.

packages/babel-plugin/src/css-map/index.ts
```diff
@@ -9,6 +9,15 @@ import { transformCssItems } from '../utils/transform-css-items';
...
 import { mergeExtendedSelectorsIntoProperties } from './process-selectors';
 
+/**
+ * @experimental Options for cssMap are not part of the public API and may change without notice.
+ * The `hashStrategy` option is intentionally omitted from the TypeScript type signature of cssMap.
+ * Internal consumers can opt in using `@ts-ignore`, it's highly risky.
+ */
+const KNOWN_OPTIONS = ['hashStrategy'];
+
+const VALID_HASH_STRATEGIES = ['default', 'enhanced', 'max'];
```
kylorhall-atlassian: Should this be typechecked somewhere against another type?

`VALID_HASH_STATEGIES: HashStrategy[] = […]` (or more of a proper type to make it an explicit same-length array of unions, not sure the syntax to do that.

I do not care if upstream/downstream this might fail.

packages/babel-plugin/src/utils/transform-css-items.ts
```diff
@@ -7,6 +7,10 @@ import { compressClassNamesForRuntime } from './compress-class-names-for-runtime
 import { getItemCss } from './css-builders';
 import type { CssItem } from './types';
 
+type TransformOptions = {
+  hashStrategy?: string;
```
kylorhall-atlassian: Should this be explicit with the type `hashStrategy?: 'default'|'enhanced'|'max';` (whatever it is)?

packages/react/src/css-map/__tests__/types.typecheck.ts
kylorhall-atlassian: Is this a test or a fixture?  Feels a bit odd.

packages/react/src/css-map/__tests__/types.typecheck.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export { styles, stylesWithHashStrategy };
```
kylorhall-atlassian, line 11: Exporting `styles` seems incorrect, this "test" is a bit weird.

packages/react/src/runtime/__tests__/ax.test.ts
kylorhall-atlassian, line 1: Wait don't we need a lot more tests for `ax` with the new variable hash lengths?

I think we should document how each of these hash strategies conflict with each other, even if it's like a known failure.

Eg. `max` vs. `default` will NOT get the last atomic declaration.

.changeset/grouped-atomic-rules.md
```diff
@@ -0,0 +1,19 @@
...
+
+Add an opt-in `group` option to `cssMap` that emits one shared atomic class per nested selector group within a variant, instead of one atomic class per CSS property. This reduces the number of atomic classes generated for variants with many nested rules under the same selector, which in turn lowers hash-collision pressure and reduces CSS bundle size for style-heavy components.
+
+```ts
+const styles = cssMap(
+  {
+    one: { '> span': { paddingBottom: '2px' } },
+    two: { outlineWidth: '2px' },
+  },
+  { group: true }
```
kylorhall-atlassian: This PR doesn't do anything with grouping, this changeset seems incorrect.

stories/css-map-grouping.tsx
kylorhall-atlassian, line 1: This seems unrelated to the PR.

kylorhall-atlassian (changes requested):

packages/react/src/css-map/__tests__/types.typecheck.ts
xtan-atlas: a type test. This is the test without me removing that @ts-expect-error
Screenshot 2026-06-05 at 12 00 15 pm

packages/babel-plugin/src/utils/transform-css-items.ts
xtan-atlas: ah, missed this. I don't want to add the type into a shared repo like @compiled/utils, as it's just for editor internal use. 

I might just add a internal type shared only in babel-plugin.

packages/react/src/css-map/__tests__/types.typecheck.ts
xtan-atlas: probably not needed, suggested by AI haha

.changeset/grouped-atomic-rules.md
ssysoev-atl: IMHO it's a major for `compiled/react` and minor for a plugin, plugin doesn't break backwards compatibility, the cssMap does :) but it's tricky...

packages/react/src/runtime/__tests__/ax.test.ts
```diff
@@ -31,27 +45,142 @@ describe('ax', () => {
       '_aaaaeeee',
     ],
     [
-      'should ensure the last atomic declaration of many multi groups with short class name wins',
```
xtan-atlas, line 34: The changes in `ax` will stop supporting mix of shorter and longer classes. Because previously we use first 4 chars (exclude _) as group hash, now we use full length - 4(value hash length) as group hash.

Not sure if we have different length of classes in production? If yes, it'd be major version bump, otherwise minor is okay.

packages/react/src/css-map/__tests__/types.typecheck.ts
xtan-atlas: the export is used to tell typescript it's a module not script. So it's still needed. I renamed the file to typecheck.ts so it's less confusing.

packages/react/src/runtime/__tests__/ax.test.ts
xtan-atlas: added more tests and special describe clause to cover the limitations.

.changeset/grouped-atomic-rules.md
xtan-atlas: thanks, cleaned it up.

stories/css-map-grouping.tsx
xtan-atlas: thanks, I've cleaned it up

packages/babel-plugin/src/css-map/index.ts
xtan-atlas: improved types now, should be good I think.

.changeset/hash-strategy-experimental.md
xtan-atlas: enriched the changeset with more info now

packages/css/src/plugins/atomicify-rules.ts
xtan-atlas, line 23: It seems caused by the regex, but it's pre-existing from master, not related to my change. The alert shows up because this file has been touched.

So I think it'd be fine to dismiss that alert.

Image

packages/eslint-plugin/src/rules/no-css-map-options/index.ts
kylorhall-atlassian, line 1: To turn it on globally, I think we just add it to `@atlaskit/ui-styling-standard/recommended` like we do some other rules. Once it's in, let me know if you need help.  Shouldn't fail anything, so easy rule to rollout!

packages/eslint-plugin/src/rules/no-css-map-options/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  meta: {
+    docs: {
+      recommended: true,
+      description:
+        'Disallows passing options to `cssMap`. Configuration options like `hashStrategy` are experimental and for internal use only.',
+      url: 'https://github.com/atlassian-labs/compiled/tree/master/packages/eslint-plugin/src/rules/no-css-map-options',
+    },
+    messages: {
+      noCssMapOptions:
+        'cssMap does not accept options. Configuration options like hashStrategy are experimental and for internal use only, only use with extream caution.',
```
kylorhall-atlassian: Typo, not important:
```suggestion
        'cssMap does not accept options. Configuration options like hashStrategy are experimental and for internal use only, only use with extreme caution.',
```

packages/eslint-plugin/src/index.ts
```diff
@@ -26,6 +28,8 @@ export const rules = {
   'no-emotion-css': noEmotionCssRule,
   'no-exported-css': noExportedCssRule,
   'no-exported-keyframes': noExportedKeyframesRule,
+  'no-css-map-options': noCssMapOptionsRule,
+  'no-mixed-css-map-options': noMixedCssMapOptionsRule,
```
kylorhall-atlassian: I don't think you added this.


```suggestion
```

packages/eslint-plugin/src/index.ts
```diff
@@ -11,6 +12,7 @@ import { noExportedKeyframesRule } from './rules/no-exported-keyframes';
 import { noInvalidCssMapRule } from './rules/no-invalid-css-map';
 import { noJavaScriptXCSSRule } from './rules/no-js-xcss';
 import { noKeyframesTaggedTemplateExpressionRule } from './rules/no-keyframes-tagged-template-expression';
+import { noMixedCssMapOptionsRule } from './rules/no-mixed-css-map-options';
```
kylorhall-atlassian:
```suggestion
```

kylorhall-atlassian (dismissed):

.changeset/grouped-atomic-rules.md
kylorhall-atlassian: It might not be a `major` if it's "a new feature that's unavailable and blocked by typechecking. do not use" I think a `minor` to both are fine, but we need to bite the bullet one day…technically `0.x` minors are majors kinda sorta?

Heck we could call it `UNSAFE_DO_NOT_USE_hashStrategy` if we really watched—then it'd be easy to ratchet against too 😄

JakeLane: Definitely not major, minor is good imo

PR #1906, opened 2026-06-10
feat: non atomic option - cssMapScoped API ( experimental )

xtan-atlas:
## What is this change?

Introduces a new experimental `cssMapScoped` API that switches to **non-atomic output mode**.

When `cssMapScoped` is used, all CSS declarations for a variant are grouped under a single `cc-<hash>` class instead of being split into individual atomic `_xxx` classes.

```tsx
import { cssMapScoped } from '@compiled/react';
const styles = cssMapScoped({
  danger: { color: 'red', '&:hover': { color: 'darkred' } },
  success: { color: 'green' },
});

// Emits: .cc-7gkmyp { color: red }  .cc-7gkmyp:hover { color: darkred }
//        .cc-1j0429u { color: green }
// Instead of: ._syaz13q2 { color: red }  ._180h1b2c:hover { color: darkred }  etc.
```

The `cc-` prefix (no `_`) means `ax()` treats these as opaque plain classes and does not attempt to deduplicate them against atomic groups.

The class name is derived from `hash(filename + ':' + variableName + ':' + variantKey)` — computed cheaply in the babel plugin without hashing the full CSS content string.

If consumer want to override the styles, they need to override by using nested styles, just like how to override normal css in old days.

### Notes:
* The `cssMapScoped` is experimental, so not exported to be use in public for now. You have to create a shims declaration file to bypass the type error, as `cssMapScoped` not available from `@compiled/react`.
* A `no-css-map-scoped` eslint rule is also created to guard the use of this experimental feature.
* Checked website - but intentionally not add public docs as the feature is experimental and use for internal only.

## Why are we making this change?

The Editor team is migrating editor styles to Compiled CSS using a large `cssMap` object. With atomic output, a single cssMap with many variants and properties generates thousands of atomic classes, all applied to a single DOM element. This causes excessive layout recalculation and negatively impacts interactivity (INP). The `cssMapScoped` reduces the per-element class count from potentially thousands down to the number of variants actually applied (typically 3–10).

## How are we making this change?

---

### PR checklist

Don't delete me!

I have...


packages/css/src/plugins/sort-style-sheet.ts
```diff
@@ -93,26 +104,40 @@ export const sortAtomicStyleSheet = (config: {
...
+        ...nonAtomicStyles,
+        ...catchAll,
+        ...atomicRules,
+        ...atomicAtRules.map((atRule) => atRule.node),
+      ];
+
+      // Clear raws.before on the first node to prevent a leading newline in the
+      // output when non-atomic rules (which may have raws.before='\n' from being
+      // mid-string in the joined input) are moved to the front of the stylesheet.
+      if (root.first) {
```
xtan-atlas, line 138: There was an existing issue that after sort, the first line of style sheet could be `\n` -> empty line.
The fix is to remove the leading `\n`. This explains why the sort-xxx-xxx.test.ts for atomic styles also got affected. 

This is why lots of sort-xxx.test.ts got updated on snapshots, which removed the first empty line.

packages/babel-plugin/src/utils/css-map.ts
xtan-atlas, line 1: allowing error message working for both cssMap and cssMapScoped

packages/css/src/transform.ts
xtan-atlas, line 1: changes are making it allowing for both cssMap and cssMapScoped working

packages/css/src/plugins/at-rule-lists.ts
xtan-atlas, line 1: reused in both atomicify-rules and non-atomicify-rules

packages/css/src/plugins/non-atomicify-rules.ts
xtan-atlas, line 1: most critical part of the feature :)

JakeLane (dismissed):

packages/babel-plugin/src/css-map/index.ts
```diff
@@ -9,41 +10,91 @@ import { transformCssItems } from '../utils/transform-css-items';
...
+    ? property.key.name
+    : t.isStringLiteral(property.key)
+    ? property.key.value
+    : undefined;
+
+  if (variantKey === undefined) return undefined;
+
+  // Use the absolute filename for uniqueness per file. The filename is always
+  // the absolute path set by Babel.
+  const fileKey = state.filename ?? '';
```
ssysoev-atl: should this be a relative path?

ssysoev-atl: does it even exist?

packages/css/src/transform.ts
```diff
@@ -61,26 +97,51 @@ export const transformCss = (
...
+        ...(flattenMultipleSelectorsOption
+          ? [flattenMultipleSelectors(), discardDuplicates()]
+          : []),
+        ...(opts.increaseSpecificity ? [increaseSpecificity()] : []),
+        sortStyleSheet({
+          sortAtRulesEnabled: opts.sortAtRules,
+          sortShorthandEnabled: opts.sortShorthand,
+        }),
+        ...tailPlugins,
+      ]).process(css, { from: undefined }).css;
```
ssysoev-atl: need to restore the comment, on why nothing is returned, and result is unsued...

packages/react/src/runtime/sheet.ts
```diff
@@ -201,3 +201,27 @@ export default function insertRule(css: string, opts: StyleSheetOpts): void {
...
+  const style = lazyAddStyleBucketToContainer('', opts);
+
+  if (process.env.NODE_ENV === 'production') {
+    const sheet = style.sheet as CSSStyleSheet;
+
+    try {
+      sheet.insertRule(css, sheet.cssRules.length);
+    } catch {}
+  } else {
+    style.appendChild(document.createTextNode(css));
```
ssysoev-atl, line 225: we probably should not add this path... as it's slow...

packages/react/src/runtime/style.tsx
```diff
@@ -2,11 +2,25 @@ import React from 'react';
...
 import { useCache } from './style-cache.js';
 import { StyleContainerProvider, useStyleContainer } from './style-container.js';
 import type { Bucket, StyleSheetOpts } from './types.js';
 
+/**
+ * Returns true if the CSS sheet string is a non-atomic rule generated by
+ * `cssMapScoped`. Non-atomic rules use the `cc-` class prefix
+ * (i.e. the sheet starts with `.cc-`).
+ */
+const isNonAtomicSheet = (sheet: string): boolean =>
```
ssysoev-atl: startsWith?

ssysoev-atl (commented):
did a first pass, will have another look later today...

packages/react/src/runtime/style.tsx
xtan-atlas: I think the existing pattern using charCodeAt was for perf optimization. For our cssMapScoped, much less calls needed for this function, so startWith is fine. I will update this.

packages/react/src/runtime/sheet.ts
xtan-atlas: ah yeah, I'll remove appendChild for non atomic

packages/babel-plugin/src/css-map/index.ts
xtan-atlas: yeah filename is existed. verified in afm. 
I tried with relative path but had trouble with extraction tests due to some test setup I think. I will look into this deeper again.

packages/css/src/transform.ts
xtan-atlas: added back the comment

packages/babel-plugin/src/css-map/index.ts
xtan-atlas: changed to relative filename, and verified fine in AFM.

packages/react/src/runtime/sheet.ts
xtan-atlas: Just found out I can't use sheet.insertRule because I joined the rules under each variant together to reduce byte size, `const _N` variable per variant, to improve perf:  `const sheets = css.sheets.length > 1 ? [css.sheets.join('')] : css.sheets;` from `packages/babel-plugin/src/utils/transform-css-items.ts`.

So revised the solution to use appendData approach, suggested by AI. tested fine. 
The appendChild only called once, subsequent calls using appendData. should be quite fast.

ssysoev-atl (approved):

JakeLane (approved):

PR #1908, opened 2026-06-15
fix jsx-local-namespace type issue

xtan-atlas:
### What is this change?

Fix `LibraryManagedAttributes` to restore React 18 compatibility and `defaultProps` stripping.

1. `key` no longer accepted `null`, contradicting `React.Attributes` which defines `key?: Key | null | undefined`. This caused `TS2322` errors in React 18 codebases when passing a nullable key (e.g. `key={item?.id ?? null}`).

2. Using `P` directly instead of `React.JSX.LibraryManagedAttributes<C, P>` broke `defaultProps` stripping for class components — props covered by `defaultProps` were no longer treated as optional at the call site (`TS2740`).

The fix delegates to `React.JSX.LibraryManagedAttributes<C, P>` (which correctly strips `defaultProps`) and explicitly restores `key?: React.Key | null` to align with `React.Attributes`. This avoids the circular reference issue since `React.JSX.LibraryManagedAttributes` does not route through the global `JSX` namespace.

Also fixes the `typecheck:react18` test which used `toMatchTypeOf` (which does not catch type narrowing regressions) — replaced with `toEqualTypeOf` on `Managed['key']`.

Verified typecheck tests via:
* `yarn workspace @compiled/react typecheck:react18`
* `yarn workspace @compiled/react typecheck:react19`

The regression in type `LibraryManagedAttributes` would cause type errors when consumers bump to ^0.21.1

Give the full context for your change here.

(Optional.)

---

### PR checklist

Don't delete me!

I have...

- [/] Updated or added applicable tests
- [/] Updated the documentation in `website/`

### DIFF
--- a/.changeset/fix-library-managed-attributes-react18.md
+++ b/.changeset/fix-library-managed-attributes-react18.md
@@ -1,0 +1,15 @@
+---
+'@compiled/react': patch
+---
+
+Fix `LibraryManagedAttributes` to restore React 18 compatibility and `defaultProps` stripping.
+
+Commit `9a805f7` introduced two regressions:
+
+1. `key` no longer accepted `null`, contradicting `React.Attributes` which defines `key?: Key | null | undefined`. This caused `TS2322` errors in React 18 codebases when passing a nullable key (e.g. `key={item?.id ?? null}`).
+
+2. Using `P` directly instead of `React.JSX.LibraryManagedAttributes<C, P>` broke `defaultProps` stripping for class components — props covered by `defaultProps` were no longer treated as optional at the call site (`TS2740`).
+
+The fix delegates to `React.JSX.LibraryManagedAttributes<C, P>` (which correctly strips `defaultProps`) and explicitly restores `key?: React.Key | null` to align with `React.Attributes`. This avoids the circular reference issue since `React.JSX.LibraryManagedAttributes` does not route through the global `JSX` namespace.
+
+Also fixes the `typecheck:react18` test which used `toMatchTypeOf` (which does not catch type narrowing regressions) — replaced with `toEqualTypeOf` on `Managed['key']`.
--- a/.changeset/fix-library-managed-attributes-react18.md
+++ b/.changeset/fix-library-managed-attributes-react18.md
@@ -4,8 +4,6 @@
 
 Fix `LibraryManagedAttributes` to restore React 18 compatibility and `defaultProps` stripping.
 
-Commit `9a805f7` introduced two regressions:
-
 1. `key` no longer accepted `null`, contradicting `React.Attributes` which defines `key?: Key | null | undefined`. This caused `TS2322` errors in React 18 codebases when passing a nullable key (e.g. `key={item?.id ?? null}`).
 
 2. Using `P` directly instead of `React.JSX.LibraryManagedAttributes<C, P>` broke `defaultProps` stripping for class components — props covered by `defaultProps` were no longer treated as optional at the call site (`TS2740`).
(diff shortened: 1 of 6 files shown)


packages/react/src/jsx/jsx-local-namespace.ts
```diff
@@ -83,7 +83,8 @@ export namespace CompiledJSX {
   // intersecting P with { key?: React.Key } directly. With a correct ElementAttributesProperty,
   // TypeScript pre-extracts the props type from class instances before passing P here.
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  export type LibraryManagedAttributes<_C, P> = WithConditionalCSSProp<P> & P & { key?: React.Key };
+  export type LibraryManagedAttributes<C, P> = WithConditionalCSSProp<P> &
+    React.JSX.LibraryManagedAttributes<C, P> & { key?: React.Key | null };
```
kylorhall-atlassian, line 87: cc: @flipatlas can you review or test?

We'd changed in https://github.com/atlassian-labs/compiled/pull/1886 and this change is unclear to me without exploring the API.

.changeset/fix-library-managed-attributes-react18.md
```diff
@@ -0,0 +1,13 @@
+---
+'@compiled/react': patch
+---
+
+Fix `LibraryManagedAttributes` to restore React 18 compatibility and `defaultProps` stripping.
```
kylorhall-atlassian, line 5: @xtan-atlas could we add a test for these, eg. `defaultProps`?

packages/react/src/jsx/__tests__/jsx-local-namespace-types.test.tsx
```diff
@@ -189,3 +189,39 @@ it('correctly infers types for generic child components', () => {
   expectTypeOf<Managed>().toMatchTypeOf<{ items: { id: string }[] }>();
   expectTypeOf<Managed>().toMatchTypeOf<{ css?: unknown }>();
 });
+
+it('class components with defaultProps — props covered by defaultProps should be optional at call site', () => {
```
xtan-atlas, line 193: @kylorhall-atlassian added a test for defaultProps problem

packages/react/src/jsx/__tests__/jsx-local-namespace-types.test.tsx
```diff
@@ -17,7 +17,7 @@ it('always includes an optional key prop', () => {
   type Props = { id: string };
   type Managed = ManagedProps<React.FC<Props>, Props>;
 
-  expectTypeOf<Managed>().toMatchTypeOf<{ key?: React.Key }>();
+  expectTypeOf<Managed['key']>().toEqualTypeOf<React.Key | null | undefined>();
```
xtan-atlas, line 20: This test covers the null type problem, as React 18 allows nullable key.

Image

.changeset/fix-library-managed-attributes-react18.md
xtan-atlas: added one :)

xtan-atlas:
Pushed commits to fix some issues when I run the typecheck tests. 
Verified it works for both React 18 and React 19.

Screenshot 2026-06-15 at 12 15 45 pm

If revert the changes in packages/react/src/jsx/jsx-local-namespace.ts, then we would fail the updated tests for React18 and React19. 
Screenshot 2026-06-15 at 12 20 51 pm
Screenshot 2026-06-15 at 12 20 43 pm

packages/react/src/jsx/jsx-local-namespace.ts
kylorhall-atlassian: I think the existing + added test coverage should be sufficient.

kylorhall-atlassian (approved):
Yeah, this seems fine, hopefully we have enough coverage from the last fixup!

PR #1912, opened 2026-06-18
fix stylesheet ordering issue

xtan-atlas:
## What is this change?

Five related fixes across the `@compiled/*` toolchain to make extracted CSS output deterministic across builds and correctly preserve `cssMapScoped` (non-atomic) source order. Affected packages: `@compiled/babel-plugin-strip-runtime`, `@compiled/parcel-optimizer`, `@compiled/vite-plugin`, `@compiled/webpack-loader`, and `@compiled/react`.

## Why are we making this change?

`cssMapScoped` relies on CSS source order for cascade correctness — when two rules target the same selector at the same specificity, the one that appears later wins. Three issues were breaking this:

1. **Reversed `require()` emission** — `babel-plugin-strip-runtime` used `unshiftContainer` in a loop, reversing the `styleRules` source order in the emitted JS. This caused downstream bundlers (Webpack, Vite) to merge CSS rules in the wrong order.
2. **Non-deterministic cross-file ordering** — `parcel-optimizer` collected rules via `bundleGraph.traverseBundles` (non-deterministic traversal order) and `vite-plugin` used a flat `Set` (insertion-order dependent on transform call order). Both could produce different CSS output across builds for the same source.
3. **Misclassified at-rule-wrapped non-atomic rules** — `@compiled/react` runtime used `startsWith('.cc-')` to detect non-atomic rules, missing `@media{.cc-...}` and `@container{.cc-...}` wrapped rules. These went into the wrong style bucket at runtime.

## How are we making this change?

- **`@compiled/babel-plugin-strip-runtime`** — replaced the per-rule `unshiftContainer` loop with a single batch `unshiftContainer` call to preserve `styleRules` source order in emitted `require()` statements. Also applied `sortStyleRulesForDeterministicOutput` (non-atomic preserved, atomic sorted lexically) when writing extracted `.compiled.css` files.
- **`@compiled/parcel-optimizer`** — collect rules per asset with `filePath`, sort assets by `filePath` for cross-file determinism, then partition non-atomic vs atomic rules via `sortStyleRulesForDeterministicOutput`. Extracted as a testable `buildDeterministicStylesheet` helper.
- **`@compiled/vite-plugin`** — switched from a flat `Set<string>` to a `Map<string, string[]>` keyed by source file path. Entries are sorted by filePath before collection, then partitioned via the same `sortStyleRulesForDeterministicOutput` helper.
- **`@compiled/webpack-loader`** — sort emitted CSS assets by name for deterministic output ordering.
- **`@compiled/react`** — changed `isNonAtomicSheet` from `startsWith('.cc-')` to `includes('.cc-')` so at-rule-wrapped non-atomic rules (`@media`, `@container`) are correctly routed to the catch-all style bucket at runtime.

### Test coverage

- **40** `babel-plugin-strip-runtime` tests (4 updated, 1 new TDD test for batch unshift fix).
- **14** `webpack-loader` tests (incl. cross-file `cssMapScoped` TDD test).
- **34** `vite-plugin` tests (incl. cross-file inline-snapshot test verifying filePath sort).
- **7** `parcel-optimizer` sort-style-rules unit tests.
- **1** `@compiled/react` browser test verifying `isNonAtomicSheet` fix.

---

### PR checklist

Don't delete me!

I have...

- [/] Updated the documentation in `website/`


packages/parcel-optimizer/src/index.ts
```diff
@@ -16,6 +16,28 @@ const configFiles = [
...
+
+  for (const rule of styleRules) {
+    if (rule.includes(NON_ATOMIC_CLASS_PREFIX)) {
+      nonAtomicRules.push(rule);
+    } else {
+      atomicRules.push(rule);
+    }
+  }
+
+  return [...nonAtomicRules, ...atomicRules.sort()];
```
liamqma, line 38: Um. This might be a tricky issue. 
If we don not sort nonAtomicRules, their order will be nondeterministic. This will cause bundle churn, which in turn reduces the cache hit ratio. What happens if you sort nonAtomicRules as well? Will it cause CSS override issues?

JakeLane: @xtan-atlas I recall we discussed sorting in the sync where we need to make sure the non-atomic rules are deterministic ordered by something like filename

xtan-atlas: @liamqma sorting non-atomic rules is unsafe because cascade depends on source order, so it'd cause css override issues. The non-atomic rules inserted from styleRules is source order deterministic, so should be fine for bundle chunk and cache hit. Again, it'd be no impact for existing atomic styles :)

JakeLane: Do we have a test covering that the non-atomic rule order is identical if the transform order changes?

xtan-atlas: @JakeLane if compose non atomic styles from different files, we said we can only rely on adding nested selector to increase css specificity, to avoid css ordering issue. Just like how traditional css works. In editor we don't compose multiple cssMapScoped call, so I think it'd fine from that perspective. I can document that in the LDR page, and later update comments in cssMapScoped API.

xtan-atlas: @JakeLane @liamqma in terms of testing, I added non atomic styles tests in packages/css/src/__tests__/sort.test.ts, packages/babel-plugin-strip-runtime/src/__tests__/extract-styles.test.ts, also unit tests in this PR and updated the integration test in this PR.

packages/parcel-optimizer/src/index.ts
liamqma, line 95: @JakeLane 
What do you think of this?
```
const allAssets = [];
childBundle.traverseAssets((asset) => {
  allAssets.push(asset);
});

// Sort assets by file path first
allAssets.sort((asset) => asset.filePath);

// Then collect the style rules
allAssets.forEach((asset) => {
  for (const rule of rules) {
    styleRules.add(rule as string);
  }
});
```
This way, even though we don't sort the non-atomic CSS itself, the order of the files stays deterministic.

packages/parcel-optimizer/src/index.ts
```diff
@@ -16,6 +16,44 @@ const configFiles = [
...
+};
+
+/**
+ * Build a deterministic stylesheet from multiple assets.
+ * Assets are sorted by filePath for stable cross-file ordering.
+ * Within each asset, rules are sorted via sortStyleRulesForDeterministicOutput
+ * and then semantically sorted via @compiled/css sort().
+ */
+export const buildDeterministicStylesheet = (
```
xtan-atlas, line 48: @JakeLane @liamqma added this helper to ensure the deterministic order.

packages/parcel-optimizer/src/__tests__/sort-style-rules.test.ts
xtan-atlas, line 1: @liamqma @JakeLane added more tests to cover the deterministic order cross assets.

packages/parcel-optimizer/src/index.ts
JakeLane: yeah that should be fine

packages/parcel-optimizer/src/index.ts
liamqma: - Do we need to worry about Webpack `packages/webpack-loader/src/extract-plugin.ts`? Jira and Confluence use Atlaspack, but products such as Townsquare use Rspack. Does need to worry about townsquare?
- Have you tried out this change in AFM before merging this PR (normally we patch the node_modules, runt afm build, and sanity check compiled.css look expected)? 
- Also, for the above, could you add a console.log to measure how long this sort takes?
`[...assets].sort((a, b) => (a.filePath < b.filePath ? -1 : 1))`
Since this PR introduces an extra step, we should make sure it does not regress build time. I do not expect it to though.

xtan-atlas: > Do we need to worry about Webpack packages/webpack-loader/src/extract-plugin.ts? Jira and Confluence use Atlaspack, but products such as Townsquare use Rspack. Does need to worry about townsquare?
Not expert on this, seems don't need to be worried, based on AI's response:
Screenshot 2026-06-19 at 10 38 49 am

xtan-atlas: > Have you tried out this change in AFM before merging this PR (normally we patch the node_modules, runt afm build, and sanity check compiled.css look expected)?

Yes, I tried `yarn build @atlaskit/editor-core` with latest patches. The output of individual asset preserved order now, will check over branch deploy a bit later.

xtan-atlas: > Also, for the above, could you add a console.log to measure how long this sort takes?
> [...assets].sort((a, b) => (a.filePath < b.filePath ? -1 : 1))
very quick, so should be good:
Screenshot 2026-06-19 at 11 06 58 am

JakeLane (approved):

PR #1916, opened 2026-07-11
fix(vite-plugin): Sort extracted CSS during development

### LINKED ISSUE #1895: .compiled.css files are not sorted in dev mode, causing CSS ordering bugs with extracted packages
The Vite plugin (@compiled/vite-plugin) does not sort .compiled.css files during development. In production builds, generateBundle applies sort() from @compiled/css to CSS assets, ensuring correct rule ordering (base before @media, shorthand before longhand, LVFHA pseudo ordering). However in dev mode, Vite injects each .compiled.css import as a <style> tag in module-graph traversal order — and that order is not guaranteed to keep base rules before their @media overrides, or shorthand properties before their longhand counterparts.
This causes a family of CSS ordering bugs when consuming packages that use CSS extraction (e.g. @atlaskit/*):

 1.  Media query ordering: A base display:none atomic class wins over @media (min-width:48rem){ display:grid } because the base rule is inserted later in the merged stylesheet. This breaks responsive layouts — e.g. navigation side-nav grid-area, search bar visibility, and top-nav grid-template-columns all break.
 2.  Font shorthand vs longhand: The font shorthand (e.g. font: var(--ds-font-body, normal 400 14px/20px ...)) embeds font-weight: 400 and is inserted after an explicit font-weight: var(--ds-font-weight-bold, 700) longhand class, causing bold text to render as regular weight.
 3.  Border shorthand reset: A border-width: 0 reset class from an Atlaskit component is inserted after explicit border-width: var(--ds-border-width, 1px) declarations, causing borders to disappear.
To Reproduce

 1.  Create a Vite project that uses @compiled/vite-plugin
 2.  Import a package that ships pre-compiled .compiled.css files with both base rules and @media query rules (e.g. @atlaskit/navigation-system or @atlaskit/primitives)
 3.  Run the Vite dev server (vite dev)
 4.  Observe that responsive styles are broken — e.g. elements that should be display:grid at desktop widths remain display:none, or bold text renders as normal weight
The issue does NOT reproduce in production builds because generateBundle applies sort() to CSS assets.
Expected behavior
.compiled.css files should be sorted in dev mode the same way they are in production builds. Base rules should always appear before @media/@supports at-rules, and shorthand properties should appear before their longhands, so that CSS specificity and source-order semantics work correctly regardless of module-graph traversal order.
Desktop (please complete the following information):

•  OS: macOS
•  Browser: Chrome, Safari (all browsers — the issue is in CSS source ordering, not browser-specific)
•  Version: @compiled/vite-plugin v1.1.4, @compiled/react v0.21.0
Additional context
The fix is straightforward: add .compiled.css handling at the top of the transform hook, before the JS/TS file filter. The existing sort() function from @compiled/css is already imported and used in generateBundle — it just needs to also be applied to .compiled.css files during transform:

async transform(code: string, id: string): Promise<any> {
  // Sort .compiled.css files to ensure correct rule ordering.
  if (id.endsWith('.compiled.css') && code.includes('._')) {
    try {
      const sortedCss = sort(code, {
        sortAtRulesEnabled: options.sortAtRules,
        sortShorthandEnabled: options.sortShorthand,
      });
      return { code: sortedCss, map: null };
    } catch {
      // Fall through — serve unsorted CSS rather than breaking
    }
  }

  // ... existing JS/TS transform logic
}
The full fix with tests is available at /tmp/compiled-upstream on branch fix/sort-compiled-css-in-dev-mode (2 files changed, +106 lines, all 22 tests passing).

KSHITIZ6341:
### What is this change?

Sorts extracted `.compiled.css` modules in the Vite transform hook during development. Invalid CSS is left unchanged and produces a Vite warning.

Credit to @lewishealey for identifying the development-only ordering gap and providing concrete cascade failures.

Fixes #1895

Production bundles already pass Compiled CSS through `@compiled/css` sorting, but Vite development mode injects each extracted stylesheet in module-graph order. That can place base rules after media overrides or shorthand declarations after their longhands, changing the cascade between development and production.

The Vite plugin now applies the same configured sorter to `.compiled.css` modules during `transform`. A shared helper keeps development transforms, bundled CSS assets, and emitted extraction output on the same sorting options.

Validation:

- `jest packages/vite-plugin/src/__tests__ --runInBand --no-cache` (36 tests)
- Vite plugin plus CSS at-rule and shorthand sorter suites (78 tests)
- `ttsc --build packages/vite-plugin/tsconfig.json --pretty false`
- Targeted ESLint and Prettier checks
- Repository pre-commit hook

Not run: visual regression tests, because this change is exercised at the deterministic Vite transform and CSS sorter layers without rendering components.

---

### PR checklist

Don't delete me!

I have...

- [ ] Updated the documentation in `website/` (not required; no API or configuration change)
- [x] Added a changeset (if making any changes that affect Compiled’s behaviour)

### DIFF
--- a/packages/vite-plugin/src/index.ts
+++ b/packages/vite-plugin/src/index.ts
@@ -76,6 +76,12 @@ function compiled(userOptions: PluginOptions = {}): any {
     ...userOptions,
   };
 
+  const sortCompiledCss = (css: string): string =>
+    sort(css, {
+      sortAtRulesEnabled: options.sortAtRules,
+      sortShorthandEnabled: options.sortShorthand,
+    });
+
   // Storage for collected style rules during transformation
   // Map of filePath → array of style rules (in source order from the babel
   // transform). We sort by filePath at extraction time for cross-file
@@ -94,6 +100,21 @@ function compiled(userOptions: PluginOptions = {}): any {
     enforce: 'pre', // Run before other plugins
 
     async transform(code: string, id: string): Promise<any> {
+      if (id.endsWith('.compiled.css') && code.includes('._')) {
+        try {
+          return {
+            code: sortCompiledCss(code),
+            map: null,
+          };
+        } catch (error) {
+          const err = error as Error;
+          this.warn({
+            message: `[@compiled/vite-plugin] Failed to sort CSS in ${id}: ${err.message}`,
+          });
+          return null;
+        }
+      }
+
       // Filter out node_modules (except for specific includes if needed)
       if (id.includes('/node_modules/@compiled/react')) {
         return null;
@@ -218,16 +239,8 @@ function compiled(userOptions: PluginOptions = {}): any {
         // This is a heuristic to identify CSS that came from .compiled.css files
         if (cssContent.includes('._')) {
           try {
-            // Apply Compiled's CSS sorting and deduplication
-            const sortConfig = {
-              sortAtRulesEnabled: options.sortAtRules,
-              sortShorthandEnabled: options.sortShorthand,
-            };
-
-            const sortedCss = sort(cssContent, sortConfig);
-
             // Update the asset with sorted CSS
-            asset.source = sortedCss;
+            asset.source = sortCompiledCss(cssContent);
           } catch (error) {
             const err = error as Error;
             this.warn({
@@ -254,19 +267,13 @@ function compiled(userOptions: PluginOptions = {}): any {
 
           // Join all rules and apply CSS sorting
           const combinedCss = allRules.join('\n');
-          const sortConfig = {
-            sortAtRulesEnabled: options.sortAtRules,
-            sortShorthandEnabled: options.sortShorthand,
-          };
-
-          const sortedCss = sort(combinedCss, sortConfig);
 
           // Emit the CSS file with content-based naming
           // Vite will add a content hash to the filename automatically
           this.emitFile({
             type: 'asset',
             name: EXTRACTED_CSS_NAME,
-            source: sortedCss,
+            source: sortCompiledCss(combinedCss),
           });
 
           // Mark that we've emitted the file so transformIndexHtml can inject it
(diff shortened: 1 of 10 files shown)


packages/vite-plugin/src/index.ts
```diff
@@ -94,6 +100,21 @@ function compiled(userOptions: PluginOptions = {}): any {
     enforce: 'pre', // Run before other plugins
 
     async transform(code: string, id: string): Promise<any> {
+      if (id.endsWith('.compiled.css') && code.includes('._')) {
```
liamqma: While it's nice fix, I don't think this PR solve https://github.com/atlassian-labs/compiled/issues/1895 entirely.
The trigger is per-file (transform runs once per module). Every test calls plugin.transform on a single string. It's not guaranteed to resolve the actual reported scenario: multiple .compiled.css files whose cross-file order is wrong. As established earlier, dev serves each file as its own <style> and sort() here only orders within one file.

KSHITIZ6341:
True true, you were right that sorting each transform independently could not fix ordering across multiple `.compiled.css` modules.

I’ve replaced the per-file-only approach in `fe8f7789` with a development-only aggregation flow:

- Script imports of `.compiled.css` are routed through virtual JavaScript proxies.
- Each proxy imports Vite’s fully processed CSS using `?inline`, preventing Vite from creating separate style tags.
- Active Compiled CSS modules are registered in one shared stylesheet and sorted together using the existing `@compiled/css` sorter.
- HMR updates and module pruning update or remove registered styles correctly.
- SSR and production builds retain their existing behavior.

The new integration test deliberately imports overrides before their base rules across two separate files. It first confirms that this reproduces the media-query, font shorthand/longhand, and border shorthand/longhand failures from #1895, then verifies that the combined stylesheet has the correct order.

It also covers Vite URL processing, custom base paths, HMR, CSP nonces, SSR/build isolation, and preservation of the single-plugin return API.

could you please take another look?

PR #1917, opened 2026-07-11
fix(css): Merge duplicate nested at-rules

### LINKED ISSUE #1847: Duplicate styles in nested media queries are not merged
**Describe the bug**
When there are duplicate styles used in nested media queries, they are not merged. 

**To Reproduce**

Add and run the unit test to the file [merge-duplicate-at-rules.test.ts](https://github.com/atlassian-labs/compiled/blob/master/packages/css/src/plugins/__tests__/merge-duplicate-at-rules.test.ts):

```tsx
it('should remove duplicate children when nested at-rules', () => {
  const actual = transform`
    @supports not (height: 1lh) {
      @media (min-width:500px) {
        color: red;
        color:green;
      }
    }

    @supports not (height: 1lh) {
      @media (min-width:500px){
        color: red;
        color:blue;
      }
    }
  `;

    "
          @supports not (height: 1lh) {
            @media (min-width:500px) {
              color: red;
              color:green;
            }
            @media (min-width:500px){
              color: red;
              color:blue;
            }
          }
        "
  `);
});
```

**Expected behavior**
Duplicate styles should be merged. In the above example, it should end up with:

```tsx
@supports not (height: 1lh) {
    @media (min-width:500px) {
      color: red;
      color:green;
      color:blue;
    }
  }
```

**Additional context**
Not quite the same, but I found a related issue from 5 years ago: https://github.com/atlassian-labs/compiled/issues/316

KSHITIZ6341:
### What is this change?

Merges duplicate children recursively when equivalent at-rules are nested, so repeated declarations under the same conditional scopes are emitted once.

Credit to @michaelabrahamian for providing the nested at-rule reproduction and expected output.

Fixes #1847

`mergeDuplicateAtRules` merged matching top-level containers but treated each nested at-rule as an opaque serialized child. When nested blocks differed by even one declaration, both survived deduplication and produced redundant CSS.

The plugin now merges equivalent sibling at-rules at each nesting level, deduplicates their children, and then recurses into the merged container. Matching remains scoped to the current parent, non-atomic `cssMapScoped` rules remain untouched, and existing top-level ordering is preserved.

A patch changeset for `@compiled/css` is included.

Validation:

- `yarn test packages/css/src/plugins/__tests__/merge-duplicate-at-rules.test.ts --runInBand` (6 tests)
- `yarn test packages/css --runInBand` (297 tests, 219 snapshots)
- `yarn test --runInBand` (5,132 tests, 436 snapshots)
- `yarn lint`
- `yarn prettier:check`
- `yarn build`
- Repository pre-commit hook

Not run: visual regression tests, because this change operates on deterministic PostCSS output and has no rendered component changes.

---

### PR checklist

Do not delete me!

I have...

- [ ] Updated the documentation in `website/` (not required; no API or configuration change)
- [x] Added a changeset (if making any changes that affect Compiled behavior)

### DIFF
--- a/packages/css/src/plugins/__tests__/merge-duplicate-at-rules.test.ts
+++ b/packages/css/src/plugins/__tests__/merge-duplicate-at-rules.test.ts
@@ -36,4 +36,112 @@ describe('discard duplicate at-rule children plugin', () => {
           "
     `);
   });
+
+  it('should move merged top-level at-rules after regular rules', () => {
+    const actual = transform`
+      @media (min-width:500px) { color: red; }
+      .default { color: blue; }
+      @media (min-width:500px) { background: white; }
+    `;
+
+    expect(actual).toMatchInlineSnapshot(`
+      "
+            .default { color: blue; }
+            @media (min-width:500px) { color: red; background: white; }
+          "
+    `);
+  });
+
+  it('should remove duplicate children from nested at-rules', () => {
+    const actual = transform`
+      @supports not (height: 1lh) {
+        @media (min-width:500px) {
+          color: red;
+          color: green;
+        }
+      }
+
+      @supports not (height: 1lh) {
+        @media (min-width:500px) {
+          color: red;
+          color: blue;
+        }
+      }
+    `;
+
+    expect(actual).toMatchInlineSnapshot(`
+      "
+            @supports not (height: 1lh) {
+              @media (min-width:500px) {
+                color: red;
+                color: green;
+                color: blue;
+              }
+            }
+          "
+    `);
+  });
+
+  it('should keep nested at-rules scoped to their parent conditions', () => {
+    const actual = transform`
+      @supports (display: grid) {
+        @media (min-width:500px) {
+          color: red;
+        }
+      }
+
+      @supports (display: flex) {
+        @media (min-width:500px) {
+          color: blue;
+        }
+      }
+    `;
+
+    expect(actual).toMatchInlineSnapshot(`
+      "
+            @supports (display: grid) {
+              @media (min-width:500px) {
+                color: red;
+              }
+            }
+            @supports (display: flex) {
+              @media (min-width:500px) {
+                color: blue;
+              }
+            }
+          "
+    `);
+  });
+
+  it('should preserve nested non-atomic at-rules', () => {
+    const actual = transform`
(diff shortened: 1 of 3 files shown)


packages/css/src/plugins/__tests__/merge-duplicate-at-rules.test.ts
```diff
@@ -36,4 +36,112 @@ describe('discard duplicate at-rule children plugin', () => {
           "
     `);
   });
+
+  it('should move merged top-level at-rules after regular rules', () => {
```
liamqma, line 40: Can you please add more tests to conver different cases? The cases I can think of are

1. Recursion depth (locks the recursive walk)

•  3-level duplicated at-rules:

```
@layer l { @supports (display: grid) { @media (min-width:500px) { color: red; } } }
@layer l { @supports (display: grid) { @media (min-width:500px) { color: blue; } } }
```

2. Nested sibling duplicates within a single parent (different path from "duplicate parents"):

```
@supports (display: grid) {
  @media (min-width:500px) { color: red; }
  @media (min-width:500px) { color: blue; }
}
```

3. Mixed nested siblings  only duplicates merge, unique one untouched, order preserved:

```
@supports (display: grid) {
  @media (min-width:500px) { color: red; }
  @media (min-width:900px) { color: purple; }
  @media (min-width:500px) { color: blue; }
}
```

4. Append-then-re-dedupe (the subtlest path a merged-in child must still be deduped by the later recursion):

```
@supports (display: grid) { @media (min-width:500px) { color: red; } }
@supports (display: grid) { @media (min-width:500px) { color: red; color: blue; } }
```

5. Same name, different params must NOT merge (guards the \0 separator):

```
@media all { color: red; }
@media all and (min-width:1px) { color: blue; }
```

6. Transitive merge across a non-matching sibling (1st and 3rd duplicate, 2nd different):

```
@media (min-width:500px) { color: red; }
@media (min-width:900px) { color: purple; }
@media (min-width:500px) { color: blue; }
```

7. Mixed atomic / non-atomic at the same level  the .cc- block is skipped, the two atomic ones merge:

```
@media (min-width:500px) { .cc-1 { color: red; } }
@media (min-width:500px) { ._atomic1 { color: red; } }
@media (min-width:500px) { ._atomic2 { color: blue; } }
```

PR #1918, opened 2026-07-11
fix(css): sort nested at-rules

### LINKED ISSUE #1846: Nested at-rules are not sorted
**Describe the bug**

When at-rules are nested inside another at-rule, they are not sorted.

This is unexpected and can cause significant production issues (especially when production apps are configured differently to development environments).

Example:
```tsx
const styles = cssMap({
  root: {
    '@supports not (-moz-appearance: none)': {
      '@media (prefers-reduced-motion: no-preference) and (min-width: 64rem)': {
        color: 'blue',
      },
      '@media (prefers-reduced-motion: no-preference)': {
        color: 'green',
      },
    },
  },
});
```

In this example, we would expect `color: 'blue'` to "win" on large viewports (width than `64rem`). However this doesn't consistently happen in apps.

This appears to be because the [sortAtomicStyleSheet](https://github.com/atlassian-labs/compiled/blob/master/packages/css/src/plugins/sort-atomic-style-sheet.ts#L43C14-L43C34) function only looks at the nodes at the first level.

There is a comment here also confirming this: https://github.com/atlassian-labs/compiled/blob/master/packages/css/src/plugins/sort-atomic-style-sheet.ts#L39

```
 * Only top level CSS rules will be sorted.
```

Nesting at-rules like this is valid CSS: https://css-tricks.com/can-you-nest-media-and-support-queries/

**To Reproduce**

You can test the current behaviour by adding the following tests to [packages/css/src/plugins/__tests__/sort-at-rules.test.ts](https://github.com/atlassian-labs/compiled/blob/master/packages/css/src/plugins/__tests__/sort-at-rules.test.ts#L950):

```tsx
it.only('should support nested queries with multiple at-rules', () => {
  const actual = transform(`
    @supports not (height: 1lh) {
      color: red;
      @media (prefers-reduced-motion: no-preference) and (width > 1500px) {
        color: blue;
      }

      @media (prefers-reduced-motion: no-preference) {
        color: green;
      }
    }
  `);

    "
            @supports not (height: 1lh) {
              color: red;
              @media (prefers-reduced-motion: no-preference) and (width > 1500px) {
                color: blue;
              }

              @media (prefers-reduced-motion: no-preference) {
                color: green;
              }
            }
          "
  `);
});

it.only('should support nested queries with multiple at-rules', () => {
  const actual = transform(`
    color: red;
    @media (prefers-reduced-motion: no-preference) and (width > 1500px) {
      @supports not (height: 1lh) {
        color: blue;
      }
    }

    @media (prefers-reduced-motion: no-preference) {
      @supports not (height: 1lh) {
          color: green;
        }
      }
  `);

    "
            color: red;

            @media (prefers-reduced-motion: no-preference) {
              @supports not (height: 1lh) {
                  color: green;
                }
              }
            @media (prefers-reduced-motion: no-preference) and (width > 1500px) {
              @supports not (height: 1lh) {
                color: blue;
              }
            }
          "
  `);
});
```

**Expected behavior**
Nested media queries should be sorted.

At a minimum, some improved guidance and lint rules to make people aware would be really helpful.

**Workaround**
The current workaround we are using is just making sure all at-rules that require sorting (e.g. `@media`) are at the top level. For my specific example, flipping the order of the at-rules to place the `@media` at the top level and have `@supports` nested at the second level would make sure the media queries are sorted.

**Additional context**
Related issue about not merging duplicate styles in nested media queries: https://github.com/atlassian-labs/compiled/issues/1847

KSHITIZ6341:
### What is this change?

Sort sibling at-rules recursively when they are nested inside another rule or at-rule, instead of limiting at-rule ordering to the stylesheet root.

Good catch, @michaelabrahamian: nested media queries currently retain their source order. A broader query can therefore appear after a narrower width-constrained query and override it when both conditions match.

The root cause was that `sortStyleSheet` only collected at-rules from the PostCSS root.

- Recursively collect and sort direct at-rule children using the existing comparator.
- Replace only at-rule positions, preserving declarations and other child nodes.
- Keep non-atomic `cssMapScoped` blocks unchanged and honor the existing `sortAtRules` opt-out.
- Add a regression test for nested reduced-motion and width queries.
- Add a patch changeset for `@compiled/css`.

There are no public API or type changes.

Fixes #1846

### Checks

- Full Jest suite: 120 suites, 5,129 tests passed
- `@compiled/css` TypeScript build
- ESLint on changed TypeScript files
- Prettier and `git diff --check`

Not run: visual regression tests; the change is covered by stylesheet-output unit tests and does not alter rendered components directly.

---

### PR checklist

Don't delete me!

I have...

- [ ] Updated the documentation in `website/` (not applicable; no API or configuration change)


packages/css/src/plugins/__tests__/sort-at-rules.test.ts
```diff
@@ -893,5 +893,32 @@ describe('sort at-rules', () => {
               "
       `);
     });
+
+    it('should sort media queries nested inside another at-rule', () => {
```
liamqma, line 897: Can you please add more test coverage?

Three level nesting sorts correctly:
```
@layer l {
  @supports (display: grid) {
    @media (min-width: 400px) { color: green; }
    @media (min-width: 200px) { color: blue; }
  }
}
```
Nested at-rules inside a plain rule sort correctly:

```
.foo {
  @media (min-width: 400px) { color: green; }
  @media (min-width: 200px) { color: blue; }
}
```

Opt-out preserves nested source order (run with sortAtRulesEnabled: false, expect no reordering):
```
@supports (display: grid) {
  @media (min-width: 400px) { color: green; }
  @media (min-width: 200px) { color: blue; }
}
```
Non-atomic cssMapScoped nested blocks are not reordered:
```
.cc-x {
  @media (min-width: 400px) { color: green; }
  @media (min-width: 200px) { color: blue; }
}
```

KSHITIZ6341:
Sorry missed those, I’ve added tests for all those cases you mentioned there, and they’re all passing now.

packages/css/src/plugins/__tests__/sort-at-rules.test.ts
KSHITIZ6341: Added all four requested cases: three-level nesting, nesting inside a plain rule, disabled at-rule sorting, and non-atomic `cssMapScoped` behavior. The focused suite passes all 25 tests and 24 snapshots, and Prettier passes. Could you take another look?

PR #1920, opened 2026-07-14
New hash algorithm

pancaspe87:
### What is this change?

Adds an opt-in `collisionResistantHash` option to the Compiled babel plugin / transform that generates atomic class names using a base-62 full-width hash, eliminating hash collisions.
Default is `false` — output is byte-for-byte identical to today. This release is additive and non-breaking. Teams can opt in incrementally; a future major release flips the default and removes the legacy branch.

Current hash doesn't guarantee unique group hash, especially in large repos or libraries with several nested selectors.
When two rules collide on the group hash, `ax()` deduplicates them as if they were the same property — silently dropping styles.

`packages/utils/src/hash.ts`

- Adds `hashBase62(str, length)` — full 32-bit MurmurHash2 → base-62 (0-9a-zA-Z), zero-padded to fixed width
- Adds `ATOMIC_GROUP_HASH_LENGTH = 6` and `ATOMIC_VALUE_HASH_LENGTH = 4` constants
- Keeps existing `hash()` (base-36) unchanged — still used for keyframe names, CSS variable names, and asset filenames (deliberately, so filenames stay lowercase and avoid case-insensitive-filesystem collisions)

`packages/css/src/plugins/atomicify-rules.ts`

- Adds `collisionResistantHash?: boolean` option (default `false`)
- `true` → `_<6-char group><4-char value>` (11-char class, e.g. `_3iDTPbvLZJ`)
- `false` → existing `_<4-char group><4-char value>` (9-char class) — unchanged

`packages/react/src/runtime/ax.ts` 
Done in https://github.com/atlassian-labs/compiled/pull/1924

Instead of slicing a fixed number of chars from the front, we slice everything except the last 4 (the value hash). Because the value hash is fixed-width (4 chars) in both formats, "everything before the value" is always exactly the full group key — regardless of how long the group is.

During migration, a single page can contain both legacy (from not-yet-rebuilt packages / pre-built npm CSS) and new (from rebuilt packages) classes. This is safe because the two formats produce group keys of different lengths:

- Legacy group key: _ + 4 chars = 5 chars (e.g. `_1e0c`)
- New group key: _ + 6 chars = 7 chars (e.g. `_3iDTPb`)
A 5-char string and a 7-char string can never be equal, so a legacy class and a new class can never collide on the same map key — even if they represent the same CSS property. They coexist as separate entries; the CSS cascade (source order / specificity) resolves which wins, exactly as it would for any two unrelated atomic classes. There is no false deduplication across formats.
---

### PR checklist

I have...

[x] Updated or added applicable tests
Because the class-name format changes when `collisionResistantHash` is on, a large number of fixtures needed updating. This was done deliberately so the test suite documents the target (new) output, while a dedicated suite still guards the legacy path. 

**Additional tests:** 
- `packages/react/src/runtime/__tests__/ax.test.ts`,  
- `packages/utils/src/__tests__/hash.test.ts`

[x] Updated the documentation in `website/` 
[x] Added a changeset (if making any changes that affect Compiled's behaviour)


packages/babel-plugin/src/utils/compress-class-names-for-runtime.ts
```diff
@@ -1,7 +1,16 @@
+import { ATOMIC_VALUE_HASH_LENGTH } from '@compiled/utils';
+
 /**
  * Compress class names based on `classNameCompressionMap`.
```
pancaspe87: Hey @liamqma can you remind me whether this feature was shipped in the end? I don't see the `classNameCompressionMap` option enabled anywhere, wondering if this is legacy code that could be removed.

liamqma: No, we didn't. The entire class name compression feature and its related code are unused.

pancaspe87: @liamqma What is your recommendation? keep it as it could still be used OR remove it to simplify the codebase? Either work for me, just asking to see how to best handle this when reviewing the hash algorithm - I need to do the same change across 3 plugins ( here, atlaspack and in-source SWC)

liamqma: I suggest removing it to simplify the codebase. Thanks, Celeste! I am not sure we will ever restart the work of compressing classnames. Even if we do, I imagine the feature will be implemented differently in Atlaspack Incremental.

packages/babel-plugin/src/types.ts
```diff
@@ -111,6 +111,25 @@ export interface PluginOptions {
...
+   * variable or feature gate in the build config. Once every consumer has
+   * migrated, the default will flip to `true` and the legacy branch removed.
+   *
+   * The runtime `ax()` handles both class formats simultaneously (via
+   * length-based group-key extraction), so mixing old and new classes on a
+   * page is safe.
+   *
+   * @default false
+   */
+  collisionResistantHash?: boolean;
```
kylorhall-atlassian, line 132: Probably could use a website change as well; doesn't have to be a blocker or come in this PR.

This could be where we better document the migration path and possible conflicts with pass-through `xcss` to link in CHANGELOG / [CDAC](https://community.developer.atlassian.com/)

packages/react/src/runtime/__tests__/ax.test.ts
```diff
@@ -45,15 +45,39 @@ describe('ax - atomic', () => {
...
+    ],
+    [
+      'should not cross-dedup legacy 9-char and new 11-char classes for the same property',
+      // During version-skew, ax() receives old 9-char classes from pre-built packages
+      // and new 11-char classes from the app's own styles in the same call.
+      // Different group key lengths mean they never falsely dedup each other.
+      // CLEANUP: remove this test when collisionResistantHash becomes the default
+      ['_syaz13q2', '_1UtDYzynoA'],
+      '_syaz13q2 _1UtDYzynoA',
+    ],
```
kylorhall-atlassian, line 86: We could in theory handle this clash if we stick to base36 for both, but if we can find a way to roll this out, your approach here is best.

There will be conflicts via NPM on how this works though, here's an example:
```tsx
import { Box } from '@atlaskit/primitives/compiled/box'; // `collisionResistantHash=false` coming from NPM (so you can't re-hash it)
import { cssMap } from '@atlaskit/css'; // `collisionResistantHash=true`

const styles = cssMap({
  border: '1px solid red'
});

export default () => <Box xcss={styles.root} />
```

This will get, in psuedo-code:
```html
<style>
._abcd1234 { border: none }
._FeDcBa7890 { border: 1px solid red }
</style>
<div class="_abcd1234 _abcdef1234"></div>
```

Now this works in isolation, but if atomic styles are ordered differently due to arbitrary file imports, things go awry.

The same can be said of the inverse—we'll need to cut a `major` when we distribute every Platform package with `collisionResistantHash=true`, but I expect the impact to be **VERY MINIMAL** as it only affects a minimal set of components…but such a release requires a http://community.developers.atlassian.com/ post and it will be hard to explain it to folks.  I think this can be informational only, eg. "As of the next version, if you're using pass-through CSS like `<Box xcss>` or […], you may see styling conflicts and should opt into `collisionResistantHash` when bumping. See CHANGELOG for more details" — beyond just informing folks, no further process should be necessary — Tim Keir went through a similar, more complex process w/ TS v4 recently.

kylorhall-atlassian (dismissed):
This looks good to me from here, but rollout and internal tooling is the hardest part — eg. internal tooling may be bound to 9-digit hashes in places via regex, but that should be an easy search for eslint or otherwise, eg. I personally may have used `/_[a-z0-0]{8}/` a few times 🙈 but unclear.

This comment I raised for example https://github.com/atlassian-labs/compiled/pull/1920/changes#r3590535998 will happen both internally and in ecosystem for folks using pass-through styles (where we **will have** a mismatch of the `collisionResistantHash`, it's not if, it's when, as we're still going through issues with Emotion vs. Compiled 2 years on)

But it's the right change and it's really just more about comms and process.

packages/react/src/runtime/__tests__/ax.test.ts
pancaspe87: Good pints @kylorhall-atlassian 
> need to cut a major when we distribute every Platform package with `collisionResistantHash=true`
Yes, that is the plan

JakeLane: I think even bumping platform packages like this introduces risk consumers don't update correctly (e.g. they use the old hash internally but the DS components use the new hash, which will cause primitives to break). I'd like to avoid ecosystem as part of this rollout as it will slow us down a lot and add a lot of risk to safely getting this into product in the next few weeks.

How about this:

**1. Bump specificity on new-format rules with `:not(#\9)`**

Emit `._123456abcd:not(#\9) { … }` for new-format atomic rules. Legacy stays as `._abcd1234 { … }`. New format wins the cascade deterministically over legacy for the same property, regardless of insertion order. Uncompressed cost ~10 bytes/rule; after brotli ~0.5–1% because the token repeats across the sheet.

Change lands in `packages/css/src/plugins/atomicify-rules.ts` (this repo) and needs to be mirrored in the SWC plugin **in AFM and Atlaspack**.

**2. One-way compatibility: external-old + internal-new is allowed; the inverse is not**

- **Bundler guard**: `webpack-loader` / `parcel-transformer` / `vite-plugin` fail the build if they see new-format classes from `node_modules` while the local `@compiled/react` is below the floor. Escape hatch via env var for gradual rollout for when platform catches up.
- **Dev-only runtime guard in `ax`**: throws on first render if it sees an `_x…` class it doesn't understand. Extracted out in prod.

**3. Update `ax` to be format-aware (ship first)**

`packages/react/src/runtime/ax.ts`: detect format from `className[1]` length → slice 8 chars for the group key, otherwise slice 5 (current behaviour). Cross-format pairs land in different keys and both survive; the CSS specificity bump from (1) resolves the cascade. Same treatment for `ac()`. Extend `ax.test.ts` to cover legacy-only, new-only, and mixed inputs.

JakeLane: - fail bundling if we find hash collisions (e.g. >1 style rule per hash in the final stylesheet). Recommend using new hash algo to fix it.
- If you pull in primitives or anything using old versions of the hash, make sure they're using the newer version of `@compiled/react` at minimum to ensure ax/ac is compatible. They may need to add a `resolution`.

pancaspe87: Hey @JakeLane just a note that we removed `ac` https://github.com/atlassian-labs/compiled/pull/1921

packages/babel-plugin/src/types.ts
pancaspe87: updated also `website`
- https://github.com/atlassian-labs/compiled/pull/1920/changes#diff-a9822097e7ae159e26a637e3534c1c0d6425d58a8519f88749d43fa98d82d54a
- https://github.com/atlassian-labs/compiled/pull/1920/changes#diff-9ff5057cff8abeb63c5efd6b52d6baa1909ec23e24749d465219a8f218926b36

JakeLane (approved):

PR #1921, opened 2026-07-15
removed unused classNameCompressionMap to simplify code

pancaspe87:
### What is this change?

As discussed in https://github.com/atlassian-labs/compiled/pull/1920#discussion_r3583427805, we are removing `classNameCompressionMap`, a feature that was create over 3 years ago and was never shipped. This allows us to simplify codebase ahead of changes to the hash algorithm.

---

### PR checklist

Don't delete me!

I have...

- [x] Updated the documentation in `website/` -> will do in another PR, I am tracking this in a ticket


packages/babel-plugin/src/class-names/index.ts
```diff
@@ -139,11 +138,7 @@ export const visitClassNamesPath = (path: NodePath<t.JSXElement>, meta: Metadata
       collectedVariables.push(...builtCss.variables);
       collectedSheets.push(...sheets);
 
-      path.replaceWith(
-        t.callExpression(t.identifier(getRuntimeClassNameLibrary(meta)), [
-          t.arrayExpression(classNames),
-        ])
-      );
+      path.replaceWith(t.callExpression(t.identifier('ax'), [t.arrayExpression(classNames)]));
```
pancaspe87, line 141: `get-runtime-class-name-library.ts` is one of the files we deleted. Its only job was to decide which runtime function to emit:

```// what the deleted helper did:
return meta.state.opts.classNameCompressionMap ? 'ac' : 'ax';
```
•  `ac` = the compression-aware runtime (looks up compressed class names)
•  `ax` = the standard runtime (plain atomic class merging)
Since we removed compression entirely, this decision no longer exists — it's always `ax`.

.changeset/tidy-experiments-clean.md
```diff
@@ -0,0 +1,17 @@
+---
+'@compiled/babel-plugin': major
```
pancaspe87, line 2: feature is unused within Atlassian, setting it to `major` for other consumers

liamqma (approved):

packages/babel-plugin/src/babel-plugin.ts
```diff
@@ -195,7 +195,7 @@ export default declare<State>((api) => {
 
           preserveLeadingComments(path);
 
-          appendRuntimeImports(path, state);
+          appendRuntimeImports(path);
```
pancaspe87, line 198: We removed `state` because the only reason it needed state was to check `state.opts.classNameCompressionMap` to decide whether to import `ac` or `ax`; it is no longer needs to know anything about state in `packages/babel-plugin/src/utils/append-runtime-imports.ts`

.changeset/tidy-experiments-clean.md
kylorhall-atlassian: We finally got a `1.0` 🎉

PR #1924, opened 2026-07-20
feat(react): make ax() forward-compatible with longer atomic class names

pancaspe87:
Extract the atomic group key via length-based slicing (`className.length - 4`) instead of the hardcoded `ATOMIC_GROUP_LENGTH = 5` offset. `ax()` now correctly deduplicates both 9-char (4-char group) and upcoming 11-char (6-char group) class names. The two formats produce different-length group keys so they are structurally disjoint and never falsely dedup each other, enabling a safe mixed-hash migration window. Behaviour for existing 9-char classes is unchanged.

Ships ahead of the new hash algorithm so every consumer's runtime is ready before any product emits new-format class names.

This is in preparation of https://github.com/atlassian-labs/compiled/pull/1920; it will simplify rollout: decouples runtime readiness from the hash flip — enabling the independent per-product rollout.

---

### PR checklist

Don't delete me!

I have...

### DIFF
--- a/.changeset/ax-length-based-group-key.md
+++ b/.changeset/ax-length-based-group-key.md
@@ -1,0 +1,13 @@
+---
+'@compiled/react': minor
+---
+
+Update the `ax()` runtime to extract the atomic group key using length-based slicing
+(`className.length - 4`) instead of a hardcoded offset.
+
+This makes `ax()` forward-compatible with longer atomic class names: it now correctly
+deduplicates both the current 9-char format (`_` + 4-char group + 4-char value) and the
+upcoming 11-char format (`_` + 6-char group + 4-char value). Because the two formats
+produce group keys of different lengths, they are structurally disjoint and never falsely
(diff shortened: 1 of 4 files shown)


JakeLane (approved):

JakeLane:
| Case | master | PR #1924 | Δ |
|---|---:|---:|---:|
| `ax() single` | 119,924,263 | 128,693,831 | **+7.3%** |
| `ax() uniques (array)` | 1,220,356 | 1,217,863 | −0.2% (noise) |
| `ax() uniques (as a string)` | 1,759,023 | 1,709,661 | −2.8% |
| `ax() clashes` | 640,615 | 602,033 | **−6.0%** |
| `ax() clashes (as a string)` | 948,028 | 937,724 | −1.1% (noise) |
| `ax() random keys (no clashes)` | 93,280 | 95,901 | +2.8% |
| `ax() random keys (with clashes)` | 66,623 | 67,914 | +1.9% |

PR #1925, opened 2026-07-23
Fix compiled css filename

xtan-atlas:
### What is this change?

change compiled css file name to 'compiled.' + hash(stylesheet) + '.css'

Predictable file name pattern allows upstream to control whether to inline or not

(Optional.)

---

### PR checklist

Don't delete me!

I have...


packages/parcel-optimizer/src/index.ts
```diff
@@ -138,7 +138,7 @@ export default new Optimizer<ParcelOptimizerOpts, unknown>({
         await outputFS.mkdirp(distDir);
       }
 
-      const cssFileName = basename(bundle.displayName, '.html') + '.' + hash(stylesheet) + '.css';
+      const cssFileName = 'compiled.' + hash(stylesheet) + '.css';
```
JakeLane, line 141: this will cause collisions if you have a new entry. Maybe something like:
```suggestion
      const cssFileName = 'compiled.' + basename(bundle.displayName, '.html') + hash(stylesheet) + '.css';
```
or a configurable filename?

xtan-atlas: configurable filename would be better, I guess from parcelrc. If you don't mind I will leave it to you to do it properly :)

PR #1928, opened 2026-07-31
fix: follow up for PR 1920

pancaspe87:
### What is this change?

This is a follow-up to [#1920](https://github.com/atlassian-labs/compiled/pull/1920), which introduced the option in `@compiled/babel-plugin` and `@compiled/css` but did not expose it on `@compiled/webpack-loader`'s options schema (so webpack/rspack builds could not enable it).

---

### PR checklist

Don't delete me!

I have...

### DIFF
--- a/.changeset/webpack-loader-collision-resistant-hash.md
+++ b/.changeset/webpack-loader-collision-resistant-hash.md
@@ -1,0 +1,12 @@
+---
+'@compiled/webpack-loader': minor
+'@compiled/vite-plugin': patch
+'@compiled/parcel-transformer': patch
+'@compiled/parcel-optimizer': patch
+---
+
+Forward the `collisionResistantHash` option through bundler integrations to `@compiled/babel-plugin`.
+
+This is a follow-up to [#1920](https://github.com/atlassian-labs/compiled/pull/1920), which introduced the option in `@compiled/babel-plugin` and `@compiled/css` but did not expose it on `@compiled/webpack-loader`'s options schema (so webpack/rspack builds could not enable it). `@compiled/vite-plugin` already forwarded the option via its babel-plugin options type; this release documents it in the defaults list and package docs for consistency.
+
+Also fix `@compiled/parcel-transformer` and `@compiled/parcel-optimizer` config loading to search upward from the Parcel entry (`config.searchPath`) instead of `options.projectRoot`. In yarn/npm workspaces `projectRoot` is the monorepo root, so per-package `.compiledcssrc` files (including `collisionResistantHash`) were previously ignored.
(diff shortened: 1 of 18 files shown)


.changeset/webpack-loader-collision-resistant-hash.md
```diff
@@ -0,0 +1,12 @@
...
+'@compiled/vite-plugin': patch
+'@compiled/parcel-transformer': patch
+'@compiled/parcel-optimizer': patch
+---
+
+Forward the `collisionResistantHash` option through bundler integrations to `@compiled/babel-plugin`.
+
+This is a follow-up to [#1920](https://github.com/atlassian-labs/compiled/pull/1920), which introduced the option in `@compiled/babel-plugin` and `@compiled/css` but did not expose it on `@compiled/webpack-loader`'s options schema (so webpack/rspack builds could not enable it). `@compiled/vite-plugin` already forwarded the option via its babel-plugin options type; this release documents it in the defaults list and package docs for consistency.
+
+Also fix `@compiled/parcel-transformer` and `@compiled/parcel-optimizer` config loading to search upward from the Parcel entry (`config.searchPath`) instead of `options.projectRoot`. In yarn/npm workspaces `projectRoot` is the monorepo root, so per-package `.compiledcssrc` files (including `collisionResistantHash`) were previously ignored.
```
kylorhall-atlassian: Is this a `minor`, I actually don't know if it's at all concerned, but I figure it's all intended.

kylorhall-atlassian (dismissed):

.changeset/webpack-loader-collision-resistant-hash.md
pancaspe87: minor is the change to `@compiled/webpack-loader`, the rest is patch. Maybe I should have created 2 changesets?

kylorhall-atlassian: Yeah, maybe.  I agree with that, just the `config.searchPath` vs. `options.projectRoot` feels like it _could_ be different?

I don't really know where either of these settings come.  I fully assume it's a completely non-breaking change, no change for consumers?

If you feel it's just a patch, go with that.  I just don't know.

OscarCookeAbbott (dismissed):

.changeset/webpack-loader-collision-resistant-hash.md
pancaspe87: on a second thought, let me undo those changes. 

I can revisit the fix in another moment. It is not needed for the hash collision, it is an error that I surfaced while trying to update tests

OscarCookeAbbott (approved):

PR #1930, opened 2026-08-06
fix: make style bucketing length-agnostic for collisionResistantHash

pancaspe87:
### What is this change?

- Fix `getStyleBucketName` so runtime style bucketing works for both legacy (9-char) and `collisionResistantHash` (11-char) class names — locate the pseudo with `lastIndexOf` instead of a fixed offset at index 10.
- Align the `__perf__` sheet harness and fixtures with the same logic; fix the React 18 CS client bench; refresh the stale babel-plugin benchmark snapshot; add the missing `.js` module mapper to `benchmark.config.json` (an oversight from [#1865](https://github.com/atlassian-labs/compiled/pull/1865)).

- Surfaced the issue when enabling the new hash in a product: 2 units test started to fail due to the wrong bucketing under the the new hash: a user-cell unit test (should render Popup on hover) failed with `toBeVisible()` returning false under the new hash, while passing under the old hash. The rendered DOM and CSS declarations were byte-for-byte equivalent between old and new — only the class names differed.
- Captured the emitted stylesheet in both builds and found the base `{opacity:0}` rule and the `:hover{opacity:1}` rule had swapped relative order between old and new hash.  jsdom doesn't model interactive state (it can't know :hover isn't active) and doesn't implement specificity — it naively applies matching rules in source order, last-wins. Hence, while this won't be an issue at runtime, it flagged an oversight from previous PR where we changed the hash strategy https://github.com/atlassian-labs/compiled/pull/1920

---

### PR checklist

I have...

### DIFF
--- a/packages/babel-plugin/src/__perf__/module-traversal-cache.test.ts
+++ b/packages/babel-plugin/src/__perf__/module-traversal-cache.test.ts
@@ -29,13 +29,17 @@ const transform = (options = {}) =>
 describe('module traversal cache benchmark', () => {
   it('transforms the code correctly', () => {
     expect(transform()).toMatchInlineSnapshot(`
-      "const _4 = \\"._19ith012{border:3px solid pink}\\";
-      const _3 = \\"._bfhk32ev{background-color:pink}\\";
-      const _2 = \\"._syaz5scu{color:red}\\";
-      const _ = \\"._1wybo7ao{font-size:15px}\\";
+      "import * as React from "react";
+      import { ax, ix, CC, CS } from "@compiled/react/runtime";
+      import { colorMixin } from "../__fixtures__/mixins/objects";
+      import { secondary } from "../__fixtures__/mixins/simple";
+      const _4 = "._0KLXruy8mA{background-color:pink}";
+      const _3 = "._1UtDYzGowl{color:red}";
+      const _2 = "._4ya3eE2sAl{font-size:15px}";
+      const _ = "._30huDKzh4V{border:3px solid pink}";
       <CC>
         <CS>{[_, _2, _3, _4]}</CS>
-        {<div className={ax([\\"_1wybo7ao _syaz5scu _bfhk32ev _19ith012\\"])} />}
+        {<div className={ax(["_30huDKzh4V _4ya3eE2sAl _1UtDYzGowl _0KLXruy8mA"])} />}
       </CC>;
       "
     `);
--- a/packages/react/src/runtime/__perf__/utils/sheet.ts
+++ b/packages/react/src/runtime/__perf__/utils/sheet.ts
@@ -97,7 +97,7 @@ function lazyAddStyleBucketToHead(
  * Input:
  *
  * ```
- * "._a1234567:hover{ color: red; }"
+ * "._a1234567890:hover{ color: red; }"
  * ```
  *
  * Output:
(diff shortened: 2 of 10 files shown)


packages/babel-plugin/src/__perf__/module-traversal-cache.test.ts
```diff
@@ -29,13 +29,17 @@ const transform = (options = {}) =>
...
-      const _2 = \\"._syaz5scu{color:red}\\";
-      const _ = \\"._1wybo7ao{font-size:15px}\\";
+      "import * as React from "react";
+      import { ax, ix, CC, CS } from "@compiled/react/runtime";
+      import { colorMixin } from "../__fixtures__/mixins/objects";
+      import { secondary } from "../__fixtures__/mixins/simple";
+      const _4 = "._0KLXruy8mA{background-color:pink}";
+      const _3 = "._1UtDYzGowl{color:red}";
+      const _2 = "._4ya3eE2sAl{font-size:15px}";
+      const _ = "._30huDKzh4V{border:3px solid pink}";
```
kylorhall-atlassian, line 39: Hold on, is this right?  Is this sorted?  `border` should come first.

Maybe this isn't sorted and this is just traversal-only?

packages/react/src/runtime/__perf__/utils/sheet.ts
```diff
@@ -114,15 +114,19 @@ const getStyleBucketName = (sheet: string): string => {
...
+  const firstBracket = sheet.indexOf('{');
+
   /**
-   * We assume that classname will always be 9 character long,
-   * using this the 10th character could be a pseudo declaration.
+   * Atomic class names vary in length (9-char legacy vs 11-char collision-resistant),
+   * so locate the pseudo-selector relative to the opening bracket instead of a
+   * fixed offset. Keep this aligned with `packages/react/src/runtime/sheet.ts`.
    */
-  if (sheet.charCodeAt(10) === 58 /* ":" */) {
```
kylorhall-atlassian, line 121: Ah, yes, good example of what I was worried about with character sets.

kylorhall-atlassian (commented):
Seems fine, but question on that first sorted change.

packages/babel-plugin/src/__perf__/module-traversal-cache.test.ts
pancaspe87: This was a stale snapshot last updated in 2021 ([#920](https://github.com/atlassian-labs/compiled/pull/920)). Benchmarks aren’t run in CI, so it never got refreshed when tests defaulted to `collisionResistantHash` or when shorthand sorting became the default. I hit it locally running `yarn benchmark` after the `__perf__/utils` changes.

Putting `border` first or last in the object doesn’t change the output.

OscarCookeAbbott (approved):

liamqma (approved):

packages/babel-plugin/src/__perf__/module-traversal-cache.test.ts
pancaspe87: Also confirmed by disabling `sortShorthand`— with `sortShorthand: false`, `ax `order matches the old snapshot’s property order (font-size → color → background-color → border). 

Image

order of classes is preserved. Hence, snapshot was stale:

- 2021: Snapshot was created - see [here](https://github.com/atlassian-labs/compiled/pull/920/changes#diff-76477c270102927d3b6c203313f8631169eaaedce9bb7cdbe300edd8cc598ce5)
- 2024: Shorthand sorting was introduced - [PR](https://github.com/atlassian-labs/compiled/pull/1700); this `__perf__` snapshot was never updated

PR #1932, opened 2026-08-07
Fix/compound bucketing

pancaspe87:
### What is this change?
Follow-up to https://github.com/atlassian-labs/compiled/pull/1930. This fixes an oversight of previous patch: I didn't take into account compound pseudo selectors like `.hash:hover:active` resulting in previous fix being partially correct to preserves the catch-all behavior for compounds for both hash strategies.

Before #1930:

`compound rules → ''`

After #1930:

```
:visited:hover  → h
:visited:active → a
```

Since `''` is earlier than h and a in the runtime’s style-bucket ordering, the generated stylesheet order changed.
For many cases, browser rendering still remains the same because specificity dominates source order:

Compiled groups generated atomic rules into ordered buckets (catch-all '', then l, v, h, a, etc.). The order of these buckets in the final stylesheet was not stable when enable the new hash in a product, and changes were caught by CSS snapshot tests. This is how I found out the regression. 
To be fair about impact: in many cases the browser still renders the same, because specificity beats source order. However, we should preserve pre-existing behaviour.

There was also no characterization test for `getStyleBucketName()` before[ #1930](https://github.com/atlassian-labs/compiled/pull/1930). Worse, #1930 added a test that asserted the new (incorrect) behavior — so the code and its test agreed with each other while both had quietly changed the legacy contract.

This time I used a characterization-first approach — pin the old behavior before changing anything, so we can prove the fix preserves it:

 1.  Revert #1930 back to the pre-regression baseline.
 2.  Add a legacy (9-char) characterization suite that records the existing contract
 3.  Re-run the unchanged legacy suite to prove old-hash behavior is preserved.
 4.  Cover the 11-char case too — the same scenarios are run against collision-resistant class names and must produce identical buckets.
5. Also tested it on the failing tests of the products by patching node_modules

you can follow the changes looking at the [commits](https://github.com/atlassian-labs/compiled/pull/1932/commits)

---

### PR checklist

Don't delete me!

I have...


packages/react/src/runtime/__perf__/utils/sheet.ts
```diff
@@ -117,16 +117,20 @@ const getStyleBucketName = (sheet: string): string => {
...
-   * so locate the pseudo-selector relative to the opening bracket instead of a
-   * fixed offset. Keep this aligned with `packages/react/src/runtime/sheet.ts`.
+   * Atomic class names are either 9 chars (legacy hash) or 11 chars
+   * (collisionResistantHash). Detect the width from the boundary character —
+   * index 10 is `:`/`{` for a 9-char class, otherwise the boundary is index 12
+   * for the 11-char hash — and read the pseudo relative to it. Keep this aligned
+   * with `packages/react/src/runtime/sheet.ts`.
    */
-  const colon = sheet.lastIndexOf(':', firstBracket);
-  if (colon !== -1 && colon < firstBracket) {
```
pancaspe87, line 125: Initial code in https://github.com/atlassian-labs/compiled/pull/1930/changes#diff-a5b059e92af3bf8dbe7b242bc001f153807ff1c03c82869d416f63bdd4303a55L121


`if (sheet.charCodeAt(10) === 58 /* ":" */)`

packages/react/src/runtime/sheet.ts
```diff
@@ -129,6 +129,26 @@ function lazyAddStyleBucketToContainer(bucketName: Bucket, opts: StyleSheetOpts)
...
 
+/**
+ * Returns true when `characterCode` is a CSS selector delimiter that can appear
+ * immediately after an atomic class name — i.e. a character that marks the end of
+ * the class token. Covers the declaration block (`{`), pseudo (`:`), attribute
+ * (`[`), class/id qualifiers (`.`/`#`), combinators (space/`>`/`+`/`~`), and the
+ * selector-list separator (`,`). Operates on character codes (not substrings) to
+ * stay allocation-free on the runtime hot path.
+ */
+const isClassBoundary = (characterCode: number): boolean =>
```
pancaspe87, line 140: If this is considered an overkill, it can be simplified back to checking only:

`const classEnd = sheet.charCodeAt(10) === 58 /* ":" */ || sheet.charCodeAt(10) === 123 /* "{" */ ? 10 : 12;`

**The trade-off**
With the minimal check, certain legacy 9-char selector shapes would be bucketed differently. For example 

Image

**Why this is generally "safe"**
The difference is usually safe to render, because these selectors carry higher specificity than a plain atomic rule, so the browser resolves them by specificity regardless of stylesheet order.

Hence. happy to simplify this; eventually we'll update snapshots in products if order/bucket is different

liamqma (dismissed):

JakeLane (dismissed):

kylorhall-atlassian (dismissed):

liamqma (approved):

PR #1934, opened 2026-08-14
fix compiled runtime issue with mixing atomic and non-atomic styles

xtan-atlas:
### What is this change?

Fixes a @compiled/react production runtime bug where injecting a cssMapScoped (.cc-…) sheet wipes previously inserted atomic rules.

•  New dedicated 'cc' bucket in runtime/types.ts and runtime/sheet.ts (placed last in styleBucketOrdering so .cc- rules cascade after all atomic rules).
•  insertNonAtomicRule and the SSR path in runtime/style.tsx now route .cc- sheets to the 'cc' bucket instead of the shared '' catch-all. .cc- rules land on their own <style> element that insertRule never targets.
•  Regression tests in runtime/__tests__/style.test.tsx assert atomic rules survive a subsequent .cc- injection, and that Text.appendData never mutates a <style> owning CSSOM rules.
•  Snapshot in __tests__/browser.test.tsx updated to reflect the new isolation (atomic m bucket and cc bucket now live on separate DOM nodes).
•  Changeset: patch bump for @compiled/react. No public API change.

We hit a global style wipe in Kitsune Storybook production when the editor lazy-mounted. 

Root cause: in production, atomic rules use sheet.insertRule() (CSSOM-only — textContent stays empty), while .cc- non-atomic rules use Text.appendData(). Both routed to the same catch-all <style> element. When a large .cc- sheet is later appended, the browser reparses the sheet from text and discards every previously insertRule-inserted rule → observable UI-wide wipe. Dev mode and extracted-CSS builds (Parcel / webpack extract: true) are unaffected; runtime-injection production clients are vulnerable.

Fix: separating the two insertion strategies onto different DOM nodes makes the reparse-wipe impossible by construction, without touching the atomic insertRule fast path or the cssMapScoped source-order cascade. Products no longer need extraction to be safe from this specific bug.

(Optional.)

---

### PR checklist

Don't delete me!

I have...


packages/react/src/runtime/__tests__/style.test.tsx
xtan-atlas, line 1: added more tests to cover

kylorhall-atlassian (approved):

PR #1935, opened 2026-08-18
fix(vite-plugin): include local development CSS in sorting

lewishealey:
Builds on #1916 to include local Compiled runtime styles in the same development stylesheet as imported `.compiled.css` modules. This prevents a later runtime-injected atomic rule from overriding a responsive rule from a package stylesheet.\n\nIt also strips React `jsxDEV` output so local development styles can be collected.\n\nValidation: `yarn jest packages/babel-plugin-strip-runtime/src/__tests__/strip-runtime-transpiled-code.test.ts packages/vite-plugin/src/__tests__/plugin.test.ts packages/vite-plugin/src/__tests__/dev-css.test.ts --runInBand --watchman=false`.

### DIFF
--- a/packages/vite-plugin/src/dev-css.ts
+++ b/packages/vite-plugin/src/dev-css.ts
@@ -45,9 +45,30 @@ type DevCssServer = {
   };
 };
 
+type DevCssModuleNode = {
+  id: string | null;
+};
+
+type DevCssHotUpdateContext = {
+  file: string;
+  server: {
+    moduleGraph: {
+      getModuleById(id: string): DevCssModuleNode | undefined;
+      invalidateModule(
+        module: DevCssModuleNode,
+        seen: Set<DevCssModuleNode>,
+        timestamp: number,
+        isHmr: boolean
+      ): void;
+    };
+  };
+  timestamp: number;
+};
+
 type DevCssHooks = {
   configResolved(config: { base: string; command: string }): void;
   configureServer(server: DevCssServer): void;
+  handleHotUpdate(context: DevCssHotUpdateContext): void;
   load(id: string, options?: { ssr?: boolean }): string | null;
   resolveId(
     this: DevCssPluginContext,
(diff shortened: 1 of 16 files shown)


packages/vite-plugin/src/dev-css.ts
kylorhall-atlassian, line 1: I think CI is failing at this file.

packages/vite-plugin/src/index.ts
```diff
@@ -131,12 +167,16 @@ function compiled(userOptions: PluginOptions = {}): any {
...
         }
 
-        // Disable stylesheet extraction in development mode
-        const isDevelopment = process.env.NODE_ENV === 'development';
+        // Production extraction writes a CSS asset. In development we instead
+        // strip the runtime and register this module's rules with the virtual
+        // stylesheet that also contains pre-built package CSS.
+        const isDevelopment = process.env.NODE_ENV !== 'production';
         const extract = options.extract && !isDevelopment;
+        const stripRuntime = extract || isDevServer;
```
kylorhall-atlassian: I'd borderline just say we force extraction, that's what this code looks to be doing.

That might reduce a bit of complexity in this file.

liamqma: Addressed in 80852f69. Client-side dev-server transforms now force extraction directly, while build extraction follows the explicit extract option independently of NODE_ENV. Dev SSR remains on the existing runtime path. Regression tests cover all three boundaries.

packages/vite-plugin/src/dev-css.ts
liamqma: Fixed in 80852f69. The new dev hooks no longer import Vite named types, which resolve through Vite’s CJS compatibility declaration without those exports. They use local structural hook types instead, and both the full CJS and ESM package builds now pass.

liamqma (dismissed):

hewislealey:
@atlassian-cla-bot check

lewishealey:
@atlassian-cla-bot check

lewishealey:
@atlassian-cla-bot check

liamqma (approved):

PR #1937, opened 2026-08-24
fix: performance optimisation for insertRule, replacing appendChild with sheet.insertRule

ssysoev-atl:
### What is this change?

This change is removing special logic for inserting rule in dev mode. And makes it behave like a production build.

In dev mode compiled is using appendChild to add styles. This is extremely slow on big amount of CSS. Can lock up browser for 4-5 sec. 

---

### PR checklist

Don't delete me!

I have...

### DIFF
--- a/.changeset/heavy-pianos-behave.md
+++ b/.changeset/heavy-pianos-behave.md
@@ -1,0 +1,10 @@
+---
+'@compiled/react': major
+---
+
+Atomic rules are now injected with `sheet.insertRule()` in every environment. Previously only production used the CSSOM; development and test builds appended a text node per rule, which forces the engine to reparse the whole sheet on every insertion.
+
+Two consequences of dev and test now behaving like production:
+
+- **`<style>` elements no longer contain text.** `insertRule` is a CSSOM-only mutation, so `styleElement.textContent` and `document.head.innerHTML` come back empty and the rules must be read off `sheet.cssRules`. Tests asserting on injected CSS via the DOM's text need updating; `toHaveCompiledCss` from `@compiled/jest` handles both sources as of the accompanying patch. In devtools, dev-mode `<style>` tags now render as empty, the same as production — inspect the element's CSSOM to see the rules.
+- **Dev no longer diverges from production.** The previous fork meant CSSOM-specific bugs could only surface in production builds; both paths are now identical by construction.
--- a/packages/react/src/__tests__/utils/css-text.ts
+++ b/packages/react/src/__tests__/utils/css-text.ts
@@ -1,0 +1,12 @@
+// `sheet.cssRules` reflects rules added via `insertRule` (atomic buckets) as well as
+// rules added as text (the non-atomic `cc` bucket, SSR), so it reads back both paths.
+// The text comes back in the CSSOM's formatting, not Compiled's.
+
+/** The CSS of each `<style>` in `root`, one string per element, in document order. */
+export const getStyleCssTexts = (root: ParentNode = document.head): string[] =>
+  Array.from(root.querySelectorAll('style'), (style) =>
+    Array.from(style.sheet?.cssRules ?? [], (rule) => rule.cssText).join('')
+  );
+
+/** The CSS of every `<style>` in `root`, concatenated. */
+export const getAllCssText = (root?: ParentNode): string => getStyleCssTexts(root).join('');
--- a/packages/react/src/runtime/__tests__/style.test.tsx
+++ b/packages/react/src/runtime/__tests__/style.test.tsx
@@ -2,6 +2,7 @@ import { render } from '@testing-library/react';
 import React from 'react';
 import type { ComponentType } from 'react';
 
+import { getAllCssText, getStyleCssTexts } from '../../__tests__/utils/css-text';
 import StyleWithContainer from '../style';
 import { StyleContainerProvider } from '../style-container';
 import type * as StyleContainerModule from '../style-container';
@@ -47,7 +48,7 @@ describe('<Style />', () => {
     createIsolatedTest((Style) => {
       render(<Style>{[`.b { display: block; }`]}</Style>);
 
-      expect(document.head.innerHTML).toInclude('<style>.b { display: block; }</style>');
+      expect(getAllCssText()).toInclude('.b {display: block;}');
       expect(console.error).not.toHaveBeenCalled();
     });
   });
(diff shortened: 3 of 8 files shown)


packages/react/src/__tests__/utils/style-capture.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Importing this module installs the patch, so it has to be imported before any rule is
+ * inserted — which top-level imports guarantee, since rules are only inserted on render.
+ */
+
+/** Rules recorded per sheet, parallel to that sheet's `cssRules`. */
+const insertedRules = new WeakMap<CSSStyleSheet, string[]>();
+
+const { insertRule } = CSSStyleSheet.prototype;
+
+CSSStyleSheet.prototype.insertRule = function (rule: string, index?: number): number {
```
ssysoev-atl: this is a weird solution...

packages/react/src/runtime/__tests__/style.test.tsx
```diff
@@ -304,7 +301,7 @@ describe('<Style />', () => {
         expect(textNodes).toHaveLength(1);
 
         // Both sheets must be present in the text node
-        const text = getAllCssText();
+        const text = Array.from(document.head.querySelectorAll('style'), styleCss).join('');
```
ssysoev-atl: make it a helper as well

.changeset/heavy-pianos-behave.md
```diff
@@ -0,0 +1,10 @@
+---
+'@compiled/react': major
```
ssysoev-atl: not sure if it should be a major... going from 1 -> 2 for this is a bit weird...

packages/react/src/__tests__/utils/css-text.ts
```diff
@@ -0,0 +1,12 @@
+// `sheet.cssRules` reflects rules added via `insertRule` (atomic buckets) as well as
+// rules added as text (the non-atomic `cc` bucket, SSR), so it reads back both paths.
+// The text comes back in the CSSOM's formatting, not Compiled's.
+
+/** The CSS of each `<style>` in `root`, one string per element, in document order. */
+export const getStyleCssTexts = (root: ParentNode = document.head): string[] =>
+  Array.from(root.querySelectorAll('style'), (style) =>
+    Array.from(style.sheet?.cssRules ?? [], (rule) => rule.cssText).join('')
+  );
```
kylorhall-atlassian, line 9: Does this handle both?  Or did this always handle both, eg. if `style.textContent` is set it also has `style.sheet.cssRules`?

It also kinda feels like it doesn't need to handle both.

packages/react/src/runtime/__tests__/style.test.tsx
```diff
@@ -180,28 +183,12 @@ describe('<Style />', () => {
 
       rerender(<Style>{[`.second-render { display: block; }`]}</Style>);
 
-      expect(document.head.innerHTML).toInclude('.second-render { display: block; }');
+      expect(getAllCssText()).toInclude('.second-render {display: block;}');
```
kylorhall-atlassian, line 186: There's zero concern here of like development being used to create `<style>` which is then saved for SSR and distributed as HTML, because if React/Tesseract/etc would stringify the page as HTML, it'd capture the `<style>._abcd1234{…}</style>`, right?

I feel Emotion used `insertRule` anyways, so surely this is the case.

Also, this should've always used `insertRule` anyways?

Just thinking aloud.

packages/react/src/runtime/sheet.ts
```diff
@@ -236,17 +236,12 @@ export const getStyleBucketName = (sheet: string): Bucket => {
...
-    try {
-      sheet.insertRule(css, sheet.cssRules.length);
-    } catch {}
-  } else {
-    style.appendChild(document.createTextNode(css));
-  }
+  // Used to avoid unhandled exceptions across browsers with prefixed selectors such as -moz-placeholder.
+  try {
+    sheet.insertRule(css, sheet.cssRules.length);
+  } catch {}
```
kylorhall-atlassian, line 244: Is there any circumstance we'd throw an error here or?  That'd be breaking though, and the `moz-placeholder` example might have merit to the point where there's no point in thinking about a breaking change as it'd not be 1:1.

It sounds like even in production before…nothing happened, so no difference to development realistically?  Or development and production looks the same?

LGTM.

kylorhall-atlassian (dismissed):

packages/react/src/runtime/__tests__/style.test.tsx
```diff
@@ -114,17 +115,18 @@ describe('<Style />', () => {
...
-        <style>._i1234567:focus-within{ color: black; }</style>
-        <style>._f1234567:focus{ color: pink; }</style>
-        <style>._h1234567:focus-visible{ color: white; }</style>
-        <style>._a1234567:hover{ color: red; }</style>
-        <style>._b1234567:active{ color: blue; }</style>
-        <style>@media (max-width: 800px){ ._e1234567{ color: yellow; } }</style>
-        "
+      expect(getStyleCssTexts()).toMatchInlineSnapshot(`
+        [
+          "._d1234567 {display: block;}",
```
pancaspe87, line 120: Are these actual Compiled classes?

.changeset/heavy-pianos-behave.md
```diff
@@ -0,0 +1,10 @@
+---
+'@compiled/react': minor
```
pancaspe87: should this be `major` in case consumers will be expected to update tests as well?

packages/react/src/runtime/__tests__/style.test.tsx
ssysoev-atl: yeah, they shouldn't have changed from previous tests, only spacing is different

.changeset/heavy-pianos-behave.md
ssysoev-atl: yep :( not sure, wanted to a major, but then going from 1 -> 2 for this change sounds weird to me, but happy to bump it to major...

packages/react/src/runtime/sheet.ts
ssysoev-atl: now they look the same, where before it could succeed in dev and fail in prod...

packages/react/src/runtime/__tests__/style.test.tsx
ssysoev-atl: yeah in ssr it would output style tags (if runtime mode)

packages/react/src/__tests__/utils/css-text.ts
ssysoev-atl: IIUC, this ignores textContent

.changeset/heavy-pianos-behave.md
pancaspe87: yeahh, I had the same feeling when working on the new hash. I ended up with a `major` to highlight the expectations. `minor` seems fine, unless it could break something else

pancaspe87 (dismissed):

.changeset/heavy-pianos-behave.md
ssysoev-atl: marked as major

pancaspe87 (approved):

PR #1943, opened 2026-08-29
Support TypeScript satisfies expressions in computed CSS keys

kylorhall-atlassian:
## Summary

Support TypeScript `TSSatisfiesExpression` nodes in computed CSS object keys.

- Unwraps `satisfies` expressions for CSS and CSS Map key processing
- Adds focused regression coverage
- Includes a patch changeset for `@compiled/babel-plugin`

Example:
```tsx
import type { MediaAboveSm } from '@atlaskit/css/at-rules/media-above-sm';
import type { MediaAboveLg } from '@atlaskit/css/at-rules/media-above-lg';
import type { Container } from '@atlaskit/css/at-rules/container';

const styles = css({
  ['@media (min-width: 48rem)' satisfies MediaAboveSm]: { padding: '8px' },
});

const mapStyles = cssMap({
  root: {
    padding: '2px'
    ['@media (min-width: 48rem)' satisfies MediaAboveLg]: { padding: '8px' },
    ['@container id (width > 100px)' satisfies Container]: { padding: '4px' },
  }
});
export default () => <div css={[styles, mapStyles.root]} />;
```

### DIFF
--- a/packages/babel-plugin/src/css-map/__tests__/at-rules-and-selectors.test.ts
+++ b/packages/babel-plugin/src/css-map/__tests__/at-rules-and-selectors.test.ts
@@ -9,6 +9,32 @@ describe('css map advanced functionality (at rules, selectors object)', () => {
   const transform = (code: string, opts: TransformOptions = {}) =>
     transformCode(code, { pretty: false, ...opts });
 
+  it('should support satisfies expressions in computed at-rule keys', () => {
+    const actual = transform(`
+      import type { MediaAboveLg } from '@atlaskit/css/at-rules/media-above-lg';
+      import type { Container } from '@atlaskit/css/at-rules/container';
+      import { cssMap } from '@compiled/react';
+
+      const styles = cssMap({
+        root: {
+          padding: '2px',
+          ['@media (min-width: 48rem)' satisfies MediaAboveLg]: { padding: '8px' },
+          ['@container id (width > 100px)' satisfies Container]: { padding: '4px' },
+        },
+      });
+
+      ${EXAMPLE_USAGE}
+    `);
+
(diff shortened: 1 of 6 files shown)


packages/babel-plugin/src/css-prop/__tests__/behaviour.test.ts
```diff
@@ -13,6 +13,20 @@ describe('css prop behaviour', () => {
...
+      import type { MediaAboveSm } from '@atlaskit/css/at-rules/media-above-sm';
+      import '@compiled/react';
+
+      <div css={{
+        ['@media (min-width: 48rem)' satisfies MediaAboveSm]: { padding: '8px' },
+      }} />;
+    `);
+
+    expect(actual).toInclude('@media (min-width:48rem)');
+    expect(actual).toInclude('padding:8px');
```
kylorhall-atlassian: [for AI] I'd like if this actually did more and tested the exact output.

packages/babel-plugin/src/css-map/__tests__/at-rules-and-selectors.test.ts
```diff
@@ -9,6 +9,32 @@ describe('css map advanced functionality (at rules, selectors object)', () => {
...
+      ${EXAMPLE_USAGE}
+    `);
+
+    expect(actual).toIncludeMultiple([
+      'padding:2px',
+      '@media (min-width:48rem)',
+      'padding:8px',
+      '@container id (width>100px)',
+      'padding:4px',
+    ]);
```
kylorhall-atlassian, line 34: [for AI] I'd like if this actually did more and tested the exact output.

JakeLane (approved):

PR #1949, opened 2026-09-14
Support TypeScript satisfies in static CSS evaluation

vykimnguyen:
Support TypeScript satisfies expressions during static CSS evaluation.
Compiled now unwraps satisfies expressions in the same way as existing as expressions. This enables static evaluation of:

- cssMap values, including nested values
- Repeated or nested TypeScript assertions
- Member access through satisfies-wrapped bindings

**Testing**

- Added cssMap value coverage
- Added member-access coverage for satisfies-wrapped bindings
- Existing Babel plugin behavior tests pass

---

### PR checklist

Don't delete me!

I have...


liamqma (approved):

packages/babel-plugin/src/css-prop/__tests__/behaviour.test.ts
```diff
@@ -29,6 +29,17 @@ describe('css prop behaviour', () => {
     ]);
   });
 
+  it('should resolve member expressions through satisfies expressions', () => {
```
kylorhall-atlassian, line 32: Ah, that's good. Not sure I supported that syntax for media queries.

@JakeLane this might solve the one-off custom theming we see in a few places 👀

kylorhall-atlassian (approved):

pancaspe87 (approved):

PR #1953, opened 2026-09-16
Support TSSatisfiesExpression and TSAsExpression in ESLint plugin

JakeLane:
### What is this change?

Support ts ast nodes in the eslint plugin.

shorthand-property-sorting is crashing with these nodes.

---

### PR checklist

Don't delete me!

I have...


.changeset/two-memes-try.md
kylorhall-atlassian, line 1: Did we ask if the agent needs to change it anywhere else? 😅

vykimnguyen (approved):

packages/eslint-plugin/src/rules/shorthand-property-sorting/__tests__/rule.test.ts
```diff
@@ -7,6 +7,17 @@ const includedImports = ['@compiled/react', '@atlaskit/css'] as const;
 
 tester.run('shorthand-property-sorting', shorthandFirst, {
   valid: includedImports.flatMap((imp) => [
+    ...['satisfies string', 'as string', 'as const satisfies string'].map((assertion) => ({
+      name: `inline ${assertion} in styled callback (${imp})`,
+      code: outdent`
+        import { styled } from '${imp}';
+        const Component = styled.div<{ expanded?: boolean }>(({ expanded }) =>
+          expanded
+            ? \`width: 100%;\`
```
kylorhall-atlassian, line 16: Does this support, or should we not care, about say `[padding satisfies 'padding-left']` being sortable?

kylorhall-atlassian (approved):

